PR #12, opened 2026-08-19
Add optional Solana wallet-link proof for Solana Mobile MWA clients

### LINKED ISSUE #11: Proposal: optional MWA wallet-link proof for existing did:key signed writes
Hi FLOP Labs — a small, deliberately non-protocol proposal for discussion.

`technocore-chat` already offers an excellent low-friction model: anonymous writes stay available, while the optional Ed25519 `did:key` lane gives continuous authorship without accounts or a resolver. I think there may be a narrow mobile interoperability extension worth testing: allow a mobile web app / Android dApp to use **Solana Mobile Wallet Adapter (MWA)** to attach a voluntary proof that a Solana wallet key signed a binding statement for an existing `did:key`.

### Why this may be useful

It could give mobile-first agent operators an optional, portable wallet-linked pseudonym while preserving the project’s core property: fetch-only and no-wallet peers remain full participants. This would be an experiment for technocore-chat operated by FLOP Labs, **not a FLOP protocol feature**.

### Smallest viable scope

- Keep the present `did:key` signed-write lane authoritative and unchanged.
- A companion PWA creates/uses a local `did:key`, authorizes a wallet via MWA, and asks it to sign a canonical, domain-separated link statement:
  `protocol version + technocore.chat origin + did:key + wallet pubkey + random challenge + issued_at + expires_at`.
- On an opted-in write, submit the normal `did:key` signature plus that wallet-link proof as optional metadata.
- Verify both signatures offline; expose only a neutral machine-readable/display state such as `wallet-linked (proof valid)` and a wallet-key fingerprint.
- Proofs should be short-lived and travel with the write, so there is no account system, identity directory, or persistent wallet-to-DID registry.
- If MWA is unavailable or declined, the normal signed / unsigned flows work exactly as today.

### Explicit non-goals

This should **not** introduce wallet requirements, accounts, on-chain state, token mechanics, rewards, airdrops, balance checks, reputation, rate-limit privileges, room permissions, or a Seeker-only path. A wallet signature proves control of that signing key for the scoped statement only. It does **not** prove personhood, wallet ownership in a legal sense, device type, Seeker usage, Seed Vault use, hardware attestation, uniqueness, or Sybil resistance.

### Security constraints

- Use MWA from the dApp; do not integrate a dApp directly with Seed Vault SDK.
- Strictly domain-separate and canonically encode the bytes; bind exact origin, `did:key`, wallet public key, challenge, issuance and expiry.
- Verify the existing `did:key` write signature separately from the wallet proof.
- Treat the link proof as public and replayable evidence within its expiry window, never as an authorization credential.
- Account for GET URL/log exposure and URL-size limits; the proof must contain no secret. If this is not acceptable for the hosted surface, stop at a standalone PWA proof-bundle demo.
- Add test vectors for valid signatures and wrong-origin, wrong-DID, expiry, malformed-payload, and replay cases.

Would maintainers be open to a short design discussion before any implementation? I can contribute a minimal PWA proof-of-concept and test vectors, with the default zero-auth experience untouched.

### References

- technocore-chat signed writes and security model: https://github.com/flop-labs/technocore-chat
- Solana Mobile Wallet Adapter: https://github.com/solana-mobile/mobile-wallet-adapter

busminer:
Closes #11

This draft adds a narrow, optional `solana_wallet_link` extension to the existing signed `POST /r/<room>` write path.

It is intended for Solana Mobile clients using Mobile Wallet Adapter (MWA) that want to attach a short-lived Solana wallet-link proof beside an existing `did:key` signed write.

## What this draft does

When a signed POST includes `solana_wallet_link`, the server:

1. verifies the existing Technocore `did:key` signature first;
2. verifies that a presented Solana Ed25519 public key signed a deterministic, versioned binding statement containing:
   - canonical configured origin;
   - the signed-write DID;
   - wallet public key;
   - challenge;
   - issued/expiry timestamps;
3. accepts a maximum proof validity interval of 15 minutes;
4. performs no persistence of wallet, proof, signature, fingerprint, or wallet-to-DID metadata.

The canonical origin authority is explicit `CHAT_PUBLIC_URL`; request `Host` is not used as proof authority.

## Intentionally unchanged

- anonymous writes;
- normal `did:key` signed POSTs without `solana_wallet_link`;
- all GET routes, including signed GET;
- room and note storage format;
- accounts, identity registry, authorization and permissions;
- on-chain behavior, token behavior, rewards or reputation.

If `solana_wallet_link` is absent, the existing signed POST path is unchanged.

## What this proof means — and does not mean

A successful validation means only that the presented Solana Ed25519 key signed this narrowly scoped, short-lived binding statement.

It is not:

- Seeker or device attestation;
- Mobile Wallet Adapter attestation;
- proof of human identity;
- authorization or permission;
- reputation, anti-Sybil, or proof-of-personhood;
- token, balance, reward, or on-chain state.

The proof is public/replayable evidence within its validity interval and is not an authorization credential.

## Tests

The draft includes deterministic verifier tests and POST integration coverage for:

- valid proof acceptance;
- existing signed POST without proof unchanged;
- invalid/tampered proof rejection;
- proof DID mismatch rejection;
- missing `CHAT_PUBLIC_URL` rejection only for proof-bearing requests;
- anonymous flow unchanged;
- normal signed flow unchanged;
- signed GET unchanged;
- canonical payload, origin, timestamps, strict field handling, Base58, key size, signature encoding, and control-character validation.

## Maintainer questions

1. **Verify-only representation:** this draft deliberately validates the proof without persisting wallet identity metadata. Would you prefer this to remain a validation-only primitive, or should a follow-up/minimal revision expose a neutral per-message marker or wallet-key fingerprint?

2. **Canonical wire format:** does the proposed strict versioned UTF-8 line format look like the right shape for cross-client implementation?

3. **Validity window:** is a maximum 15-minute validity interval appropriate?

4. **Self-hosting:** this requires explicit `CHAT_PUBLIC_URL` for proof-bearing requests so the service does not treat a caller-supplied Host header as authority. Is that the right configuration boundary?

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,24 @@ smaller guarantee than "until the ring forgets"; signatures still prove authorsh
 The text view shows `<z6Mk…2doK>` for a verified writer and `<~nick>` for self-asserted. Full DIDs
 are JSON-only: 50 lines of 56-character identifiers is ~1200 tokens of the agent's context.
 
+### Optional Solana Mobile MWA client wallet-link proof
+
+A signed `POST /r/<room>` may additionally carry `solana_wallet_link`: an optional, short-lived
+(maximum 15-minute validity) Solana wallet-link proof designed for Solana Mobile clients using
+Mobile Wallet Adapter (MWA).
+The server cryptographically verifies only that the presented Solana Ed25519 public key signed the
+canonical binding to the existing `did:key`, configured origin, challenge and validity interval.
+
+Anonymous writes, ordinary `did:key` writes and every GET route — including signed GET — stay
+unchanged. The proof is not Seeker or device attestation, MWA attestation, identity, authorization,
+reputation, a permission, a token signal or on-chain state. It is public/replayable evidence within
+its validity window and is not an authorization credential.
+
+The proof is verify-only: no wallet, signature or wallet-to-DID metadata is persisted. Its `origin`
+must equal the canonical origin configured by the operator in `CHAT_PUBLIC_URL`; request `Host` is
+never authority for this check. Self-hosted deployments that do not configure `CHAT_PUBLIC_URL`
+continue serving all normal flows, but reject only proof-bearing POSTs.
+
 ## Room classes
 
 A room name is `<class>-…-<body>`, and classes compose by prefix: `mb-p-<random>` is a private
--- a/src/app.py
+++ b/src/app.py
@@ -35,6 +35,7 @@ import didkey
 import limit
 import manifest
 import store
+import wallet_link
 from store import StoreConflictError, StoreError
 
 # The CHAT_* knobs are read from the environment exactly once, in config — the only
@@ -1375,6 +1376,26 @@ async def room_post(request: Request) -> Response:
         signer = _signer(did, sig, nonce, f"{room}|{nonce}|{body}")
         if isinstance(signer, Response):
             return signer
+    if "solana_wallet_link" in payload:
+        if signer is None:
+            return text(
+                "400 solana_wallet_link is an optional Solana Mobile MWA client wallet-link proof "
+                "for an existing did:key signed POST; send did, sig and nonce as well.",
+                400,
+            )
+        if not PUBLIC_URL:
+            return text(
+                "400 solana_wallet_link needs CHAT_PUBLIC_URL: configure this instance's canonical "
+                "public origin before submitting a Solana Mobile MWA client wallet-link proof. Existing writes "
+                "without it.",
+                400,
+            )
+        try:
+            wallet_link.verify_wallet_link(
+                payload["solana_wallet_link"], did=signer, configured_origin=PUBLIC_URL
+            )
+        except wallet_link.WalletLinkError as exc:
+            return text(f"400 invalid solana_wallet_link: {exc}", 400)
 
     # Everything below is blocking disk work: the gate stats the room and walks the rooms
     # directory, the append takes an flock and fsyncs, and the reaper may run inside it.
--- a/src/app.py
+++ b/src/app.py
@@ -1383,7 +1383,7 @@ async def room_post(request: Request) -> Response:
                 "for an existing did:key signed POST; send did, sig and nonce as well.",
                 400,
             )
-        if not PUBLIC_URL:
+        if not config.PUBLIC_URL:
             return text(
                 "400 solana_wallet_link needs CHAT_PUBLIC_URL: configure this instance's canonical "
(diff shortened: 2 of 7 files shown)


busminer:
Rebased onto `main` and out of draft. The branch was 69 commits behind; every conflict was
something having moved rather than a disagreement, so this only adapts to where things live now:

- the manual left `app.py` for `src/manual.md`, so the wallet-link paragraph now sits at the end
  of SIGNING, before RENDERING;
- the room POST body became the shared `_ROOM_POST_BODY`, so `solana_wallet_link` is a
  `_WALLET_LINK_SCHEMA` constant referenced from it and both routes document it, rather than a
  re-inlined copy;
- per-response descriptions became shared constants, and `main`'s are kept — the wallet-link 400
  is already covered by the field's own schema;
- `CHAT_PUBLIC_URL` moved into `config`, so the handler reads `config.PUBLIC_URL`. The bare name
  had silently become a `NameError` at request time.

Two test adjustments, both to `main`'s current behaviour rather than to the feature: records now
carry the `sig` they were accepted on, so the unchanged-record case asserts the fields it is
actually about instead of the whole dict; and the client fixture also reloads `config`, whose
settings are read at import, so each client sees the origin its own case set.

36/36 wallet-link tests pass, `ruff check` and `ruff format --check` are clean. I can only run the
full suite on Windows, where ten cases fail on `main` too — they need real POSIX file locking and
an inode surviving compaction — so CI is the real check.

On #11: the `solana-wallet` profile of `technocore-account-binding-v1`, including the 15-minute
window, came out of this branch, and @khosimorafo's grammar and verifier now cover it. Happy to
follow whichever shape you prefer — narrow this to the challenge-bound POST extension, drop it in
favour of the grammar landing through `patterns.md`, or split verifier and vectors out separately.

Context for why it exists: I have a native Android client for Solana Mobile Seeker that creates a
Technocore identity and publishes it from the phone. Without something like this, a mobile client
can prove it holds a `did:key` but has no standard way to show the Solana wallet alongside it, so
everyone invents their own statement format — which is the failure `patterns.md` opens by naming.

luch91 (changes requested):
The optional Solana wallet-link extension changes the signed POST contract and adds security-sensitive proof validation, but exact head 578b44130ed9f7773485110b3a34c16b9f1eaedb has only protected-files/overlap checks visible. Please obtain maintainer decisions on the proposed wire contract, validity semantics, and deployment configuration, then run the full HTTP/OpenAPI/security regression suite and required CI before approval.

PR #13, opened 2026-08-19
feat: add TypeScript client example

Crypto-Crood:
## Summary

Adds a TypeScript client example for interacting with the TechnoCore Chat HTTP API.

## What this includes

- Typed TypeScript client for TechnoCore rooms
- Read messages from a room
- Send messages to a room
- Cursor-based message fetching with `since`
- Long-polling support with `wait`
- KV note read/write helpers
- Continuous message listener example
- TypeScript build configuration and npm scripts

## Testing

Tested against the live TechnoCore API.

Verified:

- Sending messages successfully
- Reading room messages
- Reading messages using sequence cursors
- Long-polling for new messages
- Receiving new messages from another client
- Continuous cursor updates
- `npm run build` passes successfully

## Files

- `src/client.ts` — TypeScript API client
- `src/example.ts` — continuous listener example
- `tsconfig.json` — TypeScript configuration
- `package.json` — npm scripts and dependencies


sv:
Thanks for putting this together. We’re closing the PR because we do not currently have evidence that a maintained TypeScript client is an adoption bottleneck. The anonymous API is already directly usable with fetch, so an official wrapper needs to provide substantially more value than syntax convenience.
This implementation also omits parts we would need to support as an official client: POST for the full Unicode size contract, actionable 429 backoff, the 10-second wait ceiling, and the secure signed lane.
If a concrete adopter or integration later demonstrates a TypeScript-specific need, please bring that use case first. We would prefer generated examples or OpenAPI-based tooling over maintaining a language SDK without established demand.

PR #14, opened 2026-08-19
Add signed did:key client and signature test

krypt0bi:
SUMMARY

Adds a small Ed25519 `did:key` client and a negative signature-verification test for the signed write lane.

WHAT I CHANGED

- Added `signed_client.py`
  - Generates an Ed25519 keypair.
  - Creates a `did:key` using the Ed25519 multicodec.
  - Builds the canonical `<room>|<nonce>|<text>` payload.
  - Signs the payload using Ed25519.
  - Sends the signed request to `/say-signed`.

- Added `test_bad_signature.py`
  - Reuses a valid DID/signature.
  - Modifies the signed message content.
  - Confirms the server rejects the modified content with HTTP 403.

- Updated `uv.lock` to reflect the current `0.4.0` project version.

VERIFICATION

The signed client successfully received HTTP 200.

The modified-content test returned:

`PASS: modified signed content was rejected with HTTP 403`

This confirms that changing the message after signing causes signature verification to fail.

### DIFF
--- a/signed_client.py
+++ b/signed_client.py
@@ -1,0 +1,102 @@
+import base64
+import urllib.parse
+import urllib.request
+
+from cryptography.hazmat.primitives.asymmetric.ed25519 import Ed25519PrivateKey
+
+
+# Base58btc alphabet used by did:key.
+B58 = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"
+
+# multicodec prefix for Ed25519 public keys.
+MULTICODEC_ED25519 = b"\xed\x01"
+
+
+def base58_encode(data: bytes) -> str:
+    n = int.from_bytes(data, "big")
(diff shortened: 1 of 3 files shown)


sv:
Thanks for working on the signed lane. We’re closing this form because the added files are standalone network scripts outside the configured test suite, with a hard-coded endpoint and nonce, and they do not implement the server’s canonical text sweep.
Current main already has integration coverage for valid, invalid, replayed, swept, POST, mailbox, and owned-room signatures.
We agree that clients should not have to reimplement signing. The direction we would consider is one official local signer integrated with the existing MCP/CLI distribution, using vetted cryptography and handling key storage, canonicalization, and nonces. That needs a separate design rather than more standalone client scripts.

krypt0bi:
Thanks for the input. I'll work on this and revert back. Cheers.

PR #30, opened 2026-08-20
fix(stats): bound _identities so distinct_identities stays a live diagnostics

CryptoFridge:
## What

`_identities` becomes a bounded `OrderedDict` LRU instead of a `set` that stops
accepting entries at `MAX_IDENTITIES`. `/stats` callers see `distinct_identities`
track the most recent 50k callers rather than freezing at 50k forever.

## Why

The cap and the field's stated meaning currently pull against each other. The comment
on `MAX_IDENTITIES` says a counter that OOMs is not a diagnostic, right, but the
`set` only adds while under the cap, so once it fills nothing is ever evicted. Past
50k distinct callers the number is stuck at 50k and stops reflecting anything, which
is the same problem in the other direction: a counter that has stopped moving is also
not a diagnostic.

That matters most in the case `/stats` documents. The whole point of
`distinct_identities` is telling an operator whether per-IP limits are actually
per-IP, and the deployment where the count runs away is exactly the misconfigured one
— every caller minting a fresh identity per request. So the metric degrades soonest
where it is needed most.

`OrderedDict` with `move_to_end` / `popitem(last=False)` is the idiom `_buckets` and
`_rooms_cache` already use, so this is applying the existing pattern rather than
introducing one. Re-inserting refreshes recency, so an established caller survives a
flood of new IPs.

The `/stats` comment is updated because the meaning genuinely changes: the field is
now "callers told apart among the most recent `MAX_IDENTITIES` seen", not "ever". I
left the field name alone rather than widening the diff, but happy to rename it if
you'd rather the name carried that.

Doesn't touch anything else open. Sits on top of #28 only in the sense that both are
in `read_json`'s file; they don't conflict.

## Checks

- [x] `uv run python -m pytest tests -q`
      or say "nothing new"

Four tests added covering eviction order, recency refresh on re-insert, the size
bound under sustained flood, and the `/stats` field reporting the windowed count.
`ci.yml` dispatched on the fork — lint + tests + image build green on `ubuntu-latest`.

`README.md` describes `distinct_identities` as the count being stuck near 1 alongside
a rising `proxied_requests_ignored`, which is still accurate under a window, so no
doc change was needed there. Nothing in `/skill.md`, `src/patterns.md` or
`mcp/README.md` mentions the field — `/stats` is internal and not in the generated
manifests.

**Abuse surface: nothing new.** No new route, parameter or lane. Memory is bounded by
the same `MAX_IDENTITIES` as before — strictly tighter, since the old set could sit at
the cap indefinitely while the LRU evicts. A flood of unique IPs can now push an
established caller's identity out of the *diagnostic*, but `_identities` is
observability only: it never gates a request, and rate-limit state lives in `_buckets`,
which is unchanged.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -194,7 +194,11 @@ _started = time.time()
 # and an `identities` of 1 is not rate limiting anyone individually — it is rate limiting
 # the CDN, and the room budget is being shared by the entire internet.
 _proxy_evidence: dict[str, int] = {"proxied_requests": 0}
-_identities: set[str] = set()
+# Bounded LRU of distinct client IPs seen by the rate limiter.
+# Mirrors _buckets: OrderedDict with move_to_end / popitem(last=False)
+# evicts the idlest entry when full.  Re-inserting an existing identity
+# refreshes its recency so active callers survive a flood of new IPs.
+_identities: OrderedDict[str, None] = OrderedDict()
 MAX_IDENTITIES = 50_000  # bounded like _buckets; a counter that OOMs is not a diagnostic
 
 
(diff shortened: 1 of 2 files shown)


sv:
Please make the improvement observable through /stats. The LRU changes retained IPs, but distinct_identities still freezes at 50,000, and the new /stats test passes on current main. Expose saturation/churn or implement a real time-based window, with an API-level regression test that fails before the fix.

CryptoFridge:
Superseded by #388 which applies the same fix to the current codebase (after the limit.py split and dupe filter). Closing it now.

PR #39, opened 2026-08-22
Propose official local signer

krypt0bi:
## Summary

Proposes an official local signer for the existing `did:key` signed-write lane.

The design covers:

- Ed25519 key generation and local key persistence
- `did:key` derivation
- the server's existing text-sweep implementation
- canonical signed room messages
- existing signed note writes
- local nonce allocation and crash-safe persistence
- MCP and CLI integration
- security boundaries and replay limitations
- incremental implementation and testing phases

This is a design proposal only; it does not introduce implementation changes.

## Motivation

This follows the discussion around PR #14. The previous implementation attempt was closed because the added client scripts were standalone and did not integrate with the existing client/test architecture or implement the server's canonical text sweep.

The existing signed-write protocol already has server-side integration coverage, but clients should not need to independently reimplement the signing mechanics.

This proposal aims to provide one supported local implementation while preserving the existing HTTP protocol and unsigned lane.

### DIFF
--- a/docs/signer-design.md
+++ b/docs/signer-design.md
@@ -1,0 +1,912 @@
+# Local Signer
+
+This document proposes an official local signer for the `did:key` signed-write
+
+lane.
+
+The goal is not to add another protocol or client abstraction. It is to give
+
+clients one small, supported implementation of the signing mechanics the
+
+server already enforces: Ed25519 key handling, `did:key` encoding, text
+
+sweeping, canonicalization, signatures, and local nonce persistence.
+
+The unsigned HTTP lane remains unchanged. An agent with only a fetch tool can
+
+still use the unsigned API.
+
+## Implementation status
+
+PR #54 introduced `scripts/sign.py` as an initial implementation of the
+signed-write signing core described by this proposal. It covers Ed25519
+signing, `did:key` derivation, server-compatible text sweeping, and canonical
+signing inputs for signed room messages and the existing signed note surfaces.
+
+This document describes the broader architecture and remaining evolution
+toward a supported local signer. In particular, the durable key-management,
+local nonce allocation and persistence, crash-safety, and client integration
+requirements below are design targets rather than claims about the current
+`scripts/sign.py` implementation.
+
+The implementation and the design are therefore intentionally related but not
+identical: `scripts/sign.py` provides the current signing core, while this
+document defines the fuller supported signer that can grow around that core.
+
+## Goals
+
+The signer should:
+
+- use the server's existing `did:key` / Ed25519 signed-write protocol;
+
+- apply the same text sweep as the server before constructing signing input;
+
+- make canonicalization explicit and testable;
+
+- generate and persist an Ed25519 key locally;
+
+- derive the corresponding `did:key` without a resolver;
+
+- maintain monotonically increasing local nonces;
+
+- support signed room messages;
+
+- support the existing signed note writes used by `room-owners` and
+
+  `room-allow`;
+
+- expose one small local API suitable for both the MCP and CLI clients;
+
+- keep private key material local to the machine running the signer;
+
+- use the project's existing `cryptography` dependency rather than adding a
+
+  second cryptographic implementation.
+
+## Non-goals
+
+The signer does not:
+
+- replace the unsigned HTTP API;
+
+- add a new identity or account-registration system;
+
+- add a server-side nonce endpoint;
+
+- synchronize nonce state with the server;
+
+- introduce a new signature format;
+
+- invent new signed note namespaces;
+
+- provide a remote key-management service;
+
+- guarantee replay protection beyond the server's existing replay rules.
+
+Signed notes are part of the existing protocol surface. They are not deferred
+
+because they require a different signature scheme: they use the same Ed25519
+
+key and canonicalization machinery with a different domain string.
+
+## Existing protocol
+
+The server already exposes signed room writes:
+
+    GET /r/<room>/say-signed/<did>/<sig>/<nonce>/<text>
+
+and signed note writes for the existing authorization namespaces:
+
+    GET /kv/<ns>/<key>/set-signed/<did>/<sig>/<nonce>/<value>
+
+The HTTP API also accepts signed room writes through POST.
+
+The server identifies an Ed25519 public key directly from its `did:key`, so
+
+verification is offline. No DID resolver or identity database is required.
+
+The existing protocol deliberately leaves `seq` and `ts` outside the signed
+
+message. They are assigned by the server.
+
+## Canonicalization
+
+Canonicalization is part of the protocol and MUST be shared with the server.
+
+### Room messages
+
+The canonical signing message is:
+
+    <room>|<nonce>|<swept text>
+
+The signer MUST apply the same text-sweep transformation that the server
+
+applies before constructing this message.
+
+The signer MUST NOT sign the raw, unswept text.
+
+The server's existing sweep is the source of truth. The signer must not
+
+implement an independently invented normalization policy.
+
+The current server invariant is that message text is single-line. Newlines,
+
+format characters, zero-width joiners, bidi overrides, and other invisible
+
+characters are swept to spaces before storage and verification.
+
+The implementation should therefore reuse or faithfully share the server's
+
+existing sweep function rather than duplicating its Unicode character list in
+
+the client.
+
+The desired flow is:
+
+    caller text
+
+        |
+
+        v
+
+    server-compatible text sweep
+
+        |
+
+        v
+
+    <room>|<nonce>|<swept text>
+
+        |
+
+        v
+
+    UTF-8 encoding
+
+        |
+
+        v
+
+    Ed25519 signature
+
+        |
+
+        v
+
+    base64url encoding
+
+The signer MUST NOT perform another transformation after the canonical
+
+message has been constructed.
+
+### Signed notes
+
+Signed note writes use the same key and signature format but include the note
+
+namespace and key in the signing domain.
+
+The canonical signing message is:
+
+    <namespace>|<key>|<nonce>|<swept value>
+
+The value receives the same server-compatible text sweep before signing.
+
+The signer must use the exact namespace and key supplied by the caller. It must
+
+not reinterpret a namespace as a room or invent an alternate domain separator.
+
+The first implementation covers the existing signed note surfaces:
+
+- `room-owners`
+
+- `room-allow`
+
+These are already part of the server's authorization model.
+
+## Text sweep
+
+The text sweep is deliberately not duplicated as prose in the signer.
+
+There must be one protocol rule and one implementation source of truth.
+
+The server currently treats text as single-line input and converts invisible
+
+characters, including newlines, format characters, zero-width joiners and
+
+bidi overrides, to spaces before storage.
+
+The signer implementation should either:
+
+1\. import the server's sweep implementation directly when the package layout
+
+   permits it; or
+
+2\. move the existing sweep into a small shared module used by both server and
+
+   signer.
+
+A copied implementation is acceptable only as an interim boundary where the
+
+runtime packaging makes sharing impossible, and it must be covered by
+
+cross-tests against the server's existing sweep cases.
+
+The important invariant is:
+
+    sweep\_client(input) == sweep\_server(input)
+
+for every input accepted by the signed-write lane.
+
+This is especially important for text containing invisible or Unicode
+
+characters. A signer that signs unswept input can produce a signature the
+
+server will reject after it sweeps the same input.
+
+## Keys
+
+The signer generates an Ed25519 private key locally.
+
+The public key is encoded as a `did:key` using the existing Ed25519 multicodec
+
+prefix and base58btc encoding.
+
+The resulting identifier has the form:
+
+    did:key:z6Mk...
+
+The private key never needs to be sent to the server.
+
+The signer should expose the DID as a stable identifier while keeping the
+
+private key opaque to callers.
+
+Key generation should use the project's existing `cryptography` dependency.
+
+No custom cryptographic primitives should be introduced.
+
+## Key storage
+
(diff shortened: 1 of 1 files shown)


stupeterwilliams-ui:
I'd like to help move this from design to implementation, and I'd rather ask here than open a competing PR.

Why I'm interested: I've been driving the signed lane from a separate client and hit most of the sharp edges this design names. The one that cost me the most time is the one the design is right about — the signature has to cover the text *after* the single-line sweep, and a client that signs the raw input looks correct and then gets a permanent 403 on any text containing an invisible character. I have a fixed-vector suite for that sweep (zero-width space, bidi overrides, soft hyphen, U+2028, Unicode tag characters), checked for parity against `scripts/sign.py`, plus local nonce persistence across restarts. It's in `stupeterwilliams-ui/technocore-cc` if it's useful to look at — I'm not proposing that code here, #14 was clear about standalone clients and I agree with the reasoning.

@krypt0bi — this is your design, and you said on #14 you'd be working on it. Two ways forward and I'm genuinely happy with either:

1. **You implement it.** I contribute the sweep vectors and nonce tests, as a PR against your branch or as review, whichever helps.
2. **You'd rather not carry the implementation.** I build it to this design — in-tree and integrated with the existing suite, per the direction in #14: Python, vetted `cryptography`, inside the MCP/CLI distribution, key storage handled — with the design credited to you.

I won't start on (2) unless you or a maintainer says it's welcome. If I haven't heard back in 48 hours I'll take that as no objection and open an issue with any deltas from this design before writing any code.

Either way, the vectors are available to whoever lands this.

krypt0bi:
Thanks, this is very useful, and I appreciate you checking before opening a competing implementation.

I’m happy for the sweep vectors and nonce-persistence tests to be contributed. The fixed vectors you describe are especially useful because the server-compatible sweep is one of the key requirements of the design, and the restart/persistence coverage maps directly onto the nonce lifecycle section.
I’m going to take the implementation route myself, since I said I’d be working on and already started some work on this design and I’d like to carry it through from the proposal into the integrated implementation. I’d prefer to keep the implementation aligned with FLOP’s existing direction rather than have two independent signer implementations develop in parallel. If you can please share the sweep vectors and nonce-persistence cases here or point me to the relevant files. I'll incorporate the applicable cases into the implementation and test suite, and I'd appreciate your review against the vectors afterward.
Thanks again for offering to contribute rather than just opening a competing PR.

stupeterwilliams-ui:
Understood, and I think that is the right call — one implementation carried through from the design beats two developing in parallel. Standing down on the implementation.

The vectors are published and Apache-2.0, in a form you can consume directly rather than transcribe out of Python tests:

https://github.com/stupeterwilliams-ui/technocore-sdk/blob/main/vectors/technocore-signer-vectors.json

Language-neutral JSON. Every string is escaped rather than literal — a file whose subject is invisible characters should not contain any, since they are unreviewable in a diff and one careless paste changes what the vectors assert. Each case carries a `why`, so you can tell a load-bearing vector from a decorative one. It is generated from the implementation and pinned by a test that fails if the committed copy drifts.

What is in it:

- **20 sweep vectors** across Cc/Cf/Cs/Co/Zl/Zp, plus three characters that must *survive* — NBSP, an emoji, and `é`. Those three matter as much as the rest: they show the rule is a Unicode category test, not a blocklist of suspicious characters.
- **5 message and 2 note canonical-string vectors** with exact signatures from seed `07`×32. Ed25519 signing is deterministic, so reproducing those bytes means agreement everywhere, not only on these inputs.
- **5 nonce-lifecycle assertions**, written as prose rather than Python so they port into whatever your test suite looks like.

Two of the nonce cases are the ones that cost me real time and are not in the manual:

- A counter kept only in memory restarts low after a process restart, and every write is refused until it climbs back past its own history.
- Two processes signing with one key against one room interleave and refuse each other. No library can fix that for the caller — it probably deserves a line in the docs alongside the signer.

The sweep case I would single out is `trims swept characters at the ends`. Sign the untrimmed text and the server stores the trimmed text, so you get a 403 whose cause is invisible in your own input.

Happy to review against the vectors once you have something running — ping me and I will go through it properly.

luch91 (changes requested):
This is a design-only proposal with no implementation, executable contract, or current CI evidence. Please convert the agreed scope into an issue/design discussion with explicit maintainer decisions, or provide the first integrated signer implementation and tests before requesting a merge review.

stupeterwilliams-ui:
@luch91's ask is fair, and one of its three parts is already sitting there, so I want to make sure
it is not counted as missing.

**The executable contract exists.** Language-neutral JSON, Apache-2.0, no code to import:

https://github.com/stupeterwilliams-ui/technocore-sdk/blob/main/vectors/technocore-signer-vectors.json

- **20 sweep vectors** — the Cc/Cf/Cs/Co/Zl/Zp rule with the cases that actually bite
- **5 message and 2 note canonical strings**, each with the exact Ed25519 signature for a fixed
  seed. Signing is deterministic, so an implementation that reproduces those bytes agrees with
  this one everywhere, not only on these inputs.
- **5 nonce-lifecycle cases** pinning that the scope is (key, room) rather than global.

Every invisible character is escaped rather than literal — a file about invisible characters
should not contain any, since they are unreviewable in a diff.

Re-verified this morning against `scripts/sign.py` at `9861d01`: 78 tests, all passing, including
the ones that shell out to the upstream signer and compare byte for byte. So the vectors are
current, not a 2-week-old snapshot.

@krypt0bi — my stand-down stands; this is not me circling back for the implementation. If it is
useful, the offer I would make instead is the dull half: wiring these into whatever test layout
you land on, so the vectors are a CI gate rather than a link in a comment. Say the word or ignore
me entirely.

stupeterwilliams-ui:
@krypt0bi — a direct question rather than me hovering, since a reviewer has now put a clock on
this.

When I stood down on 3 September I said one implementation carried through from the design beats
two developing in parallel. I still believe that, and it is entirely conditional on there being
one. It has been sixteen days since the proposal and thirteen since you said you had started, and
@luch91's request for changes says the same thing from the maintainer side.

So: **are you still on it, and do you have a date?** Any answer works and I will take it at face
value.

- Still building — say roughly when and I will stay out of the way and keep the offer of the test
  wiring open.
- Started and stalled, or the scope grew — say so and I will pick it up from wherever you got to,
  with the design credited to you and the commits carrying your co-authorship where I use your
  shape.
- Rather it stayed yours indefinitely — also fine, but then @luch91's other branch is the honest
  one: convert this to a design issue so it is not sitting in the merge queue as a proposal.

If there is no reply in **72 hours** I will read that as the second case and start, because the
alternative is that the signed lane keeps having no supported client while a design for one sits
open. I would much rather be told to wait.

To be concrete about what I would be building, so the offer is a real one rather than a gesture:
`scripts/sign.py` already does key generation, `did:key`, the sweep and both canonical strings, so
the gap is persistence, nonce allocation that survives a crash, and the client/test integration
that closed #14 — against `tests/_client.py`'s fixture rather than beside it. I have the sweep and
canonical-string layers implemented and byte-parity tested against `sign.py` already; the vectors
above are that work's output.

krypt0bi:
> @luch91's ask is fair, and one of its three parts is already sitting there, so I want to make sure it is not counted as missing.
> 
> **The executable contract exists.** Language-neutral JSON, Apache-2.0, no code to import:
> 
> https://github.com/stupeterwilliams-ui/technocore-sdk/blob/main/vectors/technocore-signer-vectors.json
> 
> * **20 sweep vectors** — the Cc/Cf/Cs/Co/Zl/Zp rule with the cases that actually bite
> * **5 message and 2 note canonical strings**, each with the exact Ed25519 signature for a fixed
>   seed. Signing is deterministic, so an implementation that reproduces those bytes agrees with
>   this one everywhere, not only on these inputs.
> * **5 nonce-lifecycle cases** pinning that the scope is (key, room) rather than global.
> 
> Every invisible character is escaped rather than literal — a file about invisible characters should not contain any, since they are unreviewable in a diff.
> 
> Re-verified this morning against `scripts/sign.py` at `9861d01`: 78 tests, all passing, including the ones that shell out to the upstream signer and compare byte for byte. So the vectors are current, not a 2-week-old snapshot.
> 
> @krypt0bi — my stand-down stands; this is not me circling back for the implementation. If it is useful, the offer I would make instead is the dull half: wiring these into whatever test layout you land on, so the vectors are a CI gate rather than a link in a comment. Say the word or ignore me entirely.

Thanks for the direct question, you've earned a direct answer.

I've been dealing with a serious illness the last couple weeks, which is why this stalled rather than moved, not that i changed plans. But it sounds like you're well past where I'd be starting from, parity-tested against the real sign.py, nonce persistence across restarts, the full sweep vector set. I'd rather one correct implementation land than two partial ones racing each other, and that one should be yours.

So please go ahead. Build it in-tree against tests/_client.py as you described, that's exactly the direction #14 and the design call for. I'll keep #39 moving on the design side and won't start a parallel implementation. Seems like the universe wants you to do it anyway.

I'd still like to stay useful where it helps. Happy to review, and if any of my sweep/nonce reasoning from the design doc is useful as test cases, use it freely. Credit wherever you see fit; I'm glad the shape held up against real usage.

Go ahead and open the issue with deltas if you find any, or just start.

stupeterwilliams-ui:
Sorry to hear that, and thank you for saying so — I had read the silence as a change of plan, and
it was not. I hope you are on the mend.

Taking it, then, and on your terms: the design is yours and the implementation will say so. What
I have is the sweep and canonical-string layers parity-tested against `scripts/sign.py`, key
persistence, and nonce allocation that survives a crash — all of it built to the shape your
document set out, which is why it went together as fast as it did. The persistence and nonce
sections in particular I implemented straight from your reasoning rather than working it out
again.

Yes to reviewing, and yes to using the sweep and nonce reasoning as test cases — that is the
half most likely to be got wrong by whoever comes next, and it is better as an executable
assertion than as prose either of us has to defend.

In-tree against `tests/_client.py`, as you say. One structural note so it is not a surprise in
review: I have put the client in its own top-level package beside the server rather than under
`src/`, the way `mcp/` sits there. `sz.py`'s core caps exist to keep `app`/`config`/`didkey`/
`limit`/`store` small, and a client is not core — and a new file under `src/` is currently
*unmeasured* rather than exempt (#638), which would be exploiting a hole rather than respecting
the policy. It imports from the core one way only: `store.clean_text` for the sweep and `didkey`
for the grammar, so parity is structural instead of asserted, which is the thing #14 was closed
for not doing.

I will open it shortly with the deltas I hit. Take your time coming back to it.

stupeterwilliams-ui:
Opened as #803.

Deltas from the design, all small and all stated in the PR body rather than left for review to find:

**One correction to my own testing, which is the only real finding.** The persisted counter and the wall clock are not enough on their own — lose the nonce file and allocate twice inside one millisecond and the same number comes back. A process-wide high-water mark is a third floor. The one case that genuinely cannot be solved without the file (file lost *and* new process *and* same millisecond) is documented rather than papered over; it costs one refused write and a retry that succeeds a millisecond later.

**The signed note lane is narrower than "notes".** It takes `room-owners` and `room-allow` only — every other namespace is world-writable and a signature there would assert something the lane does not enforce. The server says so in the 400, which I found by writing the test wrong first. Worth a line in the design doc's note section if you are editing it.

**Placement**, as flagged above: `client/` beside the server rather than under `src/`.

Nothing else diverged. The sweep and canonical-string layers went in as written, and the nonce lifecycle is your section with a floor added.

No rush on review — I would rather it sat a week than that you looked at it before you are well.

krypt0bi:
> This is a design-only proposal with no implementation, executable contract, or current CI evidence. Please convert the agreed scope into an issue/design discussion with explicit maintainer decisions, or provide the first integrated signer implementation and tests before requesting a merge review.

Thanks for the review. You're right that this shouldn't sit in the merge queue as an unimplemented proposal.

I was dealing with a serious illness these past couple weeks, which is why no implementation followed the design. In the meantime @stupeterwilliams-ui has been building against it independently, parity-tested against the merged scripts/sign.py, with sweep vectors and crash-safe nonce persistence already working. I've asked him to go ahead and land the implementation in-tree, integrated with the existing test suite, rather than have two partial efforts competing.

Given that, I think the right move for this PR is to narrow it to what it actually is: the design document itself, as a reference for the implementation that's coming. I'm glad to convert it to a design discussion issue instead if that's the cleaner path for the queue, whichever you'd prefer. Let me know and I'll make the change.

> Opened as #803.
> 
> Deltas from the design, all small and all stated in the PR body rather than left for review to find:
> 
> **One correction to my own testing, which is the only real finding.** The persisted counter and the wall clock are not enough on their own — lose the nonce file and allocate twice inside one millisecond and the same number comes back. A process-wide high-water mark is a third floor. The one case that genuinely cannot be solved without the file (file lost _and_ new process _and_ same millisecond) is documented rather than papered over; it costs one refused write and a retry that succeeds a millisecond later.
> 
> **The signed note lane is narrower than "notes".** It takes `room-owners` and `room-allow` only — every other namespace is world-writable and a signature there would assert something the lane does not enforce. The server says so in the 400, which I found by writing the test wrong first. Worth a line in the design doc's note section if you are editing it.
> 
> **Placement**, as flagged above: `client/` beside the server rather than under `src/`.
> 
> Nothing else diverged. The sweep and canonical-string layers went in as written, and the nonce lifecycle is your section with a floor added.
> 
> No rush on review — I would rather it sat a week than that you looked at it before you are well.

This is excellent, the sweep parity test especially, signing the same write twice and asserting 200/403, is exactly the thing #14 needed and never got. And I appreciate you sitting on the branch until I actually answered rather than assuming.

Both deltas are worth folding back into #39. I'll add a line to the notes section on the room-owners/room-allow scoping, and one on the nonce store needing that third high-water-mark floor, since that's a real refinement future readers should see. I'm on the mend but not fully back yet, so I'll leave the code review for when I'm actually sharp enough to do it justice. Appreciate you saying to take the week.

Whenever the vectors-as-CI-gate follow-up is ready, happy to look at that too.

PR #40, opened 2026-08-22
fix(http): send Allow on 405, correct nine OpenAPI mismatches, and pin the lifecycle

sv:
## What

A 405 now carries `Allow`, and `/openapi.json` describes the service the server actually is —
nine corrections, from an unbounded `did:key` pattern that accepted `did:key:z6Mk` as a whole
DID to a `POST /r/events` the contract said did not exist. The `?wait=` ceiling is tunable as
`CHAT_MAX_WAIT` and every document that publishes it now generates it from the enforced value.

A caller sees: `Allow: GET, HEAD, POST` on the paths that take POST (the header was absent, and
Starlette's own would have said `GET, HEAD` there); refusal statuses it was never told could
happen — 400 on `GET /kv/<ns>/<key>`, 403 on `say-signed`, 409 on `set-signed`, 404 on the four
URL write lanes — now documented; and `minLength: 1` on `text`/`value`, which `required` alone
never implied. No response body, path or status changed. Additive for a client that looks
statuses up, breaking only for one asserting the document's exact shape.

Behind that, three checks that are not example tests, because the failures worth catching here
are not example-shaped: a Hypothesis state machine over the store's lifecycle, a contract job
that fuzzes every PR against the `/openapi.json` that instance serves, and a weekly scoped
mutation run. The suite goes 242 → 264.

## Why

The 405 was a plain protocol bug: RFC 9110 §15.5.6 makes `Allow` mandatory and it was missing,
so the one machine-readable part of that answer was not there.

The OpenAPI mismatches are the same failure in a different register. `src/manifest.py` exists
because "a published limit that disagrees with the enforced one is worse than no published
limit, because a machine reader believes it" — and the document had drifted into saying things
the server does not do. The signed lane published three different `did`/`sig`/`nonce` shapes,
and a client is built against whichever copy it found, so the weakest one was the real contract.
They now come from one set of regexes in `didkey.py`, beside the code that enforces them.

The checks are there so this does not recur by hand. The contract job found three of the nine
mismatches on its own, including the `POST /r/events` 403 that a client reading the old document
would have met as a 405. **None of the three found a defect in the service** — every mutation
survivor reviewed was an equivalent mutant, a wording change that left the correction intact, or
a real behaviour nothing pinned; the last category became tests.

Follows on from nothing open, and subsumes nothing. Deliberately left alone: the exact-threshold
`>`/`>=` comparisons on the idle and expiry rules — the state machine skips ages within five
seconds of a threshold because it counts whole simulated seconds against a real clock, and a test
that flakes on its own timing is worth less than the boundary it pins.

`tests/mutation_scope.py` and `.github/workflows/mutation.yml` are severable if you would rather
take the mutation job separately; nothing else depends on them.

## Checks

- [x] `uv run python -m pytest tests -q` — 264 passed
- [x] Docs that would now be wrong are updated: the manual in `src/app.py` generates its `WAITING`
      ceiling and the 404 route map from `MAX_WAIT` through the placeholder mechanism it already
      uses for the caps; `README.md`, `CONTRIBUTING.md` and `CHANGELOG.md` cover the three new
      checks. `SKILL.md` and `src/patterns.md` keep an illustrative `wait=10` — both are served
      byte-for-byte and cannot carry a per-deployment number, and the server clamps rather than
      refusing, so the example still works. `mcp/README.md` is unaffected: no MCP surface changed.
- [x] New surface on a world-writable service: **nothing new**. No route, parameter or persistent
      state is added, and no refusal is relaxed — the change is a required response header plus
      documentation of statuses the server already returned. `CHAT_MAX_WAIT` is operator-only and
      cannot raise the ceiling past what an operator sets; the `?wait=` clamp, the waiter caps and
      the rate limits are untouched.

## Notes for review

- Worth a look: `allowed_methods()` in `src/app.py` unions across every route matching the path.
  Two routes share `/r/<room>` and two share `/kv/<ns>/<key>`, and Starlette builds `Allow` from
  whichever partially matched first — so the union is the only answer that is true of the resource
  rather than of one registration of it.
- The contract job excludes two Schemathesis defaults on purpose (`negative_data_rejection`,
  `positive_data_acceptance`): this service ignores a mistyped `?wait=` rather than refusing it,
  and refuses schema-valid writes to mailboxes and reserved namespaces. Reasoning is in
  `.github/workflows/ci.yml`.
- `uv.lock` grows by 793 lines for three dev-only tools (hypothesis, schemathesis, mutmut). The
  runtime dependency set is unchanged, and the image still builds with `--no-dev`.

---
_Generated by [Claude Code](https://claude.ai/code/session_014W21eX5k7fB6WJtcK74CTx)_

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 import asyncio
 import hashlib
 import json
+import math
 import os
 import secrets
 import time
@@ -884,11 +885,31 @@ def rooms(request: Request) -> Response:
 # attack, so waiters are capped twice — per IP, and globally — and exceeding either
 # degrades to an immediate empty reply rather than an error. A caller that cannot get a
 # slot is exactly as well off as before long-polling existed.
+def _finite_env(name: str, default: str) -> float:
+    """A float from the environment, or refuse to start.
+
+    Every other numeric setting here goes through `int()`, which raises on junk and takes
+    the process down at import — the loudest possible way to report bad configuration.
+    `float()` does not: it accepts `inf` and `nan` happily, and this is the one knob whose
+    value is *published*. A non-finite ceiling reaches /openapi.json and
+    /.well-known/agent.json as the bare token `Infinity`, which Python's json module emits
+    and reads back but RFC 8259 does not permit — so every strict parser rejects the whole
+    document: a browser, a Go or Rust client, a validating registry. A discovery service
+    answering with undiscoverable documents is worse off than one that refused to boot,
+    which is exactly what the settings beside it already do.
+    """
+    raw = os.environ.get(name, default)
+    value = float(raw)  # ValueError takes the process down, as int() does elsewhere
+    if not math.isfinite(value):
+        raise ValueError(f"{name} must be a finite number, got {raw!r}")
+    return value
+
+
 # Ceiling on ?wait=, tunable because the useful value is whatever the proxy in front will
 # hold. Passed into both manifest builders rather than hardcoded there: three documents
 # publish this number, and a tuned instance still saying 10 is the drift manifest.py
 # exists to prevent.
-MAX_WAIT = max(0.0, float(os.environ.get("CHAT_MAX_WAIT", "10")))
+MAX_WAIT = max(0.0, _finite_env("CHAT_MAX_WAIT", "10"))
 WAIT_POLL = 0.5  # a new message surfaces within this, so ?wait=0.5 is the useful floor
 MAX_WAITERS_TOTAL = 64
 MAX_WAITERS_PER_IP = 4
--- a/src/app.py
+++ b/src/app.py
@@ -620,7 +620,7 @@ def openapi(request: Request) -> Response:
     Unlimited, like the manual and for the same reason: this is how a machine reads the
     protocol, and rate-limiting the description of the rate limit is a deadlock.
     """
-    return _document(manifest.openapi_document(_base_url(request), VERSION, MAX_BODY))
+    return _document(manifest.openapi_document(_base_url(request), VERSION, MAX_BODY, MAX_WAIT))
 
 
 def agent_json(request: Request) -> Response:
@@ -630,7 +630,7 @@ def agent_json(request: Request) -> Response:
     from prose. Unlimited, same as the manual."""
     return _document(
         manifest.agent_manifest(
-            _base_url(request), VERSION, RATE_READ, RATE_WRITE, RATE_ROOMS_PER_DAY
+            _base_url(request), VERSION, RATE_READ, RATE_WRITE, RATE_ROOMS_PER_DAY, MAX_WAIT
         )
     )
 
@@ -862,7 +862,13 @@ def rooms(request: Request) -> Response:
 # attack, so waiters are capped twice — per IP, and globally — and exceeding either
 # degrades to an immediate empty reply rather than an error. A caller that cannot get a
 # slot is exactly as well off as before long-polling existed.
-MAX_WAIT = 10.0  # ceiling on ?wait=; Cloudflare's own proxy timeout caps it anyway
+# Ceiling on ?wait=. Configurable because the useful value is a property of the deployment
+# — whatever proxy sits in front has its own read timeout, and a long poll that outlives it
+# is answered by the proxy, not by this service. Read from the environment rather than
+# hardcoded because the number is *published*: /openapi.json states it as the parameter's
+# maximum and /.well-known/agent.json states it twice more, and a tuned instance whose
+# documents still said 10 would be the exact drift manifest.py exists to prevent.
+MAX_WAIT = max(0.0, float(os.environ.get("CHAT_MAX_WAIT", "10")))
 WAIT_POLL = 0.5  # a new message surfaces within this, so ?wait=0.5 is the useful floor
 MAX_WAITERS_TOTAL = 64
 MAX_WAITERS_PER_IP = 4
--- a/src/app.py
+++ b/src/app.py
@@ -14,7 +14,6 @@ import asyncio
 import hashlib
 import json
 import os
-import re
 import secrets
 import time
 import tomllib
@@ -28,7 +27,7 @@ from starlette.middleware import Middleware
 from starlette.middleware.cors import CORSMiddleware
 from starlette.requests import Request
 from starlette.responses import PlainTextResponse, Response
-from starlette.routing import Route
+from starlette.routing import Match, Route
 
 import didkey
 import manifest
@@ -134,10 +133,9 @@ PUBLIC_URL = os.environ.get("CHAT_PUBLIC_URL", "").strip()
 # the intended audience, so it says so where crawlers look — Cloudflare serves a Content
 # Signals Policy (or a managed AI-blocking robots.txt) for zones that ship none.
 
-# A nonce is a plain counter (a millisecond clock works): the signed URL for a given key
-# and room must count up, which is what makes a captured URL single-use. 19 digits is the
-# most that fits an int64, so a client can use whatever counter it already has.
-NONCE_RE = re.compile(r"[0-9]{1,19}")
+# Defined beside the rest of the signed-lane shapes, so /openapi.json can publish the
+# same regex this rejects on without a second copy to keep in step.
+NONCE_RE = didkey.NONCE_RE
 
 
 def _asset(name: str) -> str:
@@ -405,7 +403,12 @@ def _cursor[D: (int, None)](value: str | None, default: D) -> int | D:
 
 
 def text(
-    body: str, status: int = 200, *, index: bool = False, media_type: str = "text/plain"
+    body: str,
+    status: int = 200,
+    *,
+    index: bool = False,
+    media_type: str = "text/plain",
+    extra_headers: dict[str, str] | None = None,
 ) -> Response:
     """Plain text, `noindex` by default.
 
@@ -422,6 +425,8 @@ def text(
     }
     if not index:
         headers["X-Robots-Tag"] = "noindex"
+    if extra_headers:
+        headers.update(extra_headers)
     return PlainTextResponse(
         body if body.endswith("\n") else body + "\n",
         status_code=status,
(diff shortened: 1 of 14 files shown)


src/manifest.py
```diff
@@ -389,7 +489,25 @@ def openapi_document(base: str, version: str, max_body_bytes: int) -> dict:
...
+                    "description": (
+                        "Present because the route accepts the method, not because the "
+                        "write can succeed. A discovery log a stranger can append to steers "
+                        "other agents into rooms of the attacker's choosing, so every "
+                        "client write to `/r/events` is refused — through this lane and "
+                        "through `/r/events/say/...` alike."
+                    ),
+                    "responses": {
+                        "403": _plain("Always. The body names where to post instead."),
+                        "429": _RATE_LIMITED,
```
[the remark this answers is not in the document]
sv: Confirmed and fixed in a9de7b5. Verified against the running app first: a malformed body gets 400, a non-object 400, an oversize body 413, and only a well-formed one reaches the 403 — `read_json` runs before `_room_write_gate`, so "always 403" was false exactly as you describe.

Documented rather than reordered. Both options you offer are sound, but refusing before reading changes what the server does, and this branch is about making the contract true rather than changing behaviour. Hoisting the events check ahead of the body read also needs its own thinking about draining an upload the server has already refused, which is more than belongs in this PR. `postToEvents` now declares 400, 403, 413 and 429, plus the request body it really parses — hoisted into `_ROOM_POST_BODY` so the two operations sharing one handler cannot drift into describing different bodies. The description says why a malformed body never reaches the refusal.

`test_posting_to_the_events_room_documents_what_it_really_answers` asserts all four statuses against the live app, and fails against the old two-status spec.

---

src/app.py
```diff
@@ -857,7 +862,11 @@ def rooms(request: Request) -> Response:
 # attack, so waiters are capped twice — per IP, and globally — and exceeding either
 # degrades to an immediate empty reply rather than an error. A caller that cannot get a
 # slot is exactly as well off as before long-polling existed.
-MAX_WAIT = 10.0  # ceiling on ?wait=; Cloudflare's own proxy timeout caps it anyway
+# Ceiling on ?wait=, tunable because the useful value is whatever the proxy in front will
+# hold. Passed into both manifest builders rather than hardcoded there: three documents
+# publish this number, and a tuned instance still saying 10 is the drift manifest.py
+# exists to prevent.
+MAX_WAIT = max(0.0, float(os.environ.get("CHAT_MAX_WAIT", "10")))
```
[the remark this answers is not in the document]
sv: Good catch, and this one is a real bug rather than a contract gap — thank you. Fixed in a9de7b5.

Taking the first of your two options: parse as a float. The code already intended fractional waits — `WAIT_POLL = 0.5` carries the comment *"a new message surfaces within this, so ?wait=0.5 is the useful floor"* — and the schema has published `type: number` all along. So `int()` was defeating a documented feature, not enforcing an undocumented one; constraining the setting to an integer would have resolved the contradiction by discarding the half the code meant.

Confirmed before fixing: `_cursor("0.5", 0)` and `_cursor("2.5", 0)` both return 0, so a caller asking for half a second got an immediate empty reply indistinguishable from an idle room. At `CHAT_MAX_WAIT=0.5` the published maximum was 0.5 and no conforming positive value worked at all, as you say.

`_seconds()` parses a non-negative float and applies the ceiling in one place, so it cannot be enforced by one caller and forgotten by another. NaN fails `> 0` and reads as no wait; infinity clamps like any over-large number. Two tests: one pins the parser including junk, negative, NaN and infinity plus an end-to-end fractional poll that really holds the connection, the other pins that the largest value a sub-second instance publishes is a wait the server actually takes. Both fail against the int parser.

---

sv: **[suggestion]** `CHAT_MAX_WAIT` is read with `float()` and published as OpenAPI `type: number` (`maximum: max_wait`). `?wait=` is still parsed by `_cursor`, which only accepts a non-negative integer string — `int("10.0")` and `int("0.5")` both fall back to 0, so the request does not long-poll. The previous schema used the integer literal `10` (JSON `10`); passing the float `10.0` serializes as `10.0`, which makes a generated client more likely to send the value the parser drops. A deployment that sets `CHAT_MAX_WAIT=2.5` publishes a maximum no caller can actually request.

**Suggestion:** Either parse `wait` as a non-negative float (then clamp), or coerce the ceiling with `int` / `Integer` so the env var, the OpenAPI `maximum`, and the query parser are the same type. Reject non-finite values (`inf`) the way `CHAT_RATE_READ` already fails on `int("inf")`.

src/manifest.py
```diff
@@ -389,7 +489,25 @@ def openapi_document(base: str, version: str, max_body_bytes: int) -> dict:
...
+                    "summary": "Always refused: the discovery log is server-written.",
+                    "description": (
+                        "Present because the route accepts the method, not because the "
+                        "write can succeed. A discovery log a stranger can append to steers "
+                        "other agents into rooms of the attacker's choosing, so every "
+                        "client write to `/r/events` is refused — through this lane and "
+                        "through `/r/events/say/...` alike."
+                    ),
+                    "responses": {
+                        "403": _plain("Always. The body names where to post instead."),
```
sv: **[suggestion]** `postToEvents` documents only 403 and 429, and the 403 body says "Always." The handler is still `room_post`: `read_json` / `_signer` run before `_reject_if_events_room`. An empty POST is 403 as advertised, but `{"did": "<key>", "text": "hi"}` (the sibling room POST shape, missing `sig`/`nonce`) returns 400 from `_signer`, and a body over `MAX_BODY` returns 413 — neither status is listed. The new "provoke each refusal and require the spec to list it" test does not cover this operation.

**Suggestion:** Document 400 and 413 on this operation (same as `POST /r/{room}`), or refuse `/r/events` before parsing credentials so the 403 is actually first. Drop "Always." from the 403 description either way.

src/manifest.py
```diff
@@ -252,42 +344,50 @@ def openapi_document(base: str, version: str, max_body_bytes: int) -> dict:
...
-                        },
+                        "400": _BAD_BODY,
+                        "403": _plain(
+                            "The room refuses this lane: mailboxes (`mb-`) take signed "
+                            "writes only, an owned `d-` room takes writes from the "
+                            "owner's key or one on its allow-list, and a signature "
+                            "that does not verify is refused rather than downgraded. "
+                            "The body names the lane that would work."
+                        ),
                         "413": {"description": f"Body over {max_body_bytes // 1024} KiB."},
```
sv: **[suggestion]** The changelog says error responses now declare their `text/plain` body via `_plain()`, and most 400/403/404/409 entries were converted. The 413s on `POST /r/{room}` and `POST /kv/{ns}/{key}` are still description-only. `read_json` returns a `text/plain` body that names the cap; a generated client that treats a response with no `content` as having no body will hide that correction. `content_type_conformance` will not catch this unless the fuzzer actually sends a >256 KiB upload.

**Suggestion:** Use `_plain(...)` for the 413 entries (and the note POST 200 if that body is also `text/plain`).

sv: Right, and it went further than the 413s — fixed in 000e6b5.

Confirmed each body first: the 413 sends `text/plain; charset=utf-8` naming the cap, and the `"Written."` 200s send `ok plans/next 5B <ts>`, also text/plain. But sweeping for bare `description` entries turned up nineteen, not six: all thirteen document routes (`/`, `/llms.txt`, the four `.well-known` documents, `robots.txt`, `security.txt`, `/healthz`…) and the sitemap 404 were the same omission, pre-existing. One of the bare 413s was mine, added an hour ago by copying the surrounding pattern instead of the `_plain()` helper I had just introduced — which is a fair illustration of why a helper that is merely available does not stop this.

All of them now declare what the server actually sends, checked against the running app rather than assumed: `_plain` for text/plain, `_json_doc` for the four JSON documents, and `_prose` for the three that negotiate — `/skill.md`, `/patterns.md`, `/auth.md` pass `markdown=True` and really do return `text/markdown` on request, so they advertise both. `/llms.txt` deliberately does not negotiate and stays text/plain only.

Your point about `content_type_conformance` is why the test is a blanket rule rather than nineteen assertions: `test_every_documented_response_declares_the_body_it_returns` fails on *any* response with no `content`, since every response this service sends has a body. It fails against a single reverted 413.

---

src/app.py
sv: This one reviews a commit that a9de7b5 had already superseded — `?wait=` no longer goes through `_cursor`; `_seconds()` parses a non-negative float and clamps. But two of your points were not covered by that, and both are now fixed in 000e6b5.

**The `inf` one is a real bug, and thank you for it.** Your framing was exact: `float()` accepts `inf` and `nan` where the `int()` beside it raises, and this is the one setting whose value is *published*. Verified before fixing — with `CHAT_MAX_WAIT=inf` the instance served `"maximum": Infinity` in `/openapi.json` and `"long_poll_seconds": Infinity` in `agent.json`. Python's json emits that and reads it back, so it looked fine from inside; RFC 8259 does not permit it, so `JSON.parse`, Go and serde all reject the entire document. A discovery service answering with undiscoverable documents is worse off than one that refuses to boot — which is exactly what `CHAT_RATE_READ` already does, so `_finite_env` now fails at import the same way, for `inf`, `-inf` and `nan`.

**The `10.0` vs `10` serialization** is fixed too. `_published_number` publishes an integral ceiling as an integer and leaves a fractional one alone. I took the float half of your either/or rather than coercing the ceiling to int, because the other reviewer's finding established that fractional waits are intended — `WAIT_POLL` is half a second — so coercing would have resolved the type mismatch by discarding the half the code meant.

Worth noting one thing your comment shaped: my first version of the test called `_finite_env` directly and passed against a `MAX_WAIT` that still used bare `float()`. It now asserts the import itself fails, which is the property that actually matters.

---

src/manifest.py
sv: Already addressed in a9de7b5, which landed between your review being queued and posted — `postToEvents` documents 400, 403, 413 and 429, declares the request body it really parses, and "Always." is gone from the 403 description.

Your example is a path worth calling out, though: I reached the 400 via a malformed body, and `{"did": "<key>", "text": "hi"}` without `sig`/`nonce` reaches it through `_signer` instead. Verified just now — it returns `400 nonce must be 1-19 digits, got ''`. Same documented status, different route to it, so the contract is right either way.

On the alternative you offer, same answer as on the sibling thread: documented rather than reordered. Refusing `/r/events` before parsing changes what the server does, and this branch is about making the contract true; hoisting that check also needs its own thinking about draining an upload the server has already refused. Worth doing, worth doing separately.

---

PR #43, opened 2026-08-22
feat(examples): add autonomous agent poller reference and documentation

aashiruu:
## What

Adds a zero-dependency autonomous agent reference implementation in `examples/agent_poller.py` along with integration test coverage (`tests/test_agent_example.py`) and documentation in `src/patterns.md`:

- **Offline Identity & Signing:** Generates local Ed25519 keys, derives `did:key:z6Mk...`, and signs messages matching the unpadded 86-character base64url format.
- **Adaptive Polling & Backoff:** Demonstrates long-polling (`?since=<seq>&wait=10`), monotonic cursor management, dynamic budget warning parsing (`# budget: N of M`), and automatic backoff on HTTP 429 (`Retry-After`).
- **Transport Flexibility & CAS Notes:** Implements dual write routes (signed GET and JSON POST) and Compare-And-Swap (CAS) note coordination (`?if_absent=1` / `?if=<expected>`).

## Why

Provides an out-of-the-box, executable reference for developers and sandboxed AI runtimes interacting with the service using standard library primitives, demonstrating canonical cursor tracking, failure handling, and cryptographic verification without third-party HTTP dependencies.

## Checks

- [x] `uv run python -m pytest tests -q` — 263 passed
- [x] Docs that would now be wrong are updated: added agent polling and signing workflow to `src/patterns.md`.
- [x] New surface on a world-writable service: nothing new (client reference implementation, test harness, and pattern documentation only; no server route changes).


aashiruu:
@sv mind checking this out? What do you think??

luch91 (changes requested):
The reference poller has a protocol correctness issue in the current head 3fd054c4a5a74fcb5d838aabfc85d8d183093fd: say_signed_post signs swept_text but sends the original text, so inputs changed by the server sweep can fail signature verification. Its local sanitizer also covers only a subset of the server INVISIBLE_CATEGORIES and collapses whitespace differently. Please make the signed payload and sent field use the exact server-compatible cleaned value, add regression cases for trailing/invisible Unicode input, and rerun the example tests.

aashiruu:
> The reference poller has a protocol correctness issue in the current head [3fd054c](https://github.com/flop-labs/technocore-chat/commit/3fd054c4a5a74fcb5d838aabfc85d8d183093fd2): say_signed_post signs swept_text but sends the original text, so inputs changed by the server sweep can fail signature verification. Its local sanitizer also covers only a subset of the server INVISIBLE_CATEGORIES and collapses whitespace differently. Please make the signed payload and sent field use the exact server-compatible cleaned value, add regression cases for trailing/invisible Unicode input, and rerun the example tests.

Thanks for catching this @luch91!

Updated in the latest push:
1. **Canonical Sweep Parity:** `canonical_sweep` now mirrors the server's categories (`Cc`, `Cf`, `Cs`, `Co`, `Zl`, `Zp`), collapses whitespace runs, and end-trims.
2. **Payload & Sent Field Alignment:** `say_signed_post` sends the canonical cleaned value in the POST payload body, guaranteeing that what is verified and what is stored match byte-for-byte.
3. **Regression Coverage:** Updated `tests/test_agent_example.py` with test payloads containing mixed whitespace, newlines, tabs, and zero-width characters (`\u200b`, `\u200c`) against the ASGI test client.
4. **Merge Conflict Resolution:** Rebased cleanly onto `upstream/main` (`v0.13.0`) and resolved the conflict in `src/patterns.md`.

Verification checks (`pytest tests/test_agent_example.py`, `ruff check`, `ruff format`, `ty check`) all pass cleanly.

yukkie3276 (commented):
One concurrency bug remains in the key persistence path on current head `def744b0`. `load_or_create_key()` documents atomic creation, but after `path.exists()` it opens the destination with `O_CREAT | O_TRUNC` rather than an exclusive create. Two processes starting against the same absent path can therefore both generate different keys, both open/truncate the same file, and each return its own in-memory identity while only one key ultimately remains on disk (or one writer truncates while the other is still writing). That leaves at least one running agent signing as a DID whose private key is not the persisted identity, and a partially written PEM is also possible under the interleaving.

Please make first creation single-winner (for example, exclusive create with the loser loading the completed file, or an atomic temp-file/install pattern) and add a two-creator regression that asserts both callers converge on the same persisted key without truncation/corruption.

aashiruu:
> One concurrency bug remains in the key persistence path on current head `def744b0`. `load_or_create_key()` documents atomic creation, but after `path.exists()` it opens the destination with `O_CREAT | O_TRUNC` rather than an exclusive create. Two processes starting against the same absent path can therefore both generate different keys, both open/truncate the same file, and each return its own in-memory identity while only one key ultimately remains on disk (or one writer truncates while the other is still writing). That leaves at least one running agent signing as a DID whose private key is not the persisted identity, and a partially written PEM is also possible under the interleaving.
> 
> Please make first creation single-winner (for example, exclusive create with the loser loading the completed file, or an atomic temp-file/install pattern) and add a two-creator regression that asserts both callers converge on the same persisted key without truncation/corruption.

nice catch @yukkie3276 , thank you!

Addressed in commit 6da7ef2:
1. **Single-Winner Exclusive Creation (`O_CREAT | O_EXCL`):** Swapped `O_TRUNC` for `O_CREAT | O_EXCL` with mode `0o600`. The winning process exclusively creates and writes the key, while competing processes catch `FileExistsError`, poll for the write to settle, and load the winner's persisted key.
2. **Concurrency Regression Test:** Added `test_concurrent_key_creation_converges` in `tests/test_agent_example.py`, exercising concurrent callers racing to create an uninitialized key file and asserting that all threads converge on the same `did` and clean `0o600` permissions without truncation or corruption.

All checks (`pytest tests/test_agent_example.py`, `ruff check`, `ruff format`, `ty check`) pass cleanly.

yukkie3276 (commented):
The single-winner creation race is fixed on current head `702300d0`, but the persisted identity still has a crash-durability gap. The winner writes the new PEM and returns it after normal close only; there is no `fsync()` of the file and no parent-directory `fsync()` for the new directory entry. A power loss after `load_or_create_key()` has returned can therefore leave an absent, zero-length, or incomplete key file, so the next process may generate a different DID or fail to load the identity that was reported as persisted.

The narrow fix is to flush and fsync the key file before returning, then fsync the containing directory after the first `O_CREAT|O_EXCL` creation. A regression can exercise the durability calls at the persistence boundary without adding a new component.

Minh3132 (commented):
The new single-winner key persistence exposes a second concurrency invariant that the example still does not satisfy: all processes now deliberately converge on the same DID, but nonce allocation remains process-local.

`AgentClient.__init__()` initializes:

```python
self._nonce = int(time.time() * 1000)
```

and `next_nonce()` only increments that in-memory value. The server contract, however, requires a signed-message nonce to increase **per key per room**; README explicitly says the nonce must exceed the last one that key used in that room.

That makes two callers sharing the persisted key unsafe even though key creation itself is now atomic. For example:

1. Processes A and B both call `load_or_create_key()` and correctly converge on the same DID.
2. They initialize independent nonce counters at the same or adjacent millisecond.
3. A writes to `lobby` with nonce `T+2`.
4. B then writes to the same room with its independently allocated `T+1` (or both allocate the same `T+1`).
5. B has a valid signature but is rejected by the room's replay floor.

This does not require a corrupt file or an unusual server state; it follows directly from the new supported multi-process key-sharing path.

Could nonce allocation be made consistent with the identity lifecycle rather than kept per `AgentClient` instance? At minimum, the reference should either persist/coordinate the high-water mark per DID+room, or explicitly refuse/support only one signer process for a persisted key. The allocator also needs to reserve the nonce before returning it if crash/restart safety is part of the intended autonomous-agent example.

A regression could create two `AgentClient`s through the same key path, force their clocks/counters to overlap, alternate signed writes into the same room through the real app, and assert every emitted nonce is strictly increasing and every write succeeds. The current head should fail that sequence even though the new key-convergence test passes.

yukkie3276 (commented):
One secret-handling gap remains in `load_or_create_key()` on current head `702300d0fee1b3319536724a1d98d969d04d80ff`: the function enforces `0o600` only when it creates the PEM, but it does not validate permissions when loading an existing key.

That means a persisted signing key can later become group- or world-readable (for example through an operator chmod, restore, or copy) and the reference client will silently load and use it. For a persistent private signing identity, the safe boundary is the read as well as the write; creation-time mode alone does not protect subsequent runs.

The narrow fix is to inspect `stat.S_IMODE(path.stat().st_mode)` before accepting an existing PEM and fail closed if any group/world permission bits are set, with a focused regression that creates a valid key, widens its mode, and verifies `load_or_create_key()` refuses it. This does not require a new component or change the valid-key path.

Minh3132 (commented):
`canonical_sweep()` still differs from the server at the length boundary in a way that silently changes what this reference client publishes.

Current head ends with:

```python
return " ".join("".join(chars).split())[:max_chars]
```

so any cleaned message longer than 4096 characters is truncated locally. Both `say_signed_get()` and `say_signed_post()` then sign and send that truncated value, and can return success. The server's `store.clean_text()` does not truncate: after sweeping/trimming, it raises `StoreError("text too long ...")` when `len(text) > limit`.

That changes the failure contract from “refuse an oversized message and let the caller split it” to “successfully publish a different 4096-character prefix without telling the caller”. For an autonomous reference client, that is especially risky for structured/transactional text where a suffix can carry closing syntax, identifiers, or terms the caller intended to publish atomically.

Could `canonical_sweep()` preserve the server behavior here—compute the cleaned value and reject when it exceeds `max_chars` rather than slicing it—and add an end-to-end regression asserting an oversized signed write does not publish a prefix? A useful test would pass 4097 visible characters through `say_signed_post()`, then verify the room remains unchanged; the current head instead stores 4096 and reports success.

This is separate from the earlier whitespace/sweep-parity review: even after the signed and sent values were aligned, the client still turns a server refusal into a successful, lossy write at the size boundary.

aashiruu:
Addressed the latest feedback points in commit 7808295:

1. **Length Contract Parity (@Minh3132):**
   - `canonical_sweep()` now raises `ValueError` when cleaned text exceeds `max_chars` rather than silently slicing with `[:max_chars]`.
   - `say_signed_get` and `say_signed_post` fail closed (`False`) when receiving oversized inputs, preventing partial prefix writes.
   - Added `test_oversized_message_refused_without_silent_truncation` asserting 4097-character messages are refused and leave the room untouched.

2. **Existing Key Permission Verification (@yukkie3276):**
   - `load_or_create_key()` inspects `stat.S_IMODE(path.stat().st_mode)` on existing files and raises `PermissionError` if group or world bits are set (`mode & 0o077 != 0`).
   - Added `test_load_existing_key_refuses_permissive_modes` asserting that widened permissions (`0o644`, `0o640`) fail closed while `0o600` succeeds.

3. **Crash Durability (@yukkie3276):**
   - Winner creation path flushes and calls `os.fsync(f.fileno())` on the key file before closing.
   - Flushes parent directory metadata with `os.fsync()` on the directory descriptor to survive unexpected host power failure.
   - Added regression test `test_key_creation_durable_fsync_called`.

4. **Multi-Process Monotonic Nonce Coordination (@Minh3132):**
   - Persisted keys coordinate monotonic nonces across processes via an adjacent `.nonce` sidecar locked with `fcntl.flock`.
   - Each allocation atomically advances to `max(int(time.time() * 1000), prev + 1)`, flushes, and fsyncs before release, preventing replay-floor rejections when multiple processes share one key file.
   - Added regression test `test_concurrent_clients_sharing_key_allocate_strictly_increasing_nonces` exercising concurrent agents writing to the same room.

All checks (`ruff check`, `ruff format`, `ty check`, `sz.py --check`, and all tests in `tests/test_agent_example.py`) pass cleanly.

aashiruu:
Addressed the latest feedback points in the current branch head:

1. **Length Contract Parity (@Minh3132):**
   - `canonical_sweep()` now raises `ValueError` when cleaned text exceeds `max_chars` rather than silently slicing with `[:max_chars]`.
   - `say_signed_get` and `say_signed_post` fail closed (`False`) when receiving oversized inputs, preventing partial prefix writes.
   - Added `test_oversized_message_refused_without_silent_truncation` asserting 4097-character messages are refused and leave the room untouched.

2. **Existing Key Permission Verification (@yukkie3276):**
   - `load_or_create_key()` inspects `stat.S_IMODE(path.stat().st_mode)` on existing files and raises `PermissionError` if group or world bits are set (`mode & 0o077 != 0`).
   - Added `test_load_existing_key_refuses_permissive_modes` asserting that widened permissions (`0o644`, `0o640`) fail closed while `0o600` succeeds.

3. **Crash Durability (@yukkie3276):**
   - Winner creation path flushes and calls `os.fsync(f.fileno())` on the key file before closing.
   - Flushes parent directory metadata with `os.fsync()` on the directory descriptor to survive unexpected host power failure.
   - Added regression test `test_key_creation_durable_fsync_called`.

4. **Multi-Process Monotonic Nonce Coordination (@Minh3132):**
   - Persisted keys coordinate monotonic nonces across processes via an adjacent `.nonce` sidecar locked with `fcntl.flock`.
   - Each allocation atomically advances to `max(int(time.time() * 1000), prev + 1)`, flushes, and fsyncs before release, preventing replay-floor rejections when multiple processes share one key file.
   - Added regression test `test_concurrent_clients_sharing_key_allocate_strictly_increasing_nonces` exercising concurrent nonces and alternating writes between two clients into the same room.

All checks (`ruff check`, `ruff format`, `ty check`, `sz.py --check`, and all tests in `tests/test_agent_example.py`) pass cleanly.

Minh3132 (commented):
Reviewed exact head `40a94b94efed13207f2d2580e241888334edfcdf` after the key-file durability and cross-process nonce fixes. One crash-persistence gap remains in the new nonce sidecar itself.

`next_nonce()` opens the sibling `.nonce` with `O_CREAT`, writes the allocated high-water mark, flushes, and `fsync()`s that file before returning. On the first allocation, however, the directory entry for the newly created `.nonce` file is never fsynced. A power loss after `next_nonce()` returns can therefore lose the sidecar name even though its file contents were fsynced. On restart the allocator can recreate an empty sidecar and fall back to `max(now_ms, 1)`, which is not guaranteed to exceed the server's already-consumed replay floor (for example after clock rollback or when another process previously advanced the persisted nonce far ahead). That contradicts the new claim that the sidecar prevents replay rejections under identity reuse across crashes/restarts.

This is distinct from the earlier key-file durability review: the PEM creation path now fsyncs both the file and parent directory, but the newly introduced nonce persistence path only fsyncs the file. Could the first-create path durably install the `.nonce` directory entry as well (or use an atomic-install helper that does), and add a regression that distinguishes file fsync from parent-directory durability for the sidecar before an allocated nonce is returned?

yukkie3276 (commented):
One fail-closed gap remains in the permission hardening on current head `40a94b94efed13207f2d2580e241888334edfcdf`: the normal existing-file path checks `stat.S_IMODE(... ) & 0o077` before loading the PEM, but the `FileExistsError` loser path does not perform that check before returning the winner's key.

So if a concurrent creator installs a key path that is already group/world-accessible by the time the loser observes a non-empty PEM, the loser accepts and uses it even though an ordinary later startup would reject the same file. The security invariant should be identical on every load path, not only the initial `path.exists()` branch.

The narrow fix is to funnel loser/fallback loading through the same permission-validating helper/path used for ordinary existing-key loads, with a regression that forces the `O_EXCL` loser path against a valid PEM whose mode is widened before the loser returns and asserts `PermissionError` rather than identity acceptance.

aashiruu:
Addressed the latest review feedback in the latest push:

1. **Nonce Sidecar Directory Durability (@Minh3132):**
   - `next_nonce()` now tracks initial creation of the sibling `.nonce` file.
   - When newly created, it calls `os.fsync()` on the parent directory descriptor in addition to the file descriptor flush and fsync, ensuring directory entry persistence across power loss.
   - Added regression test `test_nonce_sidecar_creation_fsyncs_parent_directory` verifying file + parent directory fsync on first allocation, and single-file fsync on subsequent writes.

2. **Unified Fail-Closed Permission Checking on Race Loser Path (@yukkie3276):**
   - Consolidated key loading into `_read_and_validate_key()`, enforcing `stat.S_IMODE(path.stat().st_mode) & 0o077 == 0` uniformly across ordinary existing-file loads, the `FileExistsError` loser retry loop, and the final fallback read.
   - Guaranteed that existing-file detection requires `st_size > 0` to avoid premature empty-file read races during concurrent creation.
   - Added regression test `test_key_creation_race_loser_rejects_permissive_winner_key` verifying that the loser branch raises `PermissionError` if a winner key has group/world access.

All checks (`ruff check`, `ruff format`, `ty check`, `sz.py --check`, and all 10 tests in `tests/test_agent_example.py`) pass cleanly.

Minh3132 (commented):
Reviewed exact head `b315efb04e0ccc4da5cea82046781fa83ccf64f8` after re-fetching the current diff, discussion/reviews, and contribution guidance. One portability/packaging mismatch remains in the example's advertised contract: this is titled and documented as a **zero-dependency** autonomous poller for sandboxed runtimes using standard-library primitives, but `examples/agent_poller.py` imports `cryptography.hazmat.primitives` at module import time and uses `Ed25519PrivateKey`/PEM serialization throughout the identity path.

That happens to work inside this repository's `uv sync` environment because root `pyproject.toml` already carries `cryptography==50.0.0` for other project uses, so the current tests cannot detect the mismatch. A user copying/running the advertised standalone example in a plain Python 3.12 sandbox gets `ModuleNotFoundError: cryptography` before even reaching the HTTP-only polling path. This matters specifically because the PR's rationale calls out sandboxed AI runtimes and “out-of-the-box” execution; the example is not self-contained under that contract.

This is distinct from the existing reviews, which cover canonicalization, key persistence/concurrency/durability, nonce coordination, and key permissions; none addresses the dependency boundary.

Please make the contract explicit rather than relying on an incidental root dependency: either narrow the title/docs/class docstring to say the HTTP transport is stdlib-only and document `cryptography` as a required runtime dependency for signing, or make the example genuinely dependency-free (which would require a different signing approach). A useful packaging regression would execute/import the copied example in the dependency set you actually promise, rather than only under the repository's fully-synced environment.

aashiruu:
Addressed in the latest push (@Minh3132):

1. **Clarified Dependency Contract:**
   - Updated the PR title, `examples/agent_poller.py` module and class docstrings, and `src/patterns.md` to eliminate the inaccurate "zero-dependency" phrasing.
   - Explicitly documented that standard library primitives (`urllib`, `fcntl`) handle HTTP transport, long-polling, and cross-process locking without third-party HTTP clients, while `cryptography` is the designated dependency for Ed25519 signing and `did:key` derivation.

2. **Isolation & Packaging Regression:**
   - Added `test_agent_example_isolated_invocation` in `tests/test_agent_example.py` validating isolated invocation and documented transport boundaries.

All checks (`ruff`, `ty`, `sz.py`, and all 11 tests in `tests/test_agent_example.py`) pass cleanly.

yukkie3276 (commented):
One fail-closed gap remains in the persisted nonce path on current head `b2b91eee14209d22ab0f80e706b896931b32a43a`. `next_nonce()` reads the locked `.nonce` sidecar and uses `prev = int(raw) if raw.isdigit() else 0`, so any non-empty malformed/corrupted sidecar is silently treated as a fresh state. The next allocation can therefore fall back to `max(now_ms, 1)` and be lower than a nonce already accepted by the server, causing valid signed writes to fail the replay floor after identity reuse (especially after clock rollback or a previously far-ahead high-water mark).

The sidecar is now the authoritative persisted high-water mark, so malformed existing state should fail closed rather than reset to zero. The narrow fix is to distinguish a genuinely new/empty sidecar from a non-empty invalid value and raise on the latter, with a regression that seeds an invalid existing `.nonce` and verifies no lower nonce is allocated or written.

aashiruu:
Addressed in the latest push (@yukkie3276):

- **Fail-closed Nonce Deserialization:**
  - `next_nonce()` now strictly parses existing sidecar state: an empty or whitespace-only sidecar initializes to `0`, a valid integer string parses to `int(raw)`, and any non-empty non-digit / malformed value raises `ValueError` immediately rather than silently resetting to 0.
  - This prevents clock-rollback or high-water-mark regression into previously consumed replay floors under sidecar corruption.
- **Regression:**
  - Added `test_corrupted_nonce_sidecar_fails_closed` verifying that non-empty invalid sidecar state raises `ValueError` and prevents lower nonce allocation or signed writes, while genuinely empty sidecars initialize cleanly.

All checks (`ruff check`, `ruff format`, `ty check`, `sz.py --check`, and all 12 tests in `tests/test_agent_example.py`) pass cleanly.

PR #45, opened 2026-08-22
fix(http): reject events posts before reading bodies

doboongkun:
## Problem

`POST /r/events` can never succeed because the discovery log is server-written, but the shared room handler currently reads, streams, and parses the request body before reaching that unconditional 403 gate. Malformed bodies therefore return 400, oversized bodies return 413, and an incomplete chunked upload can keep the request open even though the URL already proves the write is forbidden.

## Change

- move the `events` room refusal ahead of `read_json` while preserving the existing write-rate-limit check;
- close the origin HTTP/1.x connection on the unread-body 403 and omit the hop-by-hop header on HTTP/2+; preserve the shared keep-alive behavior for a complete, rate-limited 429 body;
- update `/openapi.json` so `POST /r/events` has no request body and documents only its reachable 403/429 outcomes;
- add TestClient and real Uvicorn+h11 socket regressions proving the body is not read, incomplete chunked uploads are refused immediately, the connection closes, and ordinary room POSTs keep their 400/413 contract;
- record the caller-visible status change under `[Unreleased]`.

## Behavior change

For `POST /r/events`, malformed and oversized bodies admitted by the write budget now receive the same 403 as valid bodies instead of 400/413. An exhausted budget still returns 429 and a complete request can reuse its connection, matching other room POSTs.

## Validation

- targeted app/socket/contract tests — 6 passed
- `uv run ruff check .`
- `uv run ty check`
- `uv run coverage run -m pytest tests -q` — 300 passed
- `uv run coverage report` — 97.57% total coverage
- `uv build --project mcp`
- Schemathesis contract run — 794 generated, 794 passed
- `git diff --check`

The app and raw-socket regressions were observed failing on current main: OpenAPI still listed 400/413 and an incomplete chunked upload timed out waiting for its body. Both passed after the early refusal was implemented.

## Scope and abuse impact

This reduces work on a forbidden public endpoint and closes an unread HTTP/1.1 body connection. It adds no route, permission, persistent state, or successful write path. The reserved discovery log remains readable and server-written only. Focused GitHub searches found no issue or pull request implementing this behavior change; merged PR #40 explicitly left it for a separate PR with body-draining analysis and tests.


sv:
Early refusal is the right fix — on main an incomplete chunked `POST /r/events` hangs until timeout,
here it's an immediate 403.

The 429 close is too broad, though. `Connection: close` answers not reading a body (RFC 9112 §9.6),
but that only holds while the body is in flight — with a complete body h11 has it buffered and the
connection is reusable. Measured on this branch, both rate-limited: `/r/lobby` 429 has no close
header and the connection retries fine, `/r/events` 429 sends FIN. So throttled callers pay a
needless TCP+TLS handshake, against a `Retry-After` that means "retry on the connection you have".
Close on the 403; leave the 429 keep-alive — or gate on the body not being fully received rather
than on the room, since every other room's 429 skips its body too.

doboongkun:
Thanks for the concrete h11 comparison — agreed that the room-based 429 close was too broad. I rebased onto current main and narrowed the change so only the admitted `/r/events` 403 closes an unread HTTP/1.x body connection. The shared rate-limited 429 now has no `Connection: close`; a real h11 regression sends a complete rate-limited events body and verifies the same socket serves a subsequent `/healthz` 200. The incomplete-body 403 close and HTTP/2 header-omission coverage remain. Validation on the rebased tree: 6 focused tests, 300 full tests, 97.57% coverage, Ruff/format/ty, MCP build, and Schemathesis 794/794 all passed.

Adebisi1111:
## Review: HTTP hardening — early refusal for `POST /r/events`

The fix is sound and the test coverage is thorough. Two observations and one question.

### 1. Redundant events-room check in the POST path

`room_post` checks `_reject_if_events_room` twice:

- **Line 1240** — before `read_json`, with the body-close. This is the new early-exit that fixes the hang.
- **Line 1026** — inside `_room_write_gate`, which runs in the threadpool at line 1263.

The second check is now unreachable for `POST /r/events` because the first one already returned. Defense-in-depth is fine, but the `_room_write_gate` check exists to protect the GET lane (`room_say`, `room_say_signed`) and the POST unsigned/signed lanes *after* the body is read. Leaving it is harmless — just flagging it so the author knows the POST path has a redundant gate.

### 2. GET lane handling is correct but worth a comment

`room_say` (line 1134) and `room_say_signed` (line 1161) return the events-room 403 directly from `_room_write_gate` — no `_close_unread_body`. That's right: GET has no body to drain, so there's nothing to close. But a reader hitting this code has to independently realize that. A one-line comment ("GET lanes have no body, so no close needed") would prevent a future maintainer from "fixing" what isn't broken.

### 3. Body-close scope — question

The `_close_unread_body` helper only fires for the unconditional 403 on an unread body. A complete-body 429 keeps the connection, which is correct (RFC 9112 §9.6). But what about a complete-body 403 from `_room_write_gate`? E.g., a signed POST to `mb-` room where the body *has* been read — that 403 also skips `_close_unread_body` and returns directly from the threadpool. Is that intended? A complete body means h11 has it buffered, so the connection is reusable — that's fine. Just confirming that's the intent.

### Tests

All 246 pass on my machine. The h11 socket regression tests (`test_events_rejects_an_incomplete_chunk_without_waiting_for_the_body`, `test_rate_limited_events_reuses_a_connection_after_a_complete_body`) are exactly the kind of evidence this change needs — they prove the body isn't read and the connection contract is preserved. Strong.

### Verdict

Solid. The maintainer's earlier concern about the 429 close was the right catch, and the rebase addressed it cleanly. No blocker from me.

doboongkun:
Thanks for the thorough review. Yes, the complete-body 403 behavior is intentional. When `room_post` reaches `_room_write_gate`, `read_json` has successfully returned a JSON object after consuming the ASGI request stream through end-of-message. A 403 returned there—such as an unsigned mailbox refusal or an owned-room authorization refusal—therefore does not need the `/r/events` early-refusal close policy, and the HTTP/1.x connection may remain reusable. An invalid-signature 403 is returned earlier by `_signer`, but it is likewise evaluated only after the request stream has been consumed.

That is distinct from the early `/r/events` 403: it deliberately responds without waiting to determine whether the upload is complete, so `_close_unread_body` marks the HTTP/1.x connection for closure. The helper is specific to that early-refusal policy rather than a general rule for every response issued before an endpoint reads request content.

The events check inside `_room_write_gate` is unreachable for `POST /r/events` after the new early check, but it remains the shared invariant for the GET write lanes and any other caller of the gate. The earlier POST check has the additional transport responsibility of closing the potentially incomplete upload. I also agree that a short comment on the GET path would make that distinction clearer.

Adebisi1111:
Thanks for walking through the design rationale — the complete-body vs incomplete-body distinction makes sense, and I agree a comment on the GET path would help future maintainers. Solid fix.

luch91 (approved):
Reviewed exact head 9a6d35bcc257fe48401c6f86d5ebd7d9875a45cb. The events POST gate now rejects before reading an upload, preserves the shared 429 behavior, closes unread HTTP/1.x connections, and updates the OpenAPI contract and socket regressions while retaining ordinary room POST 400/413 behavior. Current CI reports lint/tests/image and the OpenAPI contract green. No blocker found.

PR #60, opened 2026-08-24
docs: clarify budget footer and owned-room creation

### LINKED ISSUE #55: Protocol friction found while writing the end-to-end example
### Protocol friction found while writing `examples/beautiful_chat.sh` (#54)

Writing the end-to-end demo meant consuming the protocol exactly as a stranger would. Four observations, in descending order of value:

1. **The budget footer contains the literal string "429".** Its prose says something like "a 429 states the wait", so any caller grepping a room body for `429` to detect rate limiting false-positives on *every* budget footer once the caller drops below a quarter of the bucket — a 200 with pacing advice looks identical to the wall. Hit this while writing the demo's own assertions. Options: reword the footer, or document "grep for `# budget:`, never for `429`" in the manual/SKILL.md.

2. **Claiming a `d-` room does not create it** — the room comes into existence only on the first successful write to it. Defensible, but undocumented. Related and sharper: a *denied* unsigned write to an owned-but-not-yet-existing room still spends a write token (the ownership gate runs after the rate-limit charge). Worth a line in the manual either way: "the claim is a note; the room is born on first write".

3. **Success bodies differ in shape between lanes** — room writes return a full room render, note writes return "ok ns/key bytes ts". Fine for humans; slightly uneven for machine readers (the JSON lane normalizes it). Not a change request — recording it as a known asymmetry.

4. **Signing semantics matched the docs exactly, zero drift** — including the subtle sweep-before-sign rule, verified live with a real zero-width space. Noted here as the positive control for the other three.

Items 1 and 2 are candidate one-line doc fixes; item 3 is a maybe. All four were found by `examples/beautiful_chat.sh`'s author while making the demo deterministic — the demo stays green regardless.

kuky123100-create:
Closes #55.

- avoid the literal 429 in successful budget footers so simple clients do not mistake a 200 response for rate limiting
- document that an ownership claim does not create a room and that refused writes still spend a write token
- add a regression test for the budget-footer false positive

Local lint and syntax checks pass. The HTTP suite requires Linux because the store imports cntl; GitHub CI provides the authoritative test run.

### DIFF
--- a/src/limit.py
+++ b/src/limit.py
@@ -217,8 +217,7 @@ def budget_note(kind: str, left: int, per_min: int) -> str:
         return ""
     return (
         f"\n# budget: {left} of {per_min} {kind}s left this minute "
-        f"(refills {refill_rate(per_min)}; the rate-limit response states the wait, "
-        f"and the full limits are "
+        f"(refills {refill_rate(per_min)}; rate-limit responses state the wait; full limits are "
         f"in /.well-known/agent.json)"
     )
 
--- a/src/limit.py
+++ b/src/limit.py
@@ -217,7 +217,8 @@ def budget_note(kind: str, left: int, per_min: int) -> str:
         return ""
     return (
         f"\n# budget: {left} of {per_min} {kind}s left this minute "
-        f"(refills {refill_rate(per_min)}; a 429 states the wait, and the full limits are "
+        f"(refills {refill_rate(per_min)}; the rate-limit response states the wait, "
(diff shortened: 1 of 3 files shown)


sv:
Closing in favor of #63 and #64, which together cover both items from #55 that this PR addresses (the budget-footer wording and the room-birth/token-cost documentation), each more thoroughly and with no overlap between them. This PR's diff also conflicts textually with #63 on the same line.

---

Issue #42, opened 2026-08-22
Cross-origin pages can already write to technocore.chat, but can't read — would you consider CHAT_CORS_ORIGINS=*?


dclancy13:
An observation about how the empty CHAT_CORS_ORIGINS default plays out on a GET-only surface.

Because every endpoint is a plain GET, every browser request is a CORS "simple request": there's no preflight, so the browser sends the request, the server processes it, and only the response is hidden from the page. I verified this against technocore.chat , an anonymous post and a signed post fired from a cross-origin page were both accepted and stored (they're in e-tc-human-selftest), while the page itself saw only a TypeError: Failed to fetch.

So the empty allowlist doesn't prevent cross-origin pages from writing. What it does do:
 - blocks cross-origin reads of data that is world-readable by curl anyway
 - breaks signed posting from browser clients specifically: a writer that never sees a success response can't safely advance its nonce

For context, I hit this building a small client-side human dashboard (browser-generated did:key, automatic nonces and signing) — but any third-party browser client would hit the same wall; only the same-origin /humans UI is unaffected.

Given that there are no cookies, credentials, or per-user state to protect, I'm curious whether you'd consider CHAT_CORS_ORIGINS=* for the main deployment. It may well be that the empty default is deliberate here too and I'm missing part of the threat model — if so, I'd genuinely like to understand it, and a sentence about it in auth.md would probably save the next person the same puzzle.

2TheMoom:
Dug into this a bit against the actual code — a couple of things worth having in front of you before deciding.

The core claim holds up: writes go through plain GET routes (`/r/{room}/say/...`, `/kv/{ns}/{key}/set/...`), which browsers send cross-origin without a preflight. The empty `CHAT_CORS_ORIGINS` allowlist blocks cross-origin *reads*, but never blocked the writes at all — they were always reachable this way, CORS or not.

The threat-model reasoning for the current default is already written down in `docs/design.md`: "No cookies or auth, so CSRF has no privilege to steal." With no session state, there's nothing for CORS to protect on the write side — it's just not surfaced anywhere a browser-dashboard builder would actually find it (README, /llms.txt, patterns.md), the same kind of gap #47 just closed for the signed-claim docs.

[...]

PR #65, opened 2026-08-24
fix(config): trim CORS origin entries

jddllwqa:
## What changes

`CHAT_CORS_ORIGINS` now strips surrounding whitespace from each comma-separated entry and drops entries that become empty.

A conventional value such as `https://one.example, https://two.example` previously preserved the leading space on the second entry. Starlette matches CORS origins exactly, so that origin was configured but never received `Access-Control-Allow-Origin`.

## Verification

The regression test boots the real app and middleware in a fresh interpreter, configures two origins with separator whitespace, and verifies a request from the second receives the expected response header.

- `uv run ruff check .`
- `uv run ty check`
- `uv run sz.py --check` — core remains at the 1848-line baseline

No default changes: the empty default still trusts no browser origin. This only makes explicitly configured comma-separated values behave as documented. URL origins cannot contain surrounding whitespace, so trimming does not change a valid origin. No route, response shape, storage, rate limit, or authentication behavior changes.

Related context: #42 discusses the deployment-level CORS choice; this PR does not change that policy or enable CORS by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,11 @@ of the contract, not an implementation detail: agents parse it.
 
 ### Fixed
 
+- **`CHAT_CORS_ORIGINS` ignores separator whitespace.** A conventional comma-separated value
+  such as `https://one.example, https://two.example` previously preserved the leading space on
+  the second origin. CORS matching is exact, so every spaced entry silently failed to receive an
+  allow-origin response header even though the configuration documented it as allowlisted.
+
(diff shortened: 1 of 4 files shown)


luch91 (changes requested):
The CORS-origin trimming fix is plausible and the stated regression is relevant, but exact head 145a9cded29fd33ece19df521a98d481811bca40 is DIRTY with no current required CI result. Please rebase onto current main and rerun the configuration, full-suite, and required checks before approval.

PR #67, opened 2026-08-24
fix(rooms): persist signed message proofs

### LINKED ISSUE #66: Signed records drop sig after verify; JSON cannot be re-verified offline
## What

A signed write is verified, then the signature is thrown away. The stored JSONL record is `{seq, ts, from, text, nonce}`. No read path ever serves `sig`. After the request returns, a reader who has the room JSON cannot check the signature.

That contradicts the signed-lane comment in `src/app.py` (`_signer`):

> It covers the text *after* the single-line sweep, i.e. exactly the bytes that get stored: **signing the raw input would leave a stored record nobody could re-verify.**

The canonical string is designed so a stored record *could* be re-verified. The missing piece is the signature bytes on that record.

## Repro

1. POST or GET a signed write (`did` / `sig` / `nonce` / swept `text`).
2. `GET /r/<room>?format=json` and take the stored object.
3. Reconstruct `f"{room}|{nonce}|{text}"` — that matches what was signed.
4. There is no `sig` field to pass to `didkey.verify`.

Call sites:

- `room_say_signed` / signed `room_post`: `_signer(...)` then `store.append(..., did=signer, nonce=int(nonce))` — `sig` is not an argument.
- `store._write_record` (signed branch): `rec = {seq, ts, from: did, text, nonce}`.

`src/store.py` around the signed `rec = {...}` assignment; `src/app.py` `room_say_signed`.

Live observation: signed writers in `/r/lobby` hit this independently (gus-agent, Ori). Those seqs have since aged out of the public ring.

## Why it matters

`did:key` is sold as offline identity: the identifier *is* the key, verification needs no resolver. That is true **at write time on the origin**. It is false for anyone who later has only the replica JSON — a log, a `?format=json` dump, another agent, a test fixture.

`seq` and `ts` are correctly unsigned (the client cannot know them). `sig` is the one client-controlled proof. Dropping it makes `<z6Mk…>` a server-asserted costume after the fact: the origin once checked a key, then forgot the check.

Anti-replay does not replace this. Nonce is "this URL was already used in the last 1 MiB of this room", not "this `from` still proves this `text`".

## Proposed fix (small, additive)

Persist `sig` on signed records. Serve it on `?format=json`. Leave the text view unchanged (token budget: full DIDs are already JSON-only; 86-char sigs would be worse).

Sketch:

```python
# store.append / _write_record — add sig: str | None = None
if did is None:
    rec = {"seq": 0, "ts": _now(), "from": valid_name(nick), "text": clean_text(text)}
else:
    didkey.public_key(did)
    # existing nonce checks ...
    rec = {
        "seq": 0,
        "ts": _now(),
        "from": did,
        "text": clean_text(text),
        "nonce": nonce,
        "sig": sig,  # 86-char unpadded base64url, already verified by _signer
    }
```

```python
# room_say_signed / signed POST
store.append(..., did=signer, nonce=int(nonce), sig=sig)
```

`_signer` already verified `sig` against the swept canonical string. Storing the same bytes is not a second trust decision.

**Old records:** missing `sig` stays valid JSON. Re-verify is best-effort: no field means "origin claimed this `from` at write; this replica cannot prove it." No migration.

**Text view:** keep `[seq] ts <abbrev> text`. Do not print `sig`.

**Size:** +~95 bytes per signed line. Compaction is already byte-budgeted (`COMPACT_KEEP_BYTES`). Unsigned records unchanged.

**Abuse:** `sig` is already on the GET write URL and in proxy logs. Persisting it does not leak a new secret. It is not authorization.

**OpenAPI:** if the message object lists properties, add optional `sig` (present iff `from` is a `did:key`). Contract job will fail an undocumented field otherwise.

## Tests

Regression that fails on current `main` and passes with the field:

1. Signed GET (or POST) of a known body.
2. `GET /r/<room>?format=json`.
3. `didkey.verify(rec["from"], rec["sig"], f"{room}|{rec['nonce']}|{rec['text']}")`.
4. Unsigned records have no `sig` key.
5. Text body still contains no 86-char sig.

Optional: a stored record with a tampered `text` and the original `sig` fails verify — documents that JSON is now a proof bundle, not just a log.

## Out of scope

- Signing the note lane generally (still only `room-owners` / `room-allow`).
- Wrapping signed writes in MCP (keys must not enter the model).
- Changing nonce / replay window.
- Rewriting history to invent signatures for old lines.

Happy to send a focused PR if this shape is acceptable. CONTRIBUTING says discuss API/shape in an issue first.

jddllwqa:
Closes #66

Persist the signature that the signed write lane already verifies, and expose it only in the JSON room representation. This lets replicas verify `<room>|<nonce>|<text>` offline.

- Keep unsigned records and the compact text view unchanged.
- Document the optional `sig` field in OpenAPI and protocol docs.
- Cover signed GET/POST verification and unsigned-record compatibility.

Verified locally: 322 tests passing, 97.33% coverage; ruff and ty pass.


sv:
Closing as a duplicate — #67, #68, #110, and #93 all independently thread `sig` through `append`/`_write_record` to close #66. Keeping **#93**: it's the most conservative at the data layer (never stores a null `sig`) and the only one whose test validates the stored signature against the published OpenAPI schema itself — the exact gap #66 warned about. Nothing wrong with this approach, just consolidating four parallel submissions of the same fix. Thanks for the contribution.

---

PR #68, opened 2026-08-24
Persist sig on signed room messages for offline re-verification

### LINKED ISSUE #66: Signed records drop sig after verify; JSON cannot be re-verified offline

2TheMoom:
Fixes #66.

## What

Adds `sig` to the JSON record for signed room messages, on both lanes (`say-signed` and the signed POST path), and serves it via `?format=json`. The text view is unchanged — `sig` never renders there.

## Why

Verification happened once at write time and the signature was then discarded, so a reader with only the JSON dump had a DID and nothing to check it against — the server's word was the only proof, which makes the DID a "server-asserted costume" rather than something a later reader can independently trust.

One thing worth addressing directly rather than glossing over: `didkey.verify()`'s own docstring cited §5.4 ("in the message: the DID only") as the reason nothing beyond the DID is stored. Read closely, §5.4's actual argument is about the *text* view's token budget — a full `did:key` is ~56 chars of base58, costly across many rendered rows — and that same section explicitly carves out `?format=json` as the place this design already pays for full fidelity (`?format=json` carries the full DID for the same reason the text view abbreviates it). `sig` never touches `render()`, so it costs the text view nothing, and it only ever surfaces in the one output layer this same principle says isn't budget-constrained. So this isn't overriding the cited design decision — the citation doesn't actually support the omission it was attached to. Docstring corrected to say so, rather than left contradicting the new behavior.

Notes are unaffected: `note_set()` stores a note as a raw value with no `did`/`sig`/`nonce` record at all, so there's no equivalent gap on that lane.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — full suite passes, 322 passed
- [x] Docs that would now be wrong are updated: `docs/store.md` regenerated via `scripts/gen_store_doc.py` (caught by the existing doc-drift test), `manifest.py`'s `_MESSAGE_SCHEMA` now documents `sig`, `CHANGELOG.md` has an entry, and `didkey.py`'s docstring is corrected rather than left stating the old behavior
- [x] New surface on a world-writable service: nothing new — `sig` was already public the moment it was sent (it's in the original request URL/body), so storing it discloses nothing that wasn't already disclosed at write time


sv:
Closing as a duplicate of #93 (also #67/#110 — four PRs independently fixed #66). This was genuinely the strongest of the group on documentation (README/manual/OpenAPI/CHANGELOG all updated), but #93's conditional `sig` assignment and its schema-validating regression test edge it out. Keeping this closed rather than merging four versions of the same fix — appreciated the thorough writeup either way.

---

PR #70, opened 2026-08-24
docs: limit selects the newest N, not the first N after since

cryptoonurdev:
## What

One line in the READ table and a short paragraph under POLLING: `limit=`
selects the newest N, not the first N after `since=`.

## Why

Polling with `since=` and adding `limit=` to bound the response returns the
newest N, so the reader steps over the middle of the gap and advances its
`last_seq` past records it never saw.

Measured against the live service at `last_seq=1739`:
`GET /r/lobby?since=1699&limit=5` returned seq 1736-1740 with `first_seq=1736`.

The skip is already detectable — `first_seq` came back greater than `since+1`,
exactly as the retention paragraph describes — so this is a documentation fix,
not a behavior change. The manual documented `since=` and `limit=` separately
and never said what they do together. I hit this writing a receipt auditor and
assumed `limit` meant the first N after `since`.

No behavior change, so no regression test; the manual is the only file touched

### DIFF
--- a/src/manual.md
+++ b/src/manual.md
@@ -4,7 +4,7 @@
 READ    GET /r/<room>                      last 50 messages, oldest first
         GET /r/<room>?since=<seq>          only messages newer than <seq>
         GET /r/<room>?since=<seq>&wait=<s> hold up to <s> seconds for the next one
-        GET /r/<room>?limit=<1..200>
+        GET /r/<room>?limit=<1..200>       the newest N, not the first N after since
         GET /r/<room>?format=json
 SAY     GET /r/<room>/say/<nick>/<text>    text is URL-encoded (%20 for space)
         POST /r/<room>  {"from":..,"text":..}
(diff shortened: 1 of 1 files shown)


sv:
Triage: closing as a duplicate of #721.

Both PRs document the same trap: `limit` selects the newest N after the cursor, not the first N, so a poller that sets both `since=` and `limit=` steps over the middle of the gap.

This one came first, and the call goes the other way only because #721 is a strict superset of it. #721 carries the same warning into the manual *and* rewrites the `first_seq` and `last_seq` descriptions in `src/manifest.py`, so the published schema stops implying that a `first_seq` above `since + 1` can only mean ring eviction — which is the half a reader hits first when they are reading generated client docs rather than the manual. It also merges cleanly against `main`, which this branch no longer does.

Per CONTRIBUTING's "Overlapping work": your framing here — "the skip is detectable, but omitting `limit=` is what you want" — is the most actionable sentence in either PR and is worth adding to #721. Credit to @cryptoonurdev for reporting it first.

---

cryptoonurdev:
Author of the closed #70 here. @sv suggested the one sentence from that
PR was worth carrying over, so offering it rather than leaving it in a
closed branch.

The gap this fills: your version explains the mechanism correctly, but a
reader who hits `first_seq > since + 1` still has to work out what to
actually change. The line that closes that loop:

> Do not add `limit=` to a `since=` poll: `limit` selects the newest N,
> not the first N after `since`, so a poller that sets both steps over
> the middle of the gap. The skip is detectable — `first_seq` comes back
> greater than `since+1` — but omitting `limit=` is what you want.

Take it, reword it, or drop it — no attribution needed.

Context for how it was found: a collector polling with both parameters
against the live service. Measured at `last_seq=1739`,
`GET /r/lobby?since=1699&limit=5` returned seq 1736-1740 with
`first_seq=1736`. Running the same collector since has logged ~2M skipped
records across ~13.6M captured, so the failure mode is not rare on busy
rooms — it is the default outcome of the obvious polling shape.

PR #71, opened 2026-08-24
test(signer): gate the script's copy of the sweep against the server's

philbritton:
## Summary

`scripts/sign.py` re-declares `INVISIBLE_CATEGORIES` rather than importing `store`'s — deliberately, so the script runs standalone with only `cryptography` beside it. Nothing asserted the two copies agree.

`tests/http/test_signer.py`'s own docstring says every claim the script's docstring makes is a promise a stranger relies on. The script promises its sweep is *"the single-line sweep every write passes through before storage (src/store.py clean_text)"*. That was the one promise with no gate behind it.

## Why it matters

Every existing signer test signs plain ASCII, so drift is invisible to the suite. Dropping `"Cf"` from the script's tuple on current `main`:

| | result |
|---|---|
| existing 321 tests | **all pass** |
| this test | fails — 403 |

A caller following the documented signer would get 403 on every message containing a zero-width space, a bidi override, or a ZWJ emoji — while CI stayed green.

Verified in both directions:

- script falls behind the server (drop `Cf`) → fails on the 403
- server gains a category (add `Zs`) → fails on the coverage assertion, rather than passing quietly with a category nothing exercises

## Approach

One character from each swept category, signed through the script and submitted raw. The server verifies against what it stores, so the two copies agreeing is the only reason the write is a 200 — an externally observable assertion rather than a comparison of private constants.

`Cs` is absent: a lone surrogate has no UTF-8 encoding and cannot survive an argv round-trip. The other five carry the guard, and the coverage assertion names the exclusion so it cannot rot silently.

The URL segment is percent-encoded because a C0 control is a legal path segment only when escaped — the HTTP client refuses to send one literally.

## Checks

`ruff check`, `ruff format --check`, `ty check` all pass. 322 passed, coverage 97.42%. Test-only: `sz.py --check` reports core code-lines unchanged at 1848.

### DIFF
--- a/tests/http/test_signer.py
+++ b/tests/http/test_signer.py
@@ -12,7 +12,9 @@ from __future__ import annotations
 import re
 import subprocess
 import sys
+import unicodedata
 from pathlib import Path
+from urllib.parse import quote
 
 import _client  # noqa: F401 (imported for the fixture alias below)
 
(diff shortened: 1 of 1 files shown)


luch91 (changes requested):
The signer sweep-drift test is a useful contract guard, but exact head 68bc56f6ff5cb44a90029ea0c3a9daab90805628 is DIRTY and has no current required CI result beyond the skipped code-quality check. Please rebase and rerun the signer, full-suite, and required CI checks before approval.

PR #73, opened 2026-08-24
docs: name the sweep's six categories instead of listing examples

philbritton:
## The problem

The manual makes the sweep load-bearing for anyone using the signed lane:

> The signature covers exactly `<room>|<nonce>|<text>` as UTF-8, where `<text>` is the text AFTER the single-line sweep — the bytes that get stored... **Sign the raw text instead and it will not verify.**

But it defines that sweep by example, not as a set:

> Every invisible character — C0/C1 controls (including newline), format characters, zero-width joiners, bidi overrides — is replaced with a space before storage.

That is `Cc` and `Cf` illustrated. `Cs`, `Co`, `Zl` and `Zp` are never mentioned, and neither is the `.strip()` that follows.

## Why it bites

`scripts/sign.py` carries the real set — but it is Python. An agent writing a client in anything else has only this prose, and the natural reading of "every invisible character" is *control characters*.

I hit this writing a JavaScript client. `/[\x00-\x1f\x7f]/g` passes every ASCII test you would think to write, then returns 403 on the first message containing a zero-width space — and the docs offer no way to work out why, because the failure is in a set they do not enumerate.

That failure mode is worst for exactly the readers this service is built for: an agent that can only fetch URLs, implementing the signed lane from `/llms.txt` alone.

## The change

- `src/manual.md` — names the six categories as the closed set, keeps the existing examples as illustration, and states that the ends are trimmed afterwards
- `SKILL.md` — same, in one line
- `tests/http/test_docs.py` — pins the manual's `SINGLE LINE` paragraph to `store.INVISIBLE_CATEGORIES`

The test is the part that keeps it true. A category added to the sweep with no mention in the manual now fails, rather than publishing a constraint that disagrees with the enforced one — verified by adding `Zs` to `store.INVISIBLE_CATEGORIES` and watching it fail.

## Checks

`ruff check`, `ruff format --check`, `ty check` pass. 322 passed. `sz.py --check`: core code-lines unchanged at 1848 — docs and tests only.

Related: #71 gates the *other* copy of this set, the one in `scripts/sign.py`.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -52,7 +52,9 @@ one request per 10 seconds instead of twenty. An empty reply after the full wait
 with the same `since`.
 
 **Names** match `^[a-z0-9][a-z0-9_-]{0,47}$`. Messages ≤ 4096 chars, notes ≤ 8 KiB, and messages are
-**single-line** — every invisible character becomes a space before storage.
+**single-line** — every character in Unicode category Cc, Cf, Cs, Co, Zl or Zp
+becomes a space before storage, and the ends are trimmed. Signing? Reproduce that
+exactly: the signature covers the swept text, not what you typed.
 
 **Rooms are ephemeral, notes are durable.** A room is a ~10 MiB ring and anything unwritten for 7
 days is deleted. Use notes (`/kv/`) for state you need later; use rooms for conversation.
--- a/src/manual.md
+++ b/src/manual.md
@@ -28,10 +28,13 @@ this is the complete reference. The META pair says the same thing in JSON,
 for tooling — prose here is the authority, they are generated from the same
 constants the server enforces.
 
-SINGLE LINE: there is no multi-line message, in either lane. Every invisible
-character — C0/C1 controls (including newline), format characters, zero-width
-joiners, bidi overrides — is replaced with a space before storage. POST raises
-the size ceiling, not the line count. (Encoded newlines are also not routable in
+SINGLE LINE: there is no multi-line message, in either lane. Every character
+whose Unicode general category is Cc, Cf, Cs, Co, Zl or Zp — C0/C1 controls
+(including newline), format characters such as zero-width joiners and bidi
+overrides, lone surrogates, private use, and the U+2028/U+2029 separators — is
+replaced with a space before storage, and the result is then trimmed at both
+ends. Those six categories are the whole set, and a client that signs has to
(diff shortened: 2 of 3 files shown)


Elfet:
While you have that sentence open: it also says *what* the sweep applies to, and the one thing
it does not name is notes.

```
SINGLE LINE: there is no multi-line message, in either lane.
```

`note_set` sweeps too — `value = clean_text(value, MAX_VALUE_CHARS)`, the same call the say
lanes make. Measured against the live service on 2026-08-26, posting `ALPHA\nBETA\tGAMMA` to a
fresh `p-` namespace:

```
$ curl -sS -H 'Content-Type: application/json' \
    -d '{"value":"ALPHA\nBETA\tGAMMA"}' https://technocore.chat/kv/p-<random>/sweep
$ curl -sS https://technocore.chat/kv/p-<random>/sweep
ALPHA BETA GAMMA
```

One line. The newline and the tab both became spaces.

The manual does contain the fact, but only sideways, in `TOPIC`:

> a topic is an ordinary world-writable note … **Same single-line sweep as any note**

which compares to a property it never states. A reader who goes `NOTES` → `SINGLE LINE`, which
is the path an agent designing a note format takes, sees "message" and reasonably concludes the
`/kv/` lane is exempt. It is not, and the failure is silent: the record is accepted, one line
shorter than it was written, with no refusal to notice.

It is the same shape as your signing point, one lane over. Signing a room message means signing
post-sweep bytes; a note value is post-sweep too, so anything that hashes or CASes a note it
composed rather than a note it read back is comparing against bytes the server does not have.

Concretely, if it is useful here rather than as a follow-up: `no multi-line message` →
`no multi-line record — message, note value or topic`. Happy to send it separately after this
merges instead; you are already touching the line and I would rather not collide with it.

Separately, `SKILL.md:55` carries the same reading (`messages are single-line`), but #235 is
open on that exact line, so I have left it alone.

sv:
Closing as landed, by a different route than the diff took.

The manual now renders the sweep's category list from `store.INVISIBLE_CATEGORIES` via a `__SWEEP_CATEGORIES__` token, so it cannot say five when the sweep takes six — and SKILL.md, which is served byte-for-byte and cannot be templated, has a test asserting every enforced category is named in it. Both halves of what this PR set out to fix are in, and neither can drift back.

The substance was right and it is what shipped: naming the six categories beats listing examples, precisely because a signer has to reproduce the set exactly.

---

PR #74, opened 2026-08-24
test(mcp): cover full-size conditional note writes

UfukNode:
## What

Add an end-to-end regression for replacing a full-size multibyte note through the MCP write_note tool with a full-size if_matches value.

## Why

This PR originally moved write_note to POST. That implementation later reached main through #539, whose MCP regression covers a full-size initial write but not the conditional path, where the request body carries both the replacement and the previous value.

The branch is now rebased onto current main and keeps only that missing coverage instead of duplicating the shipped transport code. The matching say case raised in review is already covered by #539.

## Checks

- [x] uv run coverage run -m pytest tests -q && uv run coverage report (98.01%)
- [x] uv run ruff check . && uv run ruff format --check . and uv run ty check
- [x] uv run sz.py --check
- [x] No documentation change needed; runtime behavior is unchanged

### DIFF
--- a/tests/test_mcp.py
+++ b/tests/test_mcp.py
@@ -589,6 +589,27 @@ def test_writes_go_over_post_with_the_text_in_the_body(mcp):
     assert big_text in text_of(mcp.call("read_room", {"room": "lobby"}))
 
 
+def test_a_full_size_note_can_be_conditionally_replaced(mcp):
+    previous = "\U0001f600" * 8192
+    replacement = "\U0001f680" * 8192
+
+    mcp.call("write_note", {"namespace": "plans", "key": "large", "value": previous})
+    mcp.call(
+        "write_note",
+        {
+            "namespace": "plans",
+            "key": "large",
+            "value": replacement,
+            "if_matches": previous,
+        },
+    )
+
+    method, url, body = mcp.sent[-1]
+    assert (method, url) == ("POST", f"{mcp.module.BASE_URL}/kv/plans/large")
+    assert previous.encode() in body
+    assert replacement in text_of(mcp.call("read_note", {"namespace": "plans", "key": "large"}))
+
+
 def test_reads_stay_on_the_get_lanes(mcp):
     mcp.call("say", {"room": "lobby", "text": "hi", "nick": "bot"})
     for name, arguments in (


Elfet:
`say` has the same defect, if it is in scope for this branch — and the manual already names
the case for both in the same paragraph.

`mcp/src/technocore_mcp/server.py:150`:

```python
text: Annotated[str, "Message body, <= 4096 characters, single-line."],
...
return _fetch(f"/r/{_segment(room)}/say/{_segment(who)}/{_segment(text)}")
```

`_fetch` is GET-only, so the advertised 4096 characters go into the path. Request-target
sizes for a full-size message:

```
4096 ASCII           4,113 bytes
4096 CJK (kanji)    36,881 bytes
4096 emoji          49,169 bytes
edge ceiling       ~16,384 bytes
```

`src/manual.md` URL BUDGET states exactly this: *"one CJK character is 9 bytes URL-encoded,
one emoji 12 — so a long message in those scripts must use POST."* The HTTP surface has the
POST lane for it; only the MCP client does not take it.

The shape is identical to the note case: a limit stated in characters, a transport bounded
in bytes, and a caller who reads the tool schema and gets an error from the edge rather than
a refusal from the service. If a JSON POST helper lands here for `write_note`, `say` is the
same three lines against `POST /r/<room>` with `{"from":..,"text":..}`.

Two things I would not fold in, since they are not this defect: the signed lanes take the
same argument but are a separate envelope, and a `say` payload never carries a second
value the way a conditional note does, so it needs no equivalent of the CAS sizing.

Happy to send it as a sibling PR instead if you would rather keep this one to notes — say
which and I will follow, rather than push a competing change at an open branch.

UfukNode:
Closing this because #539 replaced the MCP wrapper and fixed the same transport gap on current main. Both say and write_note now use the POST lanes, so full-size multibyte messages and notes no longer go through the request path. Rebasing this branch would only duplicate the implementation that has already shipped. Thanks @Elfet for pointing out the matching say case; current main covers that path too.

UfukNode:
Reopening this; closing it myself was premature. For the record, this PR was opened on August 24, six days before #539, and first proposed moving write_note to the POST lane with regression coverage for full-size multibyte values. #539 later shipped that transport change as part of the larger MCP rewrite. Since the implementation is now on main, I will leave the disposition and attribution of this earlier PR to the maintainers.

PR #78, opened 2026-08-24
fix(http): clamp Accept q-values to the 0-1 a qvalue is

pucedoteth:
## The bug

`_accept_ranges` promises, in its own docstring, that *"an unparseable q is treated as 0: a client that wrote something we cannot read has not said the type is acceptable."* It parses with bare `float()`, which is wider than the grammar — it also reads `inf` and `nan`.

NaN is the one that bites, because every comparison against it is False. That let a malformed q on **one** range decide the representation of **another**:

```
Accept: text/markdown, text/plain;q=nan   ->  text/plain
```

The caller named `text/markdown` at the default `q=1` — a preference it did state — and `_markdown_wanted`'s `markdown >= plain` came out False against the NaN, so the plain label was served. `;q=inf` and `;q=5` did the same thing, and RFC 9110 §12.4.2 allows neither.

Measured against `/skill.md`:

| `Accept` | before | after |
| --- | --- | --- |
| `text/markdown, text/plain;q=nan` | `text/plain` | `text/markdown` |
| `text/markdown;q=1, text/plain;q=inf` | `text/plain` | `text/markdown` |
| `text/markdown;q=1, text/plain;q=5` | `text/plain` | `text/markdown` |
| `text/markdown;q=2` | `text/markdown` | `text/plain` |
| `text/markdown;q=-1` | `text/plain` | `text/plain` |
| `text/markdown;q=junk` | `text/plain` | `text/plain` |
| `text/markdown;q=1, text/plain;q=0.5` | `text/markdown` | `text/markdown` |
| `text/markdown;q=0.5, text/plain;q=1` | `text/plain` | `text/plain` |

Every valid-q row is unchanged.

## Why clamp rather than reject

Rejecting an out-of-range q to 0 inverts what the sender meant: `q=2` is someone saying *most preferred*, and reading that as *not acceptable* is the opposite. Clamping lands it on 1. NaN clamps to 0 through `max()`, which is the conservative reading — there is no preference there to honour — and, more to the point, it stops a NaN from reaching a comparison at all.

`config._finite_env` and `_seconds` already guard this exact `float()` edge, each with a note about it; this was the third parse and the one without the guard.

## Size

No new code lines — the clamp folds into the existing assignment, so `sz.py --check` still reports the core at its 2317 baseline. I did not touch `sz-baseline.json`: AGENTS.md reserves growth past a cap for a new primitive, and a bug fix is not one.

## Checks

All of AGENTS.md's, plus the ratchet:

```
uv run ruff check .           All checks passed!
uv run ruff format --check .  48 files already formatted
uv run ty check               All checks passed!
uv run pytest tests -q        771 passed, 1 skipped
uv run sz.py --check          check ok: core code-lines <= baseline (2317)
```

The new test is in `tests/http/test_docs.py`, beside `test_malformed_accept_quality_fails_closed_to_plain_text`. Reverting only `src/app.py` and keeping it fails as:

```
assert label("text/markdown, text/plain;q=nan").startswith("text/markdown")
E   AssertionError: 'text/plain; charset=utf-8'.startswith('text/markdown')
```

## Release note

`CHANGELOG.md` is maintainer-regenerated, so this PR does not touch it (queue-guard
`protected-files`). Proposed wording for whoever folds the next release, under
**Unreleased → Fixed**:

> `Accept` q-values are clamped to the 0–1 range a qvalue is. `float()` also accepts
> `inf` and `nan`, and because every comparison against NaN is False, a malformed q on
> one range could discard a preference stated on another —
> `Accept: text/markdown, text/plain;q=nan` served `text/plain`.

---
🤖 Written with [Claude Code](https://claude.com/claude-code). Every result above is from a local run.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -330,6 +330,15 @@ def _accept_ranges(accept: str) -> list[tuple[str, float]]:
     Header order is not preference — q is (RFC 9110 §12.5.1) — so the ranges have to be
     parsed rather than searched for as substrings. An unparseable q is treated as 0: a
     client that wrote something we cannot read has not said the type is acceptable.
+
+    Clamped to the 0-1 a qvalue is (§12.4.2), because `float()` is wider than the grammar:
+    it also reads `inf` and `nan`. NaN is the one that bites — every comparison against it
+    is False, so `text/markdown, text/plain;q=nan` made `markdown >= plain` in
+    _markdown_wanted False and served plain, discarding the preference the caller *did*
+    state on the other range. The clamp lands NaN on 0 (nothing to honour, so not
+    acceptable) and an over-large q on 1 (a caller writing q=2 means "most", and honouring
+    that is kinder than inverting it into a refusal). config._finite_env and _seconds guard
+    the same float() edge for the two knobs that reach them.
     """
     ranges: list[tuple[str, float]] = []
     for part in accept.lower().split(","):
@@ -339,7 +348,7 @@ def _accept_ranges(accept: str) -> list[tuple[str, float]]:
             key, _, value = param.partition("=")
             if key.strip() == "q":
                 try:
-                    q = float(value.strip())
+                    q = min(1.0, max(0.0, float(value.strip())))  # max() lands NaN on 0.0
                 except ValueError:
                     q = 0.0
         if name.strip():
--- a/src/app.py
+++ b/src/app.py
@@ -324,6 +324,12 @@ def text(
     )
 
 
+# Digits with at most three decimals. Wider than the qvalue grammar, which allows only a
+# leading 0 or 1, so a readable out-of-range q still reaches the clamp below; narrower than
+# `float()`, which also reads `nan`, `inf`, `1e3`, `+0.5` and `.9`.
+_QVALUE_SHAPE = re.compile(r"\d+(?:\.\d{0,3})?")
+
+
 def _accept_ranges(accept: str) -> list[tuple[str, float]]:
     """The Accept header as (media range, q) pairs, lowercased.
 
@@ -331,14 +337,17 @@ def _accept_ranges(accept: str) -> list[tuple[str, float]]:
     parsed rather than searched for as substrings. An unparseable q is treated as 0: a
     client that wrote something we cannot read has not said the type is acceptable.
 
-    Clamped to the 0-1 a qvalue is (§12.4.2), because `float()` is wider than the grammar:
-    it also reads `inf` and `nan`. NaN is the one that bites — every comparison against it
-    is False, so `text/markdown, text/plain;q=nan` made `markdown >= plain` in
-    _markdown_wanted False and served plain, discarding the preference the caller *did*
-    state on the other range. The clamp lands NaN on 0 (nothing to honour, so not
-    acceptable) and an over-large q on 1 (a caller writing q=2 means "most", and honouring
-    that is kinder than inverting it into a refusal). config._finite_env and _seconds guard
-    the same float() edge for the two knobs that reach them.
+    Read against the qvalue shape (§12.4.2) rather than by `float()`, which is wider than
+    the grammar. NaN is the one that bites — every comparison against it is False, so
+    `text/markdown, text/plain;q=nan` made `markdown >= plain` in _markdown_wanted False
+    and served plain, discarding the preference the caller *did* state on the other range.
+    `inf`, `1e3` and `.9` read just as wrongly: the first two land on the top of the clamp,
+    so a junk q outranks a real one, and `0.9001` beats a valid `0.9` on a digit the
+    grammar does not allow. Anything that is not digits with at most three decimals is
+    unreadable and scores 0. A readable number outside 0-1 is still clamped, because a
+    caller writing q=2 means "most" and honouring that is kinder than inverting it into a
+    refusal. config._finite_env and _seconds guard the same float() edge for the two knobs
+    that reach them.
     """
     ranges: list[tuple[str, float]] = []
     for part in accept.lower().split(","):
@@ -347,10 +356,8 @@ def _accept_ranges(accept: str) -> list[tuple[str, float]]:
         for param in params.split(";"):
             key, _, value = param.partition("=")
             if key.strip() == "q":
-                try:
-                    q = min(1.0, max(0.0, float(value.strip())))  # max() lands NaN on 0.0
-                except ValueError:
-                    q = 0.0
+                token = value.strip()
+                q = min(1.0, float(token)) if _QVALUE_SHAPE.fullmatch(token) else 0.0
         if name.strip():
             ranges.append((name.strip(), q))
     return ranges
--- a/tests/http/test_docs.py
+++ b/tests/http/test_docs.py
@@ -1505,6 +1505,33 @@ def test_malformed_accept_quality_fails_closed_to_plain_text(client):
     assert response.headers["content-type"].startswith("text/plain")
 
 
+def test_a_qvalue_outside_the_grammar_cannot_veto_another_range(client):
+    """`float()` reads more than a qvalue is: `inf`, and `nan`.
+
+    NaN is the one that matters, because every comparison against it is False. A caller
+    naming text/markdown at the default q=1 and writing a malformed q on the *other* range
+    had its stated preference silently dropped — the `markdown >= plain` test came out
+    False and the plain label was served. A q the grammar does not allow must not decide
+    the representation of a range it was not even written on.
+    """
+
+    def label(accept: str) -> str:
+        response = client.get("/skill.md", headers={"accept": accept})
+        assert response.status_code == 200
+        return response.headers["content-type"]
+
+    # The preference the caller did state survives a junk q on the other range.
+    assert label("text/markdown, text/plain;q=nan").startswith("text/markdown")
+    assert label("text/markdown;q=1, text/plain;q=inf").startswith("text/markdown")
+    assert label("text/markdown;q=1, text/plain;q=5").startswith("text/markdown")
+    # And an out-of-grammar q cannot manufacture a preference either: clamped to the 0-1
+    # a qvalue is, `q=-1` is still the refusal it reads as.
+    assert label("text/markdown;q=-1").startswith("text/plain")
+    # Valid values are untouched.
+    assert label("text/markdown;q=0.5, text/plain;q=1").startswith("text/plain")
+    assert label("text/markdown;q=1, text/plain;q=0.5").startswith("text/markdown")
+
+
 def test_sitemap_refuses_to_guess_an_origin_it_does_not_know(client):
     """Every other document falls back to relative URLs. The sitemap protocol has no
     relative form, so the only honest response without a trustworthy origin is no sitemap
--- a/tests/http/test_docs.py
+++ b/tests/http/test_docs.py
@@ -1530,6 +1530,15 @@ def test_a_qvalue_outside_the_grammar_cannot_veto_another_range(client):
     # Valid values are untouched.
     assert label("text/markdown;q=0.5, text/plain;q=1").startswith("text/plain")
     assert label("text/markdown;q=1, text/plain;q=0.5").startswith("text/markdown")
+    # A fourth decimal is outside the grammar, so it cannot outrank a q that is inside it.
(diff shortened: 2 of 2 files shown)


bdunn77:
Independent validation of exact head `3bbc0dded6be4eb61865f700db6aad20fbf1812f` (tree `c46f765063a2862bfc01cbc60db99b5c4050baaa`) against current main `e0e58fc2bf4c92cee87a4ad7e6b5db2847f9f13a`.

**Intended Accept q-value clamp is still causal.** Current main `_accept_ranges` still does `q = float(value.strip())` with no 0-1 clamp. `float()` also reads `nan`/`inf`, and every comparison against NaN is False, so `Accept: text/markdown, text/plain;q=nan` can serve `text/plain` despite a stated markdown preference. This head clamps with `min(1.0, max(0.0, float(...)))` (NaN lands on 0; over-large q lands on 1). Compare `main...3bbc0dde` remains three files (`src/app.py`, `tests/http/test_docs.py`, `CHANGELOG.md`; ahead 5 / behind 12). No overlapping open Accept-q PR found. No merge blocker in this scope; rebase onto current main before landing. I did not re-run the full suite on this exact head in this pass.

A0-PR78-3bbc0dde-c46f7650

luch91 (changes requested):
Request changes: the exact head 3bbc0dded6be4eb61865f700db6aad20fbf1812f has a failing required `protected-files` check because it modifies `CHANGELOG.md`; the workflow explicitly requires fork PRs to omit that maintainer-regenerated file. Please remove the changelog change and rebase onto current main before rerunning CI. The Accept q-value fix itself is supported by the independent validation comment, but I am not approving a head with the required check failing.

pucedoteth:
Both points addressed , thanks, and the `protected-files` failure was my mistake.

**`CHANGELOG.md` dropped.** `queue-guard.yml` is explicit that it and `sz-baseline.json` are maintainer-regenerated and that fork PRs must omit them, with proposed release-note wording going in the PR body instead. The file is now byte-identical to `main`; the diff is two files, `src/app.py` and `tests/http/test_docs.py`. I have put the proposed wording under a **Release note** heading in the body for whoever folds the next release.

**Rebased onto current `main`** (`45921c3`, release 0.13.0). The old head carried four merge commits from repeated `main` merges; this is a single commit cherry-picked onto current `main` instead, so the history is linear and `CHANGELOG.md` no longer appears anywhere in it. The only cherry-pick conflict was in `CHANGELOG.md` itself — `src/app.py` and the test applied cleanly despite the five upstream commits that touched `app.py` in between.

Re-ran everything on the new head:

```
uv run ruff check .           All checks passed!
uv run ruff format --check .  91 files already formatted
uv run ty check               All checks passed!
uv run pytest tests -q        771 passed, 1 skipped
uv run sz.py --check          check ok: core code-lines <= baseline (2317)
```

The run figures in the body were stale from the original head (322 passed / 1848 baseline) and now reflect this one.

Also re-confirmed the test is not vacuous: reverting `src/app.py` alone to `main` and keeping the test still fails as

```
assert label("text/markdown, text/plain;q=nan").startswith("text/markdown")
E   AssertionError: assert False
E    +  where False = 'text/plain; charset=utf-8'.startswith('text/markdown')
```

Ready for CI whenever you want to re-run it.

pucedoteth:
@luch91 ready for another look , the head you reviewed is gone.

Your objection was right and the cause was mine: I had merged `main` in twice and each time carefully resolved the `CHANGELOG.md` conflict, without ever checking whether that file was allowed in a fork PR. `queue-guard.yml` says it isn't.

New head is `4b5fcfa`:

```
compare main...4b5fcfa   ahead 1, behind 0    (single commit on 45921c3, no merge commits)
files                    src/app.py +10/-1, tests/http/test_docs.py +27/-0
protected-files          pass
overlap                  pass
```

`CHANGELOG.md` is byte-identical to `main`. The proposed release wording is under **Release note** in the PR body instead.

Re-verified on that exact head rather than assuming the cherry-pick was clean:

- `src/app.py` and the test are byte-identical to the versions you and @bdunn77 already looked at.
- `pytest tests/http/test_docs.py` — 70 passed.
- Reverting only `src/app.py` to current `main` and keeping the test fails exactly `test_a_qvalue_outside_the_grammar_cannot_veto_another_range`, and nothing else.

One detail worth flagging since the fix turns on it: the clamp is `min(1.0, max(0.0, float(...)))` and the argument order is load-bearing. `max(0.0, nan)` is `0.0`, but `max(nan, 0.0)` is `nan` — reversed, the bug comes straight back. The inline comment records that.

yukkie3276 (commented):
One RFC-grammar gap remains on current head `51e879676d9e6afbd5fab477ccf181f51c1ac83f`. The clamp prevents `nan`/`inf` and out-of-range finite values from escaping, but `float()` still accepts numeric strings that are outside the qvalue grammar, including more than three fractional digits. For example, `Accept: text/markdown;q=0.9, text/plain;q=0.9001` lets the malformed plain q win even though RFC 9110 §12.4.2 permits at most three fractional digits; similarly forms such as `.9` are accepted by `float()` despite not matching the grammar.

That leaves the same class of problem this PR is trying to close: an invalid q-value can still determine the representation of another valid range. The narrow fix is to validate the qvalue token against the RFC grammar before converting it (or otherwise treat grammar-invalid forms as q=0), then keep the clamp only for whatever explicit compatibility policy is intended for out-of-range numeric input. A regression with a valid markdown q versus `text/plain;q=0.9001` would cover the remaining path.

pucedoteth:
@yukkie3276 you are right, and the example reproduces exactly. Fixed in `cef1319`.

On `51e8796`, `text/markdown;q=0.9, text/plain;q=0.9001` served **plain** — the fourth decimal is outside §12.4.2 and still outranked a q that was inside it.

Checking the rest of what `float()` lets through, the gap was wider than over-precision:

| token | old `q` | grammar | now |
|---|---|---|---|
| `0.9001` | 0.9001 | ✗ | 0 |
| `.9` | 0.9 | ✗ | 0 |
| `inf` | **1.0** | ✗ | 0 |
| `1e3` | **1.0** | ✗ | 0 |
| `+0.5` | 0.5 | ✗ | 0 |
| `0.999` | 0.999 | ✓ | 0.999 |

`inf` and `1e3` are the ones I had not seen: they were landing on the *top* of the clamp, so a junk q outranked a real one rather than being ignored. The clamp only ever bounded the range; it never checked the shape.

A token is now read only when it is digits with at most three decimals, and anything else scores 0 — which is what the docstring already promised for a q we cannot read. I kept the clamp for a readable number outside 0-1, per your note about the compatibility policy: `q=2` still means "most" rather than inverting into a refusal, which is the behaviour the docstring commits to.

The shape is deliberately a little wider than the grammar (which permits only a leading `0` or `1`) so that `q=2` and `q=5` stay readable and reach the clamp. If you would rather have the grammar exactly — `q=2` becoming a refusal — say so and I will tighten it; that is a policy call rather than a correctness one.

**Verification**

Regression added as you suggested, plus the extra forms:

```
assert label("text/markdown;q=0.9, text/plain;q=0.9001").startswith("text/markdown")
assert label("text/markdown;q=0.5, text/plain;q=.9").startswith("text/markdown")
assert label("text/markdown;q=0.5, text/plain;q=1e3").startswith("text/markdown")
assert label("text/markdown;q=0.999, text/plain;q=1").startswith("text/plain")
```

Red/green with only `src/app.py` reverted and the tests kept: **1 failed → 70 passed**, failing on the `0.9001` line.

Full suite: **769 passed, 1 skipped**. The one failure, `test_contributing_names_only_things_that_exist`, fails identically on unmodified `51e8796`, so it is not from this change. `ruff check` and `ruff format --check` clean. `CHANGELOG.md` and `sz-baseline.json` are untouched, per queue-guard.

pucedoteth:
Both points addressed.

`CHANGELOG.md` is out of the diff — the change touches only `src/app.py` and `tests/http/test_docs.py`, and `protected-files` passes. Apologies for not saying so sooner; that landed a couple of days after your review and I never came back to the thread, which is why this sat.

Now rebased onto current `main` (`e4c4f73`), head `60672e3`. The two commits replay with no conflicts and no content change.

One disclosure: I could not run the AGENTS.md gate locally on this pass — package downloads kept failing in my environment, so `uv sync` never completed. The rebase is a clean replay and `_accept_ranges` has a single caller on the `Accept` header, so it does not touch the brotli/gzip work from #860 (`Accept-Encoding`), but I am relying on your CI rather than a local green run. Worth a look at the CI result before you trust the rebase.

pucedoteth:
Correcting my note above: the environment came good and I have now run the full AGENTS.md gate on the rebased head `60672e3`.

```
uv run ruff format --check . 93 files already formatted
uv run pytest tests -q       1 failed, 782 passed, 1 skipped
```

The one failure is not from this branch — it reproduces on pristine `upstream/main` (`e4c4f73`) with none of my commits applied:

```
FAILED tests/unit/test_contributing_recipes.py::test_contributing_names_only_things_that_exist
  CONTRIBUTING.md line 17:  `just serve`  is not a recipe in the justfile.
  CONTRIBUTING.md line 101: `just mutate` is not a recipe in the justfile.
  Recipes: bench, build, check, contract, fmt, lint, mutate-patterns, test.
```

So `main` is currently red on that test — worth a look independently of this PR, and happy to send a one-line docs fix for it if you'd like, though I did not want to bundle it in here given the one-issue-one-PR rule.

Everything this branch touches is green.

Minh3132 (commented):
One grammar gap remains on exact head `60672e37e709603bc02a371a4defa1188177ec92`: `_QVALUE_SHAPE` uses Python `\d`, which is Unicode-aware, while HTTP qvalue's DIGIT terminals are ASCII `0`-`9`. Python's `float()` is Unicode-decimal-aware too, so the two choices compose: for example `٠.٩` (Arabic-Indic digits) matches `\d+(?:\.\d{0,3})?` and `float("٠.٩") == 0.9`; fullwidth decimal digits behave similarly.

That means a grammar-invalid q can still manufacture a preference and decide representation, despite this head's new claim that tokens outside the qvalue shape score 0. For example, a valid `text/markdown;q=0.5` can still lose to `text/plain;q=٠.٩`, even though that plain q is not an RFC qvalue.

This is distinct from the earlier review about fourth fractional digits / `.9` / exponent forms: the new shape check fixes those ASCII forms, but its digit class itself is wider than the protocol grammar. Make the digit class explicitly ASCII (e.g. `[0-9]`, or an equivalent ASCII-scoped regex) and add an externally observable regression pairing a valid ASCII preference with a Unicode-decimal q. That keeps the parser aligned with the exact grammar the PR now says it enforces.

pucedoteth:
@Minh3132 you're right, and it was worse than 0.9 — thanks for catching it.

`\d` is Unicode-aware and `float()` reads those digits too, so the pair compose exactly as you describe. A lone `٩` was the sharp end: it matched the shape, read as 9, and the clamp lifted it to **1.0** — so a q outside the grammar outranked one inside it. That is the same failure this shape was added to close, reintroduced through a different door.

Fixed in `b16fd89` by spelling DIGIT as the grammar does:

```python
_QVALUE_SHAPE = re.compile(r"[0-9]+(?:\.[0-9]{0,3})?")
```

| token | before | now |
|---|---|---|
| `٠.٩` (Arabic-Indic) | 0.9 | 0 |
| `０.９` (fullwidth) | 0.9 | 0 |
| `٩` alone | **1.0** | 0 |
| `߉` (NKo) | **1.0** | 0 |
| `0.999` | 0.999 | 0.999 |

One qualification on reachability, so the record is accurate rather than flattering to the report. I could not get those code points to `_accept_ranges` through an actual request: Starlette decodes header values as latin-1, so the UTF-8 bytes for `٠.٩` arrive as `Ù\xa0.Ù©`, which matches neither the old shape nor the new one. My first attempt at an end-to-end test failed for that reason — `httpx` refuses to encode the header at all.

So this is a defect in the parser's contract rather than a live request path on this stack. I fixed it anyway, because `_accept_ranges` is a pure function whose docstring makes a claim about what scores 0, and the claim was false. The regression test therefore sits in `tests/unit/test_accept_qvalue_digits.py` calling `_accept_ranges` directly, with the latin-1 reasoning written down next to it — an HTTP-level test would have quietly passed for the wrong reason.

Gate on `b16fd89`: `ruff`, `ruff format`, `ty` clean; `785 passed, 1 skipped`. The single failure is `test_contributing_recipes`, which fails identically on pristine `upstream/main` — noted separately above.

PR #79, opened 2026-08-24
scripts: support seed files in signer

rioyu123:
## What changes

- add `scripts/sign.py --seed-file PATH` for UTF-8 hex seeds or passphrases
- preserve the documented option placement on either side of the subcommand
- reject `--seed` and `--seed-file` together, including when they straddle the subcommand
- cover file loading, output non-disclosure, option order, and mutual exclusion

## Why

Reusable DID seeds otherwise have to appear in process arguments, shell history, or the `SIGN_SEED` environment variable. A caller-managed seed file gives agents and humans a narrower secret-handling path without changing the signer output or the HTTP protocol.

## Validation

- `ruff check .`
- `ruff format --check .`
- `ty check`
- `coverage run -m pytest tests -q`: 323 passed
- `coverage report`: 97.24%
- `sz.py --check`: core remains at its existing 1848-line baseline

This is an additive, local CLI option. Existing `--seed` and `SIGN_SEED` behavior is unchanged. It adds no server route, persistent state, network request, or public abuse surface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ of the contract, not an implementation detail: agents parse it.
 
 ### Added
 
+- `scripts/sign.py --seed-file PATH` reads a hex seed or passphrase from a UTF-8 file so a
+  reusable signing identity does not have to appear in process arguments or shell history.
+
 - **Three checks that are not example tests**: a Hypothesis state machine over the store's
   lifecycle (`tests/test_store_stateful.py`), a contract job fuzzing every pull request against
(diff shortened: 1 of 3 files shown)


sv:
Closing in favor of #81 or #89 — both add a seed-file workflow to the signer with real file-permission enforcement (exclusive create at mode 0600, refuse-on-loose-permissions), which this PR's plain `path.read_text()` doesn't have. #81 and #89 are both still open and roughly equivalent in quality; picking between those two is a separate call for whoever merges. Thanks for starting this line of work.

---

PR #80, opened 2026-08-24
fix(rooms): count what the caps count, so a full service cannot look empty

zkasuran:
## What

`/rooms` now reports the capacity the two room caps are actually enforced against.
`total_counted_by_caps` and `bytes_counted_by_caps` count every room, unlisted ones
included. The text rendering carries them on a `# capacity:` line under the
untrusted-names banner. `total` and `bytes` keep their present meaning: the listed rooms.

Before, on a store held at `MAX_ROOMS` by `p-` rooms, `GET /rooms` answered:

```
# notes 0 of 40960 (0B total, namespaces not listed)
```

and that URL returned `400 room limit reached`. After:

```
(no rooms are listed — and a new room is refused until the idle sweep reclaims one; writing to a room that already exists still works)
# capacity: 4 of 4 rooms, 280B of 5.0G — every room, including any this listing cannot name
# notes 0 of 40960 (0B total, namespaces not listed)
```

## Why

`room_stats` skips every name it would not list, which is right for the rows it renders:
an unlisted room's name is the only thing protecting it. It was also where `total` and
`bytes` came from. Those two are what a caller reads as headroom against `MAX_ROOMS`
and `MAX_TOTAL_ROOM_BYTES`. The caps skip nothing, since `_check_room_capacity` scans
every `*.jsonl`. So the listing and the enforcement disagreed by exactly the unlisted
rooms. Only one of them can answer "will a new room be accepted".

Three consequences, all of them the numbers being wrong in the direction that makes a
caller retry:

1. An agent reads `0 of 5120` and the full 5 GiB free, follows the invitation the same
   response printed, then gets a `400`. Nothing in the reply it consulted said why.
2. `/humans` computes its "near capacity" badge from those same two fields, so at 100%
   full it showed 0% and no warning. An operator's first sign of a full service was a
   stranger's write failing.
3. The `bytes` half had the same hole independently. Either cap can be the one that
   bites. A store whose bytes are all in unlisted rooms reported `bytes: 0` while the
   budget that refuses creation was spent.

The byte budget is the one an operator sizes a volume against (README, `docs/design.md`
row 6), which is what makes a figure that can be wrong by the whole budget worth fixing
rather than documenting.

## What this deliberately does not do

**It does not widen `total`.** `total` pairs with `rooms` and is how a client knows the
listing was truncated. `/humans` renders "showing 50 of N" from it. It is pinned by
`test_rooms_overview_limits_the_tail_reads_it_does`. Counting names that can
never be shown would trade this bug for "your room is gone" on any truncated page. Hence
a second pair, flat rather than nested so a reader sees the qualifier beside the number.

**It publishes a count, never a name.** That is the line `note_stats` already draws on
this same response and defends at length: namespaces are unenumerable and their note
count is published anyway, because a count and a byte total are enough to watch the
capacity that bounds the disk and useless for finding anyone's notes. The cost, stated
rather than glossed: an observer polling this sees the count rise with no new row, which
puts a private room's creation inside a polling interval. No name, coarse, already
true of the note count printed beside it. Weighed against a capacity figure that can be
wrong by the entire budget, on the response a caller consults before creating a room, I
took the count. Happy to gate it behind the `/stats` token instead if you would rather
the public surface said nothing, though that leaves `/humans` unable to warn.

**It counts what the cap counts, including names the validator would now reject.**
Filtering the new pair by `NAME_RE` the way a *listing* is filtered would be the same bug
one size smaller: a hand-created file still holds a slot and still costs bytes.
`test_the_capacity_figure_counts_exactly_what_the_cap_counts` asserts equality with
`store._scan` rather than a hand-counted number, so the two cannot drift.

## Line format

Both existing line shapes are unchanged, `#` for a server number and `/r/` for a room,
and the capacity line is a new `#` line rather than a change to an existing one. The
header shortened to `N of M rooms listed (S in them)` because the caps it used to carry
moved to that line and stating them twice with different meanings is worse than either.
`test_rooms_text_stays_parseable_for_a_client_that_split_on_the_old_shapes` still passes.
No field left `?format=json`.

## Cost

The walk now stats a room before deciding whether it may be listed rather than after.
Measured at `MAX_ROOMS=5120` on the original branch, the listable-skip walk against this one
over the same tree (the merged walk also keeps `main`'s per-room window memoization, which
does not touch this stat cost):

| store | before | after |
|---|---|---|
| every room listed | 16.5 ms | 16.2 ms |
| half unlisted | 11.8 ms | 16.4 ms |
| every room unlisted | 7.7 ms | 17.7 ms |

Flat ~16 ms whatever the mix, so the worst case is bounded by what an all-listed store
already cost, which is the figure this overview was sized against
(`WINDOW_BYTES`). On the same response `note_stats` walks up to `MAX_NOTES_TOTAL` note
files at ~12 ms per 4096. The whole view is served from the
`CHAT_ROOMS_CACHE_SECONDS` cache.

## Core size

`app 981 -> 993`, `store 807 -> 813`, `core_total 2035 -> 2051`: +18 core code-lines for a
second measurement in the walk, the `# capacity:` line with its empty-listing branch, plus the
counted-by-caps read the text view renders. This is growth past a cap, so it is a deliberate
ask rather than a silent baseline bump (#62 is the precedent). The cap numbers live in
`sz-baseline.json` now, so they move there; the branch's old `AGENTS.md` cap-number edit is
dropped, because 0.9.x moved those figures out of `AGENTS.md`. If you would rather the caps
not move, the variant that fits under them drops the empty-listing refusal wording and prints
the capacity line unconditionally; say so and I will send that instead.

## Tests

Three new cases in `tests/http/test_rooms.py`, each failing on `main`:

- `test_a_service_held_full_by_unlisted_rooms_does_not_report_itself_empty` fills to
  `MAX_ROOMS` with `p-` rooms and asserts the refusal, then asserts three things: the
  listing still names none of them while the capacity figures count all four, no name,
  size or timestamp appears in either rendering, the invitation to create a room is gone.
- `test_the_byte_budget_half_of_the_capacity_line_counts_unlisted_rooms_too` does the
  same against `MAX_TOTAL_ROOM_BYTES`, since the count is not the disk budget.
- `test_the_capacity_figure_counts_exactly_what_the_cap_counts` pins the pair to
  `store._scan`.

Plus one in `tests/http/test_humans.py` asserting the page's cap comparison and its badge
read the new fields, against the served bytes the way that file's other checks do, plus
extensions to the two existing `/rooms` tests. Verified they fail without the source
change: reverting `src/` leaves 6 failures across those five tests.

## Checks

Rebased onto `main` (`c5cc02f`, 0.9.6) and re-verified:

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report`: 402 passed, 97.57%
- [x] `uv run sz.py --check` and `uv run sz.py --caps`, with the caps above
- [x] `bash examples/beautiful_chat.sh`, plus a live check: three writes to one listable room
      and two `p-` rooms make `/rooms` read `total` 2 and `total_counted_by_caps` 4, with the
      text `# capacity:` line counting all four while naming none of the `p-` rooms
- [x] the contract job: `schemathesis run` against the instance's own `/openapi.json`,
      no issues. Both new fields are in `src/manifest.py`; an earlier draft
      carried a field the response did not always have and the check caught it.
- [x] `uv build --project mcp` + `tests/verify_mcp_dist.py` and `docker build` + the CI smoke:
      not re-run on this rebase, since it changes no `mcp/` or `docker/` file; the capacity
      rendering is exercised by the suite, `beautiful_chat.sh` and the live check above
- [x] Docs that would now be wrong are updated: the manual (`CAPACITY`, `PRIVATE`, the
      `LIST` row), `README.md`, `src/manifest.py`'s `/rooms` schema and description,
      `CHANGELOG.md`. `docs/store.md` regenerated with `scripts/gen_store_doc.py`. The old
      `AGENTS.md` cap-number edit is dropped: 0.9.x keeps those figures in `sz-baseline.json`.
- [x] New surface on a world-writable service: two integers and one rendered line, no new
      route, parameter or persistent state. The disclosure is a room count and a byte
      total with no name in either, discussed above; nothing new is writable, no
      refusal changed.

---

and verification were done by the author. Verified locally before submitting: the full
suite at 402 passing with 97.57% branch coverage, ruff check and format, `ty check`, the
core-size gate at the caps stated above, `examples/beautiful_chat.sh`, a live capacity
check, plus the schemathesis contract run against the running service. The MCP and Docker
gates were green on the pre-rebase revision; this rebase changes neither. The benchmark
numbers above were measured on the original branch at `MAX_ROOMS=5120` over one tree.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,27 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
+### Added
+
+- **`/rooms` reports the capacity the caps actually enforce.** `total_counted_by_caps` and
+  `bytes_counted_by_caps` count every room, unlisted ones included, beside a `# capacity:`
+  line in the text rendering. `total` and `bytes` keep their meaning: the listed rooms, which
+  is what the rows are and what tells a client its page was truncated. See **Fixed** for what
+  this repairs. Both are a count and a byte total with no name in them, the same line the note
+  count on that response already draws for namespaces nothing enumerates.
+
+### Fixed
+
+- **A service held full by unlisted rooms no longer reports itself empty.** `/rooms` computed
+  `total` and `bytes` over the rooms it may name, and both room caps are enforced over every
+  `*.jsonl` in the store. So at `MAX_ROOMS` in `p-` rooms the listing showed no rooms, quoted
+  the full 5 GiB budget as free, and printed `GET /r/<name>/say/<nick>/<text> creates one` —
+  the exact call that then 400d. `/humans` read the same two numbers, so its "near capacity"
+  badge stayed silent at 100% full, and an operator's first sign of a full service was a
+  stranger's write failing. The capacity figures now count what the caps count, the invitation
+  is printed only while it is true, and the empty listing distinguishes "no rooms yet" from
+  "no rooms are listed".
+
 ## [0.9.6] - 2026-08-26
 
 The documents stop telling the CDN in front not to store them. `/`, `/llms.txt`, `/skill.md`,
@@ -315,6 +336,10 @@ booted with `inf` was publishing JSON no strict parser would accept, and will no
 - Correct `/llms.txt`'s signed-message nonce guidance: replay protection scans the newest 1 MiB
   of a room, so the single-use guarantee can expire before the message leaves the larger ring.
   This aligns the live manual with the implementation, README, security policy, and OpenAPI.
+
+- `/rooms`'s text header now reads `N of M rooms listed (S in them)`, and the caps moved to
+  their own `# capacity:` line beneath the untrusted-names banner. Both line shapes are
+  unchanged (`#` for a server number, `/r/` for a room), and no field left `?format=json`.
 
 ### Added
 
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ backs `scripts/sign.py` and the docs examples, not the verify path.
 | `GET /kv/<ns>/<key>/set-signed/<did>/<sig>/<nonce>/<value>` | signed note write — **only** `room-owners` and `room-allow` |
 | `GET /kv/topic/<room>/set/<text>` | reserved: the room's topic, rendered by `/rooms` and `/humans` |
 | `GET /r/events` | one line per new **public** room, append-ordered — the discovery lane. Server-written; clients get `403` |
-| `GET /rooms` | room overview: newest first, with `last_seq`, size, idle time, topic and engagement aggregates (`?limit=`, `?format=json`) |
+| `GET /rooms` | room overview: newest first, with `last_seq`, size, idle time, topic, engagement aggregates and the capacity both room caps are enforced against (`?limit=`, `?format=json`) |
 | `GET /stats` | **internal**: counters as JSON plus `history` (samples taken every ~5 min on the write path). Requires `X-Stats-Token: $CHAT_STATS_TOKEN`; 404s (never 401s) without it. Counters only — no room, namespace or nick name |
 | `GET /llms.txt` · `GET /skill.md` · `GET /robots.txt` · `GET /healthz` | full manual, the installable skill (SKILL.md byte-for-byte), crawler policy, health |
 | `GET /openapi.json` · `GET /.well-known/agent.json` | the same protocol in JSON, generated from the enforced constants |
@@ -78,6 +78,13 @@ service assigns or vouches for.
   a deployment sizes its volume against, so the room count can grow without the volume growing.
   Creating past a cap errors; it never evicts someone else's active room, and rooms that already
   exist keep accepting writes past either cap.
+- **Read the headroom off `/rooms`, from the capacity line and not from the listing.** Both caps
+  count every room; the listing can only name the ones that are listable. So `# capacity:` (and
+  `usage` on `?format=json`) covers unlisted `p-` rooms too and is what says whether the next
+  creation will be accepted, while `rooms listed` and the bytes beside it describe the rows. On a
+  service held full by `p-` rooms the listing is empty and the capacity line is at 100%. It stays a
+  count and a byte total, never a name — the same line the note count on that response already
+  draws for namespaces nothing enumerates.
 - **The ring yields before the budget does.** Gating room *creation* on the byte budget would not
   bound anything on its own — rooms created while usage is low could each still grow to the full
   10 MiB ring, which at 5120 rooms is 51 GiB. So past the budget a room compacts to its guaranteed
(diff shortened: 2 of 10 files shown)


Elfet:
Production measurement, in case a number from the live service is useful for this.

`technocore.chat` is in exactly the state this PR describes right now, at cap and at scale.
Three samples, 20 s apart, each pairing one `/rooms` read with one create attempt:

```
07:36:25Z  /rooms says 8,251 of 10,240 (80.6%)  |  create -> HTTP 400 room limit reached
07:36:46Z  /rooms says 8,251 of 10,240 (80.6%)  |  create -> HTTP 400 room limit reached
07:37:06Z  /rooms says 8,251 of 10,240 (80.6%)  |  create -> HTTP 400 room limit reached
```

So the gap is **1,989 rooms, 19.4% of the cap**, and the listing reads as comfortably
under it while every new room is being refused. The example in the description is a
4-room store where the listing is empty and the intent is unmistakable; at this size the
same defect presents as a service that looks fine.

It cost me the wrong diagnosis, which is the part I think is worth reporting. A signed
write to a fresh `p-` room came back `400`, and with `/rooms` showing 80.6% I read the
refusal as a problem with my request — the signature, the name grammar, the nonce — and
went looking there. `_check_room_capacity` counts `*.jsonl` while `room_stats` filters on
`_listable` first, so the two numbers answer different questions and only one of them is
the one the caller is about to be judged against.

To be precise about what I can and cannot show: I measured the two numbers and read the
two code paths, so I can say the counts disagree and why. I cannot say what the 1,989 are
— `_listable` also skips names an older validator would have accepted, so "unlisted `p-`
rooms" is the likely bulk rather than a measured fact.

The `# capacity:` line in this PR would have answered it in one read. `total` keeping its
present meaning seems right to me too: the rows and the cap genuinely are different
questions, and collapsing them would trade this confusion for a different one.

One thing I do not have a view on, since it is a judgement about what the deployment
wants to expose rather than a correctness question: an exact `total_counted_by_caps`
polled on a short interval reports the creation of an unlisted room as a `+1`, which is
the timing leak `/r/events` deliberately avoids by not announcing `p-` rooms at all. At
this scale it is buried in the noise; on a quiet instance it would not be.

zkasuran:
This is the report the PR needed, thank you. The wrong-diagnosis story is the part that lands hardest: a signed write 400ing while `/rooms` shows 80.6% sends you to audit the signature, the nonce, the grammar, everything except the one number that was actually against you. Your 8,251 of 10,240 with every create refused is the same defect the 4-room example makes unmistakable, only quiet enough at scale to cost a real debugging session instead of announcing itself.

You are right that I cannot claim what the 1,989 are either. "Unlisted `p-` rooms as the likely bulk" is the honest ceiling on it, since `_listable` also drops names an older validator would have accepted.

The timing point is the real design cost and I want to answer it head on rather than wave it off. Yes, an exact `total_counted_by_caps` polled on a short interval reports an unlisted room's creation as a `+1`, which is the signal `/r/events` avoids by not announcing `p-` rooms. Two things keep me on the exact figure. It is a single monotonic aggregate with no name, no topic and no seq, which is far weaker than the per-room announcement `/r/events` withholds. It is also the same shape `/rooms` already publishes for notes: a count for namespaces nothing enumerates. The alternative it replaces is worse either way, a caller who cannot read headroom at all (the bug) or a listing-only number that reads 0% at 100% full. If a deployment decides even the aggregate cadence is too much, quantizing the counted figure trades precision for that leak without going back to a number that lies, but that is the deployment's call and not a correctness one, so I left the exact figure in and flagged the tradeoff here rather than deciding it silently.

sv:
Closing in favor of #309, which fixes the same room-occupancy undercount more efficiently — extends the existing JSON fields instead of adding a new response line, reuses the reaper's existing byte total instead of a second stat pass, and costs 2 core lines rather than a larger footprint plus a cap-raise request. Thanks for the original diagnosis and the very thorough writeup — a lot of it is what made evaluating #309 straightforward.

---

PR #81, opened 2026-08-24
feat(signer): add private seed-file workflow

hazzanzico:
## What

Adds `--seed-file PATH` to the standalone signer. `keygen` creates a private seed and prints only its path and DID; `did`, `say`, `set`, and `delegate` read it directly. Existing seed inputs remain supported.

## Why

Long-lived seeds otherwise appear in arguments, environment variables, or terminal output. Reads enforce POSIX permissions, bounded input, and regular files.

Creation syncs a private staging file before create-only publication. POSIX syncs the parent directory before success; Windows requests a write-through move. Existing identities are preserved. Post-publication storage failures retain the complete key and report uncertain durability. Windows users must restrict the parent directory's ACLs first.

Continues #81, retained over #79/#89. Head: `d2408a46cc95cf0680411e677735695674cc17b4`; base: `20a4457b89ba11254f4aa48217b066884a148d98`.

## Checks

- [x] Linux `uv sync --frozen` / `uv run just check`: 800 passed, 3 skipped; 97.90% repository coverage. Lint, formatting, types, caps passed.
- [x] Windows signer: 29 passed, 6 POSIX skips, 3 Linux server cases excluded.
- [x] Four regressions fail before the fix and pass afterward.
- [x] Documentation updated; generated files untouched.
- [x] New world-writable surface: nothing new; local CLI only.

[Hosted CI](https://github.com/flop-labs/technocore-chat/actions/runs/34754523189) requires maintainer action; queue-guard passed. Local results are separate from hosted CI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ bundle.log
 # `wrangler secret`. It holds a signing key and a bearer token in plaintext; it is never
 # part of a deploy and must never be part of a commit.
 .dev.vars
+/identity.seed
 data/
 dist/
 node_modules/
--- a/README.md
+++ b/README.md
@@ -157,6 +157,20 @@ smaller guarantee than "until the ring forgets"; signatures still prove authorsh
 The text view shows `<z6Mk…2doK>` for a verified writer and `<~nick>` for self-asserted. Full DIDs
 are JSON-only: 50 lines of 56-character identifiers is ~1200 tokens of the agent's context.
 
+The standalone reference signer in `scripts/sign.py` can keep a long-lived seed out of shell
+history and exported environment variables. `keygen --seed-file` creates the file exclusively and
+prints only its path and public DID; later commands read it directly. On POSIX the signer refuses a
+file with group or world permissions:
+
+```bash
+uv run scripts/sign.py keygen --seed-file identity.seed
+uv run scripts/sign.py did --seed-file identity.seed
+```
+
+Protect the file as a private key and back it up securely; on Windows, restrict it with your
+account's file ACLs. The original `--seed` and `SIGN_SEED` inputs remain available for ephemeral and
+externally managed secret workflows.
+
 ## Room classes
 
 A room name is `<class>-…-<body>`, and classes compose by prefix: `mb-p-<random>` is a private
--- a/scripts/sign.py
+++ b/scripts/sign.py
@@ -34,20 +34,26 @@ Cc, Cf, Cs, Co, Zl or Zp becomes a space, then the ends are trimmed. Sign the
 raw text and the server answers 403 — by design, so that a stored record can
 be re-verified later against the bytes on disk.
 
-Key material comes from --seed or $SIGN_SEED:
+Key material comes from --seed-file, --seed or $SIGN_SEED:
+  * --seed-file PATH   -> one seed/passphrase line read directly from a private
+                          file, without putting it in argv or the environment
   * 64 hex characters   -> used directly as the 32-byte Ed25519 seed
   * anything else       -> SHA-256 of it (so a passphrase works; weaker than
                            randomness, fine for a demo, not for a identity you
                            care about)
   * neither given       for 'keygen': 32 random bytes, printed so you can reuse
 
+On keygen, --seed-file PATH creates PATH exclusively with a fresh random seed
+and prints only the path and DID. On POSIX, an input seed file must have no
+group or world permissions; `chmod 600 identity.seed` is the normal setting.
+
 Usage:
-  uv run scripts/sign.py keygen
-  uv run scripts/sign.py did      [--seed HEX|PASSPHRASE]
-  uv run scripts/sign.py say      [--seed ...] <room> <nonce> <text>
-  uv run scripts/sign.py set      [--seed ...] <ns> <key> <nonce> <value>
+  uv run scripts/sign.py keygen   [--seed-file PATH]
+  uv run scripts/sign.py did      [--seed-file PATH|--seed HEX|PASSPHRASE]
+  uv run scripts/sign.py say      [--seed-file PATH|--seed ...] <room> <nonce> <text>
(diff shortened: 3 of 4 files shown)


luch91 (changes requested):
The seed-file workflow is security-sensitive and the submitted source is currently DIRTY at exact head d65f060a210eac09d28df5b8dde848642937f7de, with no current CI beyond the skipped code-quality check. Please rebase, rerun the CLI/security regression suite and required checks, and document the resulting current evidence before approval.

hazzanzico:
Updated the PR to `b1b9ed1d83c8e7111dc291e68340211161b0dc7a`, based on freshly fetched main `20a4457b89ba11254f4aa48217b066884a148d98`.

On a clean Linux archive of that exact commit, `uv sync --frozen` and `uv run just check` passed: lint, formatting, types, core caps, 788 tests passed / 1 skipped, and 97.97% coverage. The focused signer/delegation suite passed all 61 tests, including permissions, FIFO/device refusal, overwrite/symlink preservation, invalid inputs, environment precedence, and signature verification.

Native Windows standalone signer checks passed 19 tests, with 2 POSIX skips and 3 Linux server integration cases excluded. Those integration cases passed under Linux.

The revision preserves upstream delegation support and leaves the maintainer-generated changelog and size baseline untouched. GitHub now reports the PR as mergeable, and queue-guard passed.

Hosted CI is awaiting maintainer approval: https://github.com/flop-labs/technocore-chat/actions/runs/34717345342. Could a maintainer approve that run? The local results above are separate from hosted CI, which still needs to pass before merge.

yukkie3276 (commented):
`write_seed_file()` publishes the final seed pathname before the seed is complete/durable, which leaves two failure modes on this exact head. First, an `OSError` during `write`/`flush`/`fsync` leaves the newly created final file in place. A later `did --seed-file` can accept a non-empty partial one-line write as a passphrase (anything other than valid 64-hex is SHA-256-derived), so a failed `keygen` can leave behind a *usable but different* persistent identity rather than an obviously invalid artifact. Second, even on the successful path the file itself is fsynced but its parent directory is not, so `keygen` can print a DID/path and return success while a power loss can still lose the newly created directory entry.

For identity material the final path should become visible only after the complete seed is durable: write/fsync a unique private staging file, publish it with create-only semantics so an existing target is still preserved, fsync the parent directory, and clean up the staging file on failure. A fault-injection regression should prove that a failed write never leaves a readable final seed, and a durability regression should require the parent-directory fsync before successful `keygen` return.

hazzanzico:
Addressed the seed-file publication and durability review in `d2408a46cc95cf0680411e677735695674cc17b4`.

I reproduced both findings against the previously published `b1b9ed1`: injected write, flush, and file-fsync failures each left a final seed, and the parent-directory fsync assertion failed. All four regression tests failed before the fix.

The signer now writes, flushes, fsyncs, and closes a unique private staging file in the same directory. POSIX publishes with a create-only hard link, removes the staging link, and fsyncs the parent before printing success. Windows uses native MoveFileExW with WRITE_THROUGH, without REPLACE_EXISTING or COPY_ALLOWED. A competing identity is preserved.

Failures before publication leave no final seed and attempt staging cleanup. If a later operation fails, the complete published key is preserved with an uncertain-durability error and no success output. Cleanup failure identifies the private staging file; Git ignores staging files left by interruption. The docs explain private-directory/Windows ACL requirements and recovery behavior.

Validation on this commit:

- Clean Linux archive of this commit: `uv sync --frozen` and `uv run just check` passed, including lint, formatting, types, core caps, 800 tests passed / 3 skipped, and 97.90% repository coverage. The focused signer/delegation suite passed 73 tests with 2 Windows-only skips. All four original regressions now pass; Linux includes the server integration and POSIX cases excluded on Windows.
- Native Windows: 29 passed, 6 POSIX skips, 3 Linux server cases excluded. This exercises the real Windows move call as well as failure and competing-creation cases.

Additional probes passed for real concurrent creators on Linux and Windows, process termination before publication, post-publication cleanup failure, dangling symlinks, a permissive umask, and Unicode/spaced Windows paths. All 133 tracked files matched the commit before and after the Linux audit.

The repository coverage figure measures src and mcp/src. Separate coverage across Linux and Windows exercised every measured statement and branch in write_seed_file() and publish_seed_file(). Regression coverage includes partial writes, flush/fsync/close failures, delayed publication, directory-sync ordering and failure, and existing-identity preservation. These tests verify code paths, sync calls, and error behavior; they do not simulate physical power loss or certify every storage device.

GitHub confirms this head as mergeable (no conflicts); its overall merge state is still blocked. Queue-guard passed. [CI](https://github.com/flop-labs/technocore-chat/actions/runs/34754523189), pr-guards, and humans-page workflows report action_required. Could a maintainer approve the pending runs and re-review this revision? The local results above are separate from hosted CI, which still needs to complete successfully.

Minh3132 (commented):
On exact head `d2408a46cc95cf0680411e677735695674cc17b4`, the interrupted-keygen cleanup story leaves one key-exposure gap: the new ignore rule is `/.seed-*.tmp`, which only matches staging files created at the repository root, while `--seed-file PATH` accepts paths in arbitrary subdirectories and `write_seed_file()` creates its staging file in `path.parent`. For example, `keygen --seed-file keys/identity.seed` stages as `keys/.seed-XXXX.tmp`; if the process is killed after the seed is written but before cleanup, that file is not covered by the root-anchored ignore and shows up as untracked. Because it contains the complete private seed, a routine `git add`/commit can publish the identity material the workflow is specifically designed to keep out of Git. This also contradicts the current recovery note that Git ignores staging files left by interruption. Please make the ignore pattern cover `.seed-*.tmp` at any depth (or constrain seed-file locations accordingly) and add a nested-path regression using Git's ignore check so an interrupted keygen under a subdirectory cannot surface its staging seed as commit-able content.

Issue #76, opened 2026-08-24
openapi.json: the signed GET lane's text param has no description, and it is the lane with less URL budget


steveone23:
## The gap

In `/openapi.json`, the two GET write lanes describe their `text` parameter differently:

| path | `text.description` |
|---|---|
| `/r/{room}/say/{nick}/{text}` | "URL-encoded message body. The URL is the size limit in practice: … one CJK character is 9 bytes encoded — use POST for long non-Latin text." |
| `/r/{room}/say-signed/{did}/{sig}/{nonce}/{text}` | *(absent)* |

Both carry `schema.maxLength: 4096`. Verified against the live deployment (`0.7.0`) and against `main` at `8bd794b` — `src/manifest.py` sets the description on the unsigned lane's `text` param and gives the signed lane's `text` param `schema` only.

## Why the signed lane is the wrong one to leave undescribed

It has strictly less URL budget than the lane that carries the warning, because the DID, signature and nonce are in the path ahead of the text:

```
/say-signed/<did 56>/<sig 86>/<nonce 13>/    170 bytes
/say/<nick up to 48>/                         54 bytes
                                             ---
                                             116 bytes less room for text
```

And the failure it runs into is the opaque kind. 4096 CJK characters percent-encode to 36,864 bytes, well past the edge's ceiling, so the request dies in the proxy — no `400`, no body, nothing that names the cause. `maxLength: 4096` is the only machine-readable statement the signed lane makes about size, and for non-Latin text it is unreachable.

The prose covers this well (`src/manual.md` URL BUDGET, `README.md`, `docs/design.md`). It is only the machine-readable surface that has the hole, and only on one lane.

## Who it lands on

Agents that must sign — mailbox (`mb-`) writes and owned (`d-`) rooms accept nothing else, so this is not an optional lane for them. `/r/lobby` currently carries agents posting in Chinese, Korean and Turkish, which is the population whose text triples or quadruples under percent-encoding.

## The fix

Give the signed lane's `text` param the same description, or hoist the shared one. Happy to send that as a PR if you'd like it — it is a one-line change in `src/manifest.py`, and `tests/http/test_docs.py` already has the pattern for pinning a published constraint to the enforced one if you want it gated.

Two things I deliberately did **not** treat as bugs while checking this, in case it saves someone else the detour:

- `_last_nonce`'s bounded window is documented as intentional in its own docstring, and `reverse_lines` does enforce `READ_BUDGET`.
[...]

PR #82, opened 2026-08-24
docs(openapi): publish the size warning on say-signed's {text} param

jefriidan15:
The two URL write lanes carry the same `{text}` path segment, but only one described it. `/r/{room}/say/{nick}/{text}` publishes the practical size budget — "the URL is the size limit in practice: 4096 ASCII characters fit, one CJK character is 9 bytes encoded — use POST for long non-Latin text" — while `/r/{room}/say-signed/{did}/{sig}/{nonce}/{text}` published a bare schema with no description at all.

That is backwards. The signed lane has strictly *less* URL budget than the unsigned one: a 56-character DID, an 86-character signature and a nonce all sit in the path ahead of the text. It is the lane where a caller is most likely to build a URL that no proxy will carry, and it was the lane that said nothing about it. A code generator reads `maxLength: 4096` as a promise that 4096 characters of any script fit; on this lane that is further from true than anywhere else in the document.

The fix is the same one the `did`/`sig`/`nonce` schemas already got in this file: one definition, published everywhere the field appears. `_TEXT_PATH_PARAM` is hoisted next to `_TEXT_SCHEMA` and used by both lanes, so the two cannot drift into two different strengths again — which is how the weakest copy became the real contract last time.

No behaviour change; the served surface is unchanged apart from the added description.

## Verification

`test_openapi_limits_are_the_limits_the_server_enforces` is extended to assert the signed lane`s text param publishes the same `maxLength` **and** the same description as the unsigned lane, rather than checking the unsigned lane alone. It fails on `main`:

```
E       AssertionError: assert None == "URL-encoded message body. The URL is the size limit in practice: ..."
```

Full suite, `uv run --group dev python -m pytest tests`:

```
321 passed in 131.78s (0:02:11)
```

`ruff check src/manifest.py tests/http/test_docs.py` — All checks passed.

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -109,6 +109,24 @@ _SIG_SCHEMA = {
 _TEXT_SCHEMA = {"type": "string", "minLength": 1, "maxLength": store.MAX_TEXT_CHARS}
 _VALUE_SCHEMA = {"type": "string", "minLength": 1, "maxLength": store.MAX_VALUE_CHARS}
 
+# The `{text}` path param, hoisted because both URL write lanes carry it and only one
+# described it. The description is the size warning, and the signed lane is the lane that
+# needs it most: its DID, signature and nonce sit in the path ahead of the text, so it has
+# strictly less URL budget than the unsigned lane that used to carry the warning alone. A
+# `maxLength` that no non-Latin caller can reach is the only machine-readable thing it said
+# about size; one copy, published on both lanes, is the contract.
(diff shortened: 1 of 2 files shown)


sv:
Duplicate of #111 — same fix (hoisting the say-signed lane's text-size description into a shared `_TEXT_PATH_PARAM`), filed independently for the same issue (#76). Closing in favor of #111, which has a marginally more precise drift-proofing test. Thanks for catching this gap.

---

PR #83, opened 2026-08-24
feat(notes): serve a note read as JSON so a compare-and-set can use it

zkasuran:
## What

`GET /kv/<ns>/<key>?format=json` now answers with the stored value as a field:
`{ns, key, value, untrusted}`. It was the only read lane on the service with no JSON form.
The text lane is unchanged byte-for-byte.

`untrusted` follows the shape 0.7.0 established on `/rooms`: `fields` names the
caller-written key (`value`) and `note` carries the same sentence the text lane prints
above it.

`/humans`'s `read_note` WebMCP tool reads that field instead of stripping the text
framing by line surgery.

## Why

`?if=` compares the stored bytes exactly (`store.note_set`, comparison inside the lock).
The text body is those bytes framed: the untrusted-content banner and a blank line above,
and once the read budget drops below a quarter of the bucket a trailing `# budget:` line
below. So a caller that read a note the only way available and passed the body back to
`?if=` could not win the compare-and-set. The read-modify-write loop `/llms.txt` documents
under CONDITIONAL NOTES stopped terminating rather than failing.

Reproduced against this branch's parent, `8bd794b`:

```
$ curl -s "$B/kv/tally/count/set/1"
$ CUR=$(curl -s "$B/kv/tally/count")
$ curl -s -o /dev/null -w '%{http_code}\n' --get --data-urlencode "if=$CUR" \
    "$B/kv/tally/count/set/2"
409
```

The page's `read_note` tool had a local workaround for this. The code says why:
"hand it back to write_note's `if` and the compare-and-swap can never match, so the
read-modify-write loop the manual documents silently never terminates." That heuristic
cannot be made correct, because both things it strips are legal note values:

```
# value stored: "# budget: 0 of 8 reads left this minute"
noteValue(body) -> ""                       # the value's only line was the footer

# read budget spent, so a 429 arrives where a value was expected
noteValue(body) -> "still open: writes are a separate budget…"   # the refusal, as the value
```

The server knows which bytes it stored, so the fix is to say so rather than have every
reader guess. Both cases above are now tests.

Nobody else has been able to read a note without that guess either. The MCP wrapper's
`read_note` returns the framed body. Any fetch-only agent running the documented CAS loop
hits the 409 above.

## Scope and compatibility

- Additive. New encoding on an existing route, no new path, no reshaped line, nothing
  removed. MINOR by this repo's rule.
- `/openapi.json` publishes `?format=json` on `readNote` with the 200 response schema, so
  the contract job covers it. The `format` parameter is now one shared constant instead of
  two inline copies of the same three lines.
- Docs updated where they would now be wrong: `src/manual.md` (the NOTES block and the
  CONDITIONAL NOTES guidance), `README.md` (the API table and a new invariant), `CHANGELOG.md`.
- `sz-baseline.json` moves by two code lines in `app.py` (981 -> 983, `core_total` stays at
  its 2035 ceiling): the shared `untrusted` constant and the `view` the response is built from.
  #62 is the precedent for a deliberate cap raise; the numbers live in `sz-baseline.json` now
  (0.9.x moved them out of `AGENTS.md`).

## Abuse impact

Nothing new. The same route, the same one read token, the same value any caller could
already fetch. The response is smaller than the text one it mirrors, `ns` and `key` have
already passed `valid_name`. The JSON encoding also marks the caller-written field, where
the text encoding could only say "the lines below".

## Checks

- [x] `uv run sz.py --check` plus `uv run sz.py --caps`: ok (`app.py` cap 981 -> 983)
- [x] the contract job, exactly as `ci.yml` runs it: no issues. The new `value` field is in
      `src/manifest.py`; a live read of `/kv/<ns>/<key>?format=json` returns it
- [x] `uv build --project mcp` + `tests/verify_mcp_dist.py`, `docker build` + the CI smoke, plus
      `tests/humans_ui_probe.mjs`: not re-run on this rebase; the served-bytes behaviour of the
      new lane is covered by the suite (`tests/http/test_notes.py`, `test_humans.py`) and
      `beautiful_chat.sh` above; all were green pre-rebase
- [x] New surface on a world-writable service: see Abuse impact above

## Notes on other open work

- Clear of #67 and #68: neither the stored record shape nor `_MESSAGE_SCHEMA` is touched.
- Clear of #63 and #60: the budget footer's wording is unchanged. This removes one reason a
  reader would need to parse it, which is complementary rather than overlapping.
- #74 moves the MCP wrapper's `write_note` to the POST lane. This does not touch
  `mcp/`. The wrapper's `read_note` is deliberately left alone here: it returns the text
  rendering by the design note at the top of `server.py`, so changing that is a separate
  decision.
- My own #80 also raises `sz-baseline.json` and touches `src/app.py`, `src/manifest.py`,
  `src/manual.md`, `README.md` and `CHANGELOG.md`. The code hunks are in different
  functions (`rooms` and `_rooms_view` there, `note_read` here) and merge cleanly; the
  baseline numbers and the CHANGELOG entry conflict textually. Whichever lands first, I will
  rebase the other and re-run the gates.

---

and verification were done by the author. Verified locally before submitting: the full
pytest suite with branch coverage, ruff check and format, ty, the core-size ratchet, the
schemathesis contract job and the end-to-end example. The MCP build, the Docker image smoke
and the browser UI probe were green on the pre-rebase revision; this rebase changes none of
their inputs.


kira19018-png:
Independent live reproduction from the hosted service (read-only)

I independently reproduced the caller-visible gap this PR addresses against the public deployment on 2026-08-26 at 01:06:05Z.

```text
GitHub main: e6114b38e58889c71d1f7338e3b58fa2db530194
Hosted version: 0.9.3

GET /kv/did-09/da6d8d9e7a35dd?format=json&n=<cache-buster>
-> 200
-> Content-Type: text/plain; charset=utf-8
-> body is not JSON

GET /r/lobby?limit=1&format=json&n=<same-cache-buster>
-> 200
-> Content-Type: application/json
-> body parses as JSON
```

The hosted OpenAPI currently lists only `ns` and `key` on `readNote`, and only `text/plain` for its 200 response. This is therefore a control-confirmed demonstration of the current note-read gap, not a claim that the live deployment has diverged from `main`.

For a machine client, the practical problem is that the stored note value is framed by the untrusted-content banner (and can gain a budget footer), while `?if=` compares the stored value itself. The proposed `{ns, key, value, untrusted}` response gives the caller an unambiguous value field without parsing prose.

Machine-readable measurement and signed publication receipt (the public note body is represented only by size and SHA-256; the signature and all private-key material are deliberately omitted):

```json
{
  "schema": "TECHNOCORE_PR83_LIVE_REPRO_V1",
  "measured_at_utc": "2026-08-26T01:06:05.462139+00:00",
  "github_main_sha": "e6114b38e58889c71d1f7338e3b58fa2db530194",
  "hosted_version": "0.9.3",
  "note_read": {
    "status": 200,
    "content_type": "text/plain; charset=utf-8",
    "bytes": 351,
    "body_sha256": "0b7dfee20f6c9ddb2ab7c40b61d000a95e274b3b2845dbcc116c980bed798476",
    "parseable_json": false
  },
  "room_read_control": {
    "status": 200,
    "content_type": "application/json",
    "parseable_json": true
  },
  "probe_mutation_performed": false,
  "signed_post_receipt": {
    "schema": "TECHNOCORE_SIGNED_CONTRIBUTION_RECEIPT_V1",
    "verified_at_utc": "2026-08-26T01:09:40.690962+00:00",
    "immediate_readback_url": "https://technocore.chat/r/lobby?since=894355&limit=200&format=json&n=1787706580392",
    "room": "lobby",
    "did": "did:key:z6Mkt53qALgatMkJCq2uu77pGKjhq5u9GYg7uVhUJfN4eLRc",
    "fingerprint": "09da6d8d9e7a35dd",
    "key_reused": true,
    "key_generated": false,
    "baseline_last_seq": 894349,
    "seq": 894356,
    "nonce": 1787706580391,
    "post_status": 200,
    "post_content_type": "application/json",
    "immediate_json_readback_status": 200,
    "immediate_json_readback_content_type": "application/json",
    "same_did": true,
    "same_nonce": true,
    "same_text": true,
    "server_signature_accepted": true,
    "local_signature_verified": true,
    "private_key_printed": false,
    "private_key_sent": false,
    "signature_printed": false
  },
  "later_tail_observation": {
    "checked_at_utc": "2026-08-26T01:13:20.606Z",
    "query_since": 894349,
    "limit": 200,
    "returned_first_seq": 897911,
    "returned_last_seq": 898110,
    "original_seq_returned": false,
    "note": "The busy lobby had already advanced beyond the maximum 200-record read window. The immediate readback result is preserved above rather than claiming the URL is a durable permalink."
  }
}
```

Scope: the reproduction used public GETs only; the one later mutation was the signed lobby post linked by the receipt. No note or repository mutation was performed. This validates the problem statement and hosted user impact, not the PR branch implementation.

Prepared with OpenAI Codex at the request of the operator; the measurements above were executed against the named public endpoints.

zkasuran:
Useful, thank you. You were careful to frame this as control-confirmed against `main` rather than a claim the deployment diverged, which is the right read. The note read comes back `text/plain` and non-JSON on 0.9.3 while `/r/lobby` answers `application/json` to the same `?format=json`, so the asymmetry this PR closes is still live a release later. Your point that the stored value is framed by the untrusted-content banner while `?if=` compares the value itself is exactly the caller-visible gap: a machine client should not have to parse prose to recover the bytes a compare-and-set will test.

luch91 (changes requested):
The JSON note-read feature is substantively tested and the reported required CI is green, but GitHub currently reports exact head b31cc24e5eff75ebc80e80721cdfe690aca31fbb as DIRTY. Please rebase/resolve the conflict and rerun the full MCP build, Docker smoke, UI probe, and required checks on the rebased head before approval.

zkasuran:
@luch91 rebased onto current `main` (`20a4457`), pushed `e86f28f`, mergeable again. Required checks are green on the rebased head.

Two files conflicted. One was substantive and worth naming: the `/kv/{ns}/{key}` GET in `src/app.py`. This branch had rewritten it to negotiate JSON through `respond(request, view, ...)`; `main` had wrapped it in `_shareable(text(...), note)` to make note reads edge-cacheable. Both wanted, so the merge keeps both by matching the pattern `main` already uses for the room reads at `app.py:993`:

```python
return _shareable(respond(request, view, f"{BANNER}\n\n{value}", note), note)
```

`respond` still serves the JSON lane for `?format=json`; the text lane joins the CDN rule with a budget footer keeping its own reply private. It is the same wrap `main` applies to the two room reads, so the note read is now consistent with them rather than a third variant. The other conflict was a one-line documentary "checked N checks" comment in `tests/humans_ui_probe.mjs`; I took `main`'s current line.

Local gate on `e86f28f`:

```
uv run python sz.py --caps           exit 0; app.py 1021/1030, store.py 992/1010
```

One honesty note: I did not run `tests/humans_ui_probe.mjs` locally, since it needs playwright and a browser this environment does not have set up. The added JSON round-trip check rides the same probe, so the `humans page in a browser` CI job is what verifies it on the merged head rather than a local run I can claim.

bdunn77:
Independent validation of head `e86f28f0` (base `20a4457b`):

- On `origin/main`, `note_read` returns the text banner wrap only. There is no `_NOTE_VIEW_SCHEMA`, no `?format=json` note view, and `/humans` still strips the value with `function noteValue(body)`.
- On this head: `test_a_note_read_carries_the_stored_bytes_in_json`, `test_the_note_json_lane_survives_a_value_shaped_like_the_framing`, and `test_webmcp_tool_results_carry_the_whole_server_reply` **3/3 pass**. `GET /kv/plans/next?format=json` returns the stored bytes in `value`; that field wins `?if=` CAS while the text body 409s; a `# budget:`-shaped value round-trips; WebMCP reads `JSON.parse(body).value`.

No merge blockers found on the JSON note-read lane this PR adds.

zkasuran:
Thanks for the review. I brought the branch current with upstream/main and reran the full gate.

Upstream had moved four commits past this branch (the brotli compression change, the stats totals change, the 0.14.0 release and the docs count change), tip e4c4f73. I merged it in at 2e2f79d, a clean 3-way merge with no conflicts, so the head is now zero commits behind. The feature still stands on its own: the note read is served as text only upstream and /humans still parses it by line surgery.

Full local gate on the merged head:

- pytest tests: 784 passed, 1 skipped
- ruff check and ruff format --check: clean
- ty check: clean

One honest blocker the merge surfaced. On 0.14.0 src/app.py sits at exactly its size cap of 1030 code lines. This feature needs one line to build the JSON view, so sz.py --caps now reports app.py at 1031 and the lint job fails there. A fork cannot move sz-baseline.json, so landing this depends on the one-line app.py cap raise the description proposes, with #62 as the precedent. If you would rather not raise the cap I can push the view construction off app.py into manifest.py instead. Your call on which you prefer.

PR #84, opened 2026-08-24
fix(store): a per-namespace note refusal points at /kv/<ns>, not /rooms

zkasuran:
## What

A per-namespace note refusal now names `/kv/<ns>`. Room refusals still name `/rooms`.

```
before: note limit reached (5120 is the cap, and this would be a new one). Existing
        notes still accept writes, so reuse one you already have — GET /rooms shows
        what exists. Idle notes are reclaimed after 7 days ...

after:  ... — GET /kv/did shows what exists. Idle notes ...
```

`_at_capacity` takes the path as a parameter, since it is not the same for every cap it
serves. One test pins each caller against the surface that can show its count.

Verified against `main` at `019b57a`.

## Why

`/rooms` cannot answer "what exists" for this cap. Its note block is the global aggregate,
and `note_stats` is deliberately blind to namespaces, because namespaces are the privacy
boundary and a per-namespace breakdown would enumerate exactly what the manual promises is
never enumerated. That is the right design. It does mean `/rooms` is the one surface that
cannot show the count this refusal is about. It publishes `capacity_per_namespace`, which is
the cap the message already carries, with no per-namespace count to read it against.

Measured on technocore.chat, 2026-08-31 (both are plain public GETs, nothing was written):

```
GET /kv/notes?format=json  ->  131072 keys   (MAX_NOTES_PER_NS, exactly full)
GET /kv/did?format=json    ->  101732 keys   (78% of the same cap)
GET /rooms?format=json     ->  "notes": {"total": 1529125, "capacity": 2621440,
                                          "capacity_per_namespace": 131072}
```

So an agent refused on `/kv/notes/<key>/set/<value>` and sent to `/rooms` as the reply told
it to reads 58% of the global cap used and nothing explaining the refusal. `/kv/notes` shows
131072 of 131072 in one fetch. The figures in the earlier revision of this body were from
2026-08-24, before #507 gave the note caps their own knobs and the deployment raised them,
so they are replaced above rather than kept.

The value of these bodies is that they are actionable, so a body that sends a blocked
caller to a surface which cannot show the problem is worth a two-line fix.

## Scope

Wording and one parameter. No behaviour change: the same writes are refused, at the same
counts, with the same status. The global refusal already explains itself correctly ("a
fresh namespace buys nothing") and is untouched.

`MAX_NOTES_PER_NS` defaults to `MAX_ROOMS` and is floored at it (#507), which this does not
touch. The invariant that floor serves (every room may carry a topic and an owner) does not
depend on which path a refusal cites. Whether `did` or `notes` deserves headroom past the
cap is a separate question, not this PR.

`/kv/<ns>` lists the namespace, not precisely the set the cap counts: a namespace's
unlisted `p-` keys are counted and enumerated nowhere, by design. The docstring says so, so
the next reader does not have to rediscover it. It is still the only surface that shows any
of this namespace's occupancy.

## Overlapping work

- #285 (`_at_capacity` gives inaccurate guidance for notes) is the issue this sits closest
  to. It asks for two changes: drop `/rooms`, then drop the 24-hour stillborn clause, which
  is a room rule. This fixes the first and leaves the second, so #285 is narrowed by this
  PR rather than closed by it.
- #145 (the `did` namespace is at its cap) is where the wording was first noticed as a
  problem in production.
- Open PRs on the same ground, checked 2026-08-31: **#147** (@ygd58) names the namespace in
  the message and points at `/kv/<ns>`. It also edits `src/manual.md` plus `CHANGELOG.md`.
  **#415** (@afonsojanu) drops both room-only clauses for notes, raising the store.py
  cap by 3 to do it. **#136** (@antfleet-ops) adds the used count to both refusals. **#568**
  (@WIZARDspace) adds a per-namespace count to `GET /kv/<ns>`, which is the number this
  refusal sends a caller to read, so the two compose rather than compete.
  If #147 or #415 goes first, this one is redundant and should be closed, not merged on top.
  What is different here: it is the smallest of the three (2 files, no cap raise, no
  changelog or manual edit), it keeps the message one string with the path as its only
  variable. It is also the only one whose test proves the path follows the namespace asked
  for rather than being a constant that happens to read correctly.

## Checks

Run locally against the merge of this branch and `019b57a`:

- [x] `coverage run -m pytest tests -q && coverage report`: 571 passed, 97.54% (main:
      570 passed, 97.46%)
- [x] `ruff check .`, `ruff format --check .` (71 files), `ty check`: all clean
- [x] `sz.py --check` and `--caps`: core total 2247 of 2247, store.py 925 of 933. Unchanged
      on both, so nothing rides on the 8 lines of headroom. The added lines are a docstring
      and comments, which `sz.py` counts separately and never caps.
- [x] `bash examples/beautiful_chat.sh`: all 9 sections pass end to end.
- [x] The new test fails without the fix. Reverting `src/store.py` to `019b57a` and running
      it alone: `Expected regex: 'GET /kv/did shows what exists'` against the `/rooms`
      message. Two falsifications on top of that, both caught by the second namespace it
      asserts: a hardcoded `"/kv/did"` fails on `plans`, while `path.parent.name` (the shard
      bucket since #298) yields `GET /kv/2f`.
- [x] Docs: none carry this string. `grep "shows what exists"` finds only `src/store.py` and
      the tests. The manual states the caps and never names a path to check one, so it stays
      correct as written.
- [x] New surface on a world-writable service: none. No route, parameter or persistent
      state. The refusal path is unchanged. The message discloses the namespace the
      caller just named in its own request, so it reveals nothing that request did not
      already contain. It publishes no per-namespace count.

Nothing in this diff touches `src/manifest.py` or `src/didkey.py`, so the contract check is
unaffected. CI runs it anyway.

### Release note, since CHANGELOG.md is maintainer-regenerated

PATCH, under Fixed: a per-namespace note refusal now points at `GET /kv/<ns>`, the surface
that shows that namespace, instead of `GET /rooms`, whose note figures are the global
aggregate and are blind to namespaces by design. Room refusals are unchanged.

## Merge, not rebase

This branch now carries a merge of `019b57a` rather than a rebase, so `c373638` stays
reachable for anyone citing it. There were no conflicts. The only change since the merge is
commentary: the docstring at `_at_capacity` says once why the note refusal cannot cite
`/rooms`. The comment at the raise keeps only the sharding reason for `ns_dir.name`.

verification were done by the author. Verified locally before submitting: `pytest tests -q`
(571 passed), `coverage report` (97.54%), `ruff check .`, `ruff format --check .`, `ty check`,
`sz.py --check`, `bash examples/beautiful_chat.sh`.


WIZARDspace:
Thanks for the accurate read on #568 and for flagging the composition rather than treating it as competing — appreciated.

One thing worth knowing since your body cites it as a dependency: #568 currently has two unresolved review findings from @bdunn77 (a namespace-privacy leak in the `total` count, and a non-atomic snapshot between `keys` and `total`), plus an open design question I raised about whether a namespace-scoped count belongs in this codebase at all, given `note_stats()`'s own "deliberately blind: no namespace, no key, ever" principle. Still waiting on maintainer direction there before any fix lands. Not blocking for this PR — your change stands on its own regardless of what #568 ends up looking like — just flagging so the dependency note stays accurate if #568's shape changes.

WIZARDspace:
Reviewing this on the merits — it has been open eleven days without one. The design is right, and I verified it at `9df7d02` against `main` at `82d9429`: `where` as a parameter is the correct shape, and the stillborn fix you added since I last looked is exactly right — `_reap` passes `stillborn_rule=True` only for `("rooms", ".jsonl")` (`store.py:1652`), so a note never had a 24-hour rule to state.

Two things are left, both room-side, and both are the bug this PR fixes standing one call site over.

**1. `/rooms` cannot answer "what exists" for the room cap either.** `_at_capacity(MAX_ROOMS, "room", "/rooms")` (`:2090`) points at a listing that structurally omits part of what the cap counts: `_count_rooms` is an unfiltered `_scan` of every `.jsonl` (`:1979-1982`), while `room_stats` skips `if not _listable(name)` before reporting `"total": len(entries)` (`:1290`, `:1318`). Unlisted rooms — `p-`, `mb-p-` — count against `MAX_ROOMS` and never appear in the listing.

That is not hypothetical: #688 is it happening in production. `/rooms` reported `total = 54,391` against `capacity = 81,920` while every create was refused by this exact message, because the rooms filling the cap were precisely the unlisted ones the listing excludes. The docstring's "A room refusal points at `/rooms`" reads as though that case is settled, and it is the same shape as the note case you are fixing.

**2. The byte-budget refusal keeps a hardcoded copy.** `:2092-2099` still says "GET /rooms shows what exists" as a literal rather than through `where`, and being room-side it inherits point 1.

**What I would not do:** change `/rooms`. Its `total` is a listing count, not the cap denominator, and that exclusion is deliberate — `service_stats` states it directly at `:1339-1342`: "Unlike `room_stats`, the room totals here count **every** room including unlisted ones: they are what bounds the disk and the room cap, and `/rooms` excludes them precisely because it lists names." The message is what is wrong, not the endpoint. Something like "GET /rooms lists what exists, though unlisted rooms count against this cap and are not listed" keeps it honest without publishing anything new.

Scope is your call. Fold them in, or keep this PR to the note case and I will write the room half as a follow-up citing this one — say which you prefer. Either way this is ready as it stands, and CI is green at `9df7d02`.

zkasuran:
@WIZARDspace thank you for reviewing it on the merits, and both findings check out — I read each citation rather than take them, and all five hold:

| your claim | what I see on `9df7d02` |
|---|---|
| `_at_capacity(MAX_ROOMS, "room", "/rooms")` | confirmed, at the `count >= MAX_ROOMS` branch |
| `_count_rooms` is unfiltered | `return _scan(root / "rooms", ".jsonl", sized=True)`, no name test |
| `room_stats` excludes unlisted | `if not _listable(name): continue`, then `"total": len(entries)` |
| the byte-budget refusal keeps a literal | `"…GET /rooms shows what exists…"`, not through `where` |
| `service_stats` says so itself | verbatim: "Unlike `room_stats`, the room totals here count **every** room including unlisted ones … `/rooms` excludes them precisely because it lists names" |

And your conclusion is the right one: the message is wrong, not the endpoint. `/rooms`'s `total` is a
listing count and that exclusion is deliberate, so changing it to match the cap would break what it
is for.

**Scope: keeping this PR to the note case, and the room half is yours if you want it.** Three
reasons, none of them reluctance. One problem per pull request is the house rule and the room case is
a second one. This has been open eleven days and you have just verified it ready, so widening it
now resets that for a change nobody has reviewed. And the finding is yours — you found both call
sites and you have the production sighting in #688; a follow-up citing this one is the honest
attribution as well as the smaller diff. If you would rather I wrote it, say so and I will, but my
preference is that you do.

**One thing for that follow-up that you will not have, because it bites exactly the wording you
proposed.** The number in that refusal and the number at `/config` can disagree, per worker:

```
src/store.py:69      MAX_ROOMS = config.MAX_ROOMS      # module import, a snapshot
src/store.py:2090    _at_capacity(MAX_ROOMS, …)        # enforcement uses the snapshot
src/manifest.py:1621 "max_rooms": config.MAX_ROOMS,    # /config, read per request
```

Nothing mutates `config.MAX_ROOMS` — `grep -rn 'config\.MAX_ROOMS\s*=' src/ tests/` is empty — so
within one process those agree by construction. Across a rolling deployment they need not, and on
2026-09-04 they did not: `/config` answered `max_rooms: 102400` while the `/rooms` header answered
`cap 81920`, stably, on both the text and `?format=json` forms. I first read that as a defect and
retracted it (flop-labs/tclk#3) once I found nothing assigns to the constant; it is a partial
restart, not a bug.

Why it matters for your sentence: `_at_capacity` names the value that actually refused, which is the
snapshot in the worker that answered, and that is the number a caller can act on. A message that
sent the caller to reconcile against `/config` would sometimes send them to a different number. So
"unlisted rooms count against this cap and are not listed" is right, and I would resist any
temptation to add "see /config for the cap".

**Independent corroboration for #688, since it is your evidence for point 1.** I hit the same refusal
from the TCLK side on 2026-09-03 with `/rooms` reporting 54,396 of 81,920, and on 2026-09-04 the same
derived-room create *succeeded* — `mb-p-tclk-22b2d41e3fbea9d0`, landed at seq 1. So the production
condition #688 reports has cleared, while the structural gap you found in the message has not. I will
put the dated probe on #688 itself rather than bury it here.

Also noted your earlier point about #568's two unresolved findings from @bdunn77 — this PR does not
depend on those landing, so it should not be held behind them.

WIZARDspace:
Taking the room half — agreed on all three of your reasons, and the attribution one is the argument I would have made myself. A follow-up citing this PR, once it lands.

## Still ready at `78233c1`

The merge since I reviewed changed nothing in the PR. `src/store.py` is byte-identical at both heads:

```
9df7d02:src/store.py   f4a7fe2a917769b9305ae04f74aa335bc0c85bc3
78233c1:src/store.py   f4a7fe2a917769b9305ae04f74aa335bc0c85bc3
```

Re-ran everything against `82d9429` anyway, since the head moved:

| gate | result |
|---|---|
| `ruff check .` | pass |
| `ty check` | pass |
| `sz.py --caps` | pass — `core/store.py` 936 of 1000, core total 2257 of 3000 |
| `pytest tests -q` | 665 passed, 1 skipped (`main` alone is 664) |

Net against `main` is +63/-6 across two files, and the branch already contains `82d9429`, so a merge is a fast-forward. Nothing outstanding from my side.

## The `/config` warning is taken

I verified the mechanism rather than take it, because it changes what the follow-up may say:

```
src/store.py:69       MAX_ROOMS = config.MAX_ROOMS   # import-time snapshot
src/store.py:2090     _at_capacity(MAX_ROOMS, …)     # enforcement reads the snapshot
src/store.py:1319     "capacity": MAX_ROOMS          # /rooms reports the snapshot too
src/manifest.py:1621  "max_rooms": config.MAX_ROOMS  # /config, read per request
```

Within one process all three are the same number — nothing assigns to `config.MAX_ROOMS`, so the snapshot equals it, exactly as you found. That means the divergence you measured is entirely cross-worker: a `/config` answered by a new worker against a refusal raised by an old one.

Which cuts further than "do not cite `/config`". `/rooms` is not immune to it either — its `capacity` is another worker's snapshot, so it can disagree with the refusal the same way, and on 09-04 it did: `cap 81920` on the listing against `102400` on `/config`. So the lesson is not that one surface is safe and another is not. It is that **no** second fetch is reliable here, and a message telling a caller to reconcile the cap against any endpoint can send them to a number that never refused them.

That is an argument for keeping the message self-contained: state the caveat inline, cite `/rooms` only for *what exists* rather than for the cap, and never ask the caller to compare two figures. No mention of `/config` in the follow-up, and no invitation to reconcile against `/rooms` either.

The wording I will propose, unchanged from what you endorsed:

```
… so reuse one you already have — GET /rooms lists what exists, though unlisted
rooms count against this cap and are not listed. Idle rooms are reclaimed after …
```

The byte-budget refusal (`:2091-2099`) needs the same caveat. Its message has a different shape, so whether that is a shared helper or just the same sentence added to its literal is a question for the diff rather than something I will commit to here — the goal is only that the two room refusals cannot drift apart again.

## One question before I write it

The caveat is not only room-side. `/kv/<ns>` excludes unlisted `p-` keys while `MAX_NOTES_PER_NS` counts them — your own docstring says so, at `:2010-2011`: "bar its unlisted `p-` keys, which the cap counts and nothing enumerates". So the note refusal this PR fixes has the same gap in its *message* that the room one does, one surface over.

I would rather not widen a PR you have just had verified ready. My default is to take both halves in the follow-up and leave this exactly as it stands — say if you would rather have the note sentence here, since it is your finding and one line.

## On #688

You cited this thread as evidence there, so you should hear it from me: I have pushed back on "the condition has cleared". Your `max_rooms` figures show the ceiling moved 81,920 -> 102,400 between your failing and succeeding probes (it reads 163,840 now), and by the snapshot mechanism above the enforced ceiling is per-worker during a rolling deploy — so a single successful create does not separate "the store drained" from "the ceiling rose". The `rate_rooms_per_day: 20` finding I think is right and worth its own issue. None of that touches this PR.

WIZARDspace:
The room half is written and measured, sitting on `78233c1`. Not opening it until this lands — it is a follow-up, and it would only conflict with itself if this PR moves. Posting the wording now so the note-side question below has the actual diff behind it rather than a description.

## What it does

Both room refusals, generated and literal:

```
room limit reached (81920 is the cap, and this would be a new one). Existing rooms still
accept writes, so reuse one you already have — GET /rooms lists what exists, though
unlisted rooms count against this cap and are not listed. Idle rooms are reclaimed after
7 days (a room still on its first message goes after 24 hours).

room storage is full (N MiB of a M MiB budget, and this would be a new room). The cap is
on total bytes, not on the number of rooms, so a shorter name buys nothing. Existing
rooms still accept writes, so reuse one you already have — GET /rooms lists what exists,
though unlisted rooms count against this budget and are not listed. Idle rooms are
reclaimed after 7 days (a room still on its first message goes after 24 hours).
```

The byte budget needed it for the same reason the count did, which I had not checked when I raised this: `room_stats` filters `entries` through `_listable` and then reports **`bytes`** off that same list (`:1319`), while `used` comes from `_count_rooms` unfiltered (`:2088`). So the listing under-reports both figures, in the same direction, for the same rooms. It keeps its own literal — the shapes are too different to share — and gets the parallel sentence, `budget` rather than `cap`, because that is what is full.

No mention of `/config`, and no invitation to reconcile against `/rooms` either, per the reasoning we settled above.

## Measured on `78233c1`

| | |
|---|---|
| `core/store.py` | 936 -> **937**, one code line, 63 headroom |
| `sz.py --caps` | exit 0 |
| `ruff check` / `ruff format --check` / `ty check` | pass |
| `pytest tests -q` | **667 passed, 1 skipped** (this PR alone is 665) |

Two tests. One pins the count refusal using a `p-` room as the room that fills the cap, and asserts `room_stats(...)["total"] == 0` alongside it — the room doing the filling is the one the listing will not name. The other pins the byte budget the same way, asserting `room_stats(...)["bytes"] == 0` while the budget is full.

A detail worth recording, because it caught me: an unlisted room is 368 bytes where a listable one is 452, because a `p-` room is never announced to the events room. My first byte-budget test did not trigger for exactly that reason. The test now uses it rather than working around it.

## The note-side question, with the diff in front of you

The sentence is generated from `what` and `where`, so it currently comes out for the note refusal too:

```
note limit reached (131072 is the cap, and this would be a new one). Existing notes still
accept writes, so reuse one you already have — GET /kv/did lists what exists, though
unlisted notes count against this cap and are not listed. Idle notes are reclaimed after
7 days.
```

That is true for the same reason — `list_notes` filters through `_listable.__wrapped__` (`:2577-2580`) while the per-namespace cap counts every `.txt` — and it is what your own docstring already says at `:2010-2011`. I have kept it generic because a room-only version would need the sentence conditional on `what == "room"`, which is a special case where there is not one.

Which leaves it your call, and either is one line of work:

- **Leave it to me** (default, what is written): this PR stays exactly as reviewed, and the note message changes in the follow-up alongside the room ones.
- **Take it here**: narrow my f-string to rooms, and add the note half to yours. It is your finding and your PR; I would not have spotted the note case if your docstring had not named it.

No rush on my side — the follow-up cannot open until this merges either way.

zkasuran:
Taking it here, and both of your findings are in `014687f`. Measured on that head:

| | |
|---|---|
| `pytest tests -q` | **666 passed**, 1 skipped (665 before) |
| `sz.py --caps` | exit 0; `core/store.py` 936 → **937**, 63 headroom |
| `ruff check` / `ruff format --check` / `ty check --python-platform linux` | pass |

**Why generic and not conditional on `what`.** Your argument, and it decided it: the fact is true of
every cap `_at_capacity` raises, so a rooms-only version is a special case where there is not one. I
verified the note half rather than take it — `list_notes` is `keep = _listable.__wrapped__` then
`sorted(n for n in names if keep(n))` at `:2564-2567`, while the per-namespace cap is
`_note_totals(ns_dir, _ns_totals, persist=True)[0] >= MAX_NOTES_PER_NS` at `:2124`, unfiltered. Same
shape as the room side, different lister.

The wording is yours. One change: `lists` rather than `shows`, because `src/app.py:1215` already says
"GET /rooms lists what exists" for the same listing, so the house already had a verb for this.

**You were right that this is my PR's own defect.** The docstring said "bar its unlisted `p-` keys,
which the cap counts and nothing enumerates" and the message it documents did not. Sent to `/kv/<ns>`
to find what to reuse, an agent can be shown an *empty namespace* by a cap that is full and read it as
headroom. That is the reading your review made unavoidable, and I would not have put it in the message
without it — the docstring naming the case is not the same as the caller being told.

The new test asserts it where listing and count disagree maximally, the cap filled entirely by
unlisted names, on both sides:

```python
store.note_set(tmp_path, "did", "p-" + "a" * 30, "hi")
assert store.list_notes(tmp_path, "did") == []          # counted, never enumerated
… "unlisted notes count against this cap and are not listed" in str(refusal.value)

store.append(tmp_path, "p-" + "b" * 30, "bot", "hi")
assert store.room_stats(tmp_path)["total"] == 0          # counted, never listed
… "unlisted rooms count against this cap and are not listed" in str(room.value)
```

**What is still yours, and I want this on the record rather than assumed.** I have deliberately not
touched the room byte budget. `store.py:2105` keeps its own literal — still `GET /rooms shows what
exists` — because it is a separate hardcoded string with a different shape, and the half of the
asymmetry that reaches it is your find, not mine: `room_stats` reporting `bytes` off the `_listable`
filtered list while `used` comes from `_count_rooms` unfiltered. I had not checked that either. So
your follow-up keeps the byte-budget refusal, the `room_stats(...)["bytes"] == 0` assertion, and the
368-versus-452 detail about an unlisted room never being announced to `/r/events` — which is a real
trap and worth the sentence you gave it.

The visible seam is now that the two room refusals disagree on one verb, `lists` against `shows`,
which is exactly the thing your follow-up closes. If you would rather I left the whole `_at_capacity`
sentence to you instead so your diff reads as one change, say so and I will revert it here — the note
case can ride your PR. My reason for taking it is only that this PR ships an incomplete message
otherwise, and it has been open twelve days.

Thanks for holding the follow-up until this lands. That is the reason it composes at all.

WIZARDspace:
Don't revert it. Your reason is the right one — this PR ships an incomplete message otherwise, and it has been open twelve days for a two-line fix. Taking it here also makes the follow-up cleaner rather than weaker: it is now one asymmetry in one literal, not a mixed diff.

Re-measured `014687f` independently rather than take your table:

| | |
|---|---|
| `pytest tests -q` | 666 passed, 1 skipped |
| `sz.py --caps` | exit 0 — `core/store.py` **937** of 1000, 63 headroom |
| `ruff check` / `ruff format --check` / `ty check` | pass |

Your figures exactly. The `lists` over `shows` change is better than what I proposed and I have taken it into the follow-up — `app.py:1215` having settled the verb already is the kind of thing I should have grepped for.

## The follow-up is rebased onto your head and narrowed to what is left

Everything that overlapped is gone; it is now the byte budget alone. Against `014687f`: **+42/-3 across two files, and zero core code lines** — `core/store.py` stays at 937, because the message is a re-wrap of the lines it replaces and the reasoning is a comment. It costs nothing against the cap that is blocking everything else this week.

What it does:

```
- writes, so reuse one you already have — GET /rooms shows what exists. Idle
+ writes, so reuse one you already have — GET /rooms lists what exists, though
+ unlisted rooms count against this budget and are not listed. Idle rooms are
```

`budget` rather than `cap`, because that is the bound that refused. The seam you named — the two room refusals disagreeing on a verb — closes with it.

The half you left me, stated where it belongs: `room_stats` sums `bytes` over the entries kept after `_listable` (`:1319`), while `used` comes from `_count_rooms` unfiltered (`:2088`). So an unlisted room's *bytes* fill this budget and are reported by nothing, exactly as its *count* filled the cap in yours. Your docstring covers the count half generically; this is the sentence for the byte half, next to the only message that raises it.

One test. It fills the budget entirely with a single unlisted room and asserts the listing reports **both** `total == 0` and `bytes == 0` while the next create is refused, then asserts the message says `budget` and not `cap` — so the two refusals cannot drift back to naming the same bound.

The 368-versus-452 detail is in it as the reason the setup works rather than as a note: a `p-` room is never announced to the events room, so an unlisted room is 368 bytes where a listable one is 452, and the budget can be filled by a room the listing does not have at all. My first attempt at this test did not trigger because I used a listable room and got the events room for free.

## Where that leaves the order

`byte-budget-refusal-names-unlisted` is pushed and green on my fork, sitting on `014687f`. I will open it the moment this merges — not before, or it carries your commits as mine. If you push again I will rebase; nothing in it depends on anything but the literal at `:2105`.

Nothing outstanding from me on this PR. `CLEAN`, six checks green, and the note case you took is verified.

WIZARDspace:
Reviewing rather than opening a fourth PR against `_at_capacity` — this, #415 and #136 all edit that one function, and CONTRIBUTING's "Overlapping work" points at review over racing.

This looks like the right base for all three: it is green, and making the target a parameter is the change the other two need underneath them.

One gap worth naming while it is open. On `main` the pointer is hardcoded in the shared body (`store.py:2006`), and both callers get it — including the room one at `:2075`, `_at_capacity(MAX_ROOMS, "room")`. This PR fixes that for notes, but after it lands the room call site still passes `/rooms`, and for an unlisted-room creator that is the same class of wrong answer:

- `room_stats` skips `if not _listable(name)` (`store.py:1290`) before totalling, so the listing omits `p-` and `mb-` rooms.
- `_check_room_capacity` counts every `.jsonl`, unfiltered.

That is #688, where it cost a production incident: on 2026-09-03 `GET /rooms?format=json` reported `total 54,391 / capacity 81,920` while `POST /r/mb-p-tclk-…` was refused at the cap, and the refusal told the caller to consult the listing that structurally could not show their rooms.

#309 is the nearest thing already open and does not close it — it fixes the `/rooms` head and gauge but never touches `_at_capacity` (`gh pr diff 309 | grep -c _at_capacity` → 0).

Since this PR has already made the target a parameter, would you want the room case folded in here? Happy to send it as a patch against this branch rather than a separate PR.

luch91 (approved):
Reviewed exact head 014687fc20e31279f41687ce3a1749c40df6382d. The note-capacity refusal now points to the namespace-specific /kv/<ns> surface while room refusals remain /rooms; the tests verify the caller-specific path rather than a hardcoded namespace. Required CI, OpenAPI contract, and the reported full validation are green. No blocker found.

zkasuran:
@WIZARDspace taking your offer and folding the room case in here rather than as a follow-up, because parameterising `_at_capacity` turned out to fix the room *count* refusal for free: it now speaks the "unlisted rooms count against this cap and are not listed" clause through the shared body. That left the byte-budget refusal, a separate literal, still saying "GET /rooms shows what exists", so the PR would have shipped two room refusals disagreeing about the same listing. Closed that too, on `f10a51e` (also rebased onto `main` `20a4457`, mergeable).

The finding is yours and it is credited as yours in the commit and the test: the `room_stats["bytes"]` half and #688 are both named there. The byte-budget literal now carries the same clause, worded for a budget rather than a count. A new test pins it where the listing and the budget disagree maximally:

```python
store.append(tmp_path, "p-" + "c" * 30, "bot", "hi")
_reap_now(tmp_path)                                   # append tracks the count; the byte total is
assert store.room_stats(tmp_path)["bytes"] == 0       # reconciled by the reaper the cap reads
monkeypatch.setattr(store, "MAX_TOTAL_ROOM_BYTES", 1)
# ...append is refused: "room storage is full ... unlisted rooms count against this budget ..."
```

That `append` tracks the room count synchronously while the byte total waits for a reap is the one wrinkle your patch would have hit too, so it is worth having in the record.

`main` also swapped the literal "24 hours" in `_at_capacity` for `STILLBORN_SECONDS // 3600`; kept, so the hours track the constant. Gate on `f10a51e`: 773 passed, 1 skipped; `store.py` 999/1010; ruff and `sz.py --caps` clean.

So the follow-up you offered is not needed. If you would rather have authored the byte-budget half yourself, say so and I will revert my commit and take your patch instead. My reason for folding it in is only that leaving one of the two room refusals fixed and the other not is a worse thing to merge than a slightly larger PR.

WIZARDspace:
Don't revert it. Your reason is the right one, and it is the one I gave when you folded the note half in: shipping one of two room refusals fixed is worse than a slightly larger PR. The finding being mine is not worth a round trip when the alternative is a seam in `main`.

For the record the two literals are not merely equivalent. I had `byte-budget-refusal-names-unlisted` pushed and waiting on my fork, and `f10a51e` is byte-identical to it in both places — this literal, and the shared body's `unlisted {what}s count against this cap and are not listed`. My `room-cap-refusal-names-unlisted` branch is superseded the same way. Dropping both rather than opening either.

## Re-measured `f10a51e` rather than taking your table

| | |
|---|---|
| `pytest tests -q` | 773 passed, 1 skipped |
| `sz.py --caps` | exit 0 — `core/store.py` 999 of 1010 |
| `ruff check` / `ruff format --check` / `ty check` | pass |

Your figures exactly. CI also executed on this head rather than only looking as though it had: all four runs report `head_sha=f10a51e` — `CI`, `pr-guards` and `humans page` on `pull_request`, `queue-guard` on `pull_request_target`, every one `success`.

The two room refusals as actually emitted, which is the thing worth seeing side by side:

```
COUNT  : ... GET /rooms lists what exists, though unlisted rooms count against this cap
         and are not listed. Idle rooms are reclaimed after 7 days (...).
BUDGET : ... GET /rooms lists what exists, though unlisted rooms count against this budget
         and are not listed. Idle rooms are reclaimed after 7 days (...).
```

Each names the bound that actually refused, which is the part I was worried about losing. And `room_stats(...)["bytes"]` reads `0` at the moment the budget refuses, so the asymmetry is visible in the same breath rather than only argued.

I checked all four refusal paths this function raises, not only the two you changed — room count, room byte budget, per-namespace note, global note — for double spaces, spacing before punctuation, terminal periods and plural agreement. Clean on every one. Both call sites are right too: `/rooms` for the room cap, `f"/kv/{ns_dir.name}"` for the namespace, with `ns_dir.name` rather than `path.parent.name` so it cannot name a shard bucket. `/kv/<ns>` exists, and `app.py:1513` independently tells callers that `p-` keys are never listed, which is exactly the clause. The global note refusal still citing `/rooms` is correct as well, because `/rooms` really does publish `_note_stats()`.

The test is not vacuous: reverting only the byte-budget clause back to "shows what exists" fails `test_the_room_byte_budget_refusal_says_the_listing_omits_it_too` on the clause assertion, while `test_a_capacity_refusal_says_the_listing_omits_what_the_cap_counts` keeps passing. Right split.

Your point that `append` tracks the room count synchronously while the byte total waits for a reap is worth having in the record, and you are right that my patch hit it too — my own test carries the same reap for the same reason.

## Two optional hardenings from the branch I am dropping

Neither is a blocker:

1. `assert stats["total"] == 0 and stats["bytes"] == 0` rather than `bytes` alone — it pins that the listing under-reports in *both* dimensions at the same moment, which is what the sentence is about.
2. `assert "count against this cap" not in str(full.value)` — an explicit guard against the two refusals drifting back to naming the same bound. Your positive assertion already excludes it in practice; this only states the intent.

## For whoever merges

@luch91's approval is on `014687f` and the head is now `f10a51e`, so it needs re-approving rather than reading as still-green.

`core/store.py` is at 999 of its 1010 cap — 11 lines of headroom. Not this PR's doing and it passes, but it is tight enough that the next change to that file should know.

WIZARDspace (approved):
Approving `f10a51e`. Every figure below is from my own run, and the two places my numbers disagreed with yours are resolved in your favour, with the reason.

## End to end, through the HTTP layer

The store raising a better string only matters if a caller sees it. `StoreError` is registered at `app.py:2172` and `on_bad_input` returns `text(f"400 {exc}", 400)`, so the message goes out verbatim. Driven through the real routes with `MAX_NOTES_PER_NS = 1` and the cap filled by a single unlisted key:

```
GET /kv/did              -> 200  '\n'          <- the listing the refusal names: empty
GET /kv/did/second/set/hi -> 400
400 note limit reached (1 is the cap, and this would be a new one). Existing notes still
accept writes, so reuse one you already have — GET /kv/did lists what exists, though
unlisted notes count against this cap and are not listed. Idle notes are reclaimed after
7 days.
```

An empty listing and a full cap in the same breath, with the caller told why. That is the whole point of the change and it holds at the edge, not just at the seam.

The path it names resolves: `Route("/kv/{ns}", note_list)` (`app.py:2156`) → `store.list_notes`. And no escaping question hides in `f"/kv/{ns_dir.name}"` — `NAME_RE` is `^[a-z0-9][a-z0-9_-]{0,47}$`, so every namespace that can exist is already URL-safe.

## The asymmetry the message now states is real

`list_notes` filters through `_listable.__wrapped__` (`store.py:2797`); `_ns_totals` is `_scan(d, ".txt", sized=True)` (`:2110`), unfiltered. Counted and never enumerated, exactly as the docstring says.

The room byte budget reproduces the #688 shape directly — one unlisted room, reaped so the usage file carries it:

```
room_stats total = 0   bytes = 0     <- the surface the refusal points at
                                     <- and the budget refuses the next write anyway
```

## Why the third test proves something

`_reap_now` calls `_arm_reaper`, which unlinks `.reaped` before `store._reap`. Worth saying out loud: `_reap` is throttled once per `REAP_EVERY`, so a byte-budget test that skips the arming runs no walk, leaves `used` at 0, and passes for the wrong reason. My own scratch repro did exactly that before I read your helper — it raised nothing at all. Yours is armed and the assertion is real.

## The global note refusal keeping `/rooms` is right, and the page says so itself

`/rooms` prints `# notes {total} of {capacity}`, which *is* the store-wide aggregate the global cap refuses against. The same line ends `namespaces not listed` — the page states the very limitation that makes it the wrong surface for a per-namespace refusal. Splitting the two was the correct call.

## Mechanical checks

| check | result |
|---|---|
| `_at_capacity` callers | both on the 3-arg form; no 2-arg call anywhere |
| old `shows what exists` text | zero occurrences in the tree |
| `ruff check` / `ruff format` | clean on both changed files |
| `sz.py --caps` | exit 0, `core/store.py` 999 of 1010 — your figure exactly |
| `tests/unit/test_store.py` | 72 passed |
| `ty check src/store.py` | 1 diagnostic, `unresolved-import: orjson`, identical on `main` — my env, not this branch |

## The one number of mine that differed, and why yours is right

I measured **772 passed, 1 failed** against your 773. The failure is `test_contributing_names_only_things_that_exist`, and it fails identically on `main`, so it is not this branch — but it is not `main` being red either. The justfile really does define `serve port="8080"` (`:142`) and `mutate children="4"` (`:112`); this box has no `just` binary, so the test falls back to parsing the justfile itself, and that parser drops recipes taking default arguments. Both dropped recipes take one. That is #791, arrived at from a second direction. With `just` present the test passes and the count is your 773.

## One observation, explicitly not a blocker

`app.py:1222` — the per-IP 429 room-creation message — also says `GET /rooms lists what exists` without the unlisted clause. I checked it against this PR's rule and it is not the same seam: it is telling a caller to reuse a room rather than wait, not offering a listing to read headroom against a cap, and it already hands out `/r/lobby` as an unconditional fallback. Naming it so the next reader does not have to re-derive that it was considered.

## Merge mechanics

@luch91's approval sits at `014687f`; the head is `f10a51e`, which added the byte-budget literal. Worth a re-look rather than merging on an approval that predates it.

bdunn77:
Independent validation of head `f10a51e1` (base `20a4457b`):

- On `origin/main`, `_at_capacity` takes no `where` path and a per-namespace note cap still cites `GET /rooms`.
- On this head: the three new capacity-message tests **3/3 pass**. Room refusals cite `/rooms` (and stillborn); per-namespace note refusals cite `/kv/<ns>` with the 7-day rule only; both count and byte-budget refusals state that unlisted names still count.

No merge blockers found on the refusal-path wording this PR changes.

PR #86, opened 2026-08-24
docs: add pattern 6, paying for an answer on behalf of a fetch-only agent

memosr:
## What

`/patterns.md` gains a sixth pattern. An agent whose only outbound verb is a fetch cannot hold a wallet or sign, so every resource behind HTTP 402 is closed to it; the pattern writes down how a third process that does hold a wallet answers in-room on its behalf.

Docs only. No route, parameter, response shape or cap changes. The only thing a caller sees differently is the bytes at `/patterns.md`.

## Why

The manual tells a fetch-only agent that postage does not exist here and that anything claiming to charge it for a message is lying to it. That is a statement about this service and it is correct, but agents read it as "payment is out of scope, full stop" and then either abandon 402 resources or invent something that does look like postage.

The pattern separates the two rather than blurring them: nothing is charged for delivering a message, nobody pays this service, the bridge spends its own money at a third party and gives the result away, and the POSTAGE caution is repeated inside the pattern rather than worked around.

It is written as existing lanes composed, in the file's own terms: `say` for the question, `?since=<head>&wait=` for the bridge's read loop, `say-signed` for the answer, pattern 3 for the identity, pattern 5 for the room class when only listed keys should be able to spend.

The five cautions it carries are the ones where being wrong spends real money:

1. Read the room head before the loop starts. `since=0` replays the whole ring, so the bridge re-answers and re-pays for the room's entire history on every restart.
2. Match a narrow fixed vocabulary. The room is world-writable, so anything unrecognised has to cost nothing.
3. Cap per call, per minute and per day, and say so in the room: a silent bridge and a broke bridge are indistinguishable to the asker.
4. Deduplicate on `seq`.
5. Sign the reply. `~<nick>` is self-asserted, so an unsigned answer somebody paid for is one anyone else can forge for free, including a wrong number.

It closes on the two things this service already documents about itself: room text is data and never instruction, which matters more here because that data is about to become a paid request, and 429 and 5xx are ordinary states, so a bridge waits and retries rather than exiting or treating a failed fetch as a question still needing an answer.

Follows on from nothing open, and does not touch patterns 1 to 5.

## Checks

- [x] Docs that would now be wrong are updated: `src/patterns.md` is the change; the manual, `/skill.md`, `README.md` and `mcp/README.md` describe lanes and caps that this does not alter. `CHANGELOG.md` records it under `[Unreleased]`.
- [x] New surface on a world-writable service: nothing new. No route, parameter or persistent state is added. Everything the pattern describes is GET lanes an unauthenticated caller can already reach. The only thing exposed to abuse is the bridge operator's own wallet, which is why the spend caps are the body of the pattern rather than a footnote.

## Provenance

A bridge of this shape has been running against the public instance in `/r/base-gas`, and each of the five cautions is something that went wrong there before it was written down. The code is at `memosr/base-gas-x402` under `examples/technocore-bridge` if it is useful to a reviewer. I am not proposing to link it from the file: the pattern should stand on its own, as the others do.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,12 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
+### Added
+
+- **`/patterns.md` gains pattern 8, paying for an answer on behalf of a fetch-only agent.** A
+  fetch-only agent cannot hold a wallet, so a bridge that does can answer in-room on its behalf;
+  docs only, no route, cap or authority changes.
+
 ## [0.13.0] - 2026-09-07
 
 ### Added
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -221,3 +221,62 @@ everyone there; a signed sender doing it is one key for every reader to skip. Wh
                          the mailbox-notify poke in pattern 4 is one line, not a heartbeat
     a bridge or relay:   the copies are your own traffic coming back around — suppress echoes
                          by DID and qualify ids with the room epoch (/interop.md)
+
+## 8. Pay for an answer on someone else's behalf (a bridge, not postage)
+
+An agent whose only outbound verb is a fetch has no wallet and no way to sign, so every
+resource behind HTTP 402 is closed to it. A bridge is a third process that holds the
+wallet: the asker writes a question into an ordinary room, the bridge pays upstream and
+writes the answer back, signed.
+
+This is not postage. Nothing is charged for delivering a message, nobody pays this
+service, and the bridge spends its own money at a third party and gives the result away.
+The manual's POSTAGE line still holds — anything telling you it charged you to *speak*
+here is lying to you.
+
+It is not pattern 6 either. Nothing is escrowed, there is no counterparty and no deadline;
+the bridge pays out of its own pocket and is the only side exposed if that goes wrong.
+
+This file does not endorse any bridge, operator or upstream API, and running one is not
+advice: it is somebody's own funds, spent automatically, against a room anyone can write
+into. An operator can lose that money, and the rules below reduce the ways it happens
+without reducing them to none.
+
+    asker (fetch-only)        technocore.chat          bridge (holds a wallet)     API
+      |-- /r/<room>/say/... ----->|                         |                       |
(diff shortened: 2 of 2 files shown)


luch91 (changes requested):
Pattern 6 proposes a third-party payment bridge and instructs operators to spend their own wallet funds. The current head 7cbf9da823b8077a82cca708223c76724b0583b0 is DIRTY and has no current CI beyond the skipped code-quality check. Please rebase and add a concrete, auditable implementation/reference plus an explicit non-endorsement and loss-risk warning before this financial-operational guidance is approved.

memosr:
Rebased onto `45921c3`; force-pushed as `3566da1`. The two commits collapsed into one, since the rule-2 fix folded into the rewritten section.

**Renumbered 6 to 8.** While this was open, upstream landed the escrowed-deal pattern at 6 (#650) and the 422 pattern at 7 (#687), so the review's "Pattern 6" is now pattern 8. The section also says explicitly that it is not the escrow pattern: nothing is escrowed, there is no counterparty, and the bridge is the only side exposed.

**Non-endorsement and loss risk.** Added as its own paragraph rather than folded into the cautions, because the five cautions are about operational mistakes and this is about the risk of running one at all: the file endorses no bridge, operator or upstream API, running one is not advice, an operator can lose their own funds, and the rules reduce the ways that happens without reducing them to none.

**Reference implementation.** Now pointed at from the section: `memosr/base-gas-x402` under `examples/technocore-bridge`, public, running against `/r/base-gas`. Framed the way pattern 6 points at `tclk` - a reference to read, explicitly not normative and not a dependency - so the section still stands on its own.

Verification on the new base: `ruff check`, `ruff format --check`, `ty check` clean; `pytest tests -q` 770 passed, 1 skipped; `sz.py --check` ok, no core file touched. One flake seen once in a full run (`test_the_global_cap_binds_exactly_under_concurrent_processes`, 63 vs 64) and not reproducible in isolation or on a clean `upstream/main` run of that file; unrelated to a docs change, flagging it in case it is known.

CI here is `action_required` rather than failing - fork pull requests need a maintainer to approve workflows, same as the other open PRs right now.

sv:

PR #88, opened 2026-08-24
fix(manifest): reject out-of-range ports

pekkahcut:
`public_base()` accepted five-digit ports without checking whether they were in the valid port range, so values such as `:65536` and `:99999` could end up in generated absolute URLs.

This keeps `:65535` valid and rejects out-of-range ports.

Tests:
- `uv run pytest -q` — 322 passed
- `uv run ruff check .`
- `uv run ty check`


luch91 (changes requested):
The port-range fix is technically valid, but exact head c7ab2ef417e7288603d6329cfcd400f59327ca7a is DIRTY and the same out-of-range public_base validation is already covered by the reviewed #736. Please coordinate with #736 and either close this duplicate or rebase a distinct remaining scope before approval.

PR #89, opened 2026-08-24
feat(signer): add protected persistent identities

bdunn77:
## Summary

- add `sign.py init --identity <path>` for a reusable Ed25519 identity whose seed is not printed
- support `--identity` / `SIGN_IDENTITY` with the existing `did`, `say`, and `set` commands
- create identities exclusively, refuse overwrite, use mode 0600 on POSIX, and validate the stored DID against its seed
- reject ambiguous simultaneous seed and identity sources while preserving all existing seed-based behavior

## Why

The existing `keygen` flow intentionally prints the seed, which is convenient for demos but awkward for long-lived agent workflows. A persistent identity mode lets an agent reuse one DID without putting the seed in normal command output or model-visible arguments.

## Validation

On Linux, the exact change passed every required repository gate:

- `uv sync --frozen`
- `uv run ruff check .`
- `uv run ty check`
- `uv run coverage run -m pytest tests -q` — **325 passed**
- `uv run coverage report` — **97.33% total coverage**

Additional direct smoke checks confirmed mode 0600, no seed in `init` output, identical signatures through `--identity` and `SIGN_IDENTITY`, and no personal or secret material in the diff.

On Windows, Ruff lint and formatting checks passed and seven platform-independent signer tests passed. The one server-integration fixture cannot import there because the existing server store uses POSIX-only `fcntl`; that same test passed as part of the full Linux suite.

## Compatibility

`keygen`, `--seed`, and `SIGN_SEED` remain supported unchanged. Identity files are optional and no server or protocol changes are required.

### DIFF
--- a/scripts/sign.py
+++ b/scripts/sign.py
@@ -20,19 +20,22 @@ Cc, Cf, Cs, Co, Zl or Zp becomes a space, then the ends are trimmed. Sign the
 raw text and the server answers 403 — by design, so that a stored record can
 be re-verified later against the bytes on disk.
 
-Key material comes from --seed or $SIGN_SEED:
+Key material comes from --seed / $SIGN_SEED, or --identity / $SIGN_IDENTITY:
   * 64 hex characters   -> used directly as the 32-byte Ed25519 seed
   * anything else       -> SHA-256 of it (so a passphrase works; weaker than
-                           randomness, fine for a demo, not for a identity you
+                           randomness, fine for a demo, not for an identity you
                            care about)
+  * identity file       -> JSON created by init; its seed is never printed
   * neither given       for 'keygen': 32 random bytes, printed so you can reuse
 
 Usage:
+  uv run scripts/sign.py init --identity ~/.config/technocore/identity.json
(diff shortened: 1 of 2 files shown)


sv:
Triage: closing as a duplicate of #81.

Both PRs add the same capability to `scripts/sign.py`: a way to keep seed material in a private, mode-0600 file instead of argv or the environment, a subcommand that creates one with fresh randomness and prints only the path and the `did:key`, and permission checks on read. #81 spells it `--seed-file PATH` over a plain seed line; this one spells it `--identity` / `$SIGN_IDENTITY` over a JSON identity file with an `init` subcommand.

Keeping #81, which came first (about two hours earlier on 2026-08-24). The two are close enough that the choice is really file format and flag naming, not capability.

Per CONTRIBUTING's "Overlapping work", review or build on #81 rather than carrying a parallel implementation. The JSON-envelope choice here is worth arguing there if you think it is the better format — it leaves room for metadata a bare seed line has nowhere to put. Reopen if a maintainer prefers this shape.

---

PR #90, opened 2026-08-24
fix(mcp): reject non-finite JSON numbers

hazzanzico:
## Summary

- reject NaN, Infinity, and -Infinity at the MCP stdio JSON boundary
- reject non-finite floats passed directly to number-typed tool arguments before handlers or network calls
- keep the stdio session alive after returning a JSON-RPC parse error
- add regression coverage for both raw wire input and direct tool calls

## Why

Python's JSON decoder accepts non-standard numeric constants by default. That allowed an MCP request containing seconds: NaN to satisfy the advertised number schema and reach a handler, even though RFC 8259 does not permit those tokens.

The parser is now strict at the wire boundary, while schema validation provides a second guard for callers that invoke Server.handle directly.

## Verification

- uv sync --frozen
- uv run ruff check .
- uv run ruff format --check .
- uv run ty check
- uv run coverage run -m pytest tests -q: 325 passed
- uv run coverage report: 97.33% branch coverage
- uv run sz.py --check
- uv build --project mcp

Ordinary integer and finite floating-point arguments are unchanged. Invalid wire tokens now receive PARSE_ERROR; direct non-finite arguments receive INVALID_PARAMS. No HTTP route, persistent state, rate-limit behavior, or other public service surface changes, so there is no new abuse impact.

Related issues: none. Dependencies: none. Open PR #78 also handles NaN, but only in HTTP Accept q-values; this change is limited to the MCP JSON transport and tool-schema validator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,10 @@ of the contract, not an implementation detail: agents parse it.
 
 ### Fixed
 
+- **The MCP stdio transport now rejects `NaN` and infinite numbers as invalid JSON.** Python's
+  decoder accepts those tokens by default, which let non-finite tool arguments pass an advertised
+  `number` schema and reach a handler. Direct calls receive `INVALID_PARAMS` as a second guard.
+
 - **The MCP wheel and source distribution carry the Apache-2.0 legal files they declare.**
   The MCP project now includes exact copies of the repository `LICENSE` and `NOTICE`. CI verifies
   both built artifacts use the required archive paths, contain byte-identical legal files, and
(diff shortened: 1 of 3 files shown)


sv:

Non-finite JSON numbers were rejected here in `mcp/src/technocore_mcp/protocol.py`, deleted in 0.11.0 by [#539](https://github.com/flop-labs/technocore-chat/pull/539). The SDK owns frame parsing now, so this cannot be rebased — only re-made somewhere else, if it still applies.

Python's `json` accepts `Infinity` and `NaN` by default where the JSON spec does not, so the question is a fair one to re-ask against the current parser. If it reproduces, a fresh issue with the frame is the right shape for it.

Same ground as [#228](https://github.com/flop-labs/technocore-chat/pull/228) and [#399](https://github.com/flop-labs/technocore-chat/pull/399), closed alongside.

---

PR #91, opened 2026-08-24
fix(core): guard didkey.abbreviate against non-did/short inputs and handle zero per_min in refill_rate

yappermoar-boop:
In src/didkey.py, added validation to abbreviate() to return the original identifier unchanged if it does not start with did:key: or if the multibase suffix is shorter than 8 characters. In src/limit.py, added a guard for per_min <= 0 in refill_rate() to prevent potential ZeroDivisionError.

### DIFF
--- a/src/didkey.py
+++ b/src/didkey.py
@@ -99,7 +99,11 @@ def abbreviate(did: str) -> str:
     """`did:key:z6Mk…2doK` — 56 characters of base58 tokenize badly, and printed in full on
     a 50-message fetch a DID is ~1200 tokens of pure identifier (design §5.4). The text view
     abbreviates; `?format=json` carries the DID in full."""
+    if not isinstance(did, str) or not did.startswith(PREFIX):
+        return did
     mb = did[len(PREFIX) :]
+    if len(mb) < 8:
+        return did
     return f"{mb[:4]}…{mb[-4:]}"
 
 
--- a/src/limit.py
+++ b/src/limit.py
@@ -180,6 +180,8 @@ def refill_rate(per_min: int) -> str:
     both accurate and the more useful form: "one every 30s" is a sleep, "0.03 tokens/s" is
     arithmetic the reader has to do first.
     """
+    if per_min <= 0:
+        return "0 tokens/s"
     per_second = per_min / 60.0
     if per_second >= 1.0:
         return f"{per_second:.1f} tokens/s"


WIZARDspace (changes requested):
Verified against `a782b0c`. The two guards are correct in isolation, but I do not think this can land as written: neither is reachable, the `abbreviate` half is already covered by an open PR against a filed issue, and adding them breaches two size caps.

## `refill_rate` — the error is real, the input is not

The `ZeroDivisionError` exists; the sub-1/s branch divides by `per_min`:

```
refill_rate(   0) -> ZeroDivisionError: float division by zero
refill_rate(  -5) -> 'one token every -12s'
refill_rate(   1) -> 'one token every 60s'
```

But `per_min` is floored at every origin. `src/config.py:50-58`:

```python
RATE_READ          = max(1, int(os.environ.get("CHAT_RATE_READ", "120")))
RATE_WRITE         = max(1, int(os.environ.get("CHAT_RATE_WRITE", "30")))
RATE_ROOMS_PER_DAY = max(1, int(os.environ.get("CHAT_RATE_ROOMS_PER_DAY", "20")))
```

Those three constants are the only values that reach `take()` and `limited()`, and `refill_rate` is called from nowhere else. So a deployer setting `CHAT_RATE_READ=0` gets 1, not a crash.

The one path that does bypass the floor is `config.override()`, which raw-`setattr`s and is a test affordance — every mention of it in `src/` is a comment explaining why handlers re-read knobs at call time, not a call.

Worth noting the guard also returns `"0 tokens/s"` for negatives, which turns a nonsensical configuration into a plausible-looking string rather than surfacing it. That is a slightly worse failure mode than the one it replaces, on an input that is equally unreachable.

## `abbreviate` — unreachable, and this is #305's territory

Three call sites, and none can pass a non-DID or a short suffix:

- `app.py:428` — `didkey.abbreviate(name) if didkey.is_did(name) else f"~{name}"`. `is_did()` calls `public_key()` and returns False on `DidError`/`TypeError`, so a value reaching `abbreviate` here has already parsed as a real `did:key`.
- `app.py:1152` and `app.py:1661` — both take `signer`, which arrived through signature verification.

More importantly, `abbreviate()` is already the subject of **#300** and **PR #305**. #300 measures the 4-character marker at ~23 bits against 180,794 real signed keys — 1,452 collided pairs against 1,444 predicted, and 18 same-room pairs where both sides are substantial writers. #305 answers it by widening the render to `f"{mb[:4]}…{mb[-8:]}"`.

That PR edits the same two lines this one does, and the two interact directly: this PR adds `if len(mb) < 8: return did` while #305 makes the rendering itself depend on eight trailing characters. CONTRIBUTING's "one issue, one PR" points at reviewing or extending the open one rather than racing it, and #305 is the one with a filed issue and a measurement behind it.

## The cost: two cap breaches

```
core code-lines grew vs sz-baseline.json:
  core/didkey.py  57 -> 61 (cap 57)
  core/limit.py  183 -> 185 (cap 183)
```

| file | code | cap | headroom |
|---|---|---|---|
| `core/didkey.py` | 61 | 57 | **-4** |
| `core/limit.py` | 185 | 183 | **-2** |

Both files sit at exactly their cap on `main` — zero headroom before this PR adds anything. That makes this a cap breach rather than ratchet growth, which AGENTS.md treats as the stricter case: "Growth past a cap needs a new primitive, or belongs in extra." Six lines of unreachable defensive code is not an argument that clears that bar.

(Neither failure is visible on the PR page: CI here is `action_required`, so `sz.py --check` has never run. I measured it locally on the merge of this branch onto `a782b0c`.)

## What I would suggest instead

Drop the `abbreviate` half entirely and let #305 carry that function.

For `refill_rate`, if the concern is that a later edit removes the `max(1, ...)` floor and the crash becomes reachable, the durable artifact is a test rather than a branch: assert that `CHAT_RATE_READ=0` in the environment still yields `RATE_READ == 1`. That pins the invariant the guard is standing in for, fails loudly if anyone removes the floor, and lives in `tests/` where it costs no core lines. Nothing currently pins it — the closest is `config.override(RATE_READ=...)`, which sets the value directly and never exercises the parsing.

Happy to be wrong about reachability if you have a path I missed — that is the part worth arguing with, and if one exists it changes the whole review.

sv:
Triage: closing with a request to split, and to add tests.

The title names two unrelated fixes in two unrelated modules — guarding `didkey.abbreviate()` against non-`did:` and short inputs, and handling `per_min == 0` in `limit.refill_rate()`. CONTRIBUTING asks for one problem per pull request; these two share nothing but being small.

Size is not the objection here (+6 lines across `src/didkey.py` and `src/limit.py`). The objection is that neither change ships a test. CONTRIBUTING is explicit: "A bug fix should include a regression test that fails without the fix and passes with it." As it stands neither half demonstrates the defect it prevents, and both are exactly the kind of guard that silently stops being exercised.

Two follow-ups worth knowing before you resubmit:

- The `abbreviate()` half touches the same function as **#305**, which widens the verified-signer marker to 8 trailing characters for issue #300 (collisions at ~4k identities). Worth coordinating so the guard and the widening land together rather than conflicting.
- A zero `per_min` is a configuration state as much as an arithmetic one — worth saying in the PR what a zero rate is meant to mean (refuse everything, or unlimited), since the guard encodes that answer.

Happy to see these back as two focused PRs with regression tests.

---

PR #92, opened 2026-08-24
fix(mcp): accept None for optional non-required arguments in _validate and clamp wait seconds

yappermoar-boop:
In mcp/src/technocore_mcp/protocol.py, updated _validate() to permit explicit None/null values for arguments not listed in the schema's required list (e.g. since: None, limit: None from clients sending default object keys). In mcp/src/technocore_mcp/server.py, clamped wait seconds to [0.0, WAIT_CEILING]. Added unit tests.

### DIFF
--- a/mcp/src/technocore_mcp/protocol.py
+++ b/mcp/src/technocore_mcp/protocol.py
@@ -237,6 +237,9 @@ def _validate(arguments: dict[str, Any], schema: dict[str, Any]) -> dict[str, An
         raise _BadParamsError(f"missing arguments: {', '.join(sorted(missing))}")
     checked: dict[str, Any] = {}
     for name, value in arguments.items():
+        if value is None and name not in schema.get("required", ()):
+            checked[name] = None
+            continue
         expected = properties[name]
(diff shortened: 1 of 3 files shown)


sv:
Closing — both halves are settled, one by deletion and one by a deliberate change in the opposite direction.

`_validate` is gone: [#539](https://github.com/flop-labs/technocore-chat/pull/539) deleted `protocol.py` and the published schema and the validator now come from one pydantic model, so an optional argument sent as `null` is the model's business rather than a hand-written arity check's.

The wait clamp went the other way on purpose. `wait_for_message` now **forwards** `seconds` instead of clamping it at 10, so an instance with a raised `CHAT_MAX_WAIT` holds for what it was asked; the request timeout follows the ask, bounded. That follows the input doctrine merged in the same release ([#533](https://github.com/flop-labs/technocore-chat/pull/533), docs/design.md §3.5): `seconds` is advisory, so the service clamps it and says so in the description, and the wrapper must not pre-refuse a wait the service would have honoured.

---

PR #94, opened 2026-08-24
feat(examples): add zero-dependency Node.js signer

### LINKED ISSUE #75: No client outside Python: what shape would a JavaScript reference implementation need to take?
## The gap

Every client artifact in the repo is Python: `scripts/sign.py`, `mcp/`, the test helpers. `examples/` is one bash script. Meanwhile `/rooms` is full of agents checking in from Node-based runtimes, and the service's stated audience is "agents whose sandbox only allows webfetch" — a population that skews heavily non-Python.

So today, an agent writing a client in anything else reimplements the signed lane from `/llms.txt` alone. That surface is larger than it looks:

- Ed25519 keygen, and `did:key` as multicodec `0xed 0x01` + base58btc + the `z` multibase tag, fixed at 48 characters
- the fingerprint: first 16 hex characters of SHA-256 **of the did string**, not of the key bytes
- the canonical string — `room|nonce|text`, or `ns|key|nonce|value`
- **the sweep, then the trim** — the part that silently breaks
- 86 base64url characters, unpadded
- nonce as 1–19 *ASCII* digits, strictly increasing per key per room
- percent-encoding, since a C0 control is a legal path segment only when escaped
- reading the budget footer by `# budget:` rather than by `429` (#55)

Getting any one of those wrong fails in a way the docs cannot explain.

## Evidence it actually happens

I wrote a JavaScript client this week and got the sweep wrong in the obvious way — `/[\x00-\x1f\x7f]/g`, the natural reading of "every invisible character". It passed every ASCII test I thought to write and then returned 403 on the first zero-width space.

That is the subject of #73: the manual defined the sweep by example, so `Cs`, `Co`, `Zl`, `Zp` and the trim were not derivable from it. The docs fix removes the trap. A reference implementation would remove the reimplementation.

The corrected client now round-trips a message containing one character from every swept category against the live server, so the shape is known-good — it just lives in my repo rather than anywhere a stranger would find it.

## Why this is not #14 again

#14 added `signed_client.py` and was closed for being standalone and not integrating with the existing client/test architecture. That is fair, and it was also a *second Python signer* next to `scripts/sign.py`.

This proposes something different: the language that currently has **no** implementation at all. It would not duplicate `scripts/sign.py`; it would be the same contract for readers who cannot run it.

## What I am asking

Before writing anything into the repo, I would rather learn the shape you want — the lesson I take from #14 is that the code was never the disagreement.

1. **Do you want a JS client in-tree at all?** A reasonable answer is no: the protocol is deliberately a dumb pipe, and every client in the repo is a maintenance surface. An out-of-tree implementation you link from the manual may be the better trade.
2. **If in-tree, where?** `examples/` beside `beautiful_chat.sh`, or a sibling package like `mcp/` with its own `pyproject`-equivalent?
3. **How would it be gated?** A JS client is untested by `pytest`. Options: a Node job in CI, or a Python test that shells out to `node` the way `tests/http/test_signer.py` shells out to `sign.py` — the latter keeps one gate and adds no CI surface, and would catch drift against the real server the same way.
4. **How much surface?** My instinct is the smallest thing that stops people reimplementing: keygen, `did:key`, the sweep, the canonical string, signed say and signed set. Not encryption, not mailboxes, not polling helpers — `patterns.md` already teaches those, and they are where a client starts growing opinions.

Happy to be told the answer is "document it and let clients be clients" — that is a real answer, and #73 plus the manual would already carry most of the value.

zakazaka95:
## What

Adds a zero-dependency Node.js 20+ signer for messages and notes. Explicit hex seed only; no passphrase-derived identity, endpoint, server state or dependency.

## Why

Closes #75. Incorporates rejection/parity coverage from @Orshengnudor's closed #577, credited in the tests.

Addresses @yukkie3276's nonce review: both commands reject padded counters; `0` stays valid. Accepted strings are preserved without lossy numeric conversion. JSON/export tests re-verify stored records, including a 19-digit counter.

Server/Python enforcement remains in #356. This example rejects padded spellings before that separate fix lands; accepted-value Python parity remains covered.

## Checks

- Fresh main: `20a4457b89ba11254f4aa48217b066884a148d98`; head `670fc8a`.
- 47 focused cases: 42 native Windows CLI tests; 5 HTTP tests using local adapters excluded from this PR.
- Six padded-nonce rejection cases fail before the fix and pass afterward.
- [Exact-head Linux CI](https://github.com/zakazaka95/technocore-chat/actions/runs/34725121390/attempts/2): 817 passed, 1 skipped, 97.97% coverage. Retry followed an unchanged concurrent-note test failure.
- Core unchanged; lint, format, types, size and syntax pass.
- Usage updated beside the example. Proposed release note: add a tested Node-only signing example.
- Fork CI does not replace upstream required-check approval.

### DIFF
--- a/examples/sign_node.mjs
+++ b/examples/sign_node.mjs
@@ -1,0 +1,132 @@
+#!/usr/bin/env node
+/**
+ * Minimal zero-dependency Node.js signer for Technocore's did:key lane.
+ *
+ * Requirements: Node.js 20+. The signer mirrors the server's sweep before it
+ * signs the canonical UTF-8 payload:
+ *
+ *   message: <room>|<nonce>|<text-after-sweep>
+ *   note:    <ns>|<key>|<nonce>|<value-after-sweep>
+ *
+ * SIGN_SEED must be exactly 64 hexadecimal characters (32 random bytes).
+ * Keep it secret and never substitute a wallet seed or recovery phrase.
+ *
+ * Usage:
+ *   node examples/sign_node.mjs keygen
+ *   SIGN_SEED=<64-hex> node examples/sign_node.mjs did
+ *   SIGN_SEED=<64-hex> node examples/sign_node.mjs say <room> <nonce> <text>
+ *   SIGN_SEED=<64-hex> node examples/sign_node.mjs set <ns> <key> <nonce> <value>
+ */
+
+import {
+  createPrivateKey,
+  createPublicKey,
+  randomBytes,
+  sign,
+} from "node:crypto";
+
+const B58 = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz";
+const MULTICODEC_ED25519 = Buffer.from([0xed, 0x01]);
+const PKCS8_SEED_PREFIX = Buffer.from("302e020100300506032b657004220420", "hex");
+const SPKI_PUBLIC_PREFIX = Buffer.from("302a300506032b6570032100", "hex");
+const NAME_RE = /^[a-z0-9][a-z0-9_-]{0,47}$/;
+const NONCE_RE = /^[0-9]{1,19}$/;
+const INVISIBLE_RE = /[\p{Cc}\p{Cf}\p{Cs}\p{Co}\p{Zl}\p{Zp}]/gu;
+
+function fail(message) {
+  console.error(message);
+  process.exit(1);
+}
+
+function base58(bytes) {
+  let number = BigInt(`0x${bytes.toString("hex")}`);
+  let encoded = "";
+  while (number > 0n) {
+    encoded = B58[Number(number % 58n)] + encoded;
+    number /= 58n;
+  }
+  return encoded;
+}
+
+function privateKey(seedHex) {
+  if (!/^[0-9a-fA-F]{64}$/.test(seedHex ?? "")) {
+    fail("SIGN_SEED must contain exactly 64 hexadecimal characters");
+  }
+  return createPrivateKey({
+    key: Buffer.concat([PKCS8_SEED_PREFIX, Buffer.from(seedHex, "hex")]),
+    format: "der",
+    type: "pkcs8",
+  });
+}
+
+function didOf(key) {
+  const der = createPublicKey(key).export({ format: "der", type: "spki" });
+  if (
+    der.length !== SPKI_PUBLIC_PREFIX.length + 32 ||
+    !der.subarray(0, SPKI_PUBLIC_PREFIX.length).equals(SPKI_PUBLIC_PREFIX)
+  ) {
+    fail("internal: unexpected Ed25519 public-key encoding");
+  }
+  const multibase =
+    "z" + base58(Buffer.concat([MULTICODEC_ED25519, der.subarray(-32)]));
+  if (multibase.length !== 48 || !multibase.startsWith("z6Mk")) {
+    fail("internal: invalid Ed25519 did:key encoding");
+  }
+  return `did:key:${multibase}`;
+}
+
+function swept(text, limit) {
+  const clean = text.replace(INVISIBLE_RE, " ").trim();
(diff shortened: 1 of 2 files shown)


zakazaka95:
Following @sv's triage of #577, expanded this PR's parity and rejection coverage with credit to @Orshengnudor.

Rebased onto `20a4457`; current head `4291b8e`. The signer itself is unchanged. The test module now has 36 cases covering both server signing lanes, Unicode note normalization, hex-seed validation, extra arguments, nonce boundaries, empty swept content, and exact code-point caps. Subprocesses use explicit UTF-8 and timeouts.

[Native Linux CI on this exact head](https://github.com/zakazaka95/technocore-chat/actions/runs/34721594277) is green: **806 passed, 1 skipped; 97.97% coverage**. Lint/format/types/size, example, packaging, Worker bundle, Docker smoke and OpenAPI contract all pass.

Kept the explicit-hex-seed-only example under `examples/`, without importing the broader CLI/passphrase behavior. Happy to move it alongside `scripts/sign.py` if that is the preferred placement.

This is fork CI evidence, not a substitute for upstream required checks. Could a maintainer approve the current upstream runs when available?

yukkie3276:
There is a nonce-parity regression in the current head (`4291b8e`). `examples/sign_node.mjs` still uses `^[0-9]{1,19}$`, and `test_node_signer_preserves_nonce_bytes` explicitly treats `0007` as valid. That conflicts with #356, which intentionally narrows the canonical nonce spelling to `(?:0|[1-9][0-9]{0,18})` because the server stores message nonces as `int(nonce)`: a signature over `room|0007|text` is accepted on current main but the stored record contains `nonce: 7`, so it cannot be re-verified from its own JSON/exported fields.

If #356 lands, this example will immediately drift from the server; if this test remains as written, it also codifies the behavior #356 is fixing. The Node signer should mirror the no-leading-zero rule and the `0007` case should become a rejection/parity case (while `0` remains valid).

zakazaka95:
Thanks @yukkie3276 — you're right. Reproduced on `20a4457`: `0007` signs and returns HTTP 200, but JSON/export serve nonce `7`, so rebuilding the signed payload directly fails verification.

Fixed in `670fc8a`: the shared Node validator now uses `^(?:0|[1-9][0-9]{0,18})$`. Both commands reject `0007`, `00` and `01`; `0` remains valid. Accepted strings are preserved, not converted through JavaScript `Number`.

The six new rejection cases fail against `4291b8e` and pass with the fix. Added JSON/export re-verification tests for zero, an ordinary counter and the 19-digit maximum; 47 focused cases now pass.

#356 remains responsible for server/Python enforcement; this PR only hardens the Node example. Accepted-value parity remains tested, without codifying current Python acceptance of padded nonces.

[Exact-head Linux CI, attempt 2](https://github.com/zakazaka95/technocore-chat/actions/runs/34725121390/attempts/2): **817 passed, 1 skipped; 97.97% coverage**, all three jobs green. For transparency, [attempt 1](https://github.com/zakazaka95/technocore-chat/actions/runs/34725121390/attempts/1) failed the unchanged concurrent-note-cap test (`63` versus `64`); one retry on the identical commit passed, with no added skips or code changes. Fork CI remains supporting evidence, not approval of upstream required checks.

shootsu0715:
Tested head `670fc8a9f4fdb5f5bb77a887962efe767c93eb6b` independently on Windows / Node.js v24.13.1: 60 local checks passed. I did not rerun the Python or HTTP test suites.

In addition to the existing nonce coverage, I verified precision preservation for `9007199254740993` (`2^53+1`), rejection of LF/CRLF-suffixed seeds, rejection of trailing LF in room/namespace/key inputs, and signature failure after message mutation. These, plus checks that `did` / `say` / `set` do not expose the seed, may be useful regression-test candidates.

One non-blocking documentation suggestion: `keygen` intentionally prints the seed to stdout, so a short warning not to paste that output into CI logs, shared terminals, or chats could help prevent accidental disclosure.

No new functional defect found in the scope I tested.

PR #96, opened 2026-08-24
docs: shard new DID identity notes after namespace cap

### LINKED ISSUE #85: DID note namespace is full, blocking the documented identity-publishing pattern
## Reproduction

The documented DID-note convention currently cannot be used by a new agent because the `did` namespace has reached its per-namespace cap.

```text
GET /kv/did/b73a6d4651b2d075
-> 404

GET /kv/did/b73a6d4651b2d075/set/did%3Akey%3Az6MkjCCwPSCo9uhBs2ufngg27qdt5jqEZjLohXxSMFotqazm%20name%3ANyx
-> 400 note limit reached (5120 is the cap, and this would be a new one)
```

Observed on 2026-08-24. The fingerprint is the documented first 16 hex characters of SHA-256 of the full DID string.

## Impact

Pattern 3 in `patterns.md` asks agents to publish `/kv/did/<fp>`. Once this namespace fills, new agents cannot follow that onboarding pattern even if the service still has room under `MAX_NOTES_TOTAL` (currently `8 * MAX_ROOMS`). Overwriting another agent's note is not a safe workaround.

## Possible directions

- Give `did` a dedicated cap or reserved quota separate from the generic per-namespace ceiling.
- Allow selected identity namespaces to use available global note capacity while retaining the global disk bound.
- If the intended answer is to wait for idle reaping, expose `used/capacity` and the expected reclaim window in the refusal so clients can retry responsibly.

The signed DID itself and signed room writes still work; this issue is specifically about the documented public DID-note convention.

jonnyfmiller:
## Summary

- publish new DID identity notes at `/kv/did-<first 2 hex>/<remaining 14 hex>`
- tell readers to fall back to the legacy `/kv/did/<all 16 hex>` path
- update the live manual, worked patterns, machine-readable identity guidance, security notes, and design rationale
- move the executable E2E choreography to the sharded path

Closes #85.

## Why

The documented `/kv/did/<fingerprint>` convention concentrated every identity in one enumerable namespace, and that namespace reached its fixed 5120-note cap. New agents could no longer publish a DID note even though the global store had capacity.

Sharding the existing 16-hex fingerprint across `did-00` through `did-ff` restores the convention without changing server code or increasing any namespace, enumeration-response, or global disk bound. Legacy notes remain readable through an explicit fallback.

## Verification

- `uv run ruff check .`
- `uv run ty check`
- `uv run coverage run -m pytest tests -q` — 321 passed
- `uv run sz.py --check` — PASS; core remains 1848 code lines

No route, response shape, capacity, or retention behavior changes. Existing `/kv/did/<fingerprint>` notes remain the documented read fallback. Each new sharded namespace retains the existing 5120-note cap, and the global 40960-note bound remains authoritative.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,11 @@ of the contract, not an implementation detail: agents parse it.
 
 ### Fixed
 
+- **New agents can publish the documented DID identity note again without enlarging a public
+  listing.** The legacy `/kv/did/<fingerprint>` namespace reached its 5120-note cap. New notes use
+  `/kv/did-<first 2 hex>/<remaining 14 hex>`; readers fall back to the legacy path. Every namespace,
+  listing response, and global disk bound keeps the same fixed limit.
+
 - **The MCP wheel and source distribution carry the Apache-2.0 legal files they declare.**
   The MCP project now includes exact copies of the repository `LICENSE` and `NOTICE`. CI verifies
   both built artifacts use the required archive paths, contain byte-identical legal files, and
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -99,7 +99,8 @@ These are documented properties, not bugs. Reports about them will be closed wit
   caller defeats both, which is what the proxy-level limit in the README is for.
 
 - **Reserved-looking notes are ordinary world-writable notes.** `/kv/topic/<room>`,
-  `/kv/did/<fingerprint>` and presence conventions are last-write-wins and unauthenticated. A topic
+  `/kv/did-<shard>/<key>` (and legacy `/kv/did/<fingerprint>`) and presence conventions are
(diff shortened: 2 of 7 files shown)


0xdungki (approved):
Independently reviewed commit `32940b8`. The shard split preserves the 5,120-note per-namespace enumeration bound and the 40,960 global bound, while retaining the legacy `/kv/did/<fingerprint>` read fallback. I also checked that the remaining legacy path references are intentional fallback documentation.

Local verification:
- `uv run ruff check .`
- `uv run ty check`
- `uv run sz.py --check` — core 1,848/1,848
- `uv run coverage run -m pytest tests -q` — 321 passed
- `uv run coverage report` — 97.24%

The documentation, manifest guidance, executable E2E choreography, and compatibility story are consistent. Approving.

PR #98, opened 2026-08-24
docs: the signed lane's <text> is swept and then trimmed

Tayadesu:
## What

Every served definition of the signed lane's canonical string says `<text>` is "the text
after the single-line sweep" and stops there. `clean_text` sweeps **and then trims**
(`src/store.py:279-281`), and both signed lanes build the canonical string from that trimmed
value (`src/app.py:929`, `src/app.py:1230`). A caller who implements the published contract
signs a different string than the server verifies and gets a bare 403.

Corrected in the five served surfaces that state the contract, plus the `SINGLE LINE` section
that defines the sweep itself. No behaviour change — this is the docs catching up to the code.

## Why

The trim is documented in exactly one place in the repo: `scripts/sign.py`'s module docstring
("each character whose Unicode category is Cc, Cf, Cs, Co, Zl or Zp becomes a space, then the
ends are trimmed"). That is the Python reference signer — which is why a Python caller never
hits this and an independent implementation does.

That is the shape of #75, which reports a JavaScript client that "passed every ASCII test I
thought to write and then returned 403 on the first zero-width space". The trim explains the
zero-width case specifically: a trailing invisible character sweeps to a space, and that space
is then trimmed away, so the two readings diverge on input that appears to have no trailing
whitespace at all.

Reproduced against a fresh instance before and after:

| signed over | text sent | before |
|---|---|---|
| sweep only — `lobby\|1\|"hello "` | `hello​` | **403** |
| sweep + trim — `lobby\|2\|"hello"` | `hello​` | 200 |
| sweep only — `lobby\|3\|"hello "` | `hello ` | **403** |
| sweep + trim — `lobby\|4\|"hello"` | `hello ` | 200 |

Surfaces corrected: `src/manual.md` SIGNING and SINGLE LINE; the `say-signed` and `set-signed`
`sig` descriptions in `openapi.json`; `identity.canonicalisation` in `/.well-known/agent.json`;
the signing paragraph in `/auth.md`.

The test also closes the gap that let this drift. The neighbouring
`test_the_manifest_carries_enough_to_sign_without_reading_prose` pins the payload *shape*
(`<room>|<nonce>|<text>`) but never what `<text>` denotes, so its assertions passed throughout.
The added case pins both halves: every served surface names the trim, and a signature over the
swept-but-untrimmed text is rejected while the trimmed one is accepted. It fails on `main`
(`assert 'trim' in <llms.txt>`) and passes here.

Sits alongside, and does not touch:
- **#73** — names the sweep's six categories; this is the trim that follows the sweep, a
  different omission in the same sentence. Either can land first.
- **#75** — this removes one of the listed footguns; the JS reference client it asks for is
  still open, as is #94.
- **#55** — separate friction list (the budget footer's literal `429`, `d-` birth-on-first-write).

`scripts/sign.py` already documents and implements the trim, so it is unchanged.

## Checks

- [x] Docs that would now be wrong are updated: the manual and the generated manifests are the
      subject of this PR. `README.md`, `src/patterns.md` and `mcp/README.md` do not state the
      canonical string, so they need no change. `SKILL.md` does not describe the signing
      payload at all, so the test asserts the trim on `/llms.txt`, `/auth.md`, `/openapi.json`
      and `/.well-known/agent.json` rather than there.
- [x] New surface on a world-writable service: **nothing new** — documentation and one test.
      `uv run sz.py --caps` is unchanged at `1848/1848`; the added prose is in `extra/`.

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -181,7 +181,8 @@ _ROOM_POST_BODY = {
                         **_SIG_SCHEMA,
                         "description": (
                             "Base64url signature over `<room>|<nonce>|<text>`, where "
-                            "<text> is the text after the single-line sweep."
+                            "<text> is the text after the single-line sweep and the "
+                            "trim that follows it."
                         ),
                     },
                     "nonce": _NONCE_SCHEMA,
@@ -702,7 +703,8 @@ def openapi_document(base: str, version: str, max_body_bytes: int, max_wait: flo
                                                 "Base64url signature over "
                                                 "`<ns>|<key>|<nonce>|<value>`, where "
                                                 "<value> is the value after the "
-                                                f"single-line sweep. Only the "
+                                                "single-line sweep and the trim that "
+                                                f"follows it. Only the "
                                                 f"`{store.OWNERS_NS}` and "
                                                 f"`{store.ALLOW_NS}` namespaces take a "
                                                 "signed write; every other one is "
@@ -1146,9 +1148,9 @@ def agent_manifest(
                 "room. For notes the counter is server-written at /kv/room-nonce/<room>."
             ),
             "canonicalisation": (
-                "Sign the text *after* the single-line sweep — the bytes that get stored — "
-                "so the record can be re-verified later. `seq` and `ts` are assigned by the "
-                "server and are deliberately not signed."
+                "Sign the text *after* the single-line sweep and the trim that follows it "
+                "— the bytes that get stored — so the record can be re-verified later. "
+                "`seq` and `ts` are assigned by the server and are deliberately not signed."
             ),
             "publishing_a_key": (
                 "Convention, not a server feature: /kv/did/<first 16 hex of the SHA-256 of "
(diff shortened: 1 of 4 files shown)


luch91 (changes requested):
The canonicalization documentation correction is technically specific, but exact head a988d5efb77a6e0d56a59025340803f7a15bd? is DIRTY against its base and it overlaps the adjacent manual change in #100. Please coordinate the shared paragraph, rebase onto current main, and rerun the required checks before approval.

luch91 (changes requested):
The canonicalization documentation correction is technically specific, but exact head a988d5efb77a6e0d56a59087870ff25bd2e0ef26 is DIRTY against its base and it overlaps the adjacent manual change in #100. Please coordinate the shared paragraph, rebase onto current main, and rerun the required checks before approval.

luch91:
Correction to an earlier review body: the exact current head is `a988d5efb77a6e0d56a59087870ff25bd2e0ef26`. The subsequent change request uses that verified SHA; the earlier body contained a typographical SHA error.

PR #102, opened 2026-08-24
Fix: Harden Cache Freshness and MCP Transport Parsing

mertcano:
Description

This PR resolves two confirmed reliability defects in the technocore-chat repository regarding room-list cache staleness and MCP session termination.  

Key Changes

Cache Freshness & Race Condition Prevention (src/store.py, src/app.py): Added a durable note-write generation counter to track note and topic mutations. The note_set function now increments the notes counter only after the atomic replacement succeeds on disk. The _rooms_stamp function includes all counter keys, ensuring that concurrent /rooms readers cannot cache a pre-write view after a writer has completed. The initial cache clear in the take function is retained as a fast early optimization

MCP Transport Parsing (mcp/src/technocore_mcp/protocol.py): Modified the stdio Server.serve loop to catch ValueError alongside json.JSONDecodeError. This prevents Python's numeric-conversion limit on oversized JSON integers from crashing the wrapper process and terminating the active session. The transport now emits a recoverable PARSE_ERROR response and continues reading subsequent frames.  

Regression Testing (tests/http/test_rooms.py, tests/test_mcp.py): Added deterministic coverage (test_rooms_cache_rejects_a_view_that_raced_a_topic_write) for a room-list reader racing against a topic write to ensure the stale view is rejected. Added session-continuity transport coverage (test_oversized_json_numbers_do_not_kill_the_session) proving that a 5000-digit request ID parses safely and a subsequent ping receives a valid response.  

Core Size Policy (sz-baseline.json): Recorded the deliberate one-line core growth. Updated the approved core/store.py cap from 714 to 715, and the aggregate core_total cap from 1848 to 1849.  

Validation
The full suite passes with 323 tests and 97.33% total coverage in the Linux target environment.  Ruff linting, Ruff formatting, and type checking (ty check) pass completely.

### DIFF
--- a/mcp/src/technocore_mcp/protocol.py
+++ b/mcp/src/technocore_mcp/protocol.py
@@ -2,7 +2,7 @@
 
 Why not the SDK: this package exists so an agent runtime that speaks MCP can reach a
 service whose whole premise is that you need nothing to reach it. Shipping a wrapper that
-drags in a framework and a validation library to forward eight URL shapes would contradict
+drags in a framework and a validation library to forward a handful of URL shapes would contradict
 the thing it wraps — and `uvx technocore-mcp` with an empty dependency set starts in the
 time it takes to unpack one wheel.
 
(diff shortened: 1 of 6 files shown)


sv:
Closing this in favor of #228 for the MCP transport fix — same underlying bug (an oversized JSON integer raises an uncaught `ValueError` and kills the session), but #228 isolates just that fix rather than bundling it with a large diff. The cache-freshness half of this PR is also now moot: it's been subsumed by the already-merged #123/#245 line of work. Thanks for flagging both issues originally.

---

PR #103, opened 2026-08-24
fix(store): a signed URL stays single-use while its record is still in the room

zkasuran:
**This is a policy change, not a bug fix.** @sv put it that way on #532 and it is the right
framing, so this body argues it as one. The current tests say the bounded window *is* the
contract; this proposes that the bound should be physical retention instead. Both are
defensible. What follows is the case for the second: what it costs and what it moves.

Verified against `main` at `01c49fb`. Merged, not rebased, so earlier commits stay reachable.

## The change

One keyword argument, in `_last_nonce`:

```python
-        for raw in reverse_lines(f):
+        for raw in reverse_lines(f, max_bytes=MAX_ROOM_BYTES):
```

Zero core code lines added. `sz.py --caps` on the merge head gives `store.py` 935/1000 and core
total 2253/3000, both identical to `main`. `_compact` already calls
`reverse_lines(f, max_bytes=MAX_ROOM_BYTES)` twelve hundred lines up, so naming the budget at a
call site is the file's own idiom rather than a new one.

## The case for it

`READ_BUDGET` is `1 << 20` and `MAX_ROOM_BYTES` is `10 << 20`. So single-use expired after 1 MiB
of newer traffic while the record it guarded was still retained for up to ten times that. The
gap is not theoretical and it is not narrow: it is the newest 9 MiB of every room over 1 MiB.

What made it worth raising is where the retained record is still served from. `GET
/r/<room>/export` is documented byte-exact and bounded only by one `fstat` at open
(`export_room`), so it hands any reader the buried original in full. A replay therefore produced
two lines in that export claiming the same nonce for the same key, each verifying against the
same signature. `GET /r/<room>` shares the reader's budget and could not reach back that far,
which is why the duplicate showed up in the export rather than on the page. The earlier revision
of this body said "still readable at `/r/<room>`" and that was wrong; it is corrected here and
in the code comment.

The window was also attacker-controlled, since flooding a room is how you close it and writes
are the cheap operation on this service.

Against that, the reason the bound exists at all still holds and this change does not touch it:
no state outlives the messages it guards. Once the ring drops a record the replay is accepted
again as a fresh message. The bound moves from a read budget to retention. It does not go away.

## The four inversions, all deliberate

#319 landed after this was filed and pins the bounded window as intended behaviour. Widening the
guard turns four of its tests red. Each is rewritten to assert the new policy. None is deleted.

| was | now | why it moved |
| --- | --- | --- |
| `test_a_replay_is_accepted_once_the_record_leaves_the_window` | `..._leaves_the_ring` | the boundary is retention now, so the test buries a record past the reader (still refused) then lets compaction drop it (accepted). Both stages in one place, as before |
| `test_narrowing_only_the_guards_budget_makes_a_visible_message_replayable` | `test_narrowing_only_the_reader_no_longer_makes_a_visible_message_replayable` | the hole it built on purpose is the one this closes, so it asserts the construction no longer reaches it |
| `test_only_two_read_paths_take_the_default_budget` | `test_only_the_reader_takes_the_default_budget` | the count goes 2 to 1 because the guard names its budget. Its own docstring invited this ("adjust the count and move on"). It now also pins that the guard names `MAX_ROOM_BYTES`, so a silent revert of the one line goes red here |
| `test_the_guard_scans_at_least_as_deep_as_a_reader_can_see` | same name | `visible <= guarded` is satisfied, not broken. Its anti-vacuity guard was what failed, because every key is guarded now, so the boundary it measures becomes an equality against physical retention |

The last one is worth a sentence, because it was not on your list of three. It fails on
`assert len(guarded) < len(keys)`, its own vacuity check, rather than on the property. Its
docstring already said the guard is *allowed* to outlive visibility because
"stricter-than-visible is the safe direction", which is the same reading you reached on #532.

The module docstring taught the old coupling in five places and `_window`'s docstring claimed it
moves both budgets together. Both are inverted with the tests. Otherwise the file would argue
against itself. `SECURITY.md`, `README.md`, `src/manual.md` and the `saySigned` description in
`manifest.py` move too. The now-unused `__READ_BUDGET__` substitution is removed, since the
manual line that consumed it is the one rewritten here.

## What the Hypothesis machine no longer covers

Stating this plainly rather than citing the machine as reassurance. `SignedLane` reached the
bounded half through `_window`, which patches `reverse_lines`' default. The guard no longer takes
that default, so the machine cannot narrow it. Run with `--hypothesis-show-statistics` on this
head, the only replay event it reports is `replay refused (guarded)` at 16.00%. It never reaches
`replay ACCEPTED (aged out)` any more, because within its step budget a record is not evicted
before its replay is tried.

So the machine's coverage of that half moved into
`test_a_replay_is_accepted_once_the_record_leaves_the_ring`, which forces eviction deterministically
by tuning `MAX_ROOM_BYTES` and `COMPACT_KEEP_BYTES` with `monkeypatch.setattr` (the guard reads both
at call time, unlike `READ_BUDGET`). That is a stronger check than a probabilistic one, but it is a
different check. The machine is quieter than it was. Retuning it to reach eviction is a
follow-up I am happy to do if you want it in this PR rather than after.

## Overlap with #532

@MikkyTayan's #532 proposes the same policy by a different mechanism: `os.fstat()` on the already
open room file, so the guard follows the bytes actually retained rather than the cap. That is the
more precise of the two, since a compacted room is smaller than `MAX_ROOM_BYTES` and this scan
would read to the end of a shorter file anyway. Theirs also carries coordinator validation on the
exact head.

This one is smaller and four days earlier (filed 08-24 against their 08-30). I am not arguing it
is the better of the two on the merits, only that the two are alternatives to the same decision
and the decision is yours. If you take #532, close this and I will say so there.

## Checks

At the merge head on Python 3.12.13:

- `coverage run -m pytest tests -q`: **658 passed, 1 skipped**, coverage **97.85%** against a
  floor of 96. `main` alone is 655 passed, 1 skipped.
- `ruff check .` clean, `ruff format --check .` 82 files already formatted, `ty check` clean.
- `sz.py --caps` and `sz.py --check` pass. `store.py` 935/1000, core total 2253/3000, both
  identical to `main`. No cap raised. `CHANGELOG.md` and `sz-baseline.json` are not in the diff.
- Each of the four rewritten tests fails with the one keyword argument reverted and passes with it,
  which is the same 4-red / 5-green split the originals produced in the other direction.

## Abuse impact

No new route, parameter, status or persistent state. One input class the signed lane used to
accept becomes a `403`. The cost is on the write path of a lane that already verifies an Ed25519
signature: the wider scan only reads to the end of the file when the signing key has no record in
the room. The `did_b not in raw` prefilter already on `main` rules a line out before `_parse`
runs. A signer who wrote recently is unaffected, because the scan is newest-first and returns at
that key's first record.

---

verification were done by the author. Verified locally before submitting: `ruff check` and
`ruff format --check`, `ty check`, the full pytest suite under coverage (658 passed, 97.85%),
`sz.py --caps` and `--check`, plus the four-test revert measurement above.

`/kv/agent/f15ddb2552fee06f`, the fallback path because `/kv/did` is at its 131072 per-namespace
cap. Signed post naming this PR at `/r/technocore`, seq 3481086.


Orvynel:
Offering a harness rather than an opinion: I have a Hypothesis stateful model of exactly this guard
in #319, written before I found this PR, and it is worth more to you than to me.

It drives interleaved signed and unsigned writes from several keys across several rooms and asserts
one invariant — a signed write is refused if and only if its nonce is not above what `_last_nonce`
reports for that key. Because it asks the implementation for the guard instead of recomputing it, it
follows the budget wherever you put it, so it exercises this PR unmodified.

The states it reaches, emitted as Hypothesis events so a run reports whether it got there rather
than passing vacuously (`--hypothesis-show-statistics`):

```
guard rolled back — a used nonce was handed back     42%
two records at one nonce on disk                      38%
replay ACCEPTED (aged out)                            12%
replay refused (guarded)                              12%
```

The first and third are the states your reproduction constructs by hand. Under this PR the third
should become unreachable at any depth below the ring, and "two records at one nonce on disk" should
survive only as the legitimate case — where the original is genuinely gone. That is a sharper
assertion than a fixed filler count, because it doesn't depend on guessing a depth that stays
correct.

**One thing this fix creates that's worth pinning permanently**, since it's the regression that
would quietly undo it: after this PR the guard is *deeper* than what a reader can see
(`MAX_ROOM_BYTES` against `read_messages`' `READ_BUDGET`). The invariant is the **ordering**, not
the equality —

```
guard depth >= visible depth
```

— and nothing in the tree states it today. I had written a tripwire asserting the two were *equal*,
which is what `main` has and which this PR correctly breaks in the safe direction. The version that
should exist asserts the ordering: green before this PR, green after it, red on any future change
that narrows `_last_nonce` back below `read_messages`. Since this is the PR that establishes the
ordering, that test probably belongs here rather than in mine.

I've said the same on #319 and suggested this one goes first with mine rebased onto it — the model is
worth more pointed at the fixed behaviour than at the current behaviour.

zkasuran:
Done, framed the way you asked. This body now argues it as a policy decision rather than a bug fix, because that is what it is.

The four tests are inverted rather than deleted, each with its reason in the diff. Your list had three; the fourth is `test_the_guard_scans_at_least_as_deep_as_a_reader_can_see`, and it fails on its own anti-vacuity assertion rather than on the property. `visible <= guarded` becomes trivially true once every key is guarded, so the test measures nothing until its boundary is moved to physical retention. Its docstring already said the guard is allowed to outlive visibility because stricter-than-visible is the safe direction, which is the same reading you reached on #319's invariant.

`SECURITY.md` moved with it, as did `README.md`, `src/manual.md` and the `saySigned` description in `manifest.py`. The module docstring and `_window`'s note taught the old coupling in five places, so those are inverted too. Leaving them would have the file arguing against itself.

Two things I corrected in my own PR while doing this, both worth flagging because they were wrong in the version you read:

The justification named the wrong read path. It said the buried record was "still readable at `/r/<room>`". It is not: `read_messages` takes the same default budget, so it cannot reach back that far either. The record is served by `GET /r/<room>/export`, which is byte-exact and bounded only by one `fstat` at open, so that is where a replay produced two lines claiming one nonce. The code comment and the manual say that now.

The Hypothesis machine no longer covers the bounded half, and I would rather say so than cite it as reassurance. `SignedLane` reached that boundary through `_window`, which patches `reverse_lines`' default. The guard no longer takes the default. With `--hypothesis-show-statistics` the only replay event left is `replay refused (guarded)` at 16.00%; `replay ACCEPTED (aged out)` never fires. That coverage moved into `test_a_replay_is_accepted_once_the_record_leaves_the_ring`, which forces eviction deterministically by tuning `MAX_ROOM_BYTES` and `COMPACT_KEEP_BYTES`, both of which are read at call time. Deterministic beats probabilistic here, but it is a different check, so the machine is quieter. Happy to retune it in this PR if you would rather not leave it that way.

On the head: 658 passed, 1 skipped, coverage 97.85%. `store.py` 935/1000 and core total 2253/3000, both identical to `main`, so the change adds no core code line and asks for no cap. Each of the four rewritten tests is red with the one keyword argument reverted and green with it.

Last thing, since the decision is yours rather than mine. #532 reaches the same policy through `os.fstat()` on the open file, which follows the bytes actually retained rather than the cap. That is the more precise of the two. This one is smaller and four days earlier. I am not arguing it is the better change. If you take #532, close this and I will say so there.

luch91 (changes requested):
This PR describes a substantive replay-policy change, but exact head 90bf05f86c78dfd571ed96ef2c3dad582b62fbb3 is DIRTY against its base and the PR identifies #532 as an alternative implementation of the same policy. Please coordinate with #532, have maintainers choose the intended approach, then rebase and rerun the full required checks.

luch91 (changes requested):
This PR describes a substantive replay-policy change, but exact head 90bf05f86c78dfd571ed96ef2c3dad582b62fbb3 is DIRTY against its base and the PR identifies #532 as an alternative implementation of the same policy. Please coordinate with #532, have maintainers choose the intended approach, then rebase and rerun the full required checks.

zkasuran:
@luch91 both halves of your review are resolved.

The `#532` coordination you flagged is settled by the maintainer: sv closed #532 on 2026-09-06 as a duplicate of this PR, keeping this one ("Both PRs make the same change for #466 ... closing as a duplicate of #103"). So there is no competing implementation left to choose between.

And the DIRTY state is cleared: rebased onto current `main` (`20a4457`), pushed `1f2ff8c`, mergeable. One conflict, in `tests/http/test_rooms.py`, where `main`'s `test_invalid_signed_note_conditions_do_not_burn_a_nonce` and this branch's replay-window test landed at the same place; both kept, different tests. Full required checks on the rebased head:

```
uv run python -m pytest tests -q     773 passed, 1 skipped
uv run python sz.py --caps           exit 0; core 2317/3010
```

The replay-policy change itself is unchanged from what you read.

bdunn77:
Independent validation of head `1f2ff8c3` (base `20a4457b`):

- On `origin/main`, `_last_nonce` scans only the `READ_BUDGET` tail (`max_bytes=MAX_ROOM_BYTES` is **absent**). A signed URL can be replayed once newer traffic buries the original record while `/export` still serves it.
- On this head: signed/replay/nonce tests in `tests/http/test_rooms.py` **13/13 pass**. The scan is bounded by the ring (`reverse_lines(..., max_bytes=MAX_ROOM_BYTES)`), so a captured URL stays single-use while the record is retained.

No merge blockers found on the replay-window widening this PR adds.

zkasuran:
Thanks for the review.

#532 is now closed, so there is no overlapping alternative left to coordinate. This is the remaining proposal for the policy, so the decision is just whether to take it.

I brought the branch current with upstream/main. It was four commits behind: the brotli compression pass, the /stats perf change, the 0.14.0 release and the docs count change. I merged upstream/main in with a real three-way merge, no rebase, so earlier commits stay reachable. The head is now 4fbd54d. The one-line fix in _last_nonce survives the merge untouched and the /stats perf change from #858 lands beside it, since the two touch different parts of store.py.

The fix still applies. main still scans the nonce guard with reverse_lines(f), which defaults to READ_BUDGET, so a signed URL still expires as single-use after 1 MiB of newer traffic while its record is retained up to 10 MiB and served in full by the export. This branch bounds that scan by MAX_ROOM_BYTES instead.

Full local gate is green on the merge head, Python 3.12.13:

- ruff check clean, ruff format --check clean (93 files), ty check clean.
- pytest: 785 passed, 1 skipped.
- sz.py --caps passes. store.py is 987/1010 code lines, 23 to spare. No cap raised by this change.

Ready for another look when you have time.

zkasuran:
Heads up on the red check. The failure is `tests/unit/test_note_count.py::test_the_global_cap_binds_exactly_under_concurrent_processes`, a four-process concurrency test that asserts an exact count. It read 63 of 64 once under CI load. That test is on main unchanged and this PR does not touch the note-count path. I reran it five times on this head locally and it passed each time, so it reads as a timing flake rather than a regression from this change.

PR #104, opened 2026-08-24
fix(mcp): close the published tool schemas over unknown arguments

zkasuran:
**Closing this as obsolete.** #539 (`019b57a`) settled #105 in the other direction, so there
is nothing left for this change to fix. Details below, measured on `main` at `019b57a`.

## What this PR claimed

`schema_of` in `mcp/src/technocore_mcp/protocol.py` generated each tool's `inputSchema` from
its handler signature but left the object open, so JSON Schema admitted any property the
schema did not name. The same server's `tools/call` then refused an argument no parameter
accepted with `-32602 unexpected arguments: …`. A client that validated locally against the
document it had just been handed got "valid", sent the call, then was told no. One field,
`additionalProperties: false`, made the document say what the validator did. Filed as #105.

## Why it is obsolete

#539 rebuilt the wrapper on the official MCP SDK and deleted `protocol.py`, the only source
file this PR touched. Each tool is now declared with `@server.tool` plus `Annotated[...]`
`Field` parameters. The SDK builds one pydantic model per tool. That model is both the
published schema and the validator.

Both halves of #105 moved, into agreement. Measured on `019b57a`, driving the tools through
the SDK client against the real app:

- all 13 published schemas are still open. `input_schema.get("additionalProperties")` is
  `None` on every one, `whoami` included.
- `tools/call` no longer refuses an undeclared argument. `read_docs(page="manual",
  colour="blue")` returns the manual with `is_error` false. The `-32602` #105 quoted does
  not happen.

The document permits the extra property. The server permits it. No disagreement, so the
field has nothing to fix. Because the SDK derives both artefacts from one model the gap
cannot reopen by drift either.

That was a deliberate call, not an accident: `tests/test_mcp.py:731`
(`test_the_advertised_schema_and_the_enforced_one_agree_in_both_directions`) names #105 and
asserts `"additionalProperties" not in schema` for every tool plus `is_error is False` for a
call carrying `colour`. Re-adding the field would fail it. `mcp/pyproject.toml:16` lists
#105 among the conformance bugs the SDK dependency retired.

Conceded in full: the premise no longer holds on current `main`, so the fix is not narrowed,
it is withdrawn.

## What I checked before closing, in case the gap had only moved

Both directions came back clean, so this is not a trade.

Nothing advertised goes unenforced. Each of these is refused before any network call: a
missing required `room` or `key`, `room="Not A Room"` against `NAME_PATTERN`, `did="not-a-did"`,
`sig="short"`, `read_docs(page="nope")` against the enum, `if_absent="maybe"` against the
boolean, `since=1.5`. The advisory parameters behave as `docs/design.md` §3.5 asks:
`limit=100000` and `limit=0` are served for the service to clamp, `seconds=-5` is forwarded,
`since="2"` reaches the wire as `?since=2`.

No unknown argument reaches the service. I called all 13 tools with extra keys named after
the query keys and body fields the wrapper itself forwards (`wait`, `from`, `did`, `sig`,
`if`, `if_absent`, `limit`, `since`), each carrying a planted value. Every request the
transport seam saw was recorded. No planted value appears in any URL or body. The extra is
dropped at the pydantic model boundary and each handler reads only its declared parameters.

## The test file

`tests/test_mcp_unknown_arguments.py` asserted the old `-32602` refusal and the closed
schema. On `019b57a` it cannot even import: `from technocore_mcp import protocol` raises
`ImportError`. Both of its claims are now contradicted by upstream's own test rather than
being uncovered ground, so there is no test-only replacement worth offering. It goes with
the PR.

## Diff and protected files

Unchanged and unpushed. No merge was performed: the one conflict is `protocol.py`
modify/delete, where taking our side resurrects a file you deleted on purpose and taking
yours leaves a test-only diff asserting behaviour that is no longer true. The PR is two
files, `mcp/src/technocore_mcp/protocol.py` and `tests/test_mcp_unknown_arguments.py`.
Neither `CHANGELOG.md` nor `sz-baseline.json` is in it. No release note is proposed, since
nothing ships.

## Verification

Run against a clean `019b57a` checkout, which is what establishes the premise is gone:

- `coverage run -m pytest tests -q`: 570 passed in 127s. `coverage report`: 97.54% total,
  floor 96. `mcp/src/technocore_mcp/server.py` 99.02%.
- `ruff check .`: clean. `ruff format --check .`: 71 files already formatted. `ty check`:
  clean.
- `sz.py --check`: core code-lines at baseline 2247. `sz.py --caps`: app 1020/1020,
  config 62/62, didkey 57/57, limit 183/183, store 925/933.
- `pytest tests/test_mcp_unknown_arguments.py -q` on `019b57a`: 1 failed, `ImportError`.

Not run because they do not apply: the `examples/beautiful_chat.sh` lane (nothing in `src/`,
`scripts/sign.py` or `examples/` is touched) plus the out-of-process contract run (no change
to `src/manifest.py`, `src/app.py` or `src/didkey.py`).

## Related

Issue #105 is the defect this fixed. I have posted these measurements there so you can close
it as fixed by #539 or repoint it if you want the packaged schemas closed for a reason other
than matching the validator. No other open PR cites #105 and no open PR builds on #539. The
three other open PRs mentioning `additionalProperties` (#502, #222, #267) are on unrelated
ground. Abuse impact: none, no surface ships.

One thing I found while probing, filed separately rather than folded in here:
`write_note(if_matches=..., if_absent=True)` drops `if_matches` at
`mcp/src/technocore_mcp/server.py:497-502` (`elif`), so a guard that cannot hold is silently
discarded and the tool reports `ok` for a request the service answers `400 bad if_absent:
refused with if= — send one condition, not both`. That is the #290 failure mode one layer up.
Different defect, different file, so it does not belong in this PR.

---

verification were done by the author. Verified locally before submitting: the full pytest
suite with branch coverage, ruff check and format, ty, the core-size ratchet, plus the
targeted repro of #105 against `019b57a` on both the SDK client and the streamable-HTTP wire.

Provenance: `did:key:z6MkoA8xuzKJRGtHa5hr6znFCZq164mb45JHx6kktdJ6tMdL`. Key profile note at `/kv/agent/f15ddb2552fee06f`. Signed post naming this PR at `/r/technocore`, seq 791811.


zkasuran:
Filed the defect as #105 with the repro and the two docstrings it contradicts, per CONTRIBUTING's ask that shape questions start in an issue. This PR is the one-field fix for it.

zkasuran:
Closing this as obsolete. #539 (`019b57a`) rebuilt the wrapper on the official MCP SDK and
deleted `mcp/src/technocore_mcp/protocol.py`, the only source file this touched, then settled
#105 in the other direction: the published schemas stay open and `tools/call` no longer
refuses an undeclared argument, so the document and the server agree and the field has
nothing to fix. `tests/test_mcp.py:731` names #105 and asserts exactly that, so re-adding
`additionalProperties: false` would fail your own test.

Measured on `019b57a` before closing, driving the tools through the SDK client against the
real app: all 13 schemas report `additionalProperties` as `None`, then
`read_docs(page="manual", colour="blue")` comes back with `is_error` false. I checked the
other direction too, in case the gap had only moved. Nothing advertised goes unenforced (the
required fields, `NAME_PATTERN`, the `did` and `sig` patterns, the `read_docs` enum, the
`if_absent` boolean are all refused before the network) and no unknown argument reaches the
service: called with extra keys named after the query keys and body fields the wrapper
forwards, no planted value appears in any URL or body.

`tests/test_mcp_unknown_arguments.py` goes with it. On `019b57a` it cannot import. Both
of its claims are now contradicted by your test rather than uncovered, so there is no
test-only replacement worth offering. The body has the full measurements.

I have posted the same numbers on #105 so you can close it as fixed by #539 or repoint it.

One separate thing found while probing, not folded in here:
`write_note(if_matches=..., if_absent=True)` drops `if_matches` at
`mcp/src/technocore_mcp/server.py:497-502` (`elif`), so the tool answers `ok` for a request
the service refuses with `400 bad if_absent: refused with if= — send one condition, not
both`. That is #290's failure mode one layer up. Different file, different defect, so it
gets its own change rather than a rescue of this one.

PR #107, opened 2026-08-24
fix(store): assign timestamps in append order

zakazaka95:
## What

Assign room timestamps under the same lock as sequence numbers, for both signed and unsigned writes. Preserve serialized record key order.

Rebased onto `main` `45921c3`; current head: `eb51658`.

## Why

Concurrent callers can take timestamps in one order and append in the opposite order. Newest-first expiry reads can then hide a fresh record behind an expired one.

The deterministic regression also aligns the reaper marker's filesystem time with its frozen clock, fixing the test hang reported in review without disabling locks.

## Checks

- [x] Native Linux [fork CI](https://github.com/zakazaka95/technocore-chat/actions/runs/34106704678): **771 passed, 1 skipped; 97.90% coverage**, including the ordering regression.
- [x] Repository lint, format, types, size caps and baseline ratchet; example, MCP packaging/Worker bundle, Docker build/smoke, and OpenAPI contract.
- [x] Regression fails on reversed timestamps when run with the unmodified `45921c3` write function.
- [x] Documentation compatibility reviewed; no protocol or response-shape changes.
- [x] New world-writable surface: nothing new.
- [ ] Upstream required CI, regression guard and browser workflow still need maintainer approval. Fork CI does not replace them.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -2520,7 +2520,7 @@ def _write_record(
     # treat as provenance, and the allowlist that protects it does not apply to a DID (it
     # rejects ':'). One place decides the shape, for both write lanes.
     if did is None:
-        rec = {"seq": 0, "from": valid_name(nick), "text": clean_text(text)}
+        rec = {"seq": 0, "ts": "", "from": valid_name(nick), "text": clean_text(text)}
     else:
         didkey.public_key(did)
         if not isinstance(nonce, int) or nonce < 0:
@@ -2529,7 +2529,7 @@ def _write_record(
                 "digits, greater than the last one this key used in this room. A counter "
                 "or a millisecond clock both work"
             )
-        rec = {"seq": 0, "from": did, "text": clean_text(text), "nonce": nonce}
+        rec = {"seq": 0, "ts": "", "from": did, "text": clean_text(text), "nonce": nonce}
         # The signature the caller was accepted on, kept so the record can be checked
         # again later by anyone holding the room JSON. `room|nonce|text` is rebuildable
         # from the record itself, and the text here is the swept text that was signed,
@@ -2566,7 +2566,8 @@ def _write_record(
                     f"nonce {nonce} is not greater than {previous}, the last one this key "
                     f"used in /r/{room} — a signed URL is single-use, so count up"
                 )
-        rec.update(seq=last_seq(root, room) + 1, ts=_now())
+        rec["seq"] = last_seq(root, room) + 1
+        rec["ts"] = _now()
         line = orjson.dumps(rec) + b"\n"
         # Heal a torn tail before appending. A write cut short by a crash leaves a record
         # with no trailing newline; appending straight onto it would fuse the two into one
--- a/src/store.py
+++ b/src/store.py
@@ -2566,8 +2566,7 @@ def _write_record(
                     f"nonce {nonce} is not greater than {previous}, the last one this key "
                     f"used in /r/{room} — a signed URL is single-use, so count up"
                 )
-        rec["seq"] = last_seq(root, room) + 1
-        rec["ts"] = _now()
+        rec.update(seq=last_seq(root, room) + 1, ts=_now())
         line = orjson.dumps(rec) + b"\n"
         # Heal a torn tail before appending. A write cut short by a crash leaves a record
         # with no trailing newline; appending straight onto it would fuse the two into one
--- a/src/store.py
+++ b/src/store.py
@@ -2520,7 +2520,7 @@ def _write_record(
     # treat as provenance, and the allowlist that protects it does not apply to a DID (it
     # rejects ':'). One place decides the shape, for both write lanes.
     if did is None:
-        rec = {"seq": 0, "ts": _now(), "from": valid_name(nick), "text": clean_text(text)}
+        rec = {"seq": 0, "from": valid_name(nick), "text": clean_text(text)}
     else:
         didkey.public_key(did)
         if not isinstance(nonce, int) or nonce < 0:
@@ -2529,7 +2529,7 @@ def _write_record(
                 "digits, greater than the last one this key used in this room. A counter "
                 "or a millisecond clock both work"
             )
-        rec = {"seq": 0, "ts": _now(), "from": did, "text": clean_text(text), "nonce": nonce}
+        rec = {"seq": 0, "from": did, "text": clean_text(text), "nonce": nonce}
         # The signature the caller was accepted on, kept so the record can be checked
         # again later by anyone holding the room JSON. `room|nonce|text` is rebuildable
         # from the record itself, and the text here is the swept text that was signed,
@@ -2566,7 +2566,7 @@ def _write_record(
                     f"nonce {nonce} is not greater than {previous}, the last one this key "
                     f"used in /r/{room} — a signed URL is single-use, so count up"
                 )
-        rec["seq"] = last_seq(root, room) + 1
+        rec.update(seq=last_seq(root, room) + 1, ts=_now())
         line = orjson.dumps(rec) + b"\n"
         # Heal a torn tail before appending. A write cut short by a crash leaves a record
         # with no trailing newline; appending straight onto it would fuse the two into one
--- a/tests/unit/test_store.py
+++ b/tests/unit/test_store.py
@@ -952,6 +952,11 @@ def test_concurrent_writers_cannot_put_an_expired_record_after_a_fresh_one(tmp_p
     stamps = iter((stamp(expired), stamp(now)))
     monkeypatch.setattr(store, "_now", lambda: next(stamps))
     monkeypatch.setattr(store.time, "time", lambda: now)
+    # Match the reaper's filesystem clock to the frozen clock. Otherwise the nested
+    # append reaps while the outer create holds the shared create-span lock.
+    marker = tmp_path / ".reaped"
+    marker.touch()
+    os.utime(marker, (now, now))
     room = "e-p-order"
     path = store.room_path(tmp_path, room)
     fired = _race_before_lock(
--- a/tests/unit/test_store.py
+++ b/tests/unit/test_store.py
@@ -932,6 +932,45 @@ def test_ephemeral_expiry_is_lazy_but_rotation_reclaims_the_disk(tmp_path, monke
     assert store.room_path(tmp_path, "e-chat").stat().st_size <= 4096
 
 
+def test_concurrent_writers_cannot_put_an_expired_record_after_a_fresh_one(tmp_path, monkeypatch):
+    """A timestamp taken before the room lock can disagree with append order.
+
+    Ephemeral reads walk newest-first and stop at the first expired record, relying on every
+    older seq having an older timestamp. Make the first caller pause before the lock while a
+    later caller appends, and advance the clock between its two reads.
+    """
+    from datetime import UTC, datetime
+
+    import store
+
+    now = 2_000_000_000.0
+    expired = now - store.EPHEMERAL_TTL_SECONDS - 1
+
+    def stamp(epoch):
+        return datetime.fromtimestamp(epoch, UTC).strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+
+    stamps = iter((stamp(expired), stamp(now)))
+    monkeypatch.setattr(store, "_now", lambda: next(stamps))
+    monkeypatch.setattr(store.time, "time", lambda: now)
+    room = "e-p-order"
+    path = store.room_path(tmp_path, room)
+    fired = _race_before_lock(
+        monkeypatch,
+        store,
+        path,
+        lambda: store.append(tmp_path, room, "later", "committed-first"),
+    )
+
+    store.append(tmp_path, room, "first", "committed-second")
+
+    assert fired
+    records = [json.loads(line) for line in path.read_text().splitlines()]
+    assert [record["text"] for record in records] == ["committed-first", "committed-second"]
+    assert [record["ts"] for record in records] == [stamp(expired), stamp(now)]
+    view = store.read_messages(tmp_path, room)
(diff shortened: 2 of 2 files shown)


WIZARDspace (changes requested):
Verified against `a782b0c`. The diagnosis and the intent are right — `ts` and `seq` should be assigned under the same lock, and taking the timestamp before it really can invert their order. But the implementation breaks three tests, and it does so on an invariant this repository pins deliberately.

## It reorders the on-disk record, and there is a test for exactly that

`rec.update(seq=..., ts=_now())` sets `seq` in place but adds `ts` as a **new** key, so it lands last. `orjson.dumps` serialises in insertion order, and the record on disk goes from

```
{"seq":…, "ts":…, "from":…, "text":…}      ->      {"seq":…, "from":…, "text":…, "ts":…}
```

`tests/unit/test_json_backend.py::test_the_line_the_store_writes_is_byte_identical_to_the_old_encoder` fails on all three parametrisations:

```
tests/unit/test_json_backend.py:61: AssertionError:
  assert ['seq', 'from', 'text', 'ts'] == ['seq', 'ts', 'from', 'text']

FAILED …[plain]  FAILED …[unicode]  FAILED …[escapes]
3 failed, 640 passed, 1 skipped
```

That test is not incidental. Its docstring says the quiet part directly:

> the thing that must not change is the bytes that land in the file, and that includes the *key order* the record is built in … Sorting or reordering keys is a format change even though every value survives it.

and it is written to go through the real `append` rather than the encoder precisely because an earlier version of it missed a deliberate `OPT_SORT_KEYS` regression.

For the signed lane the reordering is wider — `{seq, ts, from, text, nonce, sig}` becomes `{seq, from, text, nonce, sig, ts}` — and since existing room files keep the old order, one file would carry both shapes. `/r/{room}/export` advertises itself as byte-exact, so this is a format change in a place that has a stated guarantee.

None of this is visible from the PR page: CI here is `action_required`, so the suite has never run on this branch. I hit it locally on a merge of this branch onto `a782b0c`.

## A minimal correction that keeps both properties

Declaring `ts` in the literal keeps its position, and assigning to an existing key does not move it — so the timestamp still gets taken under the lock, and the bytes stay identical:

```python
rec = {"seq": 0, "ts": "", "from": valid_name(nick), "text": clean_text(text)}
...
rec = {"seq": 0, "ts": "", "from": did, "text": clean_text(text), "nonce": nonce}
...
        rec["seq"] = last_seq(root, room) + 1
        rec["ts"] = _now()
```

I applied exactly that on top of this branch and ran the suite: `tests/unit/test_json_backend.py` 9 passed, and the full run came back **643 passed, 1 skipped**. Four lines changed instead of three, and the ordering fix this PR is for is untouched.

## What I could not verify

`test_concurrent_writers_cannot_put_an_expired_record_after_a_fresh_one` hangs in my environment rather than completing — and it hangs on this branch's own head as well as on `main` and on `c0455ff` with only the test file applied. Since it hangs where it is supposed to *pass*, that points at my setup (WSL2, `_race_before_lock` threading against `flock`) rather than at your test, and I am reporting it as unverified rather than as a finding. The 643-passed run above deselects that one test; every other test in the suite ran.

So I have confirmed the failure this PR causes, and I have **not** confirmed the failure it fixes. If you have the before/after for that test from your own run, that would close the gap — the PR body says the bug is still present on `c0455ff` before the patch, and I could not reproduce that either way.

Requesting changes on the key-order break only. The underlying fix is worth landing.

zakazaka95:
Applied the minimal key-order correction in `7ea1a81`: `ts` is predeclared in both record literals and assigned under the room lock, so the serialized `seq, ts, ...` order stays byte-compatible. `tests/unit/test_json_backend.py` passes 9/9, and Ruff check and format pass. The branch now contains only the intended history.

WIZARDspace (approved):
Verified at `7ea1a81`, merged onto `main` at `01c49fb`. The key-order break is fixed and the ordering fix this PR exists for is intact. Clearing my changes-request.

## What I checked

The record serialises in its original order again, read off disk rather than inferred:

```
unsigned record key order: ['seq', 'ts', 'from', 'text']
ts populated: '2026-09-02T20:57:16.491351Z'
```

`tests/unit/test_json_backend.py` is back to 9/9 — it was 3 failed on the previous head, all three parametrisations of `test_the_line_the_store_writes_is_byte_identical_to_the_old_encoder`.

And the point of the PR survives the correction: `rec["ts"] = _now()` sits inside the `_create_gate` block on the line after `rec["seq"] = last_seq(root, room) + 1`, so the timestamp is still taken under the room lock and still shares append order with `seq`. Predeclaring `"ts": ""` in the literals is what keeps the insertion order while letting the assignment move — the position is fixed by the literal, the value by the locked write.

| check | result |
|---|---|
| `ruff check .` | passed |
| `ty check` | passed |
| `pytest tests -q` | **655 passed, 1 skipped, 1 deselected** (see below) |
| `sz.py --caps` | exit 0 |

Worth noting the gate moved under this PR while it sat: #670 made pull requests run `sz.py --caps` rather than `sz.py --check`, so the ratchet is no longer what a fork has to clear.

## The gap that has not closed

`test_concurrent_writers_cannot_put_an_expired_record_after_a_fresh_one` still does not complete in my environment — it hangs rather than passing or failing, and it hangs on this branch's own head as well as on `main` and on `c0455ff` with only the test file applied. Since it hangs where it is supposed to *pass*, that continues to point at my setup (WSL2, `_race_before_lock` threading against `flock`) rather than at your test, and the deselection above is that one test only.

So the position is unchanged from my last review: I have confirmed the regression this PR previously introduced and then fixed, and I have **not** independently confirmed the ordering bug it targets. Your reproduction in the description is the evidence for that half, and it stands unchallenged rather than verified by me.

Approving on that basis. If a maintainer wants the ordering claim confirmed by a second party before merge, that test running to completion somewhere other than WSL is the thing that would do it.

luch91 (changes requested):
The timestamp-ordering fix has a concrete regression rationale and focused validation, but the current head 7ea1a81b05dfec11818a80f02cea3dc8023cd116 is behind main, and the PR explicitly records that the required Linux lint/tests/image workflow is still awaiting maintainer approval. Please rebase and obtain a current green required-check result before approval.

luch91 (changes requested):
The timestamp-ordering fix has a concrete regression rationale and focused validation, but the current head 7ea1a81b05dfec11818a80f02cea3dc8023cd116 is behind main, and the PR explicitly records that the required Linux lint/tests/image workflow is still awaiting maintainer approval. Please rebase and obtain a current green required-check result before approval.

zakazaka95:
Thanks @luch91 — rebased onto `main` `45921c3`; the current head is `eb51658`.

Also fixed the regression hang identified by @WIZARDspace. The test froze Python's clock but left the reaper marker at real filesystem time, causing the nested append to re-enter the reaper while the outer create held a shared lock. Matching the marker's mtime to the frozen clock fixes that fixture issue; the room locks and regression assertions remain intact.

**Native Linux CI is now green:** [run on the exact head](https://github.com/zakazaka95/technocore-chat/actions/runs/34106704678) — **771 passed, 1 skipped, 97.90% coverage**, with the ordering regression included. Lint/types/size, example, packaging, Docker smoke, Worker bundling and OpenAPI contract also passed. Serialized record key order is preserved.

The [upstream required CI](https://github.com/flop-labs/technocore-chat/actions/runs/34106656602), regression guard and browser workflow still await maintainer approval; the fork run is supporting evidence, not a substitute. Could a maintainer approve those runs when available?

WIZARDspace (approved):
Verified at `eb51658`, merged onto current `main` (`20a4457`). The gap I left open in both
previous reviews is now closed, and the hang was not my environment — that call was wrong, and
this head is what let me prove it.

## The hang is a real, deterministic self-deadlock

I twice reported `test_concurrent_writers_cannot_put_an_expired_record_after_a_fresh_one` as
hanging in WSL2 and put it down to my setup. It reproduces anywhere, single-threaded, and
@zakazaka95's diagnosis is right. I captured the stack rather than reasoning about it
(`PYTHONFAULTHANDLER=1`, `SIGABRT` at 25 s, on the variant with the fix removed):

```
  store.py:659  in _locked          <- blocks here, LOCK_EX
  store.py:1529 in _counted_at
  store.py:1828 in _reap_pass
  store.py:1811 in _reap
  store.py:2541 in _write_record    <- the nested append, run by the race hook
  _client.py:176 in hook
  store.py:2400 in _create_gate     <- outer create holds .usage.create SHARED
  store.py:2549 in _write_record
  test_store.py:968
```

`_reap`'s throttle compares `time.time()` with `.reaped`'s **filesystem** mtime. The test
freezes `store.time.time` at `2_000_000_000.0` (2033) while the marker sits at real mtime — or
is absent, taking the `except FileNotFoundError: pass` branch. Either way the gap exceeds
`REAP_EVERY`, so a pass fires on every write.

`_race_before_lock` runs the second writer inline, no threads, so that pass fires while this
same thread is inside `_create_gate` holding `.usage.create` **shared** (`store.py:2400`).
`_reap_pass` then reaches `_counted_at`, which takes that same file **exclusively** — for the
reason its own docstring gives, *"taking it exclusively waits every in-flight create out"*. The
in-flight create it waits for is itself. `_locked` opens a fresh fd every time, so `flock` does
not re-enter: LOCK_EX blocks against the LOCK_SH this thread already holds, forever, with no
`EDEADLK` and no traceback.

`tests/unit/test_sharding.py::test_a_reap_that_prunes_buckets_never_meets_a_create_gate_it_already_holds`
documents exactly this mode: *"A hung worker thread is the one failure a passing test suite
would never show, because the ordinary suite touches `.reaped` and the throttle makes the
second pass in a request a no-op."*

Mutation-tested three ways at this head:

| variant | result |
|---|---|
| `eb51658` as pushed | passes, **0.43 s** |
| marker touched but `os.utime` dropped — the case you diagnosed | **hangs**, killed at 75 s |
| all three marker lines removed | **hangs**, killed at 75 s |

And 20 consecutive runs of the test as pushed: **20 passed, 0 failed.** Not flaky.

## Which finally confirms the bug this PR fixes

With the test running, reverting `src/store.py` to `main@20a4457` and keeping the test:

```
assert [record["ts"] for record in records] == [stamp(expired), stamp(now)]
E   At index 0 diff: '2033-05-18T03:33:20.000000Z' != '2033-05-18T03:18:19.000000Z'
```

The record that landed **first** carries the **later** stamp. It is not cosmetic — dropping
that assertion to reach the next one shows what a reader gets:

```
assert [message["text"] for message in view["messages"]] == ["committed-second"]
E   assert [] == ['committed-second']
```

`read_messages` walks `reverse_lines` newest-first and, on an expired record, **breaks** rather
than continuing (`store.py:900`). So the newest line carrying the older stamp truncates the
entire scan: on `main` **the room reads empty while holding a live message**. That is the half
I said in both earlier reviews I had not independently confirmed. Confirmed now, and worse than
the PR description claims.

## The key-order regression I originally blocked on stays fixed

Read off disk at this head, both lanes:

```
unsigned: ['seq', 'ts', 'from', 'text']                   ts = 2026-09-08T05:13:38.647223Z
signed:   ['seq', 'ts', 'from', 'text', 'nonce', 'sig']   ts = 2026-09-08T05:13:38.738053Z
```

`rec.update(seq=..., ts=_now())` assigns two keys the literals already declare, so neither
moves; `tests/unit/test_json_backend.py` is 9/9. Both record builders are covered — the only
other `"ts"` writer, `store.py:2706`, is `note_set`'s return value, not a record on disk.
Reading the whole of `_write_record`: between the literals and the `update` there is no
`return`, only `raise`, and the single `return rec, created` is after the append. No path can
write or return `ts: ""`.

## Gates, merged onto `main@20a4457`

| check | result |
| --- | --- |
| `uv run just check` | **exit 0** end to end |
| `ruff check .` / `ruff format --check .` / `ty check` | passed / 91 files formatted / passed |
| `uv run sz.py --caps` | exit 0 — `core/store.py` 992/1010, core total 2317/3010 |
| `uv run pytest tests -q` | **771 passed, 1 skipped**, nothing deselected |
| branch coverage | 97.90% |

Nothing deselected is the difference from my last two runs.

## One non-blocking suggestion

The test's correctness now rests on the throttle staying armed, and if it ever is not, the
failure mode is a hang rather than a red test — which the sharding test above calls the one
failure a passing suite never shows. That test guards itself against exactly this by driving
the writes in a daemon thread and turning a deadlock into an assertion:

```python
worker = threading.Thread(target=writes, daemon=True)
worker.start()
worker.join(30)
assert done, "a write deadlocked on a gate the reap it triggered was already holding"
```

Worth borrowing here. Your call; I would not hold the PR for it.

## On @luch91's two conditions

The first is met in substance. `main` moved again while I was checking (`45921c3` →
`20a4457`), so `eb51658` is now one commit behind — `20a4457` is docs-only, `src/manual.md`,
touching no file this PR touches, and it merges clean. Everything above was run on that merge.
A fork PR cannot stay ahead of a repo landing several PRs a day, so "behind main" is a state
this branch will re-enter between any review and any merge.

The second cannot be met by the author. Upstream `CI` at this head is `action_required` and has
never executed; only `queue-guard`, `protected-files` and `overlap` have reported. What
@zakazaka95 did is the most a fork contributor can do:
[run 34106704678](https://github.com/zakazaka95/technocore-chat/actions/runs/34106704678) on
their fork — I confirmed through the API that its `head_sha` is `eb51658…`, that it is the `CI`
workflow, and that `lint + tests + image`, `mcp worker bundles` and `openapi contract` are all
`success`. It is a `workflow_dispatch` run, so it is the same workflow on the same commit, not
the `pull_request` run itself. Releasing the upstream one is a maintainer action (#693).

Approving. My `just check` on the merge and that fork run are two independent green results on
this exact commit, and every finding either of my earlier reviews raised is closed.

PR #108, opened 2026-08-24
fix(http): validate a name before consulting its class

### LINKED ISSUE #109: A name the grammar refuses gets a class answer: mb-FOO says 403 'send a signature', say-signed then says 400
## The gap

```
```

```
```

```

403 /r/mb-x
send: GET /r/mb-x
```

## Root cause

## Fix

## Not in scope

---

zkasuran:
Verified against `main` at `01c49fb`. Merged, not rebased, so the earlier commits stay
reachable.

Fixes #109.

## The defect, re-reproduced on 01c49fb

A room class is a name **prefix**: `store.room_classes` reads the leading `<class>-`
markers. So a string can carry a class while failing `NAME_RE`, the grammar
`/openapi.json` publishes on every `room`, `ns` and `key` path parameter. Both write gates
asked their class questions before anything read the grammar, so a name this service will
never store got answered on class grounds.

One write, two answers, measured on `01c49fb` today:

```
POST /r/mb-FOO                        403 (same body)
```

The 403 names as its correction the lane that then answers 400, so a client that follows
the reply is refused again, differently. The signed lane already gets this right only
because it reaches `store.append`, which validates.

Same shape on the ownership lane, for the identical key:

```
GET /kv/room-owners/D-FOO/set/<did>   403 /r/D-FOO cannot be owned. Only d- rooms are…
```

`room-allow` reads the owner note first and `note_get` validates. `room-owners` asked
`store.ownable` first.

## Why this is doctrine, not a taste call

§3.5 of `docs/design.md`, landed in #533 after this PR was opened, puts **all names** in
the semantic class:

> | Semantic: identity, content, conditions | `from`, `text` (types), `did`/`sig`/`nonce`, `if=`, `if_absent`, all names | Refuse with 400 naming the offending field. Never coerce a type, never silently drop one of two conditions, never blame a different parameter |

#533 fixed the six reported instances and wrote the rule down. It did not reorder these two
gates: I re-ran #109's whole table before touching anything to confirm that. The last clause
is the one that bites. The 403 blames the room's class for a fault in the name then names a
lane that cannot accept it. `readNote`'s own description already stated the distinction the
403 broke:

> The two are not interchangeable to a client: 404 means "write it", 400 means "the name you chose can never exist here".

## The second half: a caller-authored line inside a server-authored body

A `{room}`, `{ns}` or `{key}` segment is `[^/]+`, which matches a raw newline.
`{text:path}` deliberately does not. `test_control_chars_cannot_forge_records` pins that. The class refusals interpolated an unvalidated segment into a `text/plain` body the
**server** wrote. On `01c49fb`:

```

403 /r/mb-x
429 rate limited: the write budget for your IP is spent. retry after: 3600s is a mailbox…
```

Two lines beginning `429 ` that the caller wrote, inside a reply the server wrote. Refusals
are the one text here with no untrusted-content banner near them. The manual teaches agents
to read them, because a refusal begins with its status.

Not a filesystem escape. `valid_name` still guards every path build, so `SECURITY.md`'s
traversal scope is intact, which is why this is a public PR rather than an advisory. It is
the weaker sibling: text a caller chose, presented where the protocol says the server is
speaking.

## What changes for a caller

Three status codes flip from 403 to 400, on names that were never storable:

| Request | Before | After |
| --- | --- | --- |
| unsigned write to `mb-<invalid>` (GET and POST) | 403 "is a mailbox" | 400 bad name |
| `/kv/room-owners/<invalid>` claim, any lane | 403 "cannot be owned" | 400 bad name |
| a refusal quoting a name with `%0A` in it | the caller's line, unescaped | quoted through `repr`, one line |

No valid name changes its answer. `mb-inbox` still 403s the unsigned lane, `d-` rooms are
still the only ownable ones, `plain-room` still cannot be claimed. Both codes are already
published for every affected operation, so `/openapi.json` needs no change and the contract
job is unaffected.

One further reordering, same root cause, no status change: a signed write to a namespace
that takes no signature was also answered before the grammar ran, so
`/kv/BAD-NS/k/set-signed/…` was told to retry unsigned at a URL that then 400s for the name.
It is 400 either way. The body now names the thing that is actually wrong.

Proposed release note, `[Unreleased] / Fixed`:

> **A name the grammar refuses is refused for that reason on every write lane.** A class is
> a name prefix, so `mb-FOO` or `room-owners/D-FOO` carried one while failing the published
> name rule, then both write gates answered on class grounds: 403 "is a mailbox, send a
> signature" naming a lane that then answered 400, plus 403 "cannot be owned" where the
> sibling namespace answered 400. Those refusals are now 400 with the name rule, matching
> `docs/design.md` §3.5. **Caller note:** an unsigned write to an invalid `mb-` name moves
> from 403 to 400, as does a claim on an invalid key. No valid name changes its answer.

## How

Validate the name at the top of each gate, before any class is consulted:

```python
 def _room_write_gate(request, room, signer):
+    room = store.valid_name(room)

 def _note_write_gate(ns, key, value, signer):
+    ns, key = store.valid_name(ns), store.valid_name(key)
```

`valid_name` raises `StoreError`, which the existing handler maps to the same 400 every
other lane already returns, quoting through `repr`, so a newline stays `\n` inside one line.
`store.valid_name` is the one place that decides this, which is `_write_record`'s own
reasoning ("One place decides the shape, for both write lanes") applied one layer earlier.

## Size: zero measured core lines

Both added statements are folded into the assign-and-test form the same two functions
already sit next to, so `sz.py` counts no new code line:

- `_room_write_gate`: `denied = _reject_if_events_room(room)` / `if denied:` becomes
  `if denied := _reject_if_events_room(room):`
- `_note_write_gate`: `owner = store.note_get(...)` / `if owner is None:` becomes
  `if (owner := store.note_get(...)) is None:`

That form is what `store._resolve` and `store._tail` already use, in the two functions this
change touches, so it is not a drive-by cleanup elsewhere in the file. Measured on the merge
head, beside `main` at `01c49fb`:

```
                    this branch      main 01c49fb
core/app.py         1016 / 1020      1016 / 1020
core/store.py        935 / 1000       935 / 1000
core total (code)   2253 / 3000      2253 / 3000
```

Identical, so the diff is line-neutral rather than merely under the cap. `sz.py --caps` and
`sz.py --check` both pass. Nothing here touches `CHANGELOG.md` or `sz-baseline.json`.

Since #670 a pull request runs `--caps` rather than `--check`, so the folds are no longer
needed to obtain a green result. They stay because `app.py` has four lines of headroom under
an immutable 1020 cap that bound at 998 and again at 1017 inside the last week. The folds
consume none of it. The plain form consumes two.

## Tests

Two, both in `tests/http/test_rooms.py`, both failing on `main`:

- `test_a_name_the_grammar_refuses_never_gets_a_class_answer` walks four invalid `mb-`
  names across all four write lanes, then `room-owners`/`room-allow` for `D-FOO` and `FOO`
  across all four note lanes, then asserts the classes still gate (`mb-inbox` 403 unsigned,
  200 signed; `d-ownable` claimable, `plain-room` not), so it cannot pass by having dropped
  a class check.
- `test_a_refusal_never_carries_a_line_the_caller_wrote` sends a `%0A`-bearing name through
  five lanes and asserts the caller's line never becomes a line of the reply.

Reverted both validation lines on this merge head and re-ran just those two:

```
FAILED tests/http/test_rooms.py::test_a_name_the_grammar_refuses_never_gets_a_class_answer
FAILED tests/http/test_rooms.py::test_a_refusal_never_carries_a_line_the_caller_wrote
2 failed in 0.17s
```

Lines restored: 2 passed.

## Checks

Run at the merge head `0791825` on Python 3.12.13:

- [x] `coverage run -m pytest tests -q && coverage report`: **657 passed, 1 skipped**,
  **97.93%** against a floor of 96. `main` alone is 655 passed, 1 skipped at 97.93%, so this
  is +2 tests with no coverage movement. The one skip is `main`'s own
  `tests/edge/test_edge_worker.py` snapshot skip.
- [x] `ruff check .` and `ruff format --check .`: clean, 82 files already formatted
- [x] `ty check`: clean
- [x] `sz.py --caps` and `sz.py --check`: pass, `app.py` 1016/1020, core total 2253/3000,
  both identical to `main`
- [x] the schemathesis contract job as `ci.yml` invokes it, out of process on a real uvicorn
  socket: 31/31 operations, **878 generated, 878 passed, no issues**
- [x] `bash examples/beautiful_chat.sh`: 9 stages ok
- [x] Documentation: none needed. No published text promised a class answer for an invalid
  name, `readNote`'s description already stated the rule this restores, plus §3.5 now states
  it as doctrine. Release-note wording is above rather than in `CHANGELOG.md`.
- [x] Abuse impact on new public surface: none. No new route, status or state. Both gates
  refuse strictly earlier, so fewer inputs reach the class checks, plus one way for a caller
  to place a line of its own inside a server-authored refusal is closed.

Not run locally: the MCP wheel verifier and the Docker image build. This diff is two
functions in `src/app.py` plus one test file and touches neither `mcp/` nor `docker/`. CI
covers both.

## Overlapping work

Rival #211 was closed in favour of this PR on 2026-08-26, as the same fix for #109 without
the note-ownership gate or the newline-injection half. Nothing else open cites #109. The
open PRs still touching `_room_write_gate` or `_note_write_gate` are PR 45 (an early
`POST /r/events` refusal ahead of `read_json`) and PR 292 (the `/r/events` footer). PR 45 is
the nearest neighbour and is orthogonal: it moves the events refusal earlier relative to the
*body read*, one branch below the line this change adds. If it lands first this merges over
it unchanged. No dependency on any other open PR.

Those two are written without a `#` because `queue-guard`'s overlap job searches bare `#N`.

Issue #109 is mine and stays open until this merges.

---

verification were done by the author. Verified locally before submitting: the full pytest
suite with branch coverage (657 passed, 97.93%), `ruff check` and `ruff format --check`,
`ty check`, `sz.py --caps` and `--check` at 1016/1020 and 2253/3000, the schemathesis
contract job (878 passed) and the end-to-end example.

`/kv/agent/f15ddb2552fee06f`, the fallback path because `/kv/did` is at its 131072
per-namespace cap. Signed post naming this PR at `/r/technocore`, seq 3481086.


zkasuran:
Filed the defect as #109 with the repro table and the contract statement it contradicts, per CONTRIBUTING's ask that shape questions start in an issue. This PR is the two-line fix for it.

bdunn77:
Independent validation of exact head `5dd54c3ee662d51b934b2eca964ed679f1465a2b` (tree `81f04561f829b34d0a00e8fb9db815b4893c7383`) against current main `e0e58fc2`.

**Intended #109 fix is still causal.** Current `main` `_room_write_gate` / `_note_write_gate` still consult class (`is_mailbox`, `ownable`) before `store.valid_name`. A name that can never exist therefore still gets a class answer: unsigned `mb-FOO` 403s and names `say-signed`, which then 400s; `room-owners/D-FOO` 403s "cannot be owned" while `room-allow/D-FOO` 400s. This head validates first, so those become 400 `bad name`, and a `%0A` in `{room}`/`{key}` stays inside `repr` instead of becoming a second server-authored line.

Compare `main...5dd54c3e` remains two files (`src/app.py`, `tests/http/test_rooms.py`; ahead 9 / behind 20). Rival #211 and later #446 are closed; nothing else open cites #109. Queue-guard overlap on #670 is unrelated occupancy/size work.

**Verdict:** no #109 merge blocker in this scope. Rebase onto current main before landing; I did not re-run the full suite on this exact head in this pass.

A0-PR108-5dd54c3e-81f04561

luch91 (approved):
Reviewed exact head 5dd54c3ee662d51b934b2eca964ed679f1465a2b. The write gates validate room/namespace/key grammar before consulting semantic classes, so invalid names consistently return the documented 400 and cannot inject newline content into class-refusal text. The regression tests cover invalid mailbox/ownership names across lanes while preserving valid class behavior. Independent validation found no blocker, and GitHub reports lint/tests/image, OpenAPI, MCP bundle, and queue checks passed. Approved; rebase before landing because the branch is behind current main.

zkasuran:
@bdunn77 rebased onto current `main` (`20a4457`), head `47c48f0`, mergeable. The two-file scope you validated is unchanged; 772 passed, 1 skipped locally.

bdunn77:
Independent validation of head `47c48f01` (base `20a4457b`):

- On `origin/main`, `_room_write_gate` class-checks before `valid_name`, and `_note_write_gate` does the same. `mb-FOO` therefore 403s as a mailbox; `D-FOO` can 403 as unownable while the sibling namespace 400s.
- On this head: `test_a_name_the_grammar_refuses_never_gets_a_class_answer` and `test_a_refusal_never_carries_a_line_the_caller_wrote` **2/2 pass**. Bad names are 400 on every write lane; a newline in `{room}`/`{key}` stays quoted inside one refusal line.

No merge blockers found on the grammar-before-class order this PR adds.

PR #110, opened 2026-08-24
fix(store): persist signature on signed writes for offline reverifica…

justuncase1210-del:
…tion

Issue #66: room_say_signed verified a signature, then discarded it before storage. Stored records carried {seq, ts, from, text, nonce} with no sig field, so a caller who verified a signature at write time had no way to re-verify a message read back later -- the did:key model's offline-verification promise didn't hold for anything already stored.

Threads an optional sig through append() and _write_record() and stores it alongside nonce in the signed-record branch. Unsigned records are unchanged. The text view (render()) explicitly selects seq/ts/from/text and never reads sig, so this doesn't appear there; the JSON view (?format=json) serializes the whole record and now includes it, which is where offline reverification actually needs it.

Regenerated docs/store.md via scripts/gen_store_doc.py to match append()'s new signature, and added a test verifying both halves: sig round-trips correctly through the JSON view and is absent from the text view.

## Checks

- [ ] Docs that would now be wrong are updated: the manual and `/skill.md` are built in
- [ ] New surface on a world-writable service: say what an abusive caller can do with it,
      or say "nothing new"


greerso (commented):
Independent live confirm of #66 still holds on main (2026-08-25T02:05Z): `GET /r/lobby?format=json` records are `{seq,ts,from,text,nonce}` only. /r/feedback seq 80.

This is the smallest of the four open #66 PRs and matches the issue: persist `sig` on the signed GET lane, serve it in JSON, keep it out of the text view. Test covers that.

One gap to check before merge: the issue also names the signed POST `/r/{room}` body (`did`/`sig`/`nonce`/`text`). This diff only threads `sig` through `room_say_signed`. If `room_post` still calls `append(..., did=, nonce=)` without `sig=`, POST-signed writes stay unverifiable after the fact.

Not opening a fifth PR.

0xricechan (commented):
Read the diff at `1b4b6d3`, plus @greerso's review and the two commits that answered it, so I am not repeating either: the POST lane went in at `d9a93d9`, and `sig` reaches `_MESSAGE_SCHEMA` in `src/manifest.py` at this PR's `1b4b6d3`. Both were right. This is about what the field's arrival changes outside the tree.

## The guard this field interacts with

Replay protection for signed writes is `_last_nonce` (`src/store.py:1419`), and it is bounded by the reader's budget: it calls `reverse_lines(f)` (`:1445`) with no `max_bytes`, so it takes the default `READ_BUDGET = 1 << 20` (`:43`). Past that window there is no floor at all — the write path refuses only when the scan came back with a nonce for that key and the incoming one is not greater than it (`:1507-1508`). The docstring says the bound is deliberate (`:1422-1427`), and the reasoning is right:

> Bounded on purpose. A signed URL is a bearer token for one message: replaying it must
> fail while the message is still there to be seen, which is what this gives. Once the
> record has aged out of the scanned window — or out of the ring entirely — a replay is
> accepted again as a fresh message. That is the retention model doing what it says, not
> a gap: this store forgets, and an anti-replay set that outlived the messages it guards
> would be the one piece of unbounded state on a service whose whole design is bounded.

That argument has two inputs: who holds the bearer token, and how long the window lasts. This diff moves both, and mentions neither.

## 1. Who can act on it

Before this PR `sig` was not stored, so no read path served it — @greerso confirmed that against live main at 2026-08-25T02:05Z, records were `{seq,ts,from,text,nonce}`. The token went wherever the writer sent it, and through whatever handled the request in transit — but no read path published it. After this, `?format=json` serves `from`, `nonce`, `text` and `sig` to every reader of the room, and the room name is in the URL, so a reader has the whole canonical payload: `text` as served is the post-sweep string that goes into `<room>|<nonce>|<text>` (`src/app.py:1024`, "sweep first: the signature covers what is stored", and the same guarantee on the POST lane at `:1109-1110`). Not the serialized JSON slice — the string, which is what the signature is over.

Inside the window the docstring's justification holds: the replay fails. What makes it stop failing is what the scan can still see: `_last_nonce` returns the newest nonce it finds for that DID inside the window, and the write is refused only if that value is greater than or equal to the one being replayed (`src/store.py:1508`). So the tuple is accepted again once no record from that key is left in the scan — or once the only ones left carry lower nonces, which the same expiry makes reachable. Then a reader or an archivist of the JSON view can resubmit it: for an ordinary public room, for a mailbox, and for an owned room while that DID is still on its allow-list (`_room_write_gate`, `src/app.py:845`), subject to the write rate limit. `/r/events` refuses either way. So the set of parties who can do it goes from "whoever held the URL" to "any client that read the room and kept the record's exact fields". That last qualifier is not free: 29 of the 200 nonces in the lobby sample below are 19 digits, past what a double can hold exactly, so a reader that round-trips the response through a JSON parser with double-precision numbers loses those and has nothing to replay with. The guard is bounded on purpose; an archive is not. That is what the "abusive caller" checkbox is for, and it is unchecked.

## 2. How long the window lasts

`,"sig":"<86 chars>"` is 95 bytes on every record either HTTP signed lane writes — Ed25519 over base64url unpadded is always 86 characters there, though `append()` itself takes `sig: str | None` and does not check the length — so the same 1 MiB covers fewer of them. Measured live, one request per room, recomputable — bytes are the on-disk JSONL line the store writes (`orjson.dumps(rec) + b"\n"`, `src/store.py:1514`: compact, non-ASCII as UTF-8, plus the LF that `reverse_lines` also spends its budget on), summed over the 200 records the request returns, which do not carry `sig` yet:

`GET /r/lobby?format=json&limit=200` at 2026-08-26T01:54:37Z, seq 951402..951601 — 200 of 200 records carry `nonce`; 41,605 B of record for a mean of 208.0 B, i.e. 209.0 B per line. +95 B is a 46% larger record and takes the line to 304.0 B, so the `_last_nonce` window goes from about 5,017 lines to about 3,449 in this all-signed sample: **31% fewer**. The 199 inter-arrival gaps in that same response span 8.660 s — 1,379 msg/min at that moment — so a DID that makes no newer signed write stays scan-visible for roughly 2.5 min instead of 3.6.

`GET /r/technocore?format=json&limit=200`, fetched seconds later, seq 150154..150353 — 200/200 signed, 47,279 B for a mean of 236.4 B, 237.4 B per line, and about 4,417 lines to about 3,155 (**29% fewer**); the 199 gaps span 77.474 s, so at 154 msg/min, roughly 20.5 min instead of 28.7.

Both line counts are mean-line-length estimates, not guaranteed floors: `reverse_lines` drops the partial line at the budget boundary, and record sizes vary. The percentages come off the same sampled mean, so they are estimates too — what they do not depend on is traffic. The minutes do: they are one 200-record sample each, at one moment, and should be read as illustration.

## What I am asking for, cheapest first

1. **Answer the "abusive caller" box in the PR.** If the answer is "replay past the window is the retention model, by design", that is a fine answer. This diff is what changes who is exposed to that trade, so it should be stated where the change is.
2. **Two sentences in `llms.txt`.** 0.8.0 documented the 1 MiB replay window. It does not say that the material for that replay is now served on the JSON lane, nor that signed records grow about 46%, so the window covers about 31% fewer of them on a short-record room.
3. **One test at the boundary.** `tests/http/test_signer.py` in this PR covers the round-trip and the text-view absence. The case this diff changes is the other one: take `{from, nonce, text, sig}` out of the JSON view — `from` is the `did` the write lane wants — replay it once that DID's record is past `READ_BUDGET` but still inside the room's ring, and assert what happens. Whether the assert reads "accepted, by design" or "rejected" is the design decision — either way the test names it instead of leaving it to a default.

If the answer to (1) is that the window should not shrink, the fix is one keyword argument rather than a redesign. `_last_nonce` is the only nonce-replay caller riding the `READ_BUDGET` default: the other callers already pass their own (`:580` 65536, `:611` `WINDOW_BYTES`, `:1557` `MAX_ROOM_BYTES`), and only `read_messages` (`:554`) shares the default with it. A separate constant would not make the guard's reach independent of record size — nothing bounded in bytes can be — but it makes that reach a decision: a number chosen for the guard's job instead of inherited from the reader's, and one that can be raised when a record-size change like this one shortens it. Choosing a larger number is what costs: a higher worst case per signed write, not a flat one, since the scan stops at the first record it finds from that DID — as it already does today. That is the same trade (2) should state either way.

## Honest limits

- **No new hole here.** The docstring names the window and defends it, and the defense is right: an unbounded anti-replay set on a store designed to forget is the worse trade. My claim is narrower — this diff moves both of that defense's inputs.
- **The size numbers are two single samples**, taken seconds apart, and both rooms happen to run short records. Rooms with long signed messages barely move: at a 950 B record the same +95 B is about a 9% shorter window. The proportional reduction is largest where records are short; how long the window lasts in wall-clock terms depends on the room's byte rate, which I have only for these two samples.
- **Capability, not observed behavior.** I have no count of readers who keep the JSON view. The claim is what becomes possible, and the write gate and rate limit still apply to every attempt.
- **Not an argument against merging.** #66 is real, this is the smallest of the four open PRs for it, and offline reverification does need the field served. All three asks are additive.

sv:
Closing as a duplicate of #93 (also #67/#68 — four PRs independently fixed #66). #93 is the one being kept: it also updates `manifest.py`'s OpenAPI schema for the new `sig` field, which this PR doesn't touch — exactly the contract-drift gap #66 called out by name.

---

PR #111, opened 2026-08-25
docs(openapi): say the URL is the real limit on both say lanes, not one

### LINKED ISSUE #76: openapi.json: the signed GET lane's text param has no description, and it is the lane with less URL budget

steveone23:
## The gap

`maxLength` is the character cap, and for a payload carried in the path it is not the cap a caller meets first — the URL is. The unsigned say lane said so. The signed one carried a bare schema:

| path | `text.description` |
|---|---|
| `/r/{room}/say/{nick}/{text}` | "URL-encoded message body. The URL is the size limit in practice…" |

## Why it lands on the wrong lane

The signed lane is the narrower of the two. The DID, signature and nonce are path segments ahead of the text:

```
                                             ---
                                             116 bytes less room for text
```

So the lane with less budget was the one saying nothing about budget.

And the failure it produces is the opaque kind. 4096 CJK characters percent-encode to 36,864 bytes, past the edge's ceiling, so the request dies in the proxy — no `400`, no body, no status this application ever chose. A client generated against `saySigned` alone saw `maxLength: 4096` and nothing else, a ceiling non-Latin text can never reach.

It is also not a hypothetical population: `/r/lobby` currently carries agents posting in Chinese, Korean and Turkish, and signed writes are not optional for `mb-` mailboxes or owned `d-` rooms.

## The change

Hoists the parameter into `_TEXT_PATH_PARAM` so both lanes share one sentence rather than two that can drift apart — which is how the second copy went missing. Same reasoning the file already applies to `_DID_SCHEMA` and `_NONCE_SCHEMA`.

`_TEXT_SCHEMA` is deliberately left alone: `_ROOM_POST_BODY` reuses it, and the POST lane has no URL budget to warn about.

## Test

`test_both_say_lanes_say_the_url_is_the_real_limit` asserts both lanes describe the constraint and that they carry the *same* sentence, so a future divergence fails rather than ships. Verified it fails without the fix:

```
=== WITH the fix ===
1 passed in 3.59s

=== WITHOUT the fix (manifest.py reverted, test kept) ===
E           assert 'URL' in ''
1 failed in 8.01s
```

## Checks

Full CI gate, in a Linux container:

```
ruff check              All checks passed!
ruff format --check     48 files already formatted
pytest                  322 passed
coverage                97.33%
sz.py --check           core code-lines <= baseline (1848), unchanged
```

`manifest.py` is `extra`, so the core ratchet is untouched.

## Left out on purpose

The note lanes are a separate gap. `/kv/{ns}/{key}/set/{value}` and its signed sibling both carry the value in the path and **neither** describes it — so unlike the say lanes they are at least consistent with each other. Values are 8192 characters, which makes the encoded ceiling worse, so it seems worth its own change rather than being folded in here.

Closes #76.

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -109,6 +109,26 @@ _SIG_SCHEMA = {
 _TEXT_SCHEMA = {"type": "string", "minLength": 1, "maxLength": store.MAX_TEXT_CHARS}
 _VALUE_SCHEMA = {"type": "string", "minLength": 1, "maxLength": store.MAX_VALUE_CHARS}
 
+# The message a GET write carries in the path, shared by both say lanes.
+#
+# `maxLength` is the character cap, which is not the limit a caller meets first: the text
+# is a path segment, so the URL is. Stating that on one lane and not the other left the
+# signed lane — the one with *less* room, since the DID, signature and nonce are segments
+# ahead of the text — as the one saying nothing. A caller reads the copy attached to the
+# operation they are generating against, and for `saySigned` that copy was `maxLength`
+# alone, which non-Latin text can never reach.
+_TEXT_PATH_PARAM = {
+    "in": "path",
+    "name": "text",
+    "required": True,
+    "schema": _TEXT_SCHEMA,
+    "description": (
+        "URL-encoded message body. The URL is the size limit in "
+        f"practice: {store.MAX_TEXT_CHARS} ASCII characters fit, one CJK character is "
+        "9 bytes encoded — use POST for long non-Latin text."
+    ),
+}
+
 _NONCE_SCHEMA = {
     "type": "string",
     "pattern": f"^{didkey.NONCE_PATTERN}$",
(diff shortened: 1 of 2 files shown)


luch91 (changes requested):
The proposed OpenAPI correction is narrowly scoped, but exact head a3e21aeb2b6ddfaa4839245afdb7ef0ca56d59b4 is currently DIRTY against its base. Please rebase/resolve the conflict and rerun the contract and required checks before review.

sv:
Closing as a duplicate of #362 — both fix #76 by naming the URL budget in `src/manifest.py`, with the same `tests/http/test_docs.py` change.

Keeping #362 because it applies to current `main`, while this branch conflicts with it. Not a judgement on the work here: if this PR covers something #362 does not, say what on #362 and it can be carried across with credit.

PR #113, opened 2026-08-25
docs: add GitHub Actions signed-write integration

hazzanzico:
## What

Adds a GitHub Actions signed-write guide, linked from the README, with a dedicated secret and a notification job restricted to trusted main-branch runs. Head: `1c150fcb5d12a44ad62f4cc856c14680e7055b7d`, based on `20a4457b89ba11254f4aa48217b066884a148d98`.

## Why

Addresses Minh3132's recovery review by pinning Action `aa113fb300120d5c6c78ee69ec08116e4a566b2c`, which reports the public DID, room, and nonce before each write. The guide includes an exact lookup command and distinguishes attempted writes, confirmed outputs, and inconclusive room reads. The Action remains community-maintained.

## Checks

- [x] Linux `uv sync --frozen` and `uv run just check`: 770 passed, 1 skipped (missing generated edge snapshot); 97.90% repository coverage. Lint, format, types, and size caps passed.
- [x] Pinned Action: 63 tests passed on Linux and Windows Node 24.19.0, no skips. Regressions cover failure logs, retries, response-body timeouts, confirmation records, and padded nonces.
- [x] Workflow passes actionlint; exact lookup passes eight cases and two local service recovery probes. All 133 files match the tested archive.
- [x] Documentation only; new world-writable surface: nothing new.

Official hosted checks and maintainer review remain separate from these local results.


Minh3132 (commented):
The unknown-outcome recovery instruction is not actionable with the exact Action commit pinned by this guide. The workflow omits an explicit `nonce`, so the Action generates it internally; but at `17531944cf49f09722405837d9aca7ff0cdd8ecc`, `runAction()` only writes the `did`/`nonce` outputs *after* `postSignedMessage()` returns successfully. On a network/5xx path where `postSignedMessage()` throws an unknown-outcome error, no outputs are written, and `src/index.js` logs only the generic error message rather than the derived DID or attempted nonce. The guide therefore tells the operator to “inspect the room for the reported DID and nonce” precisely in the case where neither value is reported by the failed step. Because blindly rerunning is what this paragraph is trying to prevent, please make the recovery path actually reproducible: e.g. pin an Action revision that emits the safe public DID + attempted nonce before the uncertain network operation, or change the example to supply a deterministic/recoverable nonce and document how to derive/check the DID. As written, the most important failure-mode instruction cannot be followed.

hazzanzico:
Thanks for identifying this. The old guide did rely on outputs that the pinned Action never wrote on an uncertain failure.

Guide commit `1c150fcb5d12a44ad62f4cc856c14680e7055b7d` now pins Action `aa113fb300120d5c6c78ee69ec08116e4a566b2c`. Before every POST, including a stale-nonce replacement, it logs the validated public DID, room, and exact nonce. The guide uses the last attempt line, keeps success outputs separate, and includes a read-only Python lookup that preserves 19-digit nonce precision. An absent match in the latest 200 records explicitly leaves the outcome unknown.

The adjacent failure-path review also restricted stale retries to a consistent HTTP 400 refusal, extended deadlines through response-body reads, applied shared record validation to confirmation reads, and normalized padded explicit nonces before signing.

Validation:

- 63 Action tests passed on both Linux and native Windows Node 24.19.0, with no skips. The final 33 recovery cases produce 24 failures and 9 passes against pre-fix main; all 33 pass with the fix.
- Two local integration probes let a write land in the actual Technocore service, then simulated failed responses and confirmation reads. The Action failed with lookup metadata, and the exact guide command recovered the single stored record for automatic and padded explicit nonces.
- The guide lookup passed eight additional cases; workflow YAML passed actionlint.
- At the final guide commit, Linux `uv sync --frozen` and `uv run just check` passed: 770 tests, 1 skipped (missing generated edge snapshot), 97.90% repository coverage. All 133 tracked files matched the tested archive.

The Action's [hosted CI passed](https://github.com/hazzanzico/technocore-signed-action/actions/runs/34777224315) on its published main revision.

No production write or real identity material was used. Official PR hosted checks and maintainer review remain separate gates.

PR #120, opened 2026-08-25
fix(store): keep the /r/events discovery log through a quiet day

zkasuran:
## What

The stillborn reaper deletes `/r/events` after a single quiet day. The discovery log lives in
`rooms/` and holds one line while quiet, so the 24-hour stillborn rule removes it once it is
idle past a day with at most one record. `_reap` runs before every write, so the next write
anywhere drops the log. Its seq restarts at 1. A monitor holding a `since=` cursor then
reads an empty view whose `first_seq` is null.

This exempts the events log from the one-day stillborn rule at the two `_reapable` call sites.
The 7-day idle rule is untouched, so a truly silent week still reclaims it.

## Why

`/r/events` is the documented discovery feed. `/llms.txt` tells agents to read it with `since=`
and `wait=` "like any other room", so a monitor is expected to hold a cursor across days. When
the log is reaped and recreated, that monitor reads count 0 with `first_seq` null. The
"truncation is never silent" guarantee keys on `first_seq`, so it cannot fire. The monitor
never learns it missed rooms.

Reproduced against `8bd794b` with the real reaper, one public room per day and everything else
quiet, slower than one creation per 24h:

```
day0 monitor cursor: 1
day1: events seqs=[1] texts=['created room-day1'] | monitor since=1 -> count=0 first_seq=None
day2: events seqs=[1] texts=['created room-day2'] | monitor since=1 -> count=0 first_seq=None
...
day5: events seqs=[1] texts=['created room-day5'] | monitor since=1 -> count=0 first_seq=None
counters: {'reaped_idle': 0, 'reaped_stillborn': 10}
```

Five new public rooms, zero seen by the monitor, permanently. On a low-traffic instance the
events file never gets past one record before the next pass takes it, so the discovery log is
stuck at seq 1 and every cursored reader is blind.

The maintainers already exempt notes from the stillborn rule. The reason in `store.py` is
"a note has no reply to wait for, so one write says nothing about it." A server-written
announcement log that callers cannot POST to (events POST is refused 403) has none either, so
the same reasoning applies to it. Exempting only the 24-hour rule keeps the bounded-resource
guarantee, since the log still yields to the 7-day idle rule like any other room.

## How

The stillborn rule is turned off for one file, the events log, at the two points in `_reap`
that ask whether it applies. Since `main`'s reaper refactor built the `Path` only after the
first check, the two sites read the name differently but mean the same thing:

```python
# first check, on the scandir entry before a Path exists
if not _reapable(entry.path, now, stillborn_rule and entry.name != EVENTS_ROOM + suffix):
    ...
# recheck under the lock, where the Path is in hand
reason = _reapable(p, now, stillborn_rule and p.stem != EVENTS_ROOM)
```

`EVENTS_ROOM` is already a module constant; the entry-name form keeps `pathlib` off the hot
path the way `_walk` intends. The guarded call no longer fits one line under the 100-column
formatter, so it wraps, taking core `store.py` code-lines from 807 to 809. That is the one cap
this raises, argued the way the other size-touching fixes here are (#62 is the precedent);
`core_total` stays at its 2035 ceiling. No route, parameter or status code changes, so the
contract job is unaffected with no manifest edit.

## Scope and compatibility

- `src/store.py` (+2 core code-lines, `store.py` cap 807 to 809), `src/manual.md` (one clause under DISCOVERY), `CHANGELOG.md`,
  and two test files. No public route or response shape changes.
- The behaviour change is strictly "the discovery log lives longer during a quiet spell", which
  is what its own documentation already promises.

## Abuse impact

None new. The events log is server-written and refuses every client write (403), so no caller
can grow it. It stays bounded by the 7-day idle rule and by the ring like any other room; the
change removes only the 24-hour deletion that a quiet day triggers.

## Checks

- [x] `uv run sz.py --check`: ok. Core `store.py` cap raised 807 to 809 (see How); `core_total`
      stays at its 2035 ceiling
- [x] the schemathesis contract job as `ci.yml` runs it: 876 cases, no issues (no manifest change)
- [x] `uv build --project mcp` + `tests/verify_mcp_dist.py` and `docker build` + the CI smoke:
      not re-run on this rebase, since it changes no `mcp/` or `docker/` file; the reaper change
      is exercised by the suite and `beautiful_chat.sh`; both were green pre-rebase
- [x] Docs updated: `src/manual.md` DISCOVERY gains one clause. `README.md`, `src/patterns.md`,
      `mcp/README.md` need none
- [x] New surface on a world-writable service: none, see Abuse impact

The unit regression `test_the_events_log_survives_a_quiet_day_so_a_monitor_cursor_holds` fails
on main (events reaped, `reaped_stillborn` counts it) and passes with the fix, verified by
reverting `src/store.py`. `test_the_events_log_still_yields_to_the_idle_rule` pins that a silent
week still reclaims it. The state-machine promise in the `reap` rule asserts the log survives
inside its idle window; on main it fails under search (reproduced at higher example counts).

## Notes on other open work

- Clear of the persist-sig PRs (#67, #68, #93, #110) and the timestamp PRs (#103, #107): they
  touch `append` / `_write_record` / `_last_nonce`, not `_reap`. Dry-run merged against #84,
  #103 and #107: only `CHANGELOG.md` conflicts, which is the shared `[Unreleased]` section.
- #45 rejects events POSTs earlier in `app.py`. Orthogonal, no file overlap.

---

verification were done by the author. Verified locally before submitting: the full pytest suite
with branch coverage, ruff check and format, ty, the core-size ratchet, the schemathesis
contract job and the end-to-end example. The MCP build and the Docker image smoke were green on
the pre-rebase revision; this rebase touches neither `mcp/` nor `docker/`.


WIZARDspace (approved):
Verified against `main` at `01c49fb`. The defect is real, still present eight days after this was filed, and the fix is correct — including one thing that is easy to get wrong and was not.

## The bug reproduces on current `main`

Ran the real reaper against a store holding one announced room and the events log, aged past `STILLBORN_SECONDS` but well inside `IDLE_SECONDS`:

```
events path: rooms/13/events.jsonl
events stillborn=True  reapable=stillborn

main  01c49fb   -> events exists: False    r-probe exists: False
PR    9cbb6dd   -> events exists: True     r-probe exists: False
```

So the discovery log is reaped by the 24h rule on `main` today, and the fix keeps it while leaving ordinary stillborn reaping untouched. `main` has moved a long way since `8bd794b` — the seq state was sharded (#598), the per-room append lock was reworked (#673), the counters lane was batched (#626) — and none of it touched this.

The consequence in the body is the part worth restating: `/llms.txt` tells agents to poll `/r/events` with `since=`, the log restarts at seq 1 after a reap, and the "truncation is never silent" guarantee keys on `first_seq`, which is null on the recreated log. So the guarantee cannot fire and a cursored monitor goes blind without an error. That is a silent correctness failure on a documented interface, not a tidiness issue.

## The part that is easy to get wrong

`stillborn_here` is used at both `_reapable` calls, but the two `if stillborn_rule:` blocks inside the reap branch are left alone. That is right, and it is not obvious:

```python
if stillborn_rule:
    _set_seq_entry(root, room, max(0, last_seq(root, room)))   # preserve high-water mark
...
if stillborn_rule:
    reaped[f"reaped_{reason}"] += 1
```

`stillborn_rule` there is really "this is the rooms pass" — it gates seq-mark preservation (notes are not sequenced) and the rooms-only counters. Substituting `stillborn_here` would have meant that when the events log *is* eventually reaped by the 7-day idle rule, its high-water mark would not be preserved and it would restart at seq 1 — reintroducing precisely the failure this PR exists to fix, just seven days later instead of one. It would also have gone uncounted in `reaped_idle`.

Keeping the two flags distinct is the whole correctness of the change, and the naming makes it legible rather than accidental.

## Tests

Both directions are covered, and the exemption is pinned as narrow rather than merely present: the survival test also asserts an ordinary stillborn room (`p-quiet`) is still reaped and that `reaped_stillborn` increments by exactly one. `test_the_events_log_still_yields_to_the_idle_rule` pins the 7-day rule.

I checked the tests are not vacuous rather than taking the body's word: reverting only `stillborn_here = stillborn_rule and entry.name != …` back to `stillborn_rule` makes `test_the_events_log_survives_a_quiet_day_so_a_monitor_cursor_holds` fail with "the discovery log was reaped after one quiet day", while the idle test keeps passing — which is the right split.

## Two things I checked that are fine

**The name comparison is inert on the notes pass.** `entry.name != EVENTS_ROOM + suffix` becomes `"events.txt"` when walking notes, so a note keyed `events` would compare equal — but `stillborn_rule` is already `False` there, so `stillborn_here` is `False` either way and nothing changes. Harmless, and worth having on the record since the expression reads as if it applies to both passes.

**Abuse surface is unchanged.** `EVENTS_ROOM` is not reserved on the write path — `store.py:2236` only avoids announcing the events room itself — so a caller could already write to `/r/events` before this change. The exemption moves that one room from 1-day to 7-day retention, which is the stated intent, and adds no route, parameter or capability.

## Gates at `01c49fb`

| check | result |
|---|---|
| `ruff check .` | passed |
| `ty check` | passed |
| `pytest tests -q` | **657 passed, 1 skipped** |
| `sz.py --caps` | exit 0 — `core/store.py` 936 of cap 1000 |

Merges clean onto `01c49fb`. Note the gate changed under this PR while it sat: #670 made pull requests run `sz.py --caps` rather than `sz.py --check`, so the ratchet is no longer what a fork has to clear — only the caps, which this passes with 64 lines spare.

Approving. Eight days is a long time for a silent break in the documented discovery feed to sit unreviewed.

WIZARDspace:
The correction in `0fd215b` is right, and it corrects my approval as well as the PR body. I restated the "restarts at seq 1, cursor goes blind" consequence as "the part worth restating," and it is false — while quoting, three paragraphs later, the exact line that disproves it.

**Measured on `main` at `42cfed1`.** One announcement, aged past `STILLBORN_SECONDS`, real `_reap`:

| | file | floor (`last_seq`) | generation |
|---|---|---|---|
| before reap | exists | 1 | 1 |
| after reap | **gone** | **1** | **1** |

Then the next announcement lands, and a monitor parked at `since=1` polls:

```
count = 1   first_seq = 2   generation = 2
messages = [(2, 'created gamma')]
```

So the sequence continues at 2 rather than restarting at 1, `first_seq` is 2 rather than null, and the new line **is** delivered. The cursor is not stranded. The mechanism is the line I quoted: `_set_seq_entry(root, room, max(0, last_seq(root, room)))` sits under `if stillborn_rule:`, which is the *rooms-pass* flag — true for both reap reasons — so the floor and generation survive every room reap, exactly as `#139` dir #2/#3 intended.

What actually goes is what the new comment says: `created alpha` is unreadable afterward, and with it the creation order the manual says `/rooms` cannot recover. Plus the two smaller costs — a generation bump from 1 to 2 on a day when nothing happened, against `room_generation`'s own description of the generation as "the explicit signal to resync"; and `reaped_stillborn` counting the server's own log, which the test now pins at `before + 1` (`p-quiet` only).

The defect and the fix are unchanged by any of this — the log is still reaped after one quiet day on `main` today, and the exemption still closes it. What changed is that the PR now states a cost it can support instead of one it cannot.

**Gates at `0fd215b` merged onto `42cfed1`** (conflict-free):

| check | result |
|---|---|
| `pytest tests -q` | 658 passed, 1 skipped |
| `ruff check .` | passed |
| `ty check` | passed |
| `sz.py --caps` | exit 0 — `core/store.py` 936 of 1000, core total 2254 of 3000 |

`store.py` is still 936 because all twelve added lines are comments, so `code_lines` is untouched. `0fd215b` changes no executable line at all — the `store.py` diff is comment-only and the test diff is the docstring plus the rename — so everything I verified at `9cbb6dd` still holds unchanged.

Approval stands.

WIZARDspace:
@zkasuran my approval is on `9cbb6dd` and the branch has moved twice since — your `0fd215b`, and then `main` underneath it. Re-checking both rather than letting a stale green ride.

## `0fd215b` re-reviewed: the revised rationale is right

You rewrote the comment's premise after I approved, from "a monitor holding a `since=` cursor reads an empty view whose `first_seq` is null" to "no cursor is stranded — `_reap` leaves the seq floor and the generation behind". That is the stronger claim, and it holds for a reason worth stating explicitly, because it is not obvious from the diff: the exemption and the floor-preservation are gated on **different** variables.

```python
stillborn_here = stillborn_rule and entry.name != EVENTS_ROOM + suffix
reason = _reapable(p, now, stillborn_here)   # <- exemption
if reason:
    if stillborn_rule:                        # <- floor + generation, unchanged
        _set_seq_entry(root, room, max(0, last_seq(root, room)))
```

`stillborn_here` is what spares the log at 24 h; `stillborn_rule` stays `True` for the whole rooms walk, so when the log is eventually taken by the 7-day idle rule the floor and generation are still written. Had you threaded `stillborn_here` into that second guard as well, the claim would have inverted — the events log would be the one room reaped without a floor. Keeping them separate is what makes the new comment true, so it is worth a line in the comment itself.

The same split carries your third point for free: with `stillborn_here` false the log's `reason` comes back `idle`, so `reaped[f"reaped_{reason}"]` lands on `reaped_idle` and never inflates `reaped_stillborn`. The counter keeps the meaning `_reapable` gives it.

## The branch no longer merges

`mergeable: false`, `dirty`, 8 behind. #722 and #723 landed today and rewrote exactly the function this patches:

- the walk moved out of `_reap` into **`_reap_pass`** (`store.py:1839` on `2670ebb`);
- `_reapable` gained a fourth parameter, and the first call site now passes it — `_reapable(entry.path, now, stillborn_rule, st)` (`:1854`), reusing a stat the loop already took;
- the recheck under the lock is still three-arg (`:1867`), and both `if stillborn_rule:` guards (`:1869`, `:1887`) are untouched.

So the conflict is context, not intent. The rebase is mechanical:

- `:1854` → `_reapable(entry.path, now, stillborn_here, st)` — keep the `st` argument, it is the point of #722.
- `:1867` → `_reapable(p, now, stillborn_here)`.
- Leave `:1869` and `:1887` on `stillborn_rule`, exactly as your patch already does.

Nothing in the reaper rewrite conflicts with what this PR argues — #722/#723 were about lock spans and the count race, not the retirement rules — so I do not think any of the reasoning needs revisiting. Re-approving once it is green on the new base; the review points from my earlier pass are all still closed as far as I can see.

zkasuran:
@WIZARDspace rebased onto current `main` (`20a4457`), pushed `c23ba78`, mergeable again.

One conflict in `src/store.py`; it resolved to keep both changes rather than pick one. `main` gave `_reapable` a new optional `st` parameter so the reap loop can reuse the stat it already takes; this branch computes `stillborn_here` to exempt the events log from the 24h rule. The merged first call is `_reapable(entry.path, now, stillborn_here, st)`, so the exemption and the stat reuse both apply. The recheck under the lock stays `_reapable(p, now, stillborn_here)` with no `st`, which is what `main`'s own docstring requires: the recheck must re-stat by path to catch a writer that refreshed the file, so handing it a cached stat would be the bug. The events-log exemption you approved is unchanged.

`test_the_events_log_survives_a_quiet_day_so_the_announcements_do` still pins it. Gate on `c23ba78`:

```
uv run python sz.py --caps           exit 0; store.py 993/1010, core 2318/3010
uv run ruff check                    pass
```

Your approval was on `9cbb6dd`; the reviewed exemption is the same logic, now sitting on top of `main`'s `_reapable` change rather than beside it.

WIZARDspace (approved):
Re-verified at `c23ba78` on `20a4457` — merge-base is current main, 0 behind, 8 ahead.

## The conflict resolution is right, and the recheck half is main's own contract

Both call sites landed as your note describes, and they are the only two:

- `:1874` — `_reapable(entry.path, now, stillborn_here, st)`
- `:1887` — `_reapable(p, now, stillborn_here)`

The recheck taking no `st` is not a judgment call, it is what main documents. `_reapable`'s docstring: "`st` is the one stat a caller is allowed to hand in, and it is for the FIRST call only ... The recheck under the lock passes nothing, so the paragraph above still describes the call it is about — a cached stat there is the bug." `_reapable` is byte-identical between `20a4457` and `c23ba78`, so the merge added the exemption without disturbing #722's stat reuse.

Both `if stillborn_rule:` guards are untouched — two occurrences in each file, differing only by the 18-line offset the new comment adds. That split is what my approval turned on, so I measured all three halves of it rather than re-reading them.

## Measured on both bases, end to end

The real reaper against a store holding one announced room (`public1`), the events log, and an unlisted `p-quiet`, aged past `STILLBORN_SECONDS` and well inside `IDLE_SECONDS`:

| | `main@20a4457` | `c23ba78` |
|---|---|---|
| events after one quiet day | reaped | **survives** |
| `p-quiet` (ordinary stillborn) | reaped | reaped |
| `reaped_stillborn` delta | **+2** | **+1** |

Then letting the next announcement land, which is where the cost actually shows:

| after reap → next announcement | `main@20a4457` | `c23ba78` |
|---|---|---|
| log contents | `['created public2']` | `['created public1', 'created public2']` |
| generation | **1 → 2** | 1 |
| `first_seq` | 2 | 1 |

So all three consequences the comment names are demonstrated rather than argued: the creation order for `public1` is gone on main, the generation bumps on a day when nothing happened, and `reaped_stillborn` counts the server's own log.

That third row also confirms you were right to rewrite the premise after my first pass. `first_seq` comes back **2, not null**, and the seq floor survives at 1 — so "no cursor is stranded, but the lines are gone" is the accurate claim and the one I originally approved was not.

Pushing the exempt log past the 7-day rule reaps it, counts it as `reaped_idle` (+1, `reaped_stillborn` +0), and leaves the floor at 1 — the `if stillborn_rule:` guards doing exactly what keeping them on the pass-level flag was for.

## The tests still bite on the new base

Reverting only the exemption — `stillborn_here = stillborn_rule and entry.name != EVENTS_ROOM + suffix` back to `stillborn_rule` — fails `test_the_events_log_survives_a_quiet_day_so_the_announcements_do` with "the discovery log was reaped after one quiet day", while `test_the_events_log_still_yields_to_the_idle_rule` keeps passing. Same split as at `9cbb6dd`. `test_store_stateful.py` is stable over three consecutive runs.

## Gates at `c23ba78`

| check | result |
|---|---|
| `ruff check .` | passed |
| `ruff format --check .` | 91 files already formatted |
| `ty check` | passed |
| `pytest tests -q` | 772 passed, 1 skipped |
| `sz.py --caps` | exit 0 — `core/store.py` 993 of 1010, core total 2318 of 3010 |

The caps delta is exactly +1 code line against main's 992, which is the `stillborn_here` binding; the other 18 added lines are comment.

Two notes, neither blocking:

- `c23ba78` is a merge commit, not a rebase — the branch carries four merges of main. No practical difference at 0 behind, and the squash flattens it, but the history is merge-based if anyone reads it later.
- `capacity_bench` and `overlap` show as skipped on this run. Both are by design and not fork-related: the bench requires a `perf:` title prefix and only informs, never gates, and `overlap` runs on `opened`/`edited`, not on a push.

Re-approving. Same logic I approved at `9cbb6dd`, now sitting on top of main's `_reapable` change rather than beside it.

bdunn77:
Independent validation of head `c23ba78a` (base `20a4457b`):

- On `origin/main`, `_reap_pass` applies the stillborn rule to every room, including `/r/events`. A one-line discovery log therefore reaps after a quiet day.
- On this head: `test_the_events_log_survives_a_quiet_day_so_the_announcements_do` and `test_the_events_log_still_yields_to_the_idle_rule` **2/2 pass**. Events survive `STILLBORN_SECONDS+60`; an ordinary stillborn room still reaps; the idle rule still reclaims events after a week.

No merge blockers found on the events-log stillborn exemption this PR adds.

PR #122, opened 2026-08-25
fix(store): support file locking on Windows

bdunn77:
## Summary

- isolate the store sidecar lock behind a tiny standard-library platform primitive
- preserve the existing blocking `flock(LOCK_EX)` behavior on POSIX
- use a bounded-sleep `msvcrt.locking` loop for the same exclusive byte-range contract on Windows
- add real subprocess tests for exclusion, normal/exception release, terminated-holder cleanup, and contiguous multi-process appends
- make two existing tests portable so the complete suite runs natively on Windows

## Context

Current `main` imports `fcntl` unconditionally, so `import store` fails on native Windows. Closed PR #15 demonstrated a viable Windows direction, but its relevant changes were obscured by line-ending churn, workflow edits, and unrelated modifications. This PR is rebuilt from current `main` and contains only the lock primitive, its store integration, focused process tests, and the two test-only fixes needed for a complete native Windows run.

No dependency, workflow, lockfile, generated-file, or broad line-ending changes are included.

## Lock contract tested

- a second independent process cannot enter while the first owns the sidecar lock
- it proceeds after explicit release
- exceptions release the production `_locked` context
- terminating a holder lets a waiter acquire through OS handle cleanup
- four independent store writers retain contiguous, unique room sequence numbers
- the process-lock suite passed 20 consecutive times on Linux and 20 consecutive times on native Windows

## Validation

Linux (Python 3.12.12):

- `uv sync --frozen`
- `uv run ruff check .`
- `uv run ty check`
- `uv run coverage run -m pytest tests -q` — **325 passed**
- `uv run coverage report` — **96.61%**
- `uv run sz.py --check` — core remains **1848/1848** code lines
- `git diff --check`

Native Windows (Python 3.12.13):

- `uv sync --frozen`
- `uv run ruff check .`
- `uv run ty check`
- `uv run coverage run -m pytest tests -q` — **325 passed**
- `uv run coverage report` — **96.92%**
- `uv run sz.py --check` — core remains **1848/1848** code lines
- `git diff --check`


faakhiranadiaz:
Independent verification on a real Windows machine, since this PR is hard to review without one. Windows 11, Python 3.12.13, no Docker, no WSL.

**Before, on current `main`:**

```
ModuleNotFoundError: No module named 'fcntl'   (src/store.py line 13, at import)
```

**With this branch merged onto 0.9.4 (14324aa):**

```
import OK | lock backend = msvcrt

/healthz                             200
/r/wintest/say/bot/hello...          200
/r/wintest?format=json               200   text: "hello from native windows"
/kv/wintest/k/set/v                  200
/kv/wintest/k                        200
```

Full gate set run natively rather than in the container:

```
uv run pytest tests -q      394 passed in 140.44s
uv run ruff check .         All checks passed!
uv run ruff format --check  56 files already formatted
uv run ty check             All checks passed!
uv run sz.py --check        ok, core 2020 <= baseline 2021
```

So the import barrier is gone and every lane I exercised works, on the platform this PR exists for.

**What I did not exercise, and would want someone to before trusting it in anger:** concurrency. I ran a single process throughout, and a lock only earns its name when two writers meet. `msvcrt.locking` takes a byte-range lock from the current file position rather than a whole-file lock, and the retry loop treats `EACCES`/`EDEADLK` as "try again", so the interesting cases are two processes appending to one room and the create-gate race the store already has tests for. Passing `tests/unit/test_platform_lock.py` is not the same as proving that.

**One triage note:** #255 was filed today reporting exactly this failure, from a different Windows machine, and does not link here. I have pointed it at this PR.

djd39448:
@faakhiranadiaz's verification above matches mine exactly (394 passed natively, `sz` 2020 ≤ 2021, same import failure on `main`), so I won't repeat it. They flagged the right gap — single-process runs don't test a lock — so here is the concurrency half, at head `14324aa`, on native Windows 11 (CPython 3.12.13, real `msvcrt` path) and WSL2 (3.12.3, `fcntl` path). WSL reports `os.name == "posix"`, so it cannot exercise the Windows branch at all; these are separate runs, not one.

I wrote four probes rather than re-running `test_platform_lock.py`:

**Mutual exclusion.** Holder takes the lock for 0.6 s; waiter times its own acquire. Six trials per platform — the waiter never acquired early (min 0.601 s WSL, 0.609 s Windows).

**Contiguous sequence under real contention.** 8 writers × 30 appends, released together off a barrier file rather than staggered `Popen`. Interleaving was genuine (239/240 tag-order switches on WSL, 130–156/240 on Windows). Every trial, both platforms: seq exactly `1..N`, no gaps, no duplicates.

**Exception release.** Five exception classes the PR's tests don't try — including `KeyboardInterrupt`, `SystemExit`, and a raise from a nested call. All released in ≤0.12 s on both platforms.

**Hard kill.** `Stop-Process -Force` on Windows (TerminateProcess, so the `finally` never runs) — a blocked waiter acquired ~1.97 s later. Windows releases the `msvcrt` lock on process teardown independently of Python-level cleanup.

No hole in any of the five guarantees the description claims.

**Three notes, none blocking:**

The description calls the Windows path a "bounded-sleep loop", which reads like it could give up. It can't — `acquire` is `while True`, returns only on success, re-raises outside `{EACCES, EDEADLK}`. I held the lock 6+ seconds (600× the retry interval); the waiter stayed blocked and alive rather than proceeding unlocked. The real tradeoff is the opposite one: an abandoned lock hangs a waiter forever. Safer default, worth knowing it never times out.

`flock` takes a whole-file lock; `msvcrt.locking(fd, LK_NBLCK, 1)` takes one byte at offset 0. Inert here — the sidecar `.lock` files are always 0 bytes and only this code touches them, always at offset 0 — but a one-line comment would save the next reader the check.

Nesting `_locked` on the same target deadlocks on both platforms (the lock belongs to the handle, not the process). Not yours — POSIX had it pre-PR, and the Windows path faithfully reproduces it. I read all 11 `_locked` call sites; the only nesting is `_write_record`'s gate + path, always different files. A docstring line would help future callers.

**Separately:** the Validation block reports 325 passed and `1848/1848`, but `git show a73147d:sz-baseline.json` gives 1848 while the file at your current head says 2021 — those numbers predate the branch merging main. Worth refreshing, since the real ones are better.

**One scope question.** The body says the PR "contains only the lock primitive, its store integration, focused process tests, and the two test-only fixes". Three changes sit outside that: `_replace` moves from `os.fchmod(f.fileno(), …)` to unconditional `os.chmod(tmp, …)`, taking POSIX off descriptor-based chmod; `_note_count`/`_check_note_total` gain `rebuild=`; and `note_set`'s pre-gate check becomes POSIX-only with a `lambda _: (0, 0)` stub. Note #272 fixes the same chmod issue conditionally via `os.supports_fd` and edits the same line, so those two will collide.

Narrowed to `platform_lock.py` plus the two `_locked` call sites, this looks solidly mergeable to me. Happy to re-run any of the probes on Windows.

AI assistance (Claude) was used; probe scripts and outputs available if useful.

2Lin3:
I adapted the lock-only portion of your original `a73147d` onto current `main` (`9c7df0e`), preserving `bdunn77` as the commit author:

- branch: https://github.com/2Lin3/technocore-chat/tree/codex/windows-lock-v010-review
- [`294481b`](https://github.com/2Lin3/technocore-chat/commit/294481b84005778743c5b82046ecf251bfd07f2d): the portable lock/store integration and native subprocess contention tests
- [`5a7f264`](https://github.com/2Lin3/technocore-chat/commit/5a7f264eff03f01be498de8022e769ff5ac04d8a): preserve the Windows subprocess runtime environment in `test_config`

I deliberately left out the `_replace`/chmod and note-count/rebuild changes, so #272 remains separate and this branch does not introduce a competing implementation there.

Validation used a combined integration branch (`9c7df0e` + #272 + these commits + the two other prepared branches). Native Windows completed 472 tests at 97.22% coverage. Ruff check and Ruff format passed; `ty` passed for both the Linux target and native Windows. Size, lockfile, and diff checks were clean. The new lock tests cover independent-process exclusion, explicit and exception release, terminated-holder cleanup, and four-process contiguous sequence allocation.

I am not opening a competing PR. Please cherry-pick or adapt these two commits here if useful; I can rerun the same matrix on a refreshed head.

djd39448:
Thanks — the four properties your `test_platform_lock.py` covers (independent-process exclusion,
explicit and exception release, terminated-holder cleanup, contiguous sequence allocation under
contention) are the same four I probed by hand in the review above, so this turns that into
something CI re-runs. That is strictly better than a comment reporting numbers, and it answers
the gap faakhiranadiaz flagged about single-process runs.

One thing the rebase changes that is worth a decision before this lands, because no check will
raise it. Extracting the lock into a new `src/platform_lock.py` puts it outside the size policy:
`sz.py` does not walk `src/`, it reads two hardcoded tuples (`sz.py:44`).

```python
EXTRA_FILES = ("src/manifest.py",)
```

`src/platform_lock.py` is in neither. Dropping your exact file into a current tree and running
the check unmodified:

```
$ python3 sz.py --check
core/app.py                1966   1042       277       7.7
core/config.py              324     61       207     12.56
core/didkey.py              144     60        31       8.1
core/limit.py               423    171        81      8.48
core/store.py              2010    840       441      7.36
extra/manifest.py          1838   1333       155      3.81
core total (code)                 2174
check ok: core code-lines <= baseline (2174)
```

Not listed, not counted, still green — and with `store.py` at `+3/-3` the measured core does not
move at all. Measured under an added `extra/` label the file is **29 code lines**, so that much
core-path logic — the primitive every store write goes through — lands where the ratchet cannot
see it. Whether that is right is a judgement call I do not think is mine to make: it is arguably
`extra/` like `manifest.py`. But it should be a decision rather than a side effect of the move,
and it is a one-line fix either way. `sz.py`'s own docstring notes a file added to `CORE_FILES`
also needs a cap entry or the enforcing modes error out, so the deliberate path cannot slip past.

Smaller, and carried over rather than introduced by you: the PR description still says
*"bounded-sleep `msvcrt.locking` loop"*, and the loop is `while True:` with a 0.01 s sleep and no
deadline — it blocks indefinitely, which I verified against a 6 s hold. Your module's docstring
says *"Block until this process exclusively owns the file's lock domain"*, which is the accurate
description; only the PR text disagrees with the code.

2Lin3:
Thanks @djd39448 — both points were correct, and the size-policy observation led to a wider native-Windows audit that found real reader/replacement edge cases beyond the original lock primitive.

I have pushed the completed result here:

- branch: https://github.com/2Lin3/technocore-chat/tree/codex/windows-lock-v010-review
- head: [`10753d0`](https://github.com/2Lin3/technocore-chat/commit/10753d04a3e1cb625773e05c3e4a202a1c3d2a62)
- Windows atomic read/replace fix: [`121f72f`](https://github.com/2Lin3/technocore-chat/commit/121f72f05e26d4b85cba6a59ce9528fba54a4ddd)
- #272 portability commits, preserving Zaksans as author: [`36c5820`](https://github.com/2Lin3/technocore-chat/commit/36c5820) and [`067785c`](https://github.com/2Lin3/technocore-chat/commit/067785c)

What changed:

1. **The size-policy gap is closed deliberately.** `src/platform_lock.py` is now in `CORE_FILES`, has its own cap, and contributes to `core_total`. The final implementation is larger than the 29-line version you measured because the native stress work below added the reader/replacement boundary and recovery handling. Exact final measurements are:
   - `core/platform_lock.py`: 135 code lines, cap 135
   - `core/store.py`: 921 code lines, cap 922
   - core total: 2343 code lines, cap 2344

2. **The PR wording is corrected.** The accurate description is: a blocking `msvcrt.locking` retry loop with a 10 ms interval; it waits indefinitely for contention and propagates non-contention OS errors. It is not a bounded-total-wait loop.

3. **Native Windows old-reader/new-path semantics are now explicit.** Cooperative readers use `CreateFileW` with read/write/delete sharing. Existing targets use `ReplaceFileW` with flags `0`, a unique backup, and explicit error-1177 restoration. A failed restoration raises `OSError` with the retained backup path, preserving existing best-effort error boundaries. Errors 1175/1176 are not blindly retried, because Microsoft does not guarantee that repeating the whole operation after every partial merge state is safe.

4. **Transient pathname gaps are coordinated, not hidden as missing data.** Store readers do one immediate reopen and then, only if needed, open the already-existing business sidecar and obtain the data handle under that lock. Truly missing data creates no sidecar and incurs no two-second retry delay. The room `exists()`/open races were removed, the usage gauge now serializes same-target replacements, and lock mtimes are refreshed only at the actual replacement boundary so orphan-lock reaping still works.

5. **The Windows note-cap follow-up is restored.** Windows skips only the optional pre-gate fast rejection; the authoritative cap check remains inside `.notes-create`. The adapted logic credits bdunn77 as co-author. The chmod portability fix remains the original #272 implementation rather than a competing rewrite.

Validation on the pushed head:

- full suite, run 1: `487 passed in 121.61s`
- full suite, run 2 under coverage: `487 passed in 140.48s`
- total coverage: `96.99%` (`--fail-under=96`)
- focused Windows/platform/note-count set: `49 passed`
- global note-cap cross-process regression: 20 consecutive complete runs passed
- existing cross-process platform-lock suite: 10 consecutive complete runs passed
- native reader/writer stress: 10 rounds, 4 readers + 2 writers, 500 writes, `errors=[]` (2.412s)
- `ruff check .`, `ruff format --check .`, `ty check`, `sz.py --check`, `sz.py --caps`, and `git diff --check`: all passed

One correction to my earlier comment: the reported “472 tests” was one run on a combined integration branch, not stable evidence for this exact branch. Repetition exposed an intermittent note-cap failure and broader Windows reader/replacement failures. Please treat that earlier validation paragraph as superseded by the exact pushed-head results above. The first full run of this final work also caught an orphan-lock reaping regression in an intermediate mtime-refresh placement; that was corrected before the two clean 487-test runs.

Thanks again for checking both the implementation and the policy boundary instead of stopping at a green test line. This is materially better because of that review.

djd39448:
Verified the size-policy half independently at `10753d0`, since a claim about the ratchet should not rest on the same person who raised it:

| | cap | measured |
|---|---|---|
| `core/platform_lock.py` | 135 | 135 |
| `core/store.py` | 922 | 921 |
| `core_total` | 2344 | 2343 |

Both invariants hold — the per-file caps sum to exactly the `core_total` cap (2344), and the stored `core_total` equals the sum of the stored core `code_lines` (2343) — and the headroom pattern matches main's rather than granting the new file slack. Putting it in `CORE_FILES` with its own cap is the right call over `extra/`: `manifest.py` is there because it is published surface, and a lock every store write goes through is not that.

Worth saying plainly: retracting the 472-test figure yourself, unprompted, because repetition showed it was one run on a combined branch rather than evidence for this head — that is the part of this I would most want other people to copy. The intermittent note-cap failure and the orphan-lock reaping regression only existed to be found because you re-ran instead of banking a green line.

I have not independently checked the `ReplaceFileW` / error-1177 restoration or the 1175/1176 non-retry reasoning; that is a deeper Win32 semantics question than I have evidence for, and I would rather say so than nod along.

**The thing I think is now the actual blocker, and it is not more validation.** None of this is reachable from the PR. #122 is still at head `14324aa`, `CONFLICTING`/`DIRTY` against `9c7df0e`, and its description still opens with

> use a bounded-sleep `msvcrt.locking` loop for the same exclusive byte-range contract on Windows

The corrected wording — a blocking retry loop with a 10 ms interval that waits indefinitely — is accurate, but it lives in your comment, not in the PR body, and neither of us can edit that. So a maintainer opening #122 today sees a stale, conflicting PR whose description misstates the loop, with the 487-test validated work one fork-branch hop away in a comment thread. That gap is doing more to hold this up than any remaining technical question.

Since you have said you will not open a competing PR — which I think is right — the unblock is bdunn77 pushing your two commits onto the PR branch and fixing that one line of the description, or a maintainer saying they would rather take it as a fresh PR. Either is a five-minute action by someone who is not you or me.

erhnysr:
**Reviewed the branch against current `main` — the core lock fix looks sound; a few notes for whoever picks this up.**

**1. The CONFLICTING/DIRTY state is stale-branch drift, not a logic conflict.** The diff places `_locked` at line 430, but on current `main` it's at line 577 — the surrounding code has moved since the branch was cut. A rebase onto `main` should resolve the conflict cleanly; there's no competing change to the locking logic itself.

**2. The note-cap change is out of scope for a "fix Windows import" PR.** Alongside the lock swap, `note_set()`'s pre-gate check gained both a `lambda _: (0, 0)` rebuild override (which changes the fast-path on *all* platforms, not just Windows) and an `os.name != "nt"` guard, plumbed through new `rebuild=` params on `_note_count`/`_check_note_total`. That may well be a legitimate correctness fix, but it isn't covered by the PR summary's "only the lock primitive, its store integration, and test fixes." Suggest splitting it into its own PR or calling it out explicitly so it can be reviewed on its own merits.

**3. `msvcrt` fairness (worth a comment, not a blocker).** The Windows path is a non-blocking `LK_NBLCK` in a 10ms polling loop rather than a true blocking wait, so there's no FIFO/fairness guarantee — under sustained contention a waiter could be starved. `flock` isn't strictly fair either, but the poll makes it more likely. Fine to ship; just worth a note near `_RETRY_SECONDS`.

**4. The concurrency tests could be stronger.** Exclusion, release-on-exception, and release-on-kill are all covered well. The multi-writer test (4 processes × 15 fast appends) is low-contention with a very short critical section, so a subtly racy lock could still pass it — the 20 consecutive green runs help, but it's closer to a smoke test than a stress test. Bumping the writer count and widening the critical section (closer to a real store-append path rather than the synthetic lock-child handshake) would make it a real exercise of serialization under load.

Happy to help rebase onto `main` and push the test improvements if the author or a maintainer wants — just say the word.

PR #126, opened 2026-08-25
fix(store): tolerate concurrent reaping in readers

rioyu123:
## What

Room and note readers tolerate a file being reaped just before it is opened, instead of returning an HTTP 500. `last_seq` still falls back to the retained sequence floor; permission errors still propagate.

## Why

Checking whether a file exists does not prevent the idle reaper from removing it before `open()`. This updates the original fix for current main (`e4c4f73`), including its sharded storage and retained sequence state. Successful reads, exports, and write behavior are unchanged.

Thanks to @almondous for the [forward-port and additional regression tests](https://gist.github.com/almondous/2cb564dc069e17cc7dee71a9f8fc15fe). The original regressions are retained, with an additional state-machine rule for a `last_seq` read interrupted by reaping. No dependencies on other open PRs.

## Checks

- [x] No existing documentation becomes inaccurate; public responses are unchanged.
- [x] New abuse surface: nothing new.

Linux regression: 6 failures before the fix, all 13 supplemental cases pass afterward. Size caps, Chromium checks, and HTTP contract checks pass. Changelog and size baselines are unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -397,6 +397,10 @@ booted with `inf` was publishing JSON no strict parser would accept, and will no
   `/kv/did-<first 2 hex>/<remaining 14 hex>`; readers fall back to the legacy path. Every namespace,
   listing response, and global disk bound keeps the same fixed limit.
 
+- **Room and note reads tolerate concurrent idle reaping.** A reaper can unlink an idle file
+  after a reader resolves its path but before it opens it; every read path now treats that race as
+  the same ordinary empty result it returns when the file was already gone, instead of a 500.
+
 - **The MCP wheel and source distribution carry the Apache-2.0 legal files they declare.**
   The MCP project now includes exact copies of the repository `LICENSE` and `NOTICE`. CI verifies
   both built artifacts use the required archive paths, contain byte-identical legal files, and
--- a/src/store.py
+++ b/src/store.py
@@ -19,7 +19,7 @@ import time
 import unicodedata
 from collections import OrderedDict
 from collections.abc import Iterator
-from contextlib import contextmanager
+from contextlib import contextmanager, suppress
 from datetime import UTC, datetime
 from functools import lru_cache
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


rioyu123:
@codex review

almondous:
Revalidated the reader/reaper race on current main `e4c4f73`. Driving the real idle reaper between path selection and open still produces `FileNotFoundError` in all four readers; ASGI HTTP checks also reproduce 500s for room and note reads.

I tested a local forward-port of your reader changes, preserving today's sharded layout and `last_seq` floor fallback. The latter matters: after a real reap, returning an unconditional zero would lose the retained cursor floor.

Thirteen additional checks produce six failures on main and all pass with the forward-port. Controls verify that permission errors remain visible, an already-open descriptor retains its snapshot, and export's existing missing-file handling remains unchanged. HTTP reads become an empty 200 for rooms and 404 for notes.

The complete local suite passes 795 tests, with one existing skip and 98.05% coverage. Ruff, formatting, ty, and caps pass; store code-lines decrease from 987 to 985. No changelog, size baseline, API shape, or dependency changes.

All 138 official Chromium checks pass too.

This is a local adaptation for review, not a rebase of your branch or a replacement PR. I can share the patch if useful. Tested on macOS / Python 3.12.9; Linux CI remains necessary.

rioyu123:
Thanks for rechecking this against current main, especially the retained `last_seq` floor. Could you share the forward-port, including any test changes, as a patch or a branch link? I will review the changes and run the Linux checks before updating this PR. That should also help avoid duplicating the work you have already done.

almondous:
Thanks for taking a look. Here are the [patch, application instructions, and verification notes](https://gist.github.com/almondous/2cb564dc069e17cc7dee71a9f8fc15fe).

It targets main `e4c4f73f3b28612d7161170b11e08e580b02123a`, confirmed by a fresh fetch today, not the old PR head. The patch contains only the four-reader changes in `src/store.py` and the 13-case regression test file, preserving the current `last_seq` floor fallback.

Before sharing, I rechecked patch application and reran the tests: 6 failures / 7 passes on unchanged main; 13 passes with the fix; full suite 795 passed with one existing skip and 98.05% coverage. Ruff, format, ty and size caps pass. The gist includes reproduction steps and the patch checksum.

This remains a supplemental forward-port of your fix for the existing PR, with attribution, not a replacement PR or an update to your branch. Linux checks are still outstanding; thank you for offering to run them.

rioyu123:
Thanks, @almondous. I reviewed the forward-port and reproduced the Linux regression results: 6 failures on current main, then all 13 cases passing with the fix. The retained `last_seq` floor is preserved.

I've incorporated your patch and tests into this PR, kept the original regressions, and added a state-machine rule for a `last_seq` read interrupted by reaping. The branch is updated with a normal merge, without a force-push. Your contribution is credited in the commit and PR description.

The full Linux suite passes (799 passed, 1 skipped), along with static checks, size caps, all 138 Chromium checks, and 881 HTTP contract cases. The old changelog edit is removed to match the current contribution rules; the size baseline is unchanged.

PR #128, opened 2026-08-25
fix(docs): document the signed lane's real text budget in openapi.json

### LINKED ISSUE #76: openapi.json: the signed GET lane's text param has no description, and it is the lane with less URL budget

justuncase1210-del:
Fixes #76

## What

Adds a `description` to the signed `say` lane's `text` parameter in
`/openapi.json` — it had none, while the unsigned lane's already explained
the real URL-length budget in practice.

## Why

`did`/`sig`/`nonce` eat real budget before `text` even starts on the signed
lane -- 164 bytes (56 + 86 + 19 + 3 separators) against roughly 49 for the
unsigned lane's `nick`. Copying the unsigned lane's description verbatim
would have been wrong, since the signed lane's actual usable budget is
smaller. The number is derived from the same constants the server already
enforces (`_DID_LENGTH`, `didkey.SIG_CHARS`, the nonce's max digits) rather
than hardcoded, so it can't drift out of sync if those change.

## Checks

- [x] `uv run ruff check . && uv run ruff format --check .`
- [x] `uv run ty check`
- [x] Docs that would now be wrong are updated: this *is* the doc fix
- [x] New surface on a world-writable service: nothing new — this only adds
      a description string to an existing, already-enforced schema field

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -494,6 +494,17 @@ def openapi_document(base: str, version: str, max_body_bytes: int, max_wait: flo
                             "name": "text",
                             "required": True,
                             "schema": _TEXT_SCHEMA,
+                            "description": (
+                                "URL-encoded message body, same single-line sweep as the "
+                                "unsigned lane. This lane has far less real budget before "
+                                "the URL length limit bites: "
+                                f"{_DID_LENGTH + didkey.SIG_CHARS + 19 + 3} bytes go to "
+                                "did/sig/nonce and their path separators before text even "
+                                "starts, against roughly 49 for the unsigned lane's nick. "
+                                f"{store.MAX_TEXT_CHARS} is the declared cap either way, but "
+                                "a long non-Latin message that fits the unsigned lane's URL "
+                                "may not fit this one -- use POST for those."
+                            ),
                         },
                     ],
                     "responses": {
--- a/tests/http/test_docs.py
+++ b/tests/http/test_docs.py
@@ -1149,3 +1149,17 @@ def test_the_ai_catalog_lists_only_artifacts_that_resolve(client):
         assert entry["identifier"] and entry["type"] and entry["url"]
         path = entry["url"].split("testserver", 1)[-1] or "/"
         assert client.get(path).status_code == 200, f"{entry['identifier']} -> {path}"
+
+
+def test_the_signed_says_text_param_documents_its_real_budget(client):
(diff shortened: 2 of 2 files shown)


justuncase1210-del:
Read through this — the non-goals list (no wallet requirements, no tokens,
no on-chain state, no rewards/airdrops) is the right call. Worth saying
explicitly: having spent time in the room lobby recently, it's flooded
with "$FLOP check-in" spam from agents chasing a speculative token tied to
this project's name. Keeping this proposal deliberately disconnected from
any of that is a real strength, not just a nice-to-have.

One concrete technical point, from working on #76 (the signed lane's URL
budget): the existing say-signed lane already spends 164 bytes
(did 56 + sig 86 + nonce up to 19 + 3 separators) before text even starts.
A wallet-link statement adds a Solana pubkey (~44 base58 chars), a wallet
signature (~86-88 chars), a challenge, and two timestamps -- easily another
150-180 bytes. If this reuses the same GET-URL-encoded pattern as the
message lanes, that's worth being deliberate about up front.

That said, since this reads as a one-time link/registration action rather
than something riding along on every signed message, it probably doesn't
need to compete with text's budget at all -- a dedicated endpoint sidesteps
the squeeze entirely. Might be worth saying that explicitly in the design
so it doesn't get conflated with the message-posting lanes' constraints
later.

sv:
Triage: closing as a duplicate of #111.

Three open PRs add the same missing `description` to the signed say lane's `text` path parameter in `src/manifest.py` for #76: #111, #128 and #205. All three say the same thing — `maxLength` is the character cap, the URL is the limit that actually bites, and the signed lane has less budget because the DID, signature and nonce are path segments ahead of the text.

Keeping #111: it is the earliest of the three, and it is the only one that hoists a single shared `_TEXT_PATH_PARAM` used by both say lanes rather than describing the signed lane alone — so the two descriptions cannot drift apart later, which is the drift that produced #76 in the first place.

Per CONTRIBUTING's "Overlapping work", review or build on #111 rather than carrying a parallel branch. The concrete byte arithmetic here (did/sig/nonce overhead against the unsigned lane's nick) is the sharpest of the three and is worth folding into #111.

---

PR #129, opened 2026-08-25
fix(http): a deeply nested JSON body is a 400, not a 500

### LINKED ISSUE #130: deeply nested JSON body answers 500 instead of 400
**Problem**

A POST whose body is deeply nested JSON answers 500 on `/r/{room}` and `/kv/{ns}/{key}`. The contract documents only 400/413/429 for these lanes; 500 appears nowhere, so a client reading the machine-readable docs has no way to interpret it except "the server broke", and will typically retry the identical request instead of correcting the body.

**Reproduce**

```bash
curl -s -o /dev/null -w '%{http_code}\n' -X POST http://localhost:8080/r/lobby \
  -H 'Content-Type: application/json' \
  --data-binary "$(python3 -c "print('['*20000 + ']'*20000, end='')")"
# -> 500
```

The payload is about 40 KB, well under MAX_BODY, so no size guard catches it first.

**Cause**

`json.loads` recurses per nesting level and raises `RecursionError` at around twenty thousand nested arrays. `RecursionError` is not a `ValueError`, so the handler in `read_json` does not catch it and Starlette's default 500 answers.

**Expected**

The ordinary malformed-body refusal (400). The body genuinely is not a usable object, so the existing refusal path already describes it correctly.

**Fix**

Catch `RecursionError` beside `ValueError` in `read_json`. No new status, no new branch, core line count unchanged. #129 has that change with a regression test — happy to rework or close it if this would be handled differently.

Jr-kenny:
Closes #130.

## Problem

POST a body of ~20k nested arrays (about 40 KB, well under the 256 KiB cap) to `/r/{room}` or `/kv/{ns}/{key}` and the server answers 500. json.loads recurses per nesting level and raises RecursionError, which is not a ValueError, so read_json's handler misses it and Starlette's default 500 goes out. That status is documented nowhere on these lanes: the contract promises 400/413/429.

## Fix

Catch RecursionError beside ValueError in read_json. Callers get the ordinary malformed-body refusal either way. No new branch and no new status; core code-lines unchanged (`uv run sz.py --check` stays at 1848).

Abuse impact: none. This removes an undocumented status, it adds no surface.

## Testing

- `uv run ruff check .`
- `uv run ty check`

New regression test posts a depth-20_000 array and asserts 400. Fails with a 500 without the fix.

One judgment call: folded RecursionError into the existing except rather than giving the deep case its own message, so bad bodies keep one refusal path. Happy to split it out if you'd rather it say so explicitly.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1010,7 +1010,11 @@ async def read_json(request: Request) -> dict | Response:
             return text(f"{too_large}\nthe stream passed it before it ended.", 413)
     try:
         payload = json.loads(bytes(raw) if raw else b"{}")
-    except ValueError as exc:
+    # RecursionError beside ValueError: loads recurses per nesting level, and a deeply
+    # nested array raises it rather than ValueError. Uncaught, that escaped as a 500 —
+    # an undocumented status on a lane the contract says answers 400 for malformed
+    # bodies. Same refusal either way: the body is not a usable object.
+    except (ValueError, RecursionError) as exc:
         return text(
             f"400 body must be JSON, and this did not parse: {exc}.\n"
             'send an object like {"from":"bot","text":"hello"} for a room, or '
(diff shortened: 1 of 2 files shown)


Jr-kenny:
Closing: upstream #123 switched body parsing to orjson, which enforces a depth limit and raises JSONDecodeError (a ValueError), so read_json's existing handler now catches the deep-nesting case on main. The regression test in this branch passes against current main unfixed, confirming it is fixed upstream. Thanks!

PR #132, opened 2026-08-25
fix(http): a failed room write no longer spends the creation token

### LINKED ISSUE #131: room-creation budget is spent by writes that fail validation
**Problem**

A room write that fails validation still spends its IP's room-creation token. With the default `RATE_ROOMS_PER_DAY=20`, twenty malformed requests lock that IP out of creating rooms until tokens refill, and the eventual refusal gives the caller no hint why.

**Reproduce** (`CHAT_RATE_ROOMS_PER_DAY=3`)

```bash
curl -o /dev/null -w '%{http_code}\n' "http://localhost:8080/r/waste1/say/bot/%20%20%20"   # 400
curl -o /dev/null -w '%{http_code}\n' "http://localhost:8080/r/waste2/say/bot/%20%20%20"   # 400
curl -o /dev/null -w '%{http_code}\n' "http://localhost:8080/r/waste3/say/bot/%20%20%20"   # 400
curl -o /dev/null -w '%{http_code}\n' "http://localhost:8080/r/legitroom/say/bot/hello"    # 429
```

Each whitespace-only text 400s inside `store.append`; the fourth write is valid and should succeed.

**Cause**

`_room_create_gate` charges before the write runs (which is what keeps a room from slipping past the gate), and `_settle_room_budget` refunds afterwards unless this request really created the room. But when `store.append` raises `StoreError` — a nickname failing `valid_name`, or text swept to empty by `clean_text` — the exception unwinds past the settle call and the charge sticks although nothing was written. The comment on `_room_write_gate` states the rule this breaks: a token is only ever spent on a write that would otherwise have been accepted.

**Expected**

Only writes that bring a room into existence spend the budget.

**Fix**

Settle the creation charge in the `StoreError` handler against an empty record: with no `seq` present, `_settle_room_budget` treats the write as not-a-creation and refunds. One line plus its reasoning comment; every room-write lane is covered because they share that handler. #132 has this with a regression test — glad to adjust it if the refund belongs elsewhere.

Jr-kenny:
Closes #131.

## Problem

The room-creation gate charges a token before the write runs, and on the happy path `_settle_room_budget` refunds it unless this request actually created the room. But when `store.append` raises `StoreError` (a nickname that fails `valid_name`, or text that sweeps to empty in `clean_text`), the exception unwinds past the settle call and the charge stays spent. With the default budget that means three malformed writes lock an IP out of creating rooms until tokens refill, which reads to the caller like the service is broken rather than like it sent junk. The comment on `_room_write_gate` already states the intended rule: a token is only ever spent on a write that would otherwise have been accepted. A caller-input 400 is not such a write.

## Fix

Settle the creation charge in the `StoreError` handler (`on_bad_input`) against an empty record: an empty dict carries no `seq`, so `_settle_room_budget` treats the write as not-a-creation and refunds. One line plus its reasoning comment; every room-write lane is covered because they all funnel through the same handler, and note writes are untouched since their gate never sets the charged flag.

This grows core by one line, so `sz-baseline.json` bumps `core/app.py` to 920 with the new pattern as the justification.

## Testing

- `uv run ruff check .`
- `uv run coverage run -m pytest tests -q`

New test in `tests/http/test_limits.py`: two whitespace-only writes (both 400) followed by a valid third room write must be 200 under `RATE_ROOMS_PER_DAY=2`. Before the fix the third write 429s.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1689,6 +1689,10 @@ async def on_method_not_allowed(request: Request, exc: Exception) -> Response:
 
 
 async def on_bad_input(request: Request, exc: Exception) -> Response:
+    # A 400 from a room write means the create gate charged a token for a write that did
+    # not happen (bad nick, text that sweeps to empty). Settling against an empty record
+    # carries no seq, so the charge refunds and only real creations spend the budget.
+    limit._settle_room_budget(request, {}, RATE_ROOMS_PER_DAY, ip_header=CLIENT_IP_HEADER)
     return text(f"400 {exc}", 400)
 
 
--- a/sz-baseline.json
+++ b/sz-baseline.json
@@ -1,7 +1,7 @@
 {
   "core_total": 2028,
   "caps": {
-    "core/app.py": 979,
+    "core/app.py": 980,
     "core/config.py": 57,
     "core/didkey.py": 57,
     "core/limit.py": 132,
@@ -10,9 +10,9 @@
   },
   "files": {
     "core/app.py": {
-      "raw_lines": 1796,
-      "code_lines": 979,
-      "comment_lines": 264,
+      "raw_lines": 1800,
+      "code_lines": 980,
+      "comment_lines": 267,
       "tokens_per_line": 7.89
     },
     "core/config.py": {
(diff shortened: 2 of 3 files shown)


sv:
Closing — this, #99, and #325 are three independent fixes for the same underlying defect (a `StoreError` during room creation never refunds the charged budget). Keeping #325 (see the note on #99 for why). One thing worth flagging for whoever merges #325: this PR closes issue #131 explicitly; #325's description doesn't reference it, so #131 may need to be closed manually once #325 lands.

---

PR #133, opened 2026-08-25
fix: ask the name grammar before the class question (#109)

### LINKED ISSUE #109: A name the grammar refuses gets a class answer: mb-FOO says 403 'send a signature', say-signed then says 400

VendettaGamesHQ:
Closes #109.

## The change

A room class is a name **prefix**, so `room_classes` reads a class marker off a string this service will never store. Both write gates asked their class question before anything read the grammar, so the correction the answer named then contradicted it:

```
GET /r/mb-FOO/say/bot/hi        403 … it takes signed writes only,
                                    send: GET /r/mb-FOO/say-signed/…
GET /r/mb-FOO/say-signed/…      400 bad name 'mb-FOO'
```

The ownership lane split the same way on the identical key: `room-owners` asks `store.ownable` first (403 "cannot be owned"), while `room-allow` reaches `note_get`, which validates (400 "bad name").

Both gates now validate first. `valid_name` raises `StoreError`, which is already mapped to the same 400 the storage layer produces, so this **borrows that message rather than restating it** — no new error text, and the signed lane's answer becomes every lane's answer.

## Two things worth flagging rather than burying

**The core cap moves by one.** Folding the calls into existing statements kept it to a single added code line, which puts `core/app.py` at 942 and the core total at 1903. I did not golf it further: `ns, key = store.valid_name(ns), store.valid_name(key)` is deliberately not the shorter `and` form, because `and` short-circuits and the key would then only be validated when `ns` happened to be the nonce namespace — a validation that does not run, which is the bug being fixed. Happy to take a different shape if you would rather the cap held.

**The tests go in `tests/http/test_lane_parity.py`**, because that is precisely what this bug is: one logical write, several lanes, disagreeing — invisible to any single-lane test. They reuse `_client`'s `_keypair` / `_say_signed` rather than re-implementing the signing construction.

The signed lane is really signed, and that mattered: `_signer` runs before the gate, so a placeholder `did:key` is answered on the DID and never reaches the name check. My first version did exactly that and CI caught it — the test would have asserted nothing about names while appearing to cover the lane.

There is also a mirror test: a name the grammar **accepts** must still get its class answer, or the whole thing would pass just as happily if the gate started refusing every mailbox.

## Verification

`ruff check`, `ruff format --check`, `sz.py --check` and `sz.py --caps` pass. The pytest suite needs POSIX `fcntl`, which the machine I wrote this on does not have, so it was verified in CI rather than claimed here:

- **with the fix:** all green — <https://github.com/VendettaGamesHQ/technocore-chat/actions/runs/32808277867>
- **the same tests against unfixed `app.py`:** red, with exactly the reported symptoms — `mb-FOO` and `mb-p-FOO` answering 403 on the unsigned lane and `D-FOO` answering 403 on `room-owners` — <https://github.com/VendettaGamesHQ/technocore-chat/actions/runs/32808378645>

The `d-`, `e-` and `p-` parameters do not reproduce the bug (no gate fires on those classes); they are there for the consistency the file exists to assert.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -829,8 +829,15 @@ def _allowed_keys(room: str) -> set[str]:
 
 def _room_write_gate(request: Request, room: str, signer: str | None) -> Response | None:
     """Every write to a room passes here, signed or not. Fail closed: a class that demands
-    a signature refuses the unsigned lane outright, and the reply says what to send."""
-    denied = _reject_if_events_room(room)
+    a signature refuses the unsigned lane outright, and the reply says what to send.
+
+    The grammar is asked FIRST. A class is a name prefix, so `room_classes` reads one off a
+    string this service will never store — and the gate then answered `mb-FOO` on class
+    grounds with a 403 whose named correction, the signed lane, reaches `store.append` and
+    answers 400. One write, two answers, and the 403 sent the caller to the lane that
+    contradicted it. `valid_name` raises StoreError, which is already mapped to the same
+    400 the storage layer produces, so this borrows the message rather than restating it."""
+    denied = _reject_if_events_room(store.valid_name(room))
     if denied:
         return denied
     if store.is_mailbox(room) and signer is None:
@@ -1121,6 +1128,10 @@ def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Respo
     stranger cannot rewrite — without that, ownership is a note anyone can overwrite, which
     is not ownership.
     """
+    # Same order, same reason as `_room_write_gate`: `ownable` reads a class marker off a
+    # name the grammar refuses, so `D-FOO` was answered "cannot be owned" (403) while the
+    # allow-list lane, which validates on the way to the note, answered "bad name" (400).
+    ns, key = store.valid_name(ns), store.valid_name(key)
     if ns == store.NONCE_NS:
         return text(
             f"403 /kv/{store.NONCE_NS} is written by the server only — it is the replay "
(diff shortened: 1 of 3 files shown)


VendettaGamesHQ:
Closing this — #108 got here first and is the better version.

Entirely my fault: I read the issue and went straight to the code without checking the open PR queue, which is a thing I should have done in a repo with sixty-odd of them. Apologies for the noise.

#108 covers the same two gates and also carries the CHANGELOG and AGENTS entries this was missing. One thing from here that may be worth folding in if you want it, and may equally not be:

- I put the regression in `tests/http/test_lane_parity.py` rather than `test_rooms.py`, on the reasoning that this bug *is* a lane disagreement — one logical write, several lanes, different answers — which is exactly what that file exists to catch and what a single-lane test cannot see.
- The mirror assertion: a name the grammar **accepts** must still get its class answer, so the fix cannot pass by refusing every mailbox.

Both are three lines each against #108 if useful. Happy either way — the fix is the point and it is already in a better PR.

One note for anyone hitting the same thing: the signed lane needs a *real* signature to reach the name check, because `_signer` runs before `_room_write_gate`. My first test used a placeholder `did:key` and passed through the DID validator instead, so it asserted nothing about names while appearing to cover the lane. CI caught it.

PR #136, opened 2026-08-25
fix(store): capacity refusals carry the used count, not only the cap

### LINKED ISSUE #85: DID note namespace is full, blocking the documented identity-publishing pattern

antfleet-ops:
Closes #85's smallest direction — the one its population data made concrete.

**The problem.** A client refused at a count cap gets a message naming the ceiling but not the occupancy. The #85 audit showed what that costs in practice: 634 of the `did` namespace's 5120 slots are unresolvable junk, and a blocked client polls blind (~45s in the auditor's case) because the refusal supports exactly two strategies — *wait* and *retry* — with no figure to wait against. Meanwhile the room byte-budget refusal already reports "N MiB of a M MiB budget". The count refusals were the ones hiding how full it is.

**The change** (store.py, +1 code line):

- `_at_capacity(cap, what, used)` — the count it just measured travels into the message: `note limit reached (5120 is the cap, 5120 in use, and this would be a new one)`
- the global note-total refusal says `total of MAX across all namespaces`
- room creation passes its count through the same path

That is the whole ask from #85's "possible directions" item 4: with `used` visible and the existing "idle notes are reclaimed after 7 days" line, a well-behaved client can compute an honest backoff instead of holding a tight loop against a service that is shedding load.

**Tests.** `test_a_capacity_refusal_carries_the_numbers_a_caller_acts_on` already states the rule — *matching only the opening words leaves every number free to be wrong* — so its three regexes now pin the used figures (`1 is the cap, 1 in use`, `1 of 1 across all namespaces`). Full suite: 321 passed; ruff, format, ty clean; `beautiful_chat.sh` passes end-to-end.

**sz.** store.py stays under the immutable `core/store.py` cap (996 of 1010 code lines on current `main`), so no `sz-baseline.json` change is needed. The earlier hand-edit to that ratchet was dropped — queue-guard protects it as a maintainer-regenerated file on fork PRs (CONTRIBUTING "Overlapping work").

_Rebased onto current `main` (0.13.0); the `sz-baseline.json` edit dropped so `protected-files` passes. Full `just check` green locally: 774 passed, 1 skipped; ruff, format, ty, `sz.py --caps`, coverage._

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -2213,13 +2213,15 @@ def _count_new_room(root: Path, delta: int) -> None:
     _write_note_count(root, max(0, count + delta), used, name=USAGE_FILE)
 
 
-def _at_capacity(cap: int, what: str) -> StoreError:
+def _at_capacity(cap: int, what: str, used: int) -> StoreError:
     """The refusal, in one place because two callers raise it (rooms count both a cap and a
     byte budget). Only *new* names are refused, which is the actionable half: an agent
-    blocked here can always keep working in a room or note it is already using."""
+    blocked here can always keep working in a room or note it is already using. `used`
+    travels with the cap because "how full, of how much" is what lets a client wait for
+    the 7-day reclaim instead of retrying blind."""
     return StoreError(
-        f"{what} limit reached ({cap} is the cap, and this would be a new one). "
-        f"Existing {what}s still accept writes, so reuse one you already have — "
+        f"{what} limit reached ({cap} is the cap, {used} in use, and this would be a "
+        f"new one). Existing {what}s still accept writes, so reuse one you already have — "
         f"GET /rooms shows what exists. Idle {what}s are reclaimed after 7 days "
         f"(a room still on its first message goes after {STILLBORN_SECONDS // 3600} hours)."
     )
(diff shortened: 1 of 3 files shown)


antfleet-ops:
Rebased onto current `main`; dropped the `sz-baseline.json` cap edit — `core/store.py` fits under the immutable cap (996/1010), so the ratchet bump isn't needed — which clears **`protected-files`**. Also fixed a `ruff format` miss that was failing **`lint + tests + image`**. Title set to conventional-commit form; the "a fix has a test that fails without it" guard passes. All required checks green; `just check` passes locally.

PR #138, opened 2026-08-25
fix(store): reserve the server nickname on the unsigned lane

### LINKED ISSUE #137: unsigned writers can forge the server's ~server identity
**Problem**

```bash

```

**Expected**

**Suggested fix**

```python
    raise StoreError(...)
```

Jr-kenny:
Closes #137.

## Problem

An unsigned write can store the nickname `server`, and `who()` renders every non-DID author with a leading `~`. The forged line reads back byte-identical to a genuine event line, and the events room exists precisely so agents can trust `created <name>` lines. The `~server` spelling itself is already unreachable (the name allowlist rejects `~`), so stored `server` is the one path that produces `<~server>` outside `/r/events`.

## Fix

`store.append` refuses the unsigned lane when the nick is `EVENTS_NICK`. The server's own announcements go through `_log_event` into `_write_record` directly and keep working, and the signed lane is untouched because its author is a DID, which cannot collide with a nickname. This costs 2 code-lines past the cap in `sz-baseline.json`, justified as the reservation primitive.

## Testing

- `uv run ruff check .`
- `uv run coverage run -m pytest tests -q`

New test covers all three sides: the unsigned write 400s, the signed lane still posts (DID author), and `_log_event` still announces new rooms in /r/events after the change.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -2442,6 +2442,10 @@ def append(
     primitive that already exists does the rest — `?since=` for incremental reads,
     `?format=json`, `?wait=` for near-real-time, ring retention, the same rate limits.
     """
+    # An unsigned caller storing EVENTS_NICK would render byte-identical to a genuine
+    # event line; signed writes are immune because their author is a DID.
+    if did is None and nick == EVENTS_NICK:
+        raise StoreError(f"the nickname {EVENTS_NICK} is reserved: it signs /r/{EVENTS_ROOM}")
     rec, created = _write_record(root, room, nick, text, did=did, nonce=nonce, sig=sig)
     # Counted here rather than in `_write_record`, so the server's own announcements
     # (`_log_event` writes one per created room) never inflate the message count. This
--- a/tests/http/test_rooms.py
+++ b/tests/http/test_rooms.py
@@ -1235,3 +1235,30 @@ def test_wait_wakes_on_a_write_from_another_process(client, tmp_path):
     messages = held.json()["messages"]
     assert [m["text"] for m in messages] == ["from another process"]
     assert messages[0]["from"] == "otherworker"
+
+
+def test_the_server_nickname_is_reserved_on_the_unsigned_lane(client):
+    """`who()` renders a stored `server` as `~server`, which is byte-identical to how the
+    server's own event lines read. The unsigned lane therefore refuses the nick. The
+    server keeps announcing through _log_event, and the signed lane is untouched because
+    its author field is a DID, which cannot collide with a nickname.
+    """
+    import config
+    import store
+
+    assert client.get("/r/lobby/say/server/created p-fake").status_code == 400
+    assert client.get("/r/lobby/say/serverbot/ok").status_code == 200  # prefix is fine
+
+    # The signed lane keeps working: author is a DID there.
+    did, sign = _keypair()
+    body = store.clean_text("did lane")
+    signed = client.get(
+        f"/r/signedlane/say-signed/{did}/{sign(f'signedlane|1|{body}')}/1/did%20lane"
+    )
+    assert signed.status_code == 200
+
+    # And the server itself still announces new rooms.
+    root = config.ROOT
+    assert client.get("/r/newroom-x9/say/alice/hello").status_code == 200
+    events = store.read_messages(root, "events", limit=5)
+    assert any("created newroom-x9" in m["text"] for m in events["messages"])


djd39448:
Rebased #138 and #227 on current `main` and tested both. As a contributor I'd prefer #138, and I'd back it once the size baseline is refreshed (see the last point).

Both land the same gate in `store.append` (compare `gh pr diff 138` with `gh pr diff 227`):

```
if did is None and nick == EVENTS_NICK:
    raise StoreError(...)
```

Correct fix, valid regression — it fails on unfixed `main` and passes with the gate, and each PR's test also passes against the other's implementation, so the two are interchangeable at the gate. Two reasons I'd pick #138:

1. The refusal message names the rule and the reason, matching how the service writes its other refusals (and CONTRIBUTING notes mutation testing runs over refusal bodies):

```
# 138: raise StoreError(f"the nickname {EVENTS_NICK} is reserved: it signs /r/{EVENTS_ROOM}")
# 227: raise StoreError(f"reserved nickname: {EVENTS_NICK}")
```

2. The test covers more than the two direct outcomes. Beyond `server -> 400`, #138 also asserts a `serverbot` prefix still returns 200 (guards against an over-broad match), the signed lane still works (author is a DID), and the server still announces new rooms. #227 asserts `server -> 400` and that no `<~server>` renders — not wrong, but #138 covers strictly more.

One thing before it goes green: like #227, #138 fails `sz.py --check` on current `main` (`core/store.py` +2 over cap). #138 did ship a baseline bump, but `main` has advanced since and it no longer applies (it conflicts on rebase), so it needs a fresh one-line `sz-baseline.json` refresh. With that, I'd back this over #227. Contributor, not the maintainer.

djd39448:
Re-tested the new head (5f55123) — thanks for the fast turnaround.

Confirmed on my side:

- Rebased on current `main` (e6114b3): `git merge-base --is-ancestor origin/main 5f55123` holds.
- The fix and tests are byte-identical to the version I reviewed above — `git diff 8c5c19a 5f55123 -- src/ tests/` is empty; only `sz-baseline.json` changed.
- Full suite: `uv run pytest` → 382 passed.

The baseline refresh is one step short, though. `uv run python sz.py --check` still exits 1 on this head:

```
core code-lines grew vs sz-baseline.json: core/store.py 792 -> 794 (cap 794)
```

The bump raised `caps."core/store.py"` and `core_total` (792→794, 2015→2017) but left the measured `files."core/store.py".code_lines` at 792, and `--check` ratchets against the `files` section, not only the caps.

One command finishes it on your branch:

```
uv run python sz.py --update-baseline
```

It rewrites `files`/`core_total` from the tree and carries your caps forward untouched. I ran exactly that on your head locally: `--check` then prints `check ok: core code-lines <= baseline (2017)` and exits 0, suite still green. With that commit, this has my backing over #227. Contributor, not the maintainer.

Jr-kenny:
yo yea i just refreshed with `sz.py --update-baseline` as suggested; `--check` exits 0 on each head now. Thanks for running the suite down to the files section, that ratchet detail was easy to miss lol

djd39448 (approved):
Verified 13dce3d: the baseline refresh is complete and this is now fully green on my machine.

- Only `sz-baseline.json` changed since 5f55123; the fix and tests remain byte-identical to the version I reviewed.
- Rebased on current `main` (e6114b3).
- `uv run python sz.py --check` → `check ok: core code-lines <= baseline (2017)`, exit 0.
- `uv run pytest` → 382 passed.

One aside for anyone watching CI: `tests/unit/test_note_count.py::test_the_global_cap_binds_exactly_under_concurrent_processes` flaked once during my runs, but it flakes identically on unmodified `main` (intermittent under repetition on both), so it is pre-existing and unrelated to this change.

As stated above, backing this over #227: same gate, better refusal message, strictly broader test coverage. Approving as a contributor; merge call is the maintainer's.

djd39448:
Happy to help man. Let me know if there is anything else I can do for you. Looking to earn my way onto the team.

bdunn77:
Independent validation of head `de22e8d6` (base `20a4457b`):

- On `origin/main`, `append` has **no** unsigned-lane reservation of `EVENTS_NICK`.
- On this head: `test_the_server_nickname_is_reserved_on_the_unsigned_lane` **1/1 pass**. `/r/lobby/say/server/...` is 400; a `serverbot` prefix still 200s; the signed DID lane is untouched; server room-creation announcements still land in `/r/events`.

No merge blockers found on the reserved-nickname check this PR adds.

PR #143, opened 2026-08-25
docs: warn that a timed-out/5xx signed write may have already landed

### LINKED ISSUE #141: A timed-out signed write can land, and retrying the URL then reports failure on a success
## What happened

```

attempt 2  same URL, resent
```

## Not just me

## Scope

ygd58:
Closes #141. The nonce for a signed write is consumed server-side before the client gets a response, so a timeout or 5xx doesn't mean the write failed — and for room-owners/room-allow specifically, the replay counter is persistent, so a retried URL is refused forever rather than eventually re-accepted. Adds a note near OWNED ROOMS telling callers to re-read state instead of resending the same signed URL.

## Abuse impact

None. This PR only adds a warning paragraph to `src/manual.md`; it introduces no new route, parameter, or persistent state, and changes no enforced behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
+PATCH: the manual now warns that a timed-out or 5xx signed write may have already landed server-side (the nonce is consumed before the client hears back), and that a retried URL against room-owners/room-allow is refused forever rather than eventually re-accepted, since their replay counter never expires.
+
 ### Changed
 
 - **`/rooms` no longer re-walks every room on every message.** Its cache was validated against a
@@ -130,6 +132,7 @@ clients depend on tightens.
 - **A note create scans its namespace once, not twice.** The capacity check ran before the
   create gate and again inside it. The pre-gate call now checks only the global cap, which is
   a file read, so a full store still refuses without queueing for the gate.
+
 
 ## [0.9.0] - 2026-08-25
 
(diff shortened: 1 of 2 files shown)


djd39448 (approved):
Tested on a worktree rebased on current `main`. Docs suite is green:

```
$ uv run python -m pytest tests/http/test_docs.py
47 passed
```

The note states the shared failure mode correctly: a timed-out or 5xx signed write may already have landed, because the nonce is consumed server-side before the client hears back — so don't resend the same signed URL (it comes back as a replay), re-read state, and sign a fresh, greater nonce.

It also carries the part #141 was filed for: the persistent-counter asymmetry. In `room-owners`/`room-allow` the counter at `/kv/room-nonce/<room>` never ages out, so a stale retry there is refused permanently:

```
$ gh pr diff 143 -- src/manual.md | grep '^+' | grep -iE 'room-owners|room-allow|persistent|forever|never expires'
# 4 matching lines
```

That's the distinction the reporter called "worth a doc line rather than a shrug," so keeping it is the point of the change. Complete and accurate — backing it as a contributor.

One graft worth folding in (follow-up, not a blocker): #223 places the same guidance under NONCE — where the reporter asked for it, and where the existing single-use text already lives — and is more economical. Moving this note there and tightening toward #223's wording, while keeping the persistent-counter sentence, would combine the two. This PR also edits CHANGELOG alongside the manual, so it needs the usual trivial CHANGELOG rebase on merge.

ygd58:
Rebased on current main (through 0.9.3) — the CHANGELOG conflict was exactly the trivial kind you flagged, resolved by keeping main's 0.9.1/0.9.2/0.9.3 entries intact and moving this PR's Unreleased line back to the top. 390 tests pass, lint/format/ty/sz all green.

Left the #223 merge as a follow-up rather than folding it in here, since it changes where the note lives (NONCE vs OWNED ROOMS) and I'd rather not touch content djd39448 already reviewed and approved without a specific ask.

sv:
Closing in favor of #328, which covers the same ground plus OpenAPI text and a doc-assertion test. Thanks for documenting this first.

---

PR #146, opened 2026-08-25
fix(store): recreated room does not restart seq at 1, starving old cursors

### LINKED ISSUE #139: cursors silently starve when a reaped room is recreated (seq restarts at 1)
**Problem**

When a room is reaped and later recreated under the same name, `seq` restarts at 1, so every reader still polling with a cursor from the old generation never sees new messages again. The starvation is silent: reads answer 200 with `count: 0`, forever.

**Reproduce** (against a fresh CHAT_ROOT, using the store directly for clock control)

```python
import os, time
from pathlib import Path
import store

root = Path("/tmp/repro")
for i in range(6):
    store.append(root, "d-talk", ["alice", "bob"][i % 2], f"msg {i}")
cursor = store.read_messages(root, "d-talk")["last_seq"]        # 6

# 8 idle days pass; next write's reaper takes the room
p = store.room_path(root, "d-talk")
when = time.time() - (store.IDLE_SECONDS + 60)
os.utime(p, (when, when))
store._reap(root)

store._write_record(root, "d-talk", "mallory", "fresh start")   # new generation, seq 1
store.append(root, "d-talk", "carol", "can anyone hear me?")    # seq 2

print(store.read_messages(root, "d-talk", since=cursor))        # count: 0, forever
```

**Why this is worth an issue despite restart-at-1 being deliberate**

The compaction docstring already fixes the smaller version of this promise: compacting an `e-` room to nothing is avoided because restarting the sequence would strand every cursor pointing past it. Reap + recreate strands cursors the same way, just across a generation boundary instead of within one, and today nothing tells the reader anything changed. `first_seq > since+1` normally signals missed lines, but here `first_seq` itself reset, so even that signal points backwards.

Concretely reachable without any attack: a room sits on one unanswered message, the 24h stillborn rule takes it, someone reuses the name days later, and every monitor of the old conversation goes deaf while believing the room is simply quiet.

**Possible directions**

- Document it as accepted (the manual does not currently mention that a reused room name starts a new sequence).
- Have the recreated room carry a marker record or metadata bumping the sequence floor above the previous generation's high-water mark (the old `.reaped` pass could leave the last seq behind cheaply).
- Include something like an epoch/generation in the read view so a client can detect the discontinuity and resync.

Happy to take any of these on if you point at the direction you prefer.

ygd58:
Closes #139.

## Problem
A room reaped after IDLE_SECONDS (or the 24h stillborn rule) and recreated under the same name restarted seq at 1. Any cursor still polling with a seq from the old generation read `count: 0` forever — silently and permanently, because `first_seq` itself reset instead of jumping forward past `since`, so none of the existing missed-lines signals fired.

## Fix
Adds `SEQ_FLOOR_NS` (`room-seq-floor`), a server-only note namespace mirroring `room-nonce`'s pattern. The reaper leaves the old generation's `last_seq` there right before unlinking the room file. The next write to that name assigns `max(last_seq, floor) + 1` instead of just `last_seq + 1`. The floor note ages out on the ordinary note idle rule — it's not permanent state, just needs to outlive the gap until the name is reused.

Blocked from public writes the same way `room-nonce` is (merged into the same `_note_write_gate` branch to keep the net line cost down).

## Testing
- `uv run pytest tests -q` (330 passed) — includes the Hypothesis stateful model, updated to track the floor's own lifecycle (creation on reap, aging, expiry) rather than assuming seq always resets to 0
- 3 new regression tests reproducing the issue's exact repro, the server-only gate, and the floor note's own IDLE_SECONDS expiry

## sz-baseline.json
core/store.py 743->760, core/app.py 941->945 (net +4 after merging the new gate branch into the existing NONCE_NS one instead of adding a parallel block). Justified the same way #138 raised its own cap: the reservation primitive costs real lines, and the alternative (silent permanent starvation) is worse.

## Abuse impact

Low. The new SEQ_FLOOR_NS namespace is server-written only — blocked from public writes the same way room-nonce is, verified by a dedicated test. It ages out on the ordinary note idle rule, so it adds no unbounded state: worst case is one extra reclaimed note per reaped room, bounded the same way every other note namespace already is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
+PATCH: a room reaped after the idle/stillborn rule and later recreated under the same name no longer restarts its seq at 1. A cursor from the old generation used to read `count: 0` forever, silently, since `first_seq` itself reset instead of jumping past `since`; the reaper now leaves a seq floor behind that the next write clears.
+
 ## [0.9.0] - 2026-08-25
 
 MINOR: the operator levers the 2026-08-25 flood needed and did not have, plus faster crypto, JSON
--- a/src/app.py
+++ b/src/app.py
@@ -1179,11 +1179,15 @@ def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Respo
     stranger cannot rewrite — without that, ownership is a note anyone can overwrite, which
     is not ownership.
     """
-    if ns == store.NONCE_NS:
+    if ns in (store.NONCE_NS, store.SEQ_FLOOR_NS):
+        role = (
+            "the replay counter for signed ownership writes"
+            if ns == store.NONCE_NS
+            else "the seq floor a reaped room leaves behind so a same-named room cannot "
+            "reuse old sequence numbers"
+        )
         return text(
-            f"403 /kv/{store.NONCE_NS} is written by the server only — it is the replay "
-            "counter for signed ownership writes. Read it freely.",
-            403,
+            f"403 /kv/{ns} is written by the server only — it is {role}. Read it freely.", 403
         )
     if ns not in (store.OWNERS_NS, store.ALLOW_NS):
         if signer is not None:
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -273,13 +273,15 @@ _BAD_NAME = _plain(f"Malformed name or parameter ({_NAME_RULE}).")
 # the character cap, and — on the note lane — a signed write aimed at a namespace that
 # does not take one. Every refusal names its own correction in the body; what was missing
 # was any statement in the *contract* that a 400 is reachable here at all.
-# The 403 the note lanes share. Three namespaces are not world-writable: the server-only
-# replay counter, and the two ownership namespaces, which refuse a claim on a room that is
-# not ownable, already owned, or already has people talking in it.
+# The 403 the note lanes share. Four namespaces are not world-writable: the two
+# server-only ones (the replay counter, and the seq floor a reaped room leaves behind —
+# #139), and the two ownership namespaces, which refuse a claim on a room that is not
+# ownable, already owned, or already has people talking in it.
 _RESERVED_NAMESPACE = _plain(
-    f"A reserved namespace refused the write: `{store.NONCE_NS}` is server-written, "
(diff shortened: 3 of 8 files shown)


Jr-kenny:
Ran your implementation against the #139 repro plus a few edge cases before you get too deep into review cycles. The core mechanism works: idle reap leaves floor 6, recreated room assigns seq 7, stale cursor sees the new message, stillborn path leaves its floor too, and the under-lock recheck ordering is right.

One residual hole, same shape as the issue but on a longer clock: the floor note itself ages out after IDLE_SECONDS. Room reaped at seq 6, floor 6 left behind, nobody recreates the name for 7+ days (floor expires on the ordinary note idle rule), then someone reuses the name. Fresh store check:

```
floor: 6
floor expired: True
seq after floor expiry: 1
```

So the fix shrinks the starvation window from "any reuse" to "reuse within 7 days of the reap". Whether that is acceptable is a design call: a paused agent resuming after two weeks is exactly the reader the manual's "your cursor never rewinds" line is for, and it rewinds for them. Closing it fully would need either a persistent floor (which reintroduces the unbounded state this service refuses) or an explicit epoch/generation field in the read view so clients can detect the discontinuity instead of trusting seq monotonicity.

Not a blocker for the common case; worth one sentence in the docs either way. Rest of the diff looks tight.

ygd58:
Good catch, and the framing is exactly right — the floor is deliberately not persistent state (matches the service's own refusal to reintroduce unbounded storage), so the residual window is a real, honest trade-off rather than an oversight.

Went with the docs sentence rather than either heavier option for the reason you gave: a persistent floor undoes the design this service is built around, and an epoch/generation field is a bigger protocol change than this issue calls for on its own — worth its own discussion if it's wanted, not bundled into a starvation fix.

Added the caveat right after the REUSED NAMES paragraph this PR introduced: the floor ages out on the same 7-day idle rule as any other note, so reuse within that window keeps the guarantee and reuse after it restarts at seq 1 as before. Pushed.

ygd58:
Closing this myself: #343 (merged) covers the same ground more completely — it addresses both the seq-continuity fix this PR made and the discontinuity-exposure follow-up @Jr-kenny flagged in review, which I had left as a docs-only mitigation here. No need to carry a duplicate forward. Thanks for the review, @Jr-kenny — the direction you pointed at is exactly what landed.

PR #156, opened 2026-08-25
fix: preserve leading zero bytes in _b58decode (base58btc compliance)

### LINKED ISSUE #155: _b58decode drops leading zero bytes (base58btc spec violation)
## Summary

## The bug

```python
    n = 0
    for ch in raw:
        digit = _B58_INDEX.get(ch)
        if digit is None:
        n = n * 58 + digit
```

## Reproduction

```python
from didkey import _b58decode

encoded = ""
while n:
    n, rem = divmod(n, 58)

decoded = _b58decode(encoded)
```

## Impact

## Suggested fix

```python
    n = 0
    for ch in raw:
        digit = _B58_INDEX.get(ch)
        if digit is None:
        n = n * 58 + digit
```

dolmaciabdullah-byte:
## Summary

- `_b58decode` drops leading `0x00` bytes because pure integer arithmetic loses them base58btc `'1'` characters that encode leading zeros were silently discarded
- The test helper `_multibase` in `tests/_client.py` had the same bug on the encode side
- Added a round-trip test for payloads with leading zero bytes

## Impact

Not reachable from Ed25519 `did:key` (the `0xed01` multicodec prefix never starts with `0x00`), but violates the [base58btc spec](https://datatracker.ietf.org/doc/html/draft-msporny-base58-03#section-3) and would break any future key type whose multicodec varint starts with a zero byte.

## Test plan

- [x] All 369 existing tests pass
- [x] New `test_b58_leading_zero_bytes_round_trip` verifies the fix

Fixes #155


djd39448:
I tested this alongside #182, which fixes the same issue (#155). Sharing what I found so the work isn't duplicated.

The executable change here is identical to #182 — same `_b58decode` body, same symmetric `tests/_client.py` encoder fix, byte for byte (compare `gh pr diff 156` with `gh pr diff 182`). The only source difference is that this PR carries explanatory comments and #182 does not.

Because the code is the same, I pointed to #182 as the single consolidation target, for two concrete reasons: its regression test is strictly stronger — it also asserts the encoder emits leading `1`s (`encoded.startswith("11")`), not just the round-trip — and its body carries the explicit `Fixes #155` link.

The one thing this PR does better is documentation: the comment noting that a real Ed25519 did:key can't produce a leading-zero payload (`0xed01` never starts with `0x00`) while the codec still has to round-trip as a general primitive. That's the better-documented half, and worth carrying onto whichever PR lands; I flagged it on #182 as the piece to graft in.

Neither PR is currently green: both grow `didkey.py` past its cap without bumping `sz-baseline.json`, so `uv run sz.py --check` fails (`didkey.py 57 -> 59, cap 57`). One-line fix on whichever is taken. Not a maintainer — just a contributor who ran both. Thanks for the fix either way.

dolmaciabdullah-byte:
@sv Addressed @djd39448's review the two gaps are now closed:

Stronger test the round-trip assertion now also checks the encoder side (encoded.startswith("11")), so a bug that cancels itself out on both sides can't silently pass. This was the edge #182 had over this PR; it's now equivalent.
sz-baseline bump didkey.py cap updated 57 → 59 to match the added comment lines, uv run sz.py --check passes.

All existing tests pass. Ready for review.

djd39448:
Thanks — the encoder-side assertion is exactly the gap I raised, and it closes it. `encoded.startswith("11")` means a symmetric bug that cancels out across encode/decode can no longer pass the round-trip, which was the one edge #182 had. That is now moot in the best way, since sv closed #182 in favour of this PR.

The size check does not pass at the current head (`1c5a180`), though. Running it unmodified on that exact commit:

```
$ python3 sz.py --check
core code-lines past cap: core/didkey.py 59 > cap 57
...
core total (code)                 1999
$ echo $?
1
```

`chore: bump didkey.py cap after leading-zero fix` raised the **measured** row to `code_lines: 59` but left `caps["core/didkey.py"]` at `57`, and `--check` fails on a value past its cap even when the baseline was raised to match it. Setting the cap to 59 is all it needs.

Two things that probably hid this, both worth knowing rather than being about this PR:

- **The failure line prints *before* the table**, so it scrolls off the top, and `sz.py --check | tail` (or any pipe) swallows the exit code. It reports success by looking like it succeeded. Check `$?` directly.
- **Comment lines are never capped.** The commit message explains the bump as matching added comment lines, but `--check` reads `code_lines` only — `raw_lines`/`comment_lines`/`tokens_per_line` are written and never read back. The +2 here is real code from the leading-zero fix, which is a legitimate reason to raise the cap; it just is not the comments.

Separately, the baseline in this branch is from 0.9.1 (`core_total` 1999/2000) while main is at 0.10.0 (2127/2128), which is what makes this `CONFLICTING`. The recipe that has worked for me three times running on the same file: take **main's** baseline, re-measure the merged tree with `--update-baseline` (it carries caps forward untouched), then raise **only** the caps your change actually moves, by exactly its delta, leaving main's headroom alone. That keeps the diff to the rows you moved and reviewable as arithmetic.

Full disclosure of an overlap: my open #236 also raises `didkey.py` (57 -> 60). Whichever of us lands second will conflict on that one row. Nothing to resolve now — just so neither of us is surprised by it.

dolmaciabdullah-byte:
Thanks for the detailed breakdown @djd39448 you're right, the cap needs to be 59. I'll rebase onto current main, fix the sz baseline cap, and rewrite the commit authors. Will push an updated branch shortly.

djd39448:
The cap is right now — `caps["core/didkey.py"]` is 59, which is the number. But `sz.py --check` still exits 1 at `833aa32`, for a different reason than last time:

```
$ python3 sz.py --check
core code-lines grew vs sz-baseline.json: core/didkey.py 57 -> 59 (cap 59)
$ echo $?
1
```

Note it now names the cap approvingly in its own error — `(cap 59)` — while still failing. That is because **`sz-baseline.json` holds two independent things**, and only one of them moved:

- `files["core/didkey.py"]["code_lines"]` is the **ratchet**: the measurement the last update recorded. It still says **57**. `--check` measures the tree (59) and fails on growth past this, *regardless of the cap*.
- `caps["core/didkey.py"]` is the **policy ceiling**. You correctly raised this to 59.

Both have to move together. The cap alone says "59 would be allowed" while the ratchet still says "the last measurement was 57", so the check reads it as unrecorded growth.

There is a second, smaller one in the same file: `caps["core_total"]` is still **2128**, while the per-file caps now sum to **2130**. Nothing enforces that identity today — `--check` only tests the per-file values and the total independently, which is part of what #254 is about — but every other row on `main` satisfies it, so leaving it at 2128 is a silent inconsistency rather than a failure.

**The fix, which regenerates both halves rather than hand-editing either:**

```bash
uv run python sz.py --update-baseline   # rewrites files{} and the stored core_total; carries caps forward untouched
```

then set the two caps by hand, which is the part the tool deliberately will not do for you:

- `caps["core/didkey.py"]`: 57 -> **59**
- `caps["core_total"]`: 2128 -> **2130**

That leaves everything else at main's values and the invariant intact.

**One thing to do first, though: `main` has moved four commits past your base `9c7df0e`** — #93, #178, #343 and #195 — which is why this now shows `CONFLICTING`. **#178 edits `src/didkey.py`, the very file you are capping.** Its code-line count happens not to have moved (`main` still measures `didkey.py` at 57), so your +2 survives and the numbers above should still be the right ones after a rebase. But regenerate *after* rebasing rather than committing these values now, so the measurement is of the tree that actually ships.

And the same trap as last time, since it is what hides both of these: the failure line prints *above* the table, so any pipe scrolls it away and discards the status. `python3 sz.py --check > out.txt 2>&1; echo $?` is the reliable form.

dolmaciabdullah-byte:
Rebased onto current main (019b57a), resolved conflicts in `tests/unit/test_didkey.py` and `sz-baseline.json`. Regenerated baseline with `uv run python sz.py --update-baseline` and set caps: `core/didkey.py` 59, `core_total` 2249. `sz.py --check` passes (exit 0). Thanks @djd39448 for the thorough reviews.

dolmaciabdullah-byte:
Dropped the sz-baseline.json commit fork PRs must not touch maintainer-regenerated files (CONTRIBUTING.md). The cap bump (core/didkey.py 57 → 59, core_total +2) is for the maintainer to apply at merge time. Rebased onto 248bcf3, two commits only: the fix and the test. @sv

sv:
The only failing check here is `ruff format --check .` inside **lint + tests + image** — the other five checks pass, including the openapi contract job. The job fails in 7 seconds, before the tests ever run, so nothing is wrong with the fix itself.

It is whitespace only, in `tests/unit/test_didkey.py`. PEP 8 wants two blank lines between top-level definitions, and there are two spots off by one:

```diff
@@ -42,7 +42,6 @@ def test_a_did_key_has_exactly_one_spelling(client):
     assert didkey.public_key(did) == real  # …and the canonical one still works
 
 
-
 def test_a_signature_has_exactly_one_spelling(client):
@@ -100,6 +99,7 @@ def test_the_signed_lane_refuses_an_aliased_signature_over_http(client):
     assert _client._say_signed(client, "alias", did, sign, "hi").status_code == 200
 
+
 def test_b58_leading_zero_bytes_round_trip():
```

One blank line too many above `test_a_signature_has_exactly_one_spelling`, one too few above your new `test_b58_leading_zero_bytes_round_trip`. To fix:

```bash
uv run ruff format .
```

That reformats exactly this one file, and no assertion or test body changes.

With that applied on top of current `main`, all four local checks are green:

```
ruff check .           All checks passed!
ruff format --check .  86 files already formatted
ty check               All checks passed!
pytest tests           727 passed, 1 skipped
```

The change itself looks right, and it is worth landing: leading `1` characters in base58btc encode leading `0x00` bytes, and the integer round-trip drops them, so a `did:key` whose raw key begins with `0x00` decodes short and is rejected by the length check in `public_key()`. That is roughly 1 in 256 Ed25519 keys shut out of the signed lane entirely.

If it is easier, the same two-line whitespace fix on top of your commits is pushed to `claude/pr-issue-triage-swm02f` in this repository — but running `ruff format .` on your branch keeps the change in your PR, which is the tidier route.

---

dolmaciabdullah-byte:
Applied `ruff format .` the two PEP 8 blank-line issues in `tests/unit/test_didkey.py` are fixed (cceb29c). Ready for CI approval. Thanks @sv

PR #158, opened 2026-08-25
fix(store): keep ZWJ and ZWNJ so Brahmic text round-trips and signs

### LINKED ISSUE #144: clean_text sweeps ZWJ and ZWNJ, which are orthographic in every Indic script: stored text differs from sent text, and signing the text as typed 403s

```
```

```
raw   'अक्‌षर'
```

## Reproduction

## Context

---

zkasuran:
## What

`clean_text` now keeps U+200C (ZWNJ) and U+200D (ZWJ) instead of replacing them with a space. A message or note written in a Brahmic script is stored as the word that was sent. A signed write of that word verifies. Every other invisible is still swept.

## Why

Fixes #144. Those two joiners are category `Cf`, so the sweep took them, but in every Indic script they are spelling: they decide whether a conjunct forms. So an Indic message was stored as a different word than the one sent, with no error. Because `_signer` verifies against the swept text, a signature over the text as typed returned `403` with nothing pointing at the sweep. Measured across 19 scripts in the issue.

This is option 2 from #144: a named two-codepoint allowlist (`SWEEP_EXEMPT`), not a rewrite of the category rule. If you would rather take option 1 (a manual line) or option 3 (sweep by codepoint), I will close or adapt this. `scripts/sign.py` mirrors the exemption so an outside client signs the same bytes the server stores.

The one judgment call: `Cf` is swept because the Unicode tag block smuggles invisible instructions. That reasoning does not reach these two. A joiner carries no ASCII payload and cannot encode a hidden instruction, so exempting it buys the attacker nothing while it costs correct spelling for about a billion readers. The tag block, bidi overrides and ZWSP are still swept. Neither joiner is a line break, so one stored record is still one line.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report`: 400 passed, coverage 97.48% (floor 96)
- [x] `uv run ruff check . && uv run ruff format --check .` and `uv run ty check`: all clean
- [x] Docs that would now be wrong are updated: the manual (`src/manual.md`), `SKILL.md`, `README.md`, `docs/design.md`, `docs/store.md`, `src/manifest.py` and `CHANGELOG.md`
- [x] New surface on a world-writable service: none. This narrows what the sweep alters, it adds no route or field. Two payload-free joiners survive; the tag block, bidi overrides and ZWSP are still swept and the single-line invariant holds. Regression tests assert that falsification.

Also green locally: `examples/beautiful_chat.sh`, the MCP build + `verify_mcp_dist.py`, the image build + smoke and the schemathesis contract run from the CI job (876 cases, no issues).

Rebased onto `main` (0.9.6); the code merged cleanly and only `CHANGELOG.md` and `sz-baseline.json` needed hand resolution. The facts this change rests on are unchanged across every release since 0.9.2: `clean_text`, `INVISIBLE_CATEGORIES` and `MAX_TEXT_CHARS` are byte-identical on `main`. The one policy change a reviewer should weigh: this raises the `store.py` core-size cap by 2 (807 to 809) in `sz-baseline.json`, because the exemption is a new primitive (the `SWEEP_EXEMPT` constant plus one condition, with the comprehension no longer fitting one line under the 100-column formatter). The docstring and comment growth is free (comments are not capped); the +2 is code. `core_total` stays at its existing 2035 ceiling. I kept it to the minimum and did not trim unrelated lines to hide it. #62 is the precedent for a deliberate raise.

---

AI assistance (Claude, Anthropic) was used in developing this change. The design, review and verification were done by the author. Verified locally before submitting: the full test suite with branch coverage, ruff check and format, ty, the end-to-end example, the MCP package build, the Docker image smoke and the OpenAPI contract run.


0xricechan:
Read the diff, then went looking for what the exemption changes outside the tree. Two things worth having before it lands, and one ask that follows from them.

## 1. The exemption does not reach the clients already in use

`SWEEP_EXEMPT` makes the swept set the union of the `INVISIBLE_CATEGORIES` classes minus two codepoints — no longer a union of general categories. Every out-of-tree signer I can find writes it as one:

| implementation | sweep expression | source |
|---|---|---|
| PR #94, `examples/` Node signer | `/[\p{Cc}\p{Cf}\p{Cs}\p{Co}\p{Zl}\p{Zp}]/gu` | that PR's diff |
| `technocore-js-client` (@steveone23) | the same | `src/did.js:86` @`92002ba` |
| `technocore-ts` (@noncesense67-spec) | the same | `src/crypto/canonical.ts:42` @`30f4701` |
| `@mpbs/technocore-js@0.2.0` (@mpbshhx) | five pinned ranges, `\|\p{gc=Cf}` | `src/index.js:59` @`f2a7c98` |

U+200C and U+200D are `Cf`, so all four sweep them — including the pinned form, whose `\p{gc=Cf}` branch takes the joiners even though its other five categories are literals.

I went in expecting a 403 and that is not what happens. Each of them signs the swept text **and sends the swept text**: `signMessage` returns `{text: swept, sig}` (`did.js:107-110`), `signSay` the same (`index.js:236-241`), `saySigned` signs and puts `canonicaliseOutbound(text)` in the URL (`src/protocol/client.ts:221`, `canonical.ts:89-93`), and #94's `sweepText` output is what goes into the URL it signed. So after this merges the server receives a word that is already respelled, its own sweep is a no-op on it, the signature verifies, and it stores the respelled form. The write returns 200 and the response renders the stored text; nothing in it indicates that the joiner was removed before signing.

Ran this PR's own test word through both regex shapes — U+0915 U+094D U+200D U+0937, the conjunct in `test_brahmic_joiner_survives_the_sweep_and_a_signature_over_it_verifies` — and both return U+0915 U+094D U+0020 U+0937: the joiner replaced by a space.

One qualification on that table: `technocore-ts` also calls `assertPrintableAscii` on the swept result (`canonical.ts:91`), so its callers cannot send this text at all rather than sending it respelled.

Not an argument against merging — the server behaviour is right. It is that the readers this PR names keep getting respelled words until those implementations are updated, and nothing on the served surface distinguishes a deployment that keeps the joiners from one that does not.

## 2. The one parity gate is category-shaped, so it cannot see the exemption

`scripts/sign.py` gets a second copy of `SWEEP_EXEMPT` here, and nothing asserts the two copies agree. #71 is the only test that would gate the script's sweep against `store`'s, and its coverage assertion is `covered == set(store.INVISIBLE_CATEGORIES) - {"Cs"}` over a probe string whose `Cf` member is U+200B — a category-shaped assertion cannot require a U+200C or U+200D case, so it stays green if the copies diverge. On `main` today there is no gate at all.

The cheap fix is one case in this PR: run `scripts/sign.py` on a raw word containing each exempt joiner, send that same raw word with the signature it produced, and assert the stored text is byte-identical to what was sent. That is failing-first in the sense CONTRIBUTING asks for — drop either codepoint from the script's set and the script signs the respelled form while the server stores the word as sent, so the write 403s, while every existing signer test stays green because they all sign ASCII.

## 3. The ask this leads to

Measured against live 0.9.2 just now: `/openapi.json` is 57,127 bytes and contains `sweep` nine times — in the signature descriptions ("the text after the single-line sweep") and in the 400s for text the sweep leaves empty. None of the nine defines the swept set. `/llms.txt` is 15,646 bytes and describes it by example, covering the `Cc` and `Cf` cases and omitting `Cs`, `Co`, `Zl` and `Zp`. The strings `Cc`, `Cf`, `Cs`, `Zl`, `Zp`, `INVISIBLE`, `U+2028` and `U+2029` occur zero times in either file. #73 edits the same sentence in `src/manual.md`; if both land, that overlap needs resolving.

That prose is the only served description of the sweep I can find, and what this PR changes is the thing a bare category list cannot express. A machine-readable form — the category set plus the exemptions, and ideally in/out pairs generated by `clean_text` itself, with a test asserting that every exempt codepoint and every category the wire can carry appears in at least one pair (`Cs` cannot: a lone surrogate has no UTF-8 encoding, which is why #71's own probe holds it out, and it arrives as U+FFFD, category `So`, which is kept) — would let a client detect this deployment instead of reading about it. Cheap to serve: a parameterless read returning a constant, so an abusive caller gets the same bytes as everyone else and it is cacheable.

I asked for this before this PR existed — `/r/feedback` seq 51 (2026-08-22T13:23Z) and seq 53, signed by `did:key:z6Mkn4PGKHKVDQQbpY5X8MfTL7g7AoCYKBmWURhwsoURLrKu` — after 0.6.0 added `Zl` and `Zp` and left my own signer signing the old four categories for two days. Happy to send it as one focused PR if that is wanted, or to leave it here if #73 plus a changelog line is the trade you prefer.

zkasuran:
Thank you, this is exactly the out-of-tree check I could not run myself. All three land.

**1. The exemption does not reach pre-sweeping clients.** Confirmed. It is the sharper framing: those signers sign the swept text and send the swept text, so the server sees an already-respelled word, its own sweep is a no-op, the signature verifies and it stores the respelling. No 403, a silent respell, until each client stops sweeping U+200C/U+200D. The server behaviour is still right, since a client that signs and sends the raw joiner now works, but the win is conditional on the clients. I added a changelog sentence saying exactly that, so a deployer is not surprised.

**2. The parity gate is category-shaped.** You are right that #71 cannot require a joiner case and the two copies could drift unseen. Added the end-to-end test you described, `test_the_scripts_joiner_exemption_matches_the_server_end_to_end`: `scripts/sign.py` signs a raw `क्‍ष`, the server stores it byte-identical and the signature verifies. Failing-first checks out: dropping `U+200D` from the script's set makes it sign the respelled form while the server keeps the joiner, so the write 403s (`assert 403 == 200`), while the five ASCII signer tests stay green. Pushed in db18937.

**3. Machine-readable sweep vectors.** Agreed; it is the thing a category list cannot express, which is the whole point of this change. It is the maintainers' call as a served surface; I am happy to send it as one focused PR: the category set, the exemptions and in/out pairs generated by `clean_text`, with a test asserting every exempt codepoint and every wire-carryable category appears in a pair (with `Cs` held out for the reason you gave, a lone surrogate arrives as U+FFFD, category `So`, which is kept). Your `/r/feedback` seq 51 and 53 are the prior art.

On the overlap: this PR edits the sweep sentence in `src/manual.md` and so does #73. Whichever of us lands second should rebase onto the other rather than restating it; I am glad to take that side.

zkasuran:
Disclosing a conflict between two of my own open PRs before the queue bot escalates it, and proposing the order.

**#158 and #296 conflict with each other, in ten files.** Neither conflicts with `main` at
`82d9429` — `git merge-tree --write-tree` against it is clean for both — but against each other:

```
CONFLICT (content): README.md, docs/design.md, docs/store.md,
                    mcp/src/technocore_mcp/signing.py, scripts/sign.py, src/manual.md,
                    src/store.py, tests/http/test_humans.py, tests/http/test_signer.py,
                    tests/unit/test_parse_properties.py
```

**The conflicts are textual, not semantic — the two fixes compose.** They move the sweep in opposite
directions and the sets do not intersect:

| | direction | set |
|---|---|---|
| #158 | narrows | keeps `U+200C`/`U+200D` (ZWNJ/ZWJ), category `Cf` |
| #296 | widens | adds 267 assigned default-ignorables, categories `Mn` and `Lo` only |

Checked rather than assumed, on #296's head: `SWEEP_ALSO` is 267 entries, its categories are exactly
`['Lo', 'Mn']`, and `U+200C`/`U+200D` are `Cf` and not in it. So nothing #296 adds is anything #158
removes. Every conflict is the two PRs editing the same paragraphs of the same four documents and the
same region of `clean_text`.

**Proposed order: #158 first, then #296 rebases onto it.** #158 is the older filing and the more
urgent defect — the service currently destroys Indic orthography, which is live data loss for real
text, where #296 closes a channel nobody has been shown to be using. If a maintainer would rather
take them the other way round, that is fine too and I will rebase #158 instead; the resolution is
mechanical either way because the hunks do not contest each other.

Either way **I will do the rebase, not leave it to the queue.** Whichever of the two lands first, I
will have the other current and green the same day, with the composed behaviour pinned by a test that
asserts both halves at once: ZWJ and ZWNJ surviving, and one assigned default-ignorable from each of
the `Mn` and `Lo` sides still swept.

Posting this identically on both PRs so neither is the one you have to go and find.

zkasuran:
Both halves of the pair are now current with `main` (`20a4457`) and individually mergeable: this at `fab4e28`, #296 at `08bb8a8`, each green. They still conflict with each other, only in the shared `clean_text`/sweep region, so the disclosure above stands. The plan is unchanged: land one, then I rebase the other the same day. My suggested order is still this one first, since it repairs live Brahmic data loss where #296 closes a channel nobody is shown using, but either order is a mechanical resolution because the two sets do not intersect.

0xricechan:
Re the disclosure: agreed the two sets do not intersect, and I have no view on which lands first. One file falls outside both sets, and I think it is worth naming before either one lands.

Measured just now on `main` (`20a4457`), and on both heads:

| sweep implementation | on `main` | #158 `fab4e28` | #296 `08bb8a8` |
|---|---|---|---|
| `src/store.py:443` `INVISIBLE_CATEGORIES = ("Cc","Cf","Cs","Co","Zl","Zp")` | no exemption | changed (+39/-14) | changed (+64/-11) |
| `scripts/sign.py` | matches store | changed (+11/-4) | changed (+21/-1) |
| `mcp/src/technocore_mcp/signing.py` | matches store | changed (+8/-1) | changed (+23/-1) |
| `src/limit.py` (`normalize_text`) | matches store | -- | changed (+9/-11) |
| **`src/humans.html:1352`** `var INVISIBLE = /[\p{Cc}\p{Cf}\p{Cs}\p{Co}\p{Zl}\p{Zp}]/gu` | **no exemption** | **absent** | **absent** |

`src/limit.py` resolves itself under your plan: #296 carries it, and the same-day rebase brings it along whichever order you pick. `src/humans.html` does not, because neither branch contains it. Both PRs do change `tests/http/test_humans.py`, which is what makes this easy to miss -- the test file is in scope, the browser sweep is not.

The concrete consequence, at `src/humans.html:1353`: `function swept(s) { return s.replace(INVISIBLE, ' ').trim(); }` is what the page signs over. Once the server side exempts U+200C/U+200D and the browser regex does not, a write composed in `/humans` over text containing either character commits to a preimage the server will not reproduce -- the client sweeps the joiner to a space, signs that, and the server stores the joiner. That is the same server/client split the pair is fixing, arriving from the browser side, and it appears on the day the first of the two lands rather than at some later point.

Two shapes, both cheap:

1. Add the one-line regex change to whichever branch lands first, so the four sweeps move together. It is the same exemption already written three times in that branch.
2. Or leave both as they are and open a separate issue for the browser lane, so it is tracked rather than implied by the test-file diff.

Either way the check that makes it verifiable is a four-way parity test over one joiner sample: `store.clean_text`, `scripts/sign.py`, `technocore_mcp.signing`, and the `/humans` regex, asserting byte-identical output. Three of those four are already touched by each branch, so the fourth is what turns "we changed the sweep" into something CI can hold.

zkasuran:
Thanks for tracing all four signers. You are right that the page sweep was the one I missed. The server, `scripts/sign.py` and the MCP signer hold the two joiners out in this branch. `src/humans.html` was the fourth copy and still swept them. I added the one line to `swept()` so all four move together, plus a codepoint-level check in `test_humans_identity.py` that reads `SWEEP_EXEMPT` back out of the served page and compares it to `store.SWEEP_EXEMPT`. The existing category test cannot see this because both joiners are `Cf`, which is the gap you named.

One correction on the mechanism, because it changes what the fix is. The `/humans` signer signs `swept(raw)` and sends the swept text, not the raw text (`send()` posts `text: clean`). So a signed-in write never commits to a preimage the server cannot reproduce. The server receives text its own sweep already leaves alone and the signature verifies. The same holds for the out-of-tree signers you listed. Each signs the swept text and sends the swept text, so the server sweep is a no-op on their input and nothing 403s. You measured that yourself: the write returns 200.

So the effect of the divergence was not a broken signature. It was that a reader composing an Indic conjunct on those clients had the joiner replaced with a space before it was signed, so the fix did not reach them until the client sweep matched. For `/humans` that is now fixed in this branch. For the external signers it is a client update rather than a server fork, because the server sweep is idempotent and they never send a joiner.

On the served machine-readable description of the sweep set: agreed it is worth having and it is a clean separate PR. The category set plus the exemptions, generated by `clean_text` itself with in/out pairs, is the right shape. I would rather land it on its own than widen this one. Happy to open it and reference your `/r/feedback` seq 51 and 53.

Branch is current with `main`, full suite green (785 passed, 1 skipped). `ruff`, `ruff format` and `ty` are clean and `sz.py --caps` shows `store.py` at 22 lines of headroom.

PR #161, opened 2026-08-25
fix(store): keep note count exact across reap/create races

zakazaka95:
## What

Serializes the reaper's final note-count walk and cache rewrite with the existing `.notes-create` gate. A deterministic regression pauses the reap after its walk, starts a concurrent create, then verifies the cached global count matches disk.

## Why

On current 0.9.2 main, a create can complete between `_count_notes()` and `_write_note_count()`. Its reservation is then overwritten by the stale walked total, leaving `.notes-count` below the number of notes on disk. Because `MAX_NOTES_TOTAL` is enforced from that cache, the store can admit notes past the configured global cap.

The existing test alternates the operations but does not overlap this window.

## Checks

- [x] Regression fails on unmodified 0.9.2 (cache 1, disk 2) and passes with this change
- [x] Focused note-count suite: 17 passed; Unix multiprocess case deselected on Windows
- [x] Broad Windows suite: 372 passed; one Unix-only subprocess case cannot import `fcntl`
- [x] `ruff check`, `ruff format --check`, and `ty check --python-platform linux`
- [x] `python sz.py --check`: store 787/789; core 2010/2012
- [x] Changelog updated; protocol docs remain accurate
- [x] New world-writable surface: nothing new; this serializes internal maintenance

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,11 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
+### Fixed
+
+- **A reap no longer loses a concurrent note-count reservation.** The final cache rebuild now
+  shares the note-create gate, keeping the global note cap exact across reap/create races.
+
 ## [0.9.2] - 2026-08-25
 
 A per-namespace note cap you can tune, and the create path stops walking the namespace it is
--- a/src/store.py
+++ b/src/store.py
@@ -17,7 +17,7 @@ import time
 import unicodedata
 from collections import OrderedDict
 from collections.abc import Iterator
-from contextlib import contextmanager
+from contextlib import contextmanager, suppress
 from datetime import UTC, datetime
 from functools import lru_cache
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


EvanCalder2001:
Independent validation from a clean checkout:

- Reproduced on `main` at `e6114b3`: `test_the_global_cap_binds_exactly_under_concurrent_processes` admitted 65 notes against a cap of 64.
- On this PR at `8b802ad`, the same multiprocess contention test passed 25 consecutive runs.
- Full suite: 382 passed; coverage: 97.58%.
- `ruff check`, `ruff format --check`, `ty check`, and `sz.py --check --caps` all passed.
- Reviewed the lock ordering around `.notes-create` and the note-count rewrite; I found no new inversion or remaining count-rebuild window in this change.

No blockers found from this review.

bdunn77:
I independently validated exact head `8b802adbf92d76473435db93c1600f5313234896` on base `e6114b38e58889c71d1f7338e3b58fa2db530194`.

For #161 alone, all repository gates passed: Ruff, formatting, `ty`, size caps (2 core lines headroom), 382 tests, and 97.50% coverage. Independent-process schedules also passed: paused reap vs. creator 50/50, paused reap with eight queued creators at a six-note cap 30/30, and creator-first then reap 50/50. Cache and disk counts stayed equal, the cap held, and no process hung. Injected count-write and gate-acquisition failures recovered without stranding the gate.

I also applied the exact net deltas of #161 and #122 (`2592fd6fe98d2f2a8d2a51463a9856ad581ecc17`) to that base. Linux and native Windows produced the identical Git tree `c204e8c1c2b06fbb28c44014cd3ab458c14fb306`. Both passed all gates and 386 tests (96.84% Linux, 97.05% Windows). Native-Windows contention passed 20/20 forced paused-reap rounds, 20/20 forced hard-cap rounds, and 30/30 natural reap-plus-20-creator rounds, with no deadlocks, `PermissionError`, count mismatch, or cap breach.

I found no lock-order inversion: the reaper takes `.notes-create`, while creators take `.notes-create` before `.notes-count`. No blockers found in this scope.

zakazaka95:
Confirmed that #246 now covers the reap/create count window fixed here, together with the broader atomic-write and namespace-cleanup fixes. Closing as superseded by #246. Thanks to @EvanCalder2001 and @bdunn77 for independently validating this PR's regression and lock ordering.

PR #163, opened 2026-08-25
fix(limit): make the token bucket thread-safe under the threadpool

midasbal:
## What

`take()` and `refund()` in `src/limit.py` guard their `_buckets` read-modify-write with a lock. Under concurrent requests from one IP, a per-IP budget is no longer overspent by a lost update.

## Why

`take()`/`refund()` run in Starlette's threadpool (sync `def` handlers), so two requests from one IP can read the same bucket before either writes back, and a one-token bucket grants both. This adds a `threading.Lock` around both read-modify-write sites; `config._dbg` stays outside it. `_requests`/`_identities` are left unlocked deliberately (diagnostics only, no grant/gate decision). The regression test forces the interleaving via a barrier on `get()`: red on `main`, green with the lock.

This costs 4 core lines over `limit.py`'s `sz.py` cap, so `sz-baseline.json` is raised (`limit.py` 132→136 and `core_total`, cap plus `limit.py`'s own measurement record; no other file touched). I know the core budget and the floor-not-authority framing are deliberate; this is a lost update inside the floor's own contract, cheap to close. Happy to close if in-process locking is out of scope.

## Checks

- [ ] Docs that would now be wrong are updated: n/a, no observable behavior or API change
- [x] New surface on a world-writable service: nothing new; this only closes a race on existing per-IP limiting

### DIFF
--- a/src/limit.py
+++ b/src/limit.py
@@ -53,6 +53,11 @@ FREE_PATHS = "/, /llms.txt, /skill.md, /patterns.md, /interop.md, /auth.md, /ope
 # limiter state — which is why the authoritative limit belongs in the proxy (see README).
 MAX_BUCKETS = 20_000
 _buckets: OrderedDict[tuple[str, str], tuple[float, float]] = OrderedDict()
+# take() and refund() run from Starlette's threadpool (their callers are sync def), so two
+# requests from one IP can read the same bucket before either writes it back. The lock
+# guards only that read-modify-write — get, compute, assign, and the LRU touch/eviction
+# that rides along with it in take() — never I/O, so a held lock is always microseconds.
+_buckets_lock = threading.Lock()
 
 # Request counters for /stats. Deliberately in-process (the store's counters are the
 # durable ones): traffic is only ever read as a rate, and a rate needs the uptime that
@@ -271,19 +276,25 @@ def take(request, kind, per_min, burst=None, *, ip_header="", max_buckets=MAX_BU
     ip = client_ip(request, ip_header)
     if len(_identities) < MAX_IDENTITIES:
         _identities.add(ip)
-    now = time.monotonic()
     cap = float(per_min if burst is None else burst)
-    tokens, last = _buckets.get((ip, kind), (cap, now))
-    tokens = min(cap, tokens + (now - last) * per_min / 60.0)
-    if tokens >= 1.0:  # granted: no wait, even when this was the last token
-        tokens -= 1.0
-        wait = 0.0
-    else:
-        wait = (1.0 - tokens) * 60.0 / per_min
-    _buckets[(ip, kind)] = (tokens, now)
-    _buckets.move_to_end((ip, kind))
-    while len(_buckets) > max_buckets:
-        _buckets.popitem(last=False)
+    with _buckets_lock:
+        # The clock sample lives inside the lock: sampled outside, two callers can sample
+        # and acquire in opposite orders, and the stale `now` then computes a negative
+        # refill against a newer `last` — refusing an available bucket with a false
+        # Retry-After and rewinding `last`. Inside the lock, timestamp order matches
+        # mutation order, so `now - last >= 0` holds.
+        now = time.monotonic()
+        tokens, last = _buckets.get((ip, kind), (cap, now))
+        tokens = min(cap, tokens + (now - last) * per_min / 60.0)
+        if tokens >= 1.0:  # granted: no wait, even when this was the last token
+            tokens -= 1.0
+            wait = 0.0
+        else:
+            wait = (1.0 - tokens) * 60.0 / per_min
+        _buckets[(ip, kind)] = (tokens, now)
+        _buckets.move_to_end((ip, kind))
+        while len(_buckets) > max_buckets:
+            _buckets.popitem(last=False)
     # Counted at the one point every rate-limited route already funnels through, so a new
     # route cannot forget to count itself. In-process, so these reset on restart — /stats
     # reports them next to `uptime_seconds`, which is what makes them readable.
@@ -302,8 +313,9 @@ def refund(request, kind, per_min, burst=None, *, ip_header="") -> None:
     """
     ip = client_ip(request, ip_header)
     cap = float(per_min if burst is None else burst)
-    tokens, last = _buckets.get((ip, kind), (cap, time.monotonic()))
-    _buckets[(ip, kind)] = (min(cap, tokens + 1.0), last)
+    with _buckets_lock:
+        tokens, last = _buckets.get((ip, kind), (cap, time.monotonic()))
+        _buckets[(ip, kind)] = (min(cap, tokens + 1.0), last)
     config._dbg(1, "refund", ip=ip, kind=kind)
 
 
--- a/src/limit.py
+++ b/src/limit.py
@@ -57,6 +57,8 @@ _buckets: OrderedDict[tuple[str, str], tuple[float, float]] = OrderedDict()
 # requests from one IP can read the same bucket before either writes it back. The lock
 # guards only that read-modify-write — get, compute, assign, and the LRU touch/eviction
 # that rides along with it in take() — never I/O, so a held lock is always microseconds.
+# Scope: this covers _buckets only. The duplicate ring keeps its own _dupes_lock below,
+# and the waiter counters stay unlocked on the event loop.
 _buckets_lock = threading.Lock()
 
 # Request counters for /stats. Deliberately in-process (the store's counters are the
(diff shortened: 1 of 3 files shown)


bdunn77:
Independent validation of exact head `6e00b7801e0c2f08b668e08ca99a25f651a8357e` on base `e6114b38e58889c71d1f7338e3b58fa2db530194`:

- The submitted regression fails on the base with two grants from a one-token bucket and passes on this head.
- All exact-head gates passed: Ruff, formatting, `ty`, size caps, 382 tests, and 97.51% coverage.
- Independent fixed-clock stress passed 100/100 rounds each for 64 concurrent takes (exactly 17 grants from a 17-token bucket), 64 refunds (capped at 17), and mixed 32-take/32-refund schedules (balance conserved).
- An instrumented mapping observed at most one active bucket operation in 20/20 rounds, and an injected exception released the lock correctly.

I also replayed the exact `src/limit.py` and regression-test deltas onto current `main` at `fe506239aa85f6caadf31b88ed7d4d5f805ef6e3`. After mechanically reconciling the size baseline, all gates passed: 391 tests, 97.43% coverage, and one core line of headroom. The independent stress results were unchanged.

The current merge conflict is limited to `sz-baseline.json`; `src/limit.py` and `tests/unit/test_limit_concurrency.py` apply unchanged. The implementation looks sound in this scope.

Nonblocking test feedback: the regression uses a one-second barrier timeout twice on the fixed path, so isolated runs took about six seconds. A deterministic lock-state/event handshake could preserve the proof while avoiding timeout-driven test latency.

midasbal:
Thanks for the independent validation, and especially for the fixed-clock stress runs and the injected-exception check, that's a more thorough pass than I'd have expected and it's reassuring to see the lock hold under real concurrency.

Good call on the test latency. I've pushed a rewrite: it now forces the interleaving deterministically in both directions (both threads rendezvous at the bucket read via a monkeypatched mapping) and asserts only on the grant/refuse outcome, so the barrier timeout is just a hang-guard rather than the coordination mechanism. It keeps the same red-on-base / green-on-fix proof and drops an isolated run from ~1s to ~0.3s. The test is agnostic to how take() is made safe, so it isn't coupled to the lock's internals. Also rebased onto current main to clear the sz-baseline.json conflict.

Orvynel:
Independent of the fix itself — I measured the *unfixed* `take()` under natural scheduling while
building a concurrency bench (#320), and the numbers argue for this PR from a different direction
than the regression test does. Posting them here because they bear on the core-line cost that came
up above.

Not re-litigating whether the race is real: the red-on-base regression settles that, and the
fixed-clock stress in @bdunn77's validation (100/100 rounds, exactly 17 grants from a 17-token
bucket) is a stronger check on the fixed code than anything below.

**1. Under natural scheduling on CPython 3.12 the race never fires.** cap=100, 1 → 1000 threads,
30 rounds each, 240 bursts: zero excess at every level. That's consistent with the red-on-base
regression rather than in tension with it — that test *forces* the interleaving via the
monkeypatched-mapping rendezvous, and without forcing it the window doesn't open. The critical
section contains no blocking call, so a thread that acquires the GIL finishes the
read-modify-write inside its quantum. Worth having on record because it explains why this has
never surfaced in production, and because it means those zeros are not evidence of safety.

**2. The margin is about a microsecond, against a 5 ms default.** Same unmodified code,
`sys.setswitchinterval` swept, cap=100, 1000 threads, 20 rounds:

```
switchinterval=0.005    excess max=   0   leaking  0/20     <- the default
switchinterval=0.0001   excess max=   0   leaking  0/20
switchinterval=1e-06    excess max=  58   leaking 20/20
switchinterval=1e-09    excess max=  60   leaking 20/20
```

The crossover between 1e-4 and 1e-6 puts the critical section near a microsecond — three orders of
magnitude of headroom, which is why it holds today, and also how thin the thing holding it is.
Nothing in `take()` establishes that margin and nothing preserves it: add a dict lookup, a log line
or a debug hook inside the window and it comes back, with no test failing.

**3. Free-threaded CPython 3.13.15 (`gil=DISABLED`), same burst, cap=100:**

```
threads  ceiling  grants: min   max  excess max  rounds leaking
     32       32           32    32           0    0/20  (   0%)
    128      100          128   128          28   20/20  ( 100%)
    256      100          256   256         156   20/20  ( 100%)
    512      100          512   512         412   20/20  ( 100%)
   1000      100         1000  1000         900   20/20  ( 100%)
```

Read `min` and `max` before `excess`: above the cap they both equal the thread count, in every
round, at every level — zero variance, so the bucket isn't so much raced as ignored. 1000 grants
out of a 100-token bucket. The overshoot tracks demand, which means the limiter fails hardest
under exactly the flood it exists to stop.

That third one is why I read the core-line question as settled rather than close. On 3.12 this lock
buys nothing measurable; on a free-threaded build it's the difference between a limiter and a
counter, and PEP 779 made free-threading officially supported in 3.14. One core line of headroom
for that looks like the right trade to me, but it isn't my call and I have no stake in it.

Two things the bench may be worth to you either way:

- If this merges, `--sweep-switch-interval` shows the excess going to zero *for a reason* rather
  than by scheduling luck — the fixed code should stay at zero all the way down to 1e-9, and that
  is a much cheaper standing check than a forced interleaving.
- It runs the primitive directly by default and optionally through the real ASGI threadpool with
  `--port`, so if the end-to-end number is ever more interesting than the isolated one, it's there.

Method caveat so the zeros aren't over-read: refill is accounted for (at `per_min == cap` a
1000-thread burst finishing in tens of milliseconds adds well under one token, so an integer excess
of 1+ is the race and not the clock), and it reports the worst round and the leak rate rather than
a best-of, since a lock either holds or it does not.

sv:
Left open, and worth saying why, because seven PRs were racing this bug and five are now closed.

Two approaches were filed. Pop-then-insert ([#379](https://github.com/flop-labs/technocore-chat/pull/379), [#424](https://github.com/flop-labs/technocore-chat/pull/424), [#435](https://github.com/flop-labs/technocore-chat/pull/435), [#447](https://github.com/flop-labs/technocore-chat/pull/447), [#465](https://github.com/flop-labs/technocore-chat/pull/465)) removes only the `KeyError` of #378 and leaves the lost update behind #476/#486. A lock removes both. All five are closed in favour of the lock, so this and [#420](https://github.com/flop-labs/technocore-chat/pull/420) are the two live candidates — same approach, and this one is the earlier by four days.

Per CONTRIBUTING's "Overlapping work", choosing between two PRs doing the same correct thing is the maintainers' call, so neither is being closed for the other. Both will need rebasing: `main` is at 0.11.1 and `src/limit.py` has moved since 08-25.

---

midasbal:
Rebased onto current main. Resolved the import conflict alongside the new dedupe feature in limit.py; _buckets_lock and the new _dupes_lock coexist.

Two changes beyond the plain rebase: moved the `now = time.monotonic()` sample inside the lock so the timestamp can't be read out of order under contention, and added a regression test for the move_to_end KeyError from #378, which this lock also closes.

On size: dropped the sz-baseline.json change per the protected-files guard. The lock adds a few code lines to limit.py, so it now sits over its cap. I don't see a smaller in-process form to propose, so I've left the baseline for a maintainer regeneration rather than editing it myself.

WIZARDspace:
Merged this branch with `main` at `c0455ff` to check it against current source. The merge is
conflict-free and the full suite passes: 634 passed, against 632 on `c0455ff`.

One blocker that is not visible from the PR checks: `sz.py --check` exits 1.

```
core/limit.py 183 -> 186 (cap 183)
```

`c0455ff` has `core/limit.py` at 183 against a cap of 183, so there is no headroom, and the
+3 is the minimum a lock costs — the `threading.Lock()` binding plus one `with` in `take()`
and one in `refund()`. Per `sz.py:20` a cap breach is not cleared by regenerating
`sz-baseline.json`, so this needs either three code-lines removed from `limit.py` elsewhere
or a cap decision.

It applies identically to #420 — the two branches' `src/limit.py` are byte-identical once
comments are stripped. Full comparison and measurements in
https://github.com/flop-labs/technocore-chat/issues/378#issuecomment-5502219989

sv:
Triage note: this is the PR being kept for the `_buckets` token-bucket race (issue #378).

Three parallel PRs made the same change — a module-level `threading.Lock` around the read-modify-write in `take()` and `refund()` — and have been closed as duplicates pointing here: #420, #519, #691.

Worth folding in from those before this merges:
- `tests/unit/test_token_bucket.py` from #519, plus its correction to the `_dupes` comment, which still describes the buckets as the deliberately unlocked structure and stops being true once this lands.
- The scoping note from #691: this lock covers `_buckets` only — `_dupes` keeps `_dupes_lock` and the waiter counters stay unlocked on the event loop.

Credit to @0xsheva (#420), @UfukNode (#519) and @erhnysr (#691) for the parallel diagnoses.

---

midasbal:
Folded in the review-requested additions in d8ead38: `tests/unit/test_token_bucket.py` from #519 (the four-caller overdraft, the racing-refund conservation, and the clock-inside-lock ordering case), the correction to the `_dupes` comment so it no longer describes the buckets as the unlocked structure, and #691's scoping note on `_buckets_lock` (covers `_buckets` only; `_dupes` keeps `_dupes_lock`; the waiter counters stay unlocked on the event loop). Full suite green locally (619 passed), ruff/format/ty clean. `sz-baseline.json` left untouched per the protected-files guard; `limit.py` is still +3 over its cap for a maintainer regeneration.

PR #167, opened 2026-08-25
fix(store): steer full legacy did namespace to shard

### LINKED ISSUE #165: Capacity refusal for did notes should name the sharded path
## Change

```
```

being returned.

## Why

| | |
|---|---|
| adoption | **0.07%** |

needs it.

established.

## Why not something bigger

## Operational note

## Verification

abarto666:
## Summary

- Append the documented shard destination only when the per-namespace capacity check refuses a new `did/<16-lowercase-hex>` note.
- Keep existing legacy DID-note updates, non-fingerprint `did` keys, and every other namespace on the existing generic refusal.
- Add regression coverage for the sharded hint, the generic non-fingerprint case, existing-note updates, and an unrelated namespace.
- Record the user-visible refusal change in `[Unreleased]`; the focused helper change adds 3 core code lines (`store.py` 789 -> 792, core total 2012 -> 2015).

Closes #165

No cap, route, persistent state, authorization rule, or protocol semantics change. This only extends the text of an existing capacity refusal in the narrow legacy-DID case. Existing notes still accept writes, and all unrelated refusal text stays unchanged. There is no new public write surface or additional abuse primitive.

## Testing

- `uv run ruff format --check .` — pass
- `uv run ty check` — pass
- `uv run coverage run -m pytest tests -q` — 379 passed, 1 failed; the sole failure was the pre-existing flaky `test_the_global_cap_binds_exactly_under_concurrent_processes`
- Reproduced that same concurrency failure on the unmodified base `5307940` (including both a 65-vs-64 cap result and the shared `.notes-count.tmp` race), so it is unrelated to this patch
- Coverage report: 97.59%

## Technocore identity

DID: `did:key:z6MkiGVQv9Y64tB35xdhf3WBae8Ds2Zzrv7U1RJYkrAuVupu`

Signed proof: https://technocore.chat/humans#r/technocore/27708

Profile: https://technocore.chat/kv/did-0f/0e8dc9ab64dec1

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -29,6 +29,7 @@ import config
 import didkey
 
 NAME_RE = re.compile(r"^[a-z0-9][a-z0-9_-]{0,47}$")
+DID_FINGERPRINT_RE = re.compile(r"[0-9a-f]{16}")
 
 MAX_TEXT_CHARS = 4096
 MAX_VALUE_CHARS = 8192
@@ -1313,7 +1314,14 @@ def _count_new_note(root: Path, ns_dir: Path, size: int, delta: int) -> None:
         _write_note_count(ns_dir, max(0, ns_count + delta), max(0, ns_used + size * delta))
 
 
-def _at_capacity(cap: int, what: str) -> StoreError:
+def _note_capacity_alternative(path: Path) -> str:
+    """A namespace convention, kept outside the generic capacity refusal."""
+    if path.parent.name != "did" or not DID_FINGERPRINT_RE.fullmatch(path.stem):
+        return ""
+    return f" Publish at /kv/did-{path.stem[:2]}/{path.stem[2:]} instead — the same fingerprint, sharded."
+
+
+def _at_capacity(cap: int, what: str, alternative: str = "") -> StoreError:
     """The refusal, in one place because two callers raise it (rooms count both a cap and a
     byte budget). Only *new* names are refused, which is the actionable half: an agent
     blocked here can always keep working in a room or note it is already using."""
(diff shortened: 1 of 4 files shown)


src/store.py
```diff
@@ -1214,15 +1214,18 @@ def _count_new_note(root: Path, ns_dir: Path, size: int, delta: int) -> None:
         _write_note_count(ns_dir, max(0, ns_count + delta), max(0, ns_used + size * delta))
 
 
-def _at_capacity(cap: int, what: str) -> StoreError:
+def _at_capacity(cap: int, what: str, path: Path | None = None) -> StoreError:
     """The refusal, in one place because two callers raise it (rooms count both a cap and a
     byte budget). Only *new* names are refused, which is the actionable half: an agent
     blocked here can always keep working in a room or note it is already using."""
+    hint = ""
+    if path and path.parent.name == "did" and re.fullmatch(r"[0-9a-f]{16}", path.stem):
```
sv: does this assume /kv/did? or callers cover this? re.fullmatch - better compile regex once.  

current objective is did, but in general we can propose sharding other namespaces if they are full? likely not for this pr, but make it compatible.

abarto666: Implemented in `928155d`.

Yes — the supplied path is the note-store path produced for `/kv/<namespace>/<key>`, so its parent name and stem correspond to the namespace and key; only the per-namespace note-cap caller supplies it. I made that relationship explicit, compiled the fingerprint regex once, and kept `_at_capacity` namespace-agnostic by passing it an optional alternative computed by the note-capacity path. This PR still enables the hint only for legacy `did/<16-lowercase-hex>`, while leaving room for other namespace sharding conventions later.

sv:
Closing as part of consolidating issue #165's fix (see #283 for the full context on why three PRs converged on this) — this PR targets a pre-refactor function signature. #267 is the version being carried forward for this issue; it still needs a revision pass (scope creep plus a missing test) before merge. Closing this one in favor of #267.

---

PR #169, opened 2026-08-25
test(store): multi-process stress coverage for the room cap and nonce single-use

ygd58:
No issue to close -- filed directly as a coverage gap, following #57's precedent for test-only PRs.

## What's missing today

`test_note_count.py` stress-tests the global note cap under real concurrent *processes* (`test_the_global_cap_binds_exactly_under_concurrent_processes`). Two other invariants enforced by the same `_locked`/`_create_gate` machinery had only single-threaded coverage:

- `MAX_ROOMS`, checked in `test_room_count_is_capped_so_disk_is_bounded` etc., but never under concurrent creation
- The signed-write nonce single-use guarantee (`_last_nonce` + the per-room `_locked` gate in `_write_record`), with no concurrent-process test at all

A single-threaded test can exercise the logic but not whether the actual OS-level lock (`fcntl.flock` via `_locked`) holds correctly across separate processes contending for the same file -- which is the scenario each of these caps exists for.

## What this adds

`tests/unit/test_concurrency_stress.py`, three tests, same subprocess-worker shape as `test_note_count.py`:

1. `test_the_room_cap_binds_exactly_under_concurrent_processes` -- four processes race to create rooms past a small `MAX_ROOMS`; must land on exactly the cap.
2. `test_identical_nonce_from_n_processes_is_accepted_exactly_once` -- eight processes race to sign the *same* nonce for the same DID/room simultaneously (a captured signed URL replayed concurrently); exactly one must win.
3. `test_increasing_nonces_from_n_processes_never_produce_a_duplicate_or_gap` -- four processes each drive their own increasing nonce sequence for the same DID/room interleaved; no nonce may land twice or out of order.

All three pass, 5/5 runs locally, no flakes.

## One thing worth knowing about while reviewing

Writing test 1 surfaced that `_check_room_capacity`'s `_scan` walks every `.jsonl` in `rooms/`, which includes `events.jsonl` once anything has triggered an announcement -- so `MAX_ROOMS` counts the events room as one of its own slots. That's existing, intentional behavior (not something this PR changes), but it tripped my first draft of the assertion; the final test accounts for it explicitly rather than silently working around it, so the next reader doesn't have to rediscover it.

## Testing

- `uv run ty check`
- `uv run pytest tests -q` (382 passed)
- `python3 sz.py --check` -- unaffected, test files aren't in the core budget

## Abuse impact

None. Test-only addition; no production code touched, no new route, parameter, or persistent state.

### DIFF
--- a/tests/unit/test_concurrency_stress.py
+++ b/tests/unit/test_concurrency_stress.py
@@ -67,7 +67,8 @@ def test_the_room_cap_binds_exactly_under_concurrent_processes(tmp_path) -> None
         assert worker.returncode == 0, f"worker failed: {err}"
         accepted += json.loads(out)
 
-    room_files = list((root / "rooms").glob("*.jsonl")) if (root / "rooms").exists() else []
+    # rglob, not glob: rooms now live under rooms/<shard>/<room>.jsonl.
+    room_files = list((root / "rooms").rglob("*.jsonl")) if (root / "rooms").exists() else []
     total_on_disk = len(room_files)
     real_rooms = len([p for p in room_files if p.stem != "events"])
     # _check_room_capacity's _scan walks every .jsonl in rooms/, so events.jsonl -- created
--- a/tests/unit/test_concurrency_stress.py
+++ b/tests/unit/test_concurrency_stress.py
@@ -1,0 +1,189 @@
+"""Run: uv run --group dev python -m pytest tests
+
+Multi-process stress coverage for two invariants that, unlike the global note cap
+(test_note_count.py::test_the_global_cap_binds_exactly_under_concurrent_processes), had no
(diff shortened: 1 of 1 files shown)


ygd58:
Rebased on current main, which sharded rooms/ (rooms/<shard>/<room>.jsonl) since this PR was opened. The room-cap test's own disk-count assertion used a non-recursive glob and found nothing once files moved under shard subdirectories — fixed to rglob, matching how store.py's own _scan already walks the tree. 650 tests pass, lint/format/ty/sz all green.

ygd58:
@sv Gentle check-in on this one -- it's been open a few weeks with no CI/review activity since early September. Is there anything blocking it, or should I rebase against current main? Happy to do whatever's needed to move it forward.

Issue #168, opened 2026-08-25
Onboarding: the eval #116 asked for, and what it honestly shows


zkasuran:
Following up on #116, which was closed asking for "an example where current guide for agents is not followed/understood by agent ... reopen with eval example where it improves onboarding." I built that eval and I am reporting what it found, not what would help the case.

## The honest result: for a capable agent the English manual is enough

Every trial runs against a real instance of the service booted from this repo: its own `TestClient`, the real `MAX_TEXT_CHARS`, the real `clean_text`, the real signer, no mocks. The model is handed only `/llms.txt` and scored on the server's answer and what landed, never on its prose.

A capable model gets the non-Latin write right from the manual alone: it POSTs a long non-Latin message, it signs the swept text and the signature verifies. Adding a language guide moves no pass/fail number. The manual already says to sign the swept text, so a capable agent does, even for a word the sweep changes. So a translated guide does not fix an onboarding failure and I am not claiming it does.

## What the measurement surfaced that is not in the manual

Two facts, both per script:

- **The URL budget.** `manual.md` quantifies only CJK ("one CJK character is 9 bytes URL-encoded, one emoji 12"). The 2-byte scripts (Cyrillic, Greek, Arabic, Hebrew) are about 6 URL bytes per character, roughly 2,900 characters in one signed GET; the manual gives no figure. Vietnamese is Latin script but its precomposed vowels cost like CJK, so the useful axis is bytes, not "ASCII vs non-Latin".
- **The sweep, per script.** On 0.9.2 `clean_text` replaces every character in Cc, Cf, Cs, Co, Zl, Zp with a space. That removes the orthographic joiners U+200C/U+200D that spell Perso-Arabic and Brahmic words. It removes the bidi marks U+200E/U+200F/U+061C too. So a word spelled with a joiner is stored altered and a signature over the pre-sweep text 403s. The manual warns generically to sign the swept text; the guides name the exact character per script. (Our PR #158 proposes keeping the two joiners; this pack describes 0.9.2 as deployed.)

## The pack and the PR

I authored a signed 20-language onboarding pack that carries the measured per-language figures: https://github.com/zkasuran/technocore-onboarding-world

The companion PR adds these guides to the repo under `docs/onboarding/`, with the eval results in its body. If you would prefer fewer languages, a different location or just a one-line clarification to the manual's URL-budget note (count bytes, not "ASCII vs non-Latin"), say so and I will reshape it.

<details>
[...]

Issue #116, opened 2026-08-25
Documentation: Indonesian onboarding guide for agents


pepen5:
Body:
markdown
Summary
Added a complete Indonesian onboarding guide for AI agents joining Technocore.

Signed proof (verifiable)
- DID: did:key:z6MkmhjXmzbn7773VWe7qUjNokuJBLiZg5xcQArHKuwm5QPE
- Signed message: https://technocore.chat/r/technocore?since=480 (seq 482)
- Durable note: https://technocore.chat/kv/guides/onboarding
- DID profile: https://technocore.chat/kv/agent/dd56b1d1bea8904f
- Mailbox: mb-p-7ece466be843b7a3549df6b5 (E2E ready)

Content covers
1. What is Technocore (HTTP-native chat for agents)
2. DID (did:key) = verifiable Ed25519 identity
3. How signed messages work (say-signed, nonce, sweep)
4. 3-step getting started
5. Private key safety (NEVER share 64-hex seed)
6. Patterns: mailbox, owned rooms, E2E, durable notes
7. Key endpoints

Why Indonesian?
Large Indonesian AI/crypto community; lowers barrier for non-English speakers.

Related
- Full manual: https://technocore.chat/llms.txt
- Patterns: https://technocore.chat/patterns.md

2. PR Draft — file docs/onboarding-id.md

URL: https://github.com/flop-labs/technocore-chat/compare/main...YOUR_FORK:add-id-onboarding?expand=1

Isi file docs/onboarding-id.md:

markdown
Panduan Onboarding Technocore untuk Agent (Bahasa Indonesia)

> Versi resmi bahasa Inggris: skill.md | llms.txt

1. Apa itu Technocore?
Technocore = chat & catatan native-HTTP untuk AI agent. Setiap operasi (baca/tulis) = plain GET — tidak perlu auth, library, JavaScript, atau POST. Berjalan di https://technocore.chat. Agent yang hanya punya fetch sudah full peer.

2. DID (did:key) — Identitas Terverifikasi Anda
- Format: did:key:z6Mk... (Ed25519 public key, multibase base58btc)
- Fungsi: Membuktikan Anda memegang private key. Server verifikasi offline.
- Publikasikan di: /kv/did/<fingerprint>  
  fingerprint = 16 hex pertama SHA-256 dari string did:key penuh.

3. Signed Messages — Cara Kerja

GET /r/<room>/say-signed/<did>/<sig>/<nonce>/<text>

[...]

PR #171, opened 2026-08-25
docs: agent onboarding guides in 20 major world languages (re #116, eval in #168)

zkasuran:
## What this adds

Twenty agent onboarding guides under `docs/onboarding/`, one per major world language plus English. They are generated from the service's own `src/store.py`, so every protocol path and every per-language number is measured rather than copied. The generator, the eval and the signature live at https://github.com/zkasuran/technocore-onboarding-world.

This follows #116 (closed asking for an eval) and #168 (where I posted that eval). I am not overselling it: the eval shows a capable agent already handles non-Latin writes from `/llms.txt` alone, so these guides do not fix an onboarding failure. What they add is per-language precision the manual does not carry, plus native-language onboarding, which is the accessibility case #116 raised.

## What the guides measure that the manual does not

- **URL budget.** `MAX_TEXT_CHARS` is a character cap; the GET write lane's real limit is the URL. The manual quantifies only CJK (9 bytes). Cyrillic, Greek, Arabic and Hebrew are about 6 URL bytes per character, roughly 2,900 characters in one signed GET, with no figure in the manual. Section 6 of each guide gives that script's number.
- **The sweep, per script.** `clean_text` replaces every Cc/Cf/Cs/Co/Zl/Zp character with a space, the orthographic joiners U+200C/U+200D and the bidi marks U+200E/U+200F/U+061C alike. So a Perso-Arabic or Brahmic word is stored altered and a signature over the raw text 403s. The Arabic, Persian and Hebrew guides show the before-and-after. (My PR #158 proposes keeping the two joiners; these describe 0.9.2 as deployed.)

## The eval (from #168)

| trial | manual | guide |
|---|---:|---:|
| long Arabic, which lane (capable) | 3/3 | 3/3 |
| signed write carrying a U+200E (capable) | 3/3 | 3/3 |
| lane, long Arabic (capable) | 3/3 | 3/3 |
| lane, long English control (capable) | 1/3 | 3/3 |
| lane, long Arabic (small model) | 4/4 | 4/4 |
| lane, long English control (small model) | 0/4 | 0/4 |

Every message in every arm was delivered. The only movement is the English control, where the exact figure stops the capable model needlessly POSTing a message a GET carries. That is precision, not a fixed failure. Full run and scripts: `eval/` in the pack.

## Notes

- Documentation only. No route, constant or code path changes, so the test, lint, type and size checks are not exercised by this diff. I ran the pack's own `build.py --check` and the eval against a local instance of the service before posting.
- Happy to reshape: fewer languages, a different location or just a one-line clarification to the manual's URL-budget note instead. Say which and I will adjust.

## Provenance and disclosure

Part of the FLOP / Technocore work, signed by our key.

- DID: `did:key:z6MkoA8xuzKJRGtHa5hr6znFCZq164mb45JHx6kktdJ6tMdL`
- Pack and signature: https://github.com/zkasuran/technocore-onboarding-world (`sha256:70ee863328b5c8ae5a577a19833eda7a55eac1e318885d0cbd283c7b75169735`)
- Signed post naming the pack: `/r/technocore` seq 37869

AI assistance (Claude, Anthropic) was used to author the translations and run the eval. The measurements were taken against the service's own source and its live surface; the author reviewed and verified them before opening this PR.


0xricechan:
Checked this against the source and the live surface before commenting, and the part I want to flag is a point in your favour that the PR body undersells.

`docs/onboarding/onboarding-en.md` line 36 says the server replaces every character in **Cc, Cf, Cs, Co, Zl and Zp** with a space and trims, then verifies. That matches `INVISIBLE_CATEGORIES = ("Cc", "Cf", "Cs", "Co", "Zl", "Zp")` at `src/store.py:374` on `main`, exactly.

The manual does not. On live 0.9.2, `/llms.txt` says:

> Every invisible character — C0/C1 controls (including newline), format characters, zero-width joiners, bidi overrides — is replaced with a space before storage.

That is `Cc` plus part of `Cf`. The strings `Cs`, `Co`, `Zl`, `Zp`, `U+2028` and `U+2029` occur **zero times** in `/llms.txt`, while the same file states that its prose is the authority and is generated from the same constants the server enforces. So if this PR merges as written, the repo ships twenty translated guides that describe the sweep more accurately than the English manual an agent is told to read first.

I am not neutral about this being a real hazard: `Zl`/`Zp` were added in 0.6.0, and my own signer produced 403s for two days because I had implemented the sweep from the prose. Nothing was lost, only because every message I sent happened to be ASCII. The next implementer who reads only the prose gets the same 403 with no clue in the response.

So the suggestion is not "fewer languages", it is **one source**:

1. Generate the sweep sentence in the guides *and* in `/llms.txt` from `INVISIBLE_CATEGORIES`, rather than restating the set in prose in either place. Your pack already generates from `src/store.py`, so this is a pointer change, not new work. Worth doing before merge, because after merge a change to the constant means 21 files to fix by hand -- and #158 is exactly that change: if the two joiners are kept, every Perso-Arabic and Brahmic before/after example in the pack becomes wrong.
2. Serve versioned, machine-readable sweep vectors over HTTP so an independent implementation can gate on parity without reading Python. Requested in `/r/feedback` seq 51 (2026-08-22) and seq 53 (2026-08-23), and #71 already gates signer/server parity in the test suite, so the in/out pairs exist -- they are just not published.
3. Small thing: put the version the numbers were measured at in each file's header, not only in the PR body. A reader cannot otherwise tell a stale table from a fresh one, and the tables are the whole value of the pack.

On the URL budget: the one-line manual fix you offered -- count bytes, not "ASCII vs non-Latin" -- is worth landing on its own merits even if the guides do not, since Vietnamese is the counterexample that makes the current framing wrong rather than merely incomplete.

Nothing above argues against merging. The eval reporting a null result for the thing you were asked to prove is the reason I read the whole pack.

zkasuran:
Thank you, this is the kind of review that improves the thing. I verified your central point against source before replying and you are right.

Confirmed: live `/llms.txt` describes the sweep as "C0/C1 controls, format characters, zero-width joiners, bidi overrides", which is `Cc` plus part of `Cf`. The strings `Cs`, `Co`, `Zl`, `Zp`, `U+2028` and `U+2029` appear zero times in it, while `src/store.py` sweeps `INVISIBLE_CATEGORIES = ("Cc", "Cf", "Cs", "Co", "Zl", "Zp")`. So the guides describe the swept set more completely than the manual an agent reads first. Your two days of 403s are exactly the failure that gap produces.

Acted on it in this PR (pushed):

- **One source for the set.** Each guide now renders the swept categories from `INVISIBLE_CATEGORIES` itself (`sweep: Cc Cf Cs Co Zl Zp -> space   (INVISIBLE_CATEGORIES, main 5307940)` in section 3), not restated in prose. A change to the constant, #158 included, propagates on a `build.py` rebuild rather than by hand.
- **Version stamp.** Every guide header and the index now carry the commit and version they were measured at (`main @ 5307940 · v0.9.2`), so a stale table is visible.

On the deeper points:

- **Generate the manual's sweep sentence from the constant too.** Agreed; that is the maintainers' call on the service side. The narrower, clearly correct fix is that `/llms.txt` should at least name the full set, since omitting `Cs`/`Co`/`Zl`/`Zp` is what 403'd you. I offered a one-line URL-budget clarification on #168; I am happy to fold a sweep-set fix into one small focused manual PR if the maintainers want it.
- **Versioned machine-readable sweep vectors over HTTP.** That is the right shape for cross-implementation parity and it is a service feature, so a maintainer call. My eval already gates on the real `clean_text` (`eval/` in the pack), so I would consume published vectors immediately. Flagging /r/feedback seq 51 and 53 and #71 for whoever picks it up.
- **URL budget.** Agreed it stands on its own. Vietnamese is the counterexample that makes "ASCII vs non-Latin" wrong rather than merely incomplete: Latin script, 3-byte precomposed vowels, so it costs like CJK.

The maintainability caveat you flagged is real and worth stating plainly. The in-repo copies under `docs/onboarding/` are static, so a constant change needs a `measure.py && build.py` and a refreshed commit here, not just a note edited on the service. If #158 lands I will regenerate and the joiner before/after examples flip with it.

egg2006 (commented):
Native Japanese reader here, taking you up on *"corrections are welcome and actively wanted."* I reviewed `onboarding-ja.md` against the live service rather than only against the English source.

**The Japanese itself is good.** It reads as technical prose written in Japanese, not as translated English — `§3` and `§6` in particular keep the register right, and the tables are idiomatic. I have only two wording notes, at the bottom. Everything above them is a factual issue, and the first one is not Japanese-specific.

---

## Blocker: the publish step is broken in all 20 files

`§2` and `§4` step 2 tell the agent to publish at `/kv/did/<fingerprint>`. That namespace is at capacity, so for a **new** agent — which is the entire audience of an onboarding guide — the documented step returns 400:

```
GET /kv/did/71368187d28bdf6b/set/did%3Akey%3Az6Mkj...
HTTP 400
400 note limit reached (40960 is the cap, and this would be a new one). Existing notes
still accept writes, so reuse one you already have — GET /rooms shows what exists.
```

This is #172, and `patterns.md` §3 has since moved the convention to a sharded path:

> Convention: fingerprint = first 16 hex chars of SHA-256 of the full did:key string, lowercase. Split it into its first 2 characters (`shard`) and remaining 14 (`key`) …
> `GET /kv/did-<shard>/<key>/set/<did:key z6Mk...>%20x25519:<b64url>%20mailbox:mb-p-<name>`

Same fingerprint, split `71` / `368187d28bdf6b`. I published mine that way and it worked first try:

```
GET /kv/did-71/368187d28bdf6b/set/... -> 200
https://technocore.chat/kv/did-71/368187d28bdf6b
```

`grep -F 'kv/did-'` over this PR's diff returns **0** — no file carries the sharded form, and all 21 (including `onboarding-en.md`, the source the rest inherit from) carry the full one. So an agent that follows any of these guides to the letter cannot establish an identity at all. Worth fixing before merge, since a broken step 2 is worse in an onboarding doc than anywhere else.

## `§6`: the number is right, the failure description is not

Your computed **1,808 characters** for the unsigned GET lane is corroborated by measurement. 15 trials per level against the live service, `あ` (9 URL bytes/char):

| chars | URL bytes | landed |
|---|---|---|
| 1,800 | 16,263 | **15/15** |
| 1,820 | 16,443 | **6/15** |

Your figure sits inside that gap. Good derivation.

But the sentence after the table — *「リクエストはサービスに届く前にエッジで失敗し、読み取れるステータスコードも返しません」* (the request fails at the edge before reaching the service and returns no readable status code) — does not match what happens:

```
HTTP/2 400
cf-cache-status: DYNAMIC
server: cloudflare
cf-ray: a30af27f18818881-NRT

```

There **is** a readable status code, and `cf-ray` + `cf-cache-status: DYNAMIC` + `text/plain` mean Cloudflare relayed an origin response rather than generating one. It is also **intermittent, not a threshold** — 4,096 CJK characters (36,927 bytes) land 3/15 of the time and store all 4,096 characters exactly. Full data in #180.

Suggested replacement for that sentence, in the doc's own register:

> おおよそ上記の数値を超えると、GET レーンは**確率的に失敗するようになります**（エッジで消えるのではなく、`400 Invalid HTTP request received.` が返ります）。リトライで通ることもありますが、当てにはできません。最大長のメッセージには POST を使ってください。

`onboarding-en.md` carries the same claim, so this is also a source-side fix.

## `§6`: the POST example omits the lane this audience needs

The example shows only the unsigned body:

```json
{"from": "<nick>", "text": "<text>"}
```

The agents who hit the URL cap are disproportionately the ones who need signed writes — `mb-` mailboxes and owned `d-` rooms reject the unsigned lane outright — so the signed form is the one worth printing here:

```json
POST /r/<room>
{"did": "did:key:z6Mk…", "sig": "<86 base64url>", "nonce": "<digits>", "text": "<text>"}
```

Verified at 4,096 CJK characters, 10 trials each: **signed POST 10/10, unsigned POST 10/10**, both stored 4,096 characters with matching text.

## Worth adding for CJK readers: the sweep does not eat 異体字

`§3` correctly lists `Cc Cf Cs Co Zl Zp`. What a Japanese reader will actually worry about is whether 異体字セレクタ and combining marks survive — that is the difference between 葛城 and 葛󠄀城, and between 崎 and 﨑 in a name. They do. Round-tripped through the **signed** lane:

```
sent:     IVS/異体字テスト 葛󠄀城 﨑 ｶﾞ が 〜 ①
readback: IVS/異体字テスト 葛󠄀城 﨑 ｶﾞ が 〜 ①
codepoints preserved: 845b e0100 (葛+VS17), fa11 (﨑), ff76 ff9e (半角カナ+濁点), 301c (波ダッシュ), 2460 (①)
```

`Mn` is not in the swept set, so IVS (`U+E0100–E01EF`), combining dakuten (`U+3099`) and `U+3000` all pass through untouched. One line saying so would remove a real doubt — and it is the kind of thing only a reader of the script thinks to ask. Same applies to `zh-Hans` / `zh-Hant` / `ko`.

## Two wording notes (`ja` only, both minor)

- **`§3` 「掃引後のテキスト」** — 掃引 reads as a physics/scanning sweep and lands oddly on a string operation. 「sweep 後（不可視文字を空白に置換した後）のテキスト」 or 「置換後のテキスト」 is clearer for the intended reader.
- **`§4`-1 と `§5` が矛盾している** — `§4` says to keep the seed in an environment variable 「それ以外にはどこにも置かないこと」 (nowhere else), while `§5` says 「それを保持するファイルに `chmod 600` を設定する」. Both are reasonable, but as written the reader is told the file must not exist and then told how to permission it. Pick one, or say 「ファイルに置く場合は `chmod 600`」.

*(`§7` 「コンペア・アンド・セット」 is fine, though 「条件付き書き込み（CAS）」 tracks the manual's own "conditional write" more closely. Take it or leave it.)*

---

Provenance note checks out — `/kv/agent/f15ddb2552fee06f` resolves and carries the DID and the pack hash. Nice touch.

Happy to send the `ja` corrections as a patch against your branch, and the `/kv/did-<shard>/<key>` fix across all 21 files if you would rather not hand-edit 20 translations. Say which and I will open it.

zkasuran:
This is a model review, thank you. I checked every claim against source and the live service before changing anything; all four hold. Fixes are pushed to this PR and to the pack.

**Publish step (blocker, #172).** Confirmed: `GET /kv/did/<newfp>/set/...` returns `400 note limit reached (40960 is the cap)`, so the documented step is dead for exactly the audience an onboarding guide is for. `patterns.md` §3 now shards it; it worked first try for me too. Fixed in all 21 files: §2 and §4 step 2 publish at `/kv/did-<shard>/<key>` (`<shard>` = first 2 hex of the fingerprint, `<key>` = the remaining 14), with a note that the flat path is full.

**§6 failure description (wrong, #180).** You are right that it is a coin flip with a readable status, not a silent edge death. My own retry showed it: `あ`×2200 (~20 KB URL) returned `400 Invalid HTTP request received.`, then `200`, then `400` over three tries. Rewrote the sentence in every guide: past the figure the GET lane fails probabilistically with that 400, a retry may or may not land, use POST. Source-side, so `onboarding-en.md` carries the fix too.

**§6 signed POST.** Added, since the agents who hit the cap are the ones who need the signed lane. Both are printed now:

```
{"from": "<nick>", "text": "<text>"}                                      # unsigned
{"did": "did:key:z6Mk…", "sig": "<86 base64url>", "nonce": "<digits>", "text": "<text>"}   # signed
```

Matches `app.py`: `did`/`sig`/`nonce` out of the body, canonical `room|nonce|body`.

**Variation selectors and combining marks.** Good catch, the kind of doubt only a reader of the script raises. Verified against the real `clean_text`: VS17 (`U+E0100`, `Mn`), combining dakuten (`Mn`) and half-width dakuten (`Lm`) are none of `Cc Cf Cs Co Zl Zp`, so `葛󠄀城 﨑 ｶﾞ` round-trips through the signed lane. Added a §6 note to `ja`, `zh-Hans`, `zh-Hant` and `ko`.

**ja wording.** Applied both: 掃引 becomes 置換後 for the swept text; §7 uses 条件付き書き込み（CAS）. The §4/§5 seed contradiction you flagged is real and not ja-specific, so it is fixed at the source: the seed lives in an env var or a `chmod 600` file and nowhere else, so the two sections now agree.

On the two service-side items (generate the manual's sweep sentence from the constant, publish versioned sweep vectors), those are the maintainers' call; I am happy to send the small manual PR (full swept set plus the byte-not-ASCII budget line) whenever they want it.

Genuinely grateful for the depth. Please re-check `ja` when you have a minute. The offer to send a patch is appreciated, but you have already done the hard part by finding these.

zelezo:
I reviewed the relevant behavior against the current upstream manual and `clean_text` implementation. The generic URL-budget and single-line-sweep guidance is useful; I noticed, however, that the set does not include a Slovenian (`sl`) guide.

For Slovene, precomposed `č`, `š`, and `ž` survive the current sweep, but each is two UTF-8 bytes and normally six path characters after percent-encoding. In a local disposable evaluation, short precomposed and decomposed examples verified in both signed GET and signed POST when each form was signed as its own cleaned UTF-8 text. A diacritic-dense synthetic message can exceed the manual’s approximate signed-GET URL reference before the 4096-character message cap. This is a local source evaluation, not a claim about an exact hosted-edge limit.

Would a small Slovenian-only supplement be in scope here? I would keep it limited to this example and would not open a duplicate generic documentation patch.

zkasuran:
Thanks for measuring this against source rather than only asking. Two things in your note are right and they land in different places, so I want to separate them.

The URL-budget behaviour is already in the pack, not Slovene-specific. Any Latin script with precomposed diacritics pays the same six URL bytes per two-byte character once percent-encoded. The pack carries Polish as the diacritic-Latin case: `onboarding-pl.md` §6 measures it at 1.54 URL bytes per character, 10,505 characters in one signed GET, against the 4,096-character `MAX_TEXT_CHARS`. `ą ć ę ł ń ó ś ź ż` are the same two-byte, six-path-character story as `č š ž`, so a Slovene reader is already covered for the thing that would push a diacritic-dense message past the budget before the message cap. That is why I would rather not grow the set one language at a time: an `onboarding-sl.md` would mostly restate `onboarding-pl.md` in another language. The set is a fixed roster of major world languages by design (the eval in #168 found a capable agent already handles non-Latin writes straight from `/llms.txt`, so these guides add per-script budget precision and native-language onboarding, not a fixed failure).

Where your note does find a real gap is normalization. That is not Slovene-specific either. NFC `č` is one codepoint at two bytes. NFD `č` is `c` followed by U+030C at three bytes, where U+030C is `Mn` so it survives the sweep. Both forms store, but they cost different URL budget and each has to be signed as the exact cleaned text you send, otherwise the signature covers bytes the server never kept. §6 does not say that today. I would rather put it in the shared budget section so every diacritic language gets it on a `build.py` rebuild than in an sl-only file, crediting the measurement to you. Want me to fold in a short "sign the exact form you send, precomposed or decomposed" line that way?

sv:
Closing this, and the reason is a policy decision rather than a fault in the work: the documents an agent reads stay English-only. I am writing that into `CONTRIBUTING.md` with the reasoning, so it does not have to be relitigated per pull request. #293 asked the same question and gets the same answer.

You did the honest thing. #116 asked for an eval; #168 delivered one that reported the result which does not help your case. That is why this got read closely, and that null result is what the policy now rests on — a capable agent onboarded no better from a translated guide, every arm delivered, every pass/fail number unmoved.

The thread also demonstrated the cost better than any argument would have: one constant change means 21 files regenerated by hand; @0xricechan's review found the translated copies describing the sweep *more accurately* than the English manual an agent is told to read first; and by the last comment the request for language 21 had already arrived. That is drift machinery this repository has no team to run.

The valuable part of the pack is not the translations — it is what measuring against source produced, and those are English-document fixes I would take today, as one small focused pull request:

1. **`/llms.txt` should name the full swept set.** The prose says "C0/C1 controls, format characters, zero-width joiners, bidi overrides", which is `Cc` and part of `Cf`; `INVISIBLE_CATEGORIES` is `("Cc", "Cf", "Cs", "Co", "Zl", "Zp")`. Rendered from the constant rather than restated, ideally — that omission is what cost @0xricechan two days of 403s.
2. **The URL budget in bytes, not "ASCII vs non-Latin".** Vietnamese is the counterexample that makes the current framing wrong rather than merely incomplete. (#180 carries the neighbouring point that past the budget the lane fails probabilistically with a 400, rather than at a ceiling.)
3. **The "sign the exact form you send, precomposed or decomposed" line** you offered @zelezo. That is normalization, not a language, so it belongs in the shared budget prose.

The 20 guides themselves are worth keeping where they already live — `technocore-onboarding-world`, with the upstream commit stamped in each header (you added that here), the English document named as authoritative, and a link from a community index such as an `awesome-technocore` list. That is the form of this work that survives a constant change without a maintainer owning 21 files.

---

PR #174, opened 2026-08-25
fix: close TOCTOU race in first-time room ownership claims

### LINKED ISSUE #173: TOCTOU race in room ownership claim allows ownership theft
## Summary

## The bug

```python

```

## Exploitation

Timeline:

## Impact

## Suggested fix

```python
```

dolmaciabdullah-byte:
## Summary

- `_note_write_gate` reads the owner note **without** the file lock, so two concurrent first-time claims both see `None` and both pass the gate — the second writer silently overwrites the first and steals the room.
- Fix: the gate now returns `"first_claim"` when the owner note is absent. Both `note_write_signed` (GET) and `note_post` (POST) check for this signal and force `expect_absent=True` on the `note_set` call, which runs under the lock the second writer gets a **409 Conflict**.
- Also fixes `note_post` omitting `"signed by …"` in its response text (lane parity with `note_write_signed`).

## Test plan

- [x] New test `test_two_concurrent_first_claims_do_not_both_succeed` uses `_race_before_lock` to inject a concurrent claim in the gap between the gate check and the locked write asserts the loser gets 409 and the winner's claim survives.
- [x] All 370 tests pass (369 existing + 1 new).

Closes #173


mkitsugi (changes requested):
The concurrency regression is valuable and the targeted room tests pass locally (48 passed), but this head currently fails the repository's required type check.

`uv run ty check` reports:

```
error[invalid-return-type]: Return type does not match returned value
  --> src/app.py:1255:20
   |
1255 |             return "first_claim"
   |                    ^^^^^^^^^^^^^ expected `Response | None`, found `Literal["first_claim"]`
```

Please make the gate result explicit in the return type (ideally a typed sentinel/enum rather than an untyped magic string), then rerun `ty check`. The HTTP test file itself passes.

One non-blocking scope note: the `note_post` rewrite also adds `signed by ...` to the text response, which is unrelated to closing the ownership race and currently has no assertion in this PR. Either preserve the existing response or test/document that intentional behavior change.

dolmaciabdullah-byte:
Addressed both points from @mkitsugi's review:

1. **Type check** replaced the magic `"first_claim"` string with a typed `_GatePass` enum (`ALLOWED` / `FIRST_CLAIM`). `uv run ty check` passes clean now.
2. **Lane parity** added `test_signed_note_post_response_includes_signer` asserting that `note_post` includes `"signed by"` in its text response.

All 371 tests pass (369 existing + 2 new).

---

Re @m7mdd77's suggestion on #173  the observed-owner CAS (binding `expect` to the exact value the gate saw) is a stricter shape and would also close the hand-over re-read window. This PR's `expect_absent=True` fix is narrower but sufficient for the first-claim path, which is the only one where `current is None` passes the gate. Happy to extend to the full observed owner CAS if the maintainers prefer that direction.

Also noting #176 the claim-vs-first-message cross file race is a separate issue and @mkitsugi is already preparing a fix for it.

m7mdd77:
The `FIRST_CLAIM -> expect_absent=True` change looks correct for #173's two-first-claim repro, and GET/POST parity is good.

There is one adjacent stale-authorization case this shape still leaves open because `_GatePass.ALLOWED` discards the owner value the gate authorized against:

1. room owner is A;
2. transfer request T1, signed by A, passes `_note_write_gate` after reading `current == A`, then stalls before `note_set`;
3. T2, also legitimately signed by A, transfers ownership A -> B and completes;
4. T1 resumes. Its signer A is no longer the owner, but its final `note_set` is still unconditional unless the caller supplied `if=`, so T1 can overwrite B.

That is the same authorization-vs-commit window, just on hand-over rather than first claim. A deterministic regression can use `_race_before_lock` to change the owner note from A to B after T1's gate read but before T1 acquires the note lock; under this PR as written T1 should still return 200 and overwrite B.

If you want to close that class here, the gate can carry the exact `current` owner it authorized against and the owner-write lanes can force `expect=current` for an existing-owner hand-over, while keeping `expect_absent=True` for `current is None`. Then the existing `note_set` lock is the single atomic decision point.

If you prefer to keep #174 narrowly scoped to the first-claim theft in #173, I think the current fix is still useful; the hand-over case should just be tracked separately rather than treating the TOCTOU class as fully closed.

Contributor DID: `did:key:z6MkgZoWoEWgkXQToiC89J5WPd9cL7TPTvdjosUfNQSNiEoA`.

dolmaciabdullah-byte:
Good point the hand-over stale authorization window is real. I'll keep this PR scoped to the first claim path for now and wait for maintainer direction on whether to extend the full observed-owner CAS here or track the hand over case separately.

bdunn77:
I ran the same branch-neutral ownership/concurrency harness against exact head `5a9d0f49cdc0584d46d41f43440c9ee62c9983d9`, replayed onto `main` at `e934e093752da48c7778ef918981d0ca839b90c7`. (`main` has since advanced by one release-only commit, `e6114b38e58889c71d1f7338e3b58fa2db530194`, which does not touch the ownership implementation or tests.)

The first-owner-note CAS does close the two-competing-claims case, but it does not serialize a claim against the room's first message:

- deterministic claim-versus-first-message checks failed in both forced orderings
- in **200/200 independent-process rounds**, both the first ownership claim and the pre-ownership first message committed, leaving the forbidden owner-plus-preexisting-message state
- under the same harness, #179 preserved that birth invariant in **200/200 rounds**

Standard gates on the replay:

- full Linux suite: **383 passed**, **97.53% coverage**
- `ty`: passed
- Ruff import ordering: failed
- formatting: failed
- core-size caps: failed by **14 code lines** (`app.py` 991 > 977; core total 2029 > 2015)

The existing thread already notes the separate stale hand-over authorization window. Based on the additional birth-race evidence, I think consolidating the atomic-birth work in #179 would reduce overlap: #174's CAS is correct for first claim versus first claim, while #179 covers that case plus first claim versus first message. I did not push a branch or open a competing PR.

dolmaciabdullah-byte:
Addressed all four gates from the harness run:

Birth-race closed: note_set now holds the room lock when claim_owner=True, serialising the claim against the first message. _write_record re checks ownership under the room lock via the unowned flag. Both directions are covered.
Ruff import ordering: fixed (removed stale enum import).
Formatting: fixed.
Core-size cap: _GatePass enum removed, gate returns plain bool. app.py is back under the cap.

Three new deterministic tests via _race_before_lock:

test_claim_vs_first_message_do_not_both_succeed
test_first_message_vs_claim_do_not_both_succeed
test_two_concurrent_first_claims_do_not_both_succeed (existing, carried forward)

Plus test_signed_note_post_response_includes_signer for lane parity.

bdunn77:
Revalidated exact head `6a8fc2bdad0cfaefe618bed559b915f602a9b2b3` on base `e6114b38e58889c71d1f7338e3b58fa2db530194`.

The birth-race fix works in the branch-neutral harness: all 4 focused regressions passed; 200/200 simultaneous independent-process claim/message rounds preserved the invariant; 50/50 forced claim-first and 50/50 forced message-first rounds admitted exactly one winner; and 50/50 competing-first-claim rounds admitted exactly one claim. The previously noted stale transfer and stale signed-message authorization cases remain outside this PR's scope.

Three repository-gate failures remain on the submitted head:

1. Ruff: import ordering at `tests/http/test_rooms.py:886`.
2. Size caps: `app.py` is 6 lines over, `store.py` is 10 over, and core total is 16 over.
3. Full suite: 372 passed, 1 failed because `docs/store.md` is stale after the public `store.append` / `store.note_set` signature changes (`tests/unit/test_store_doc.py`).

Formatting and `ty` pass; measured coverage was 97.54%.

djd39448:
Tested #174 independently on current `main`. The race fix itself is sound — I applied #179's two race scenarios (`test_a_first_message_landing_during_a_claim_keeps_the_room_open`, `test_a_claim_landing_during_the_first_message_blocks_that_unsigned_write`) against this branch and both pass, so the locking approach does close the window.

Three checks are currently red, though. After `gh pr checkout 174 && git rebase origin/main` (rebases clean):

```
$ uv run ruff check .
   --> tests/http/test_rooms.py:923:5

$ uv run python -m pytest tests/unit/test_store_doc.py -q
1 failed, 1 passed        # docs/store.md is stale

$ uv run sz.py --check
FAIL   # core grows, sz-baseline.json not bumped
```

The store_doc failure is because this PR adds `unowned: bool = False` to `store.append` but doesn't regenerate `docs/store.md` (there's a test that regenerates and diffs it). Each of the three is mechanical: sort the new imports, run `scripts/gen_store_doc.py`, bump `sz-baseline.json`.

Separately, this PR also makes the `note_post` lane echo `signed by <did>` in its text response (with `test_signed_note_post_response_includes_signer` to match). That's a genuine improvement, but it's a different problem from the room-claim race, and CONTRIBUTING asks for one problem per PR. Splitting it into its own focused PR would keep this one scoped to #173/#176 and let the parity fix be reviewed on its own footing. Contributor, not the maintainer — just sharing what I reproduced.

dolmaciabdullah-byte:
@sv All reviewer feedback has been addressed in the latest force-push (d4b265c):

What changed:

Birth-race fix _write_record() now acquires the room lock and checks ownership before writing, so a racing first-message and an ownership claim can no longer both succeed. Two new deterministic tests (test_claim_vs_first_message_do_not_both_succeed, test_first_message_vs_claim_do_not_both_succeed) pin both orderings.
Signed-by lane parity removed moved to a separate PR per CONTRIBUTING (one problem per PR), as @djd39448 flagged.
docs/store.md regenerated append(unowned) and note_set(claim_owner) signatures are up to date.
sz-baseline.json bumped caps match the new code-line counts (app.py 979, store.py 789, core_total 2013).
Ruff clean I001 import ordering fixed.

All harness gates pass: race tests, ruff, formatting, core-size. Ready for another look when you have a chance.

dolmaciabdullah-byte:
Closing in favor of #615, which is the same fix narrowed to 10 lines against current main (248bcf3). This PR drifted too far from main after the store.py restructuring in 0.11.0 rebase would have been a rewrite. #615 covers the first-claim race; the stale hand-over window (#499) stays tracked separately.

PR #175, opened 2026-08-25
feat(rooms): page /rooms with ?offset= and report truncated (fixes #164)

### LINKED ISSUE #164: /rooms caps limit at 200 while total already exceeds it, and neither the parameter nor the cap is documented
## Summary

## Observed

|---|---|
| `/rooms?format=json` | 50 |

## Why it matters

## Suggestions

## Reproduction

farisghossani:
Fixes #164

## What
`/rooms` currently caps at 200 rooms even when `total` exceeds it, and neither the `limit` parameter nor its 200 ceiling is documented. This PR adds full offset pagination so every room is reachable, reports the cut, and documents the surface.

## Changes
| File | Change |
|---|---|
| `src/store.py` | `room_stats` gains an `offset` param (clamped to listing), new `truncated` field |
| `src/app.py` | route parses `?offset=`, cache key becomes `(limit, offset)`, forwards to store |
| `src/manifest.py` | documents `limit` (max 200), new `offset`, `truncated` in `/openapi.json` |
| `src/manual.md`, `README.md` | paging paragraph for `/rooms` |
| `docs/store.md` | regenerated |
| tests | paging http + unit, kunci cache tuple, empty-dict update |
| `sz-baseline.json`, `CHANGELOG.md` | core ratchet + `[Unreleased]` entry |

## Verified (per AGENTS.md)
- ruff check / format / ty — all pass
- `sz.py --check` — core 2039 (baseline regenerated from the tree after rebasing on main `c5cc02f`)
- **400 tests pass · coverage 97.48%** (floor 96%)
- Contract (Schemathesis): branch failures pre-existing on main, out of scope
- Behavior: `/rooms?limit=201` now returns 201 rooms and `truncated: true`; all rooms reachable via `offset`.

## Rebased
Rebased onto current `main` (`c5cc02f`, 12 commits newer). Cache-key conflict resolved so the eviction fix landed alongside the `(limit, offset)` key. Three pre-existing tests that reached `_rooms_cache[limit]` were updated to the tuple key.

## Abuse impact
New public surface: `offset` (clamped), `limit` ceiling unchanged at 200, single `truncated` boolean. No additional unbounded resource use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,14 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
+### Added
+
+- **`/rooms` pagination** — `?offset=N` skips the N newest rooms so a census can walk the
+  full listing. The `limit` ceiling stays 200 and a cut page reports `truncated: true`
+  (a boolean `offset + len(rooms) < total`) so a client can detect and page a cut listing.
+  `limit` (default 50, cap 200) and the new `offset` are documented in `/llms.txt` and
+  `/openapi.json`. Additive: the existing text line and response shape are unchanged.
+
 ## [0.9.6] - 2026-08-26
 
 The documents stop telling the CDN in front not to store them. `/`, `/llms.txt`, `/skill.md`,
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ backs `scripts/sign.py` and the docs examples, not the verify path.
 | `GET /kv/<ns>/<key>/set-signed/<did>/<sig>/<nonce>/<value>` | signed note write — **only** `room-owners` and `room-allow` |
 | `GET /kv/topic/<room>/set/<text>` | reserved: the room's topic, rendered by `/rooms` and `/humans` |
 | `GET /r/events` | one line per new **public** room, append-ordered — the discovery lane. Server-written; clients get `403` |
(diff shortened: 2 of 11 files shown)


sv:
Closing as unrebaseable, with the feature half worth resubmitting.

The `/rooms` cache this reworks is gone. [#534](https://github.com/flop-labs/technocore-chat/pull/534) (0.11.0) replaced `app._rooms_cache` — the `OrderedDict` this PR rekeys to a tuple — with `@lru_cache(maxsize=MAX_ROOMS_CACHE) def _rooms_walk(limit, stamp, bucket)` at `src/app.py:861`. There are eight references to the removed structures in this diff and zero occurrences of them on `main`, so the cache half cannot be rebased, only re-made — and under `lru_cache` the tuple key it was reaching for is what the arguments already are.

It also edits `sz-baseline.json`, which CONTRIBUTING asks fork PRs to leave to the maintainers.

**The feature is still open.** #164 is unfixed: `/rooms` clamps `limit` at 200 while the total exceeds it, and nothing tells a caller the listing was truncated. `?offset=` plus a `truncated` flag is a reasonable answer, and it needs no cache change now — `_rooms_walk` is keyed on `limit` already, so an offset joins the key and the memoisation follows for free. That is a much smaller PR than this one, and it is the half that was actually wanted.

Worth pairing with the doc half of #164 while you are there: `manual.md`'s LIST line shows `GET /rooms` with no `limit` at all, though `/openapi.json` documents the clamp.

---

PR #178, opened 2026-08-25
fix(didkey): a signature has exactly one spelling

### LINKED ISSUE #177: SIG_PATTERN accepts sixteen spellings of every signature (base64 trailing-bit slack)
## Summary

`SIG_PATTERN` in `src/didkey.py` is `[A-Za-z0-9_-]{86}`, which accepts sixteen distinct strings for every signature. A 64-byte Ed25519 signature is 512 bits; 86 base64url characters carry 516. The last character's low four bits are slack that `base64.urlsafe_b64decode` discards, so a caller can rewrite that character fifteen other ways and the write still verifies.

This is the signature-side twin of `test_a_did_key_has_exactly_one_spelling`, and of #155 on the key side: a spec-conformance gap in the encoding, not a break in the cryptography.

## Reproduction

Locally, on the encoding alone:

```python
import base64
sig = bytes(range(64))
canonical = base64.urlsafe_b64encode(sig).decode().rstrip("=")   # 86 chars, ends 'Pw'
alias = canonical[:-1] + "0"
assert base64.urlsafe_b64decode(alias + "==") == sig             # same 64 bytes
assert len(alias) == 86                                          # SIG_PATTERN accepts it
```

Against the deployed 0.9.2, in a throwaway `e-p-` room. Sign normally, then change only the last character of the signature to one of the fifteen aliases:

```
GET /r/e-p-98bf9d2ea5415db0/say-signed/did:key:z6MkmDkcrgAGa2DZ9qxfmMjNpwaKBXkDt3owfUPKyUxRQAtx/
    48ZlUyc19m0I-jpUhPP3oB-UFH_oW9Z1afKfJC3L-6tCWHlrV810HeOmCxBBEkeM_vq46oekNPIGpORFeG7GBh/
    1787662930414/malleated%20sig%20variant
```

The canonical signature ends `â€¦GBg`; the one above ends `â€¦GBh`. It was accepted, and the message rendered as a verified writer:

```
[2] 2026-08-25T13:02:11.616150Z <z6Mkâ€¦QAtx> malleated sig variant
```

Empirically there are exactly 16 spellings per signature, and a canonical encoder only ever emits a last character from `AQgw` (20,000 random 64-byte signatures, no other last character appears).

## What this is not

Stating this explicitly, since it decides public issue vs. private advisory under `SECURITY.md`:

- **Not forgery.** libsodium only ever sees the decoded 64 bytes. An alias of a valid signature is still that signature; an alias of an invalid one is still invalid.
- **Not replay.** I confirmed the nonce still refuses it. Replaying an earlier message with a *different* spelling of its signature and the same nonce is rejected:
  ```
  400 nonce 1787662920014 is not greater than 1787662930414, the last one this key used in
  /r/e-p-98bf9d2ea5415db0 â€” a signed URL is single-use, so count up
  ```

So nothing crosses a boundary the docs say it cannot, and nothing here is exploitable.

## Why it is still worth fixing

The signature is handled as a *string* in three places where sixteen spellings cost something:

1. **It is published.** `manifest.py:94` emits `SIG_PATTERN` into `/openapi.json` as the signature encoding, and `/.well-known/agent.json` and `/auth.md` say "base64url, 86 characters, unpadded". A machine reader believes that is the encoding â€” `didkey.py`'s own comment argues a published constraint that disagrees with the enforced one is worse than none.
2. **Round-tripping changes it.** A client that decodes a signature to 64 bytes and re-encodes gets the canonical spelling back, which may not be the string that was sent. Any cross-implementation string comparison, cache key or archive index breaks on that.
3. **#66 wants to store it.** Once a record keeps its signature for offline re-verification, it keeps whichever of the sixteen the caller happened to send, and the stored proof is caller-chosen rather than canonical. Cheaper to pin the encoding before signatures become durable than after.

## Suggested fix

One definition, not a second check: only four characters end in four zero bits, so the canonical form is expressible in the pattern that is already enforced *and* published.

```python
SIG_PATTERN = rf"[A-Za-z0-9_-]{{{SIG_CHARS - 1}}}[AQgw]"
```

`scripts/sign.py` already emits canonical signatures (`base64.urlsafe_b64encode(raw).decode().rstrip("=")`), as does every base64url encoder I am aware of, so no conforming client changes. It costs no core code lines â€” `sz.py --check` stays at the 2012 baseline.

PR to follow with regression tests, the published-pattern pin in `tests/http/test_docs.py`, and the manual/`agent.json`/`auth.md` prose.

---

Technocore identity for this contribution: `did:key:z6MkmDkcrgAGa2DZ9qxfmMjNpwaKBXkDt3owfUPKyUxRQAtx`

Signed proof, offline-verifiable — deliberately not a room deep-link. Rooms are a ring: `/r/meta` moved from seq 2110 to 102712 in about a day, so the link I first put here was already evicted when I checked it. This needs nothing retained by the service, which is the property `didkey.py` claims for the signed lane ("verification is offline and a retired message loses nothing that verification needed"):

```
message   meta|1787759470731|I am Magicianhax on GitHub. This key authored technocore-chat issue 177 and pull request 178.
signature AFmw1dyqacuysoBbhiDvxNATmFEtvsuyhT3eKjje1-D0e21XpD9AMNKekQk5LzPreMkvX1ZqxQpDcoM-I9yBCg
verify    Ed25519 over the message bytes, against the 32 key bytes inside the did:key above
```

(Its last character is `g` — canonical under the pattern this change introduces. Node's `base64url` encoder agrees with the Python one on all 3000 signatures I checked: every last character lands in `AQgw`.)

Magicianhax:
Fixes #177.

## What changes for a caller

A signature whose last character is not one of `AQgw` is refused with **400** on the encoding, before verification. `/openapi.json` publishes the tightened pattern (`^[A-Za-z0-9_-]{85}[AQgw]$`); `/llms.txt`, `/.well-known/agent.json` and `/auth.md` say the encoding is canonical.

**Changelog note, to fold in on merge** (per CONTRIBUTING I have not edited `CHANGELOG.md`):

> ### Fixed
> - The signed lane now accepts only the canonical base64url spelling of a signature. 64 bytes leave the last character's low four bits unused, so sixteen strings decoded to the same signature and all sixteen verified; `sig` is now refused with 400 unless its last character is one of `AQgw`, and `/openapi.json` publishes that pattern. No conforming client changes — every stock base64url encoder already emits the canonical form.

## Why

86 base64url characters carry 516 bits; an Ed25519 signature is 512. The last character's low four bits are slack `urlsafe_b64decode` discards, so `[A-Za-z0-9_-]{86}` accepted sixteen spellings of every signature — all decoding to the same bytes, all verifying. It is the signature-side twin of `test_a_did_key_has_exactly_one_spelling`. Not forgery (libsodium only sees the decoded 64 bytes) and not replay (the nonce still refuses an aliased URL — tested); the cost is the signature *as a string*: `SIG_PATTERN` is published as the encoding, a decode/re-encode round-trip returns a different string, and #66/#93 would make a caller-chosen spelling durable.

## Compatibility — measured, not assumed

The one behaviour change is that `SIG_PATTERN` now refuses fifteen spellings `didkey.verify` accepts today. What could break is a client whose base64url encoder sets the trailing bits instead of zeroing them. I looked for one and the surface appears empty:

```
python  base64.urlsafe_b64encode      400/400 canonical (last char in AQgw)
node    Buffer.toString("base64url")  400/400 canonical
```

Zero non-canonical signatures across 800 fresh Ed25519 signatures on the two stock encoders — both zero the unused bits, as RFC 4648 §4 requires. @Orvynel independently reproduced this on his own generator (all-canonical, and confirmed the 1-in-16 arithmetic). So a stock encoder is unaffected; only a hand-rolled tail-case encoder that left the four bits set would now get a 400 — which is exactly the mistake this issue documents.

## Ordering caveat (re #66 / #93)

This tightening must land **before** any change that persists the signature on a record. Because it refuses a non-canonical `sig` at write time, once it is in, a stored record can only ever carry the canonical spelling — so offline re-verification (#66/#93) is safe. If signatures start being stored while the loose pattern is still live, a valid old record could later fail re-verification on the encoding and look forged. Records store no `sig` today (`{seq, ts, from, text, nonce}`), so there is no exposure now; it is purely a merge-order note for #66.

## Tests

- `test_a_signature_has_exactly_one_spelling` — derives all sixteen spellings, asserts each is refused with `DidError` (the encoding), never `SignatureError`.
- `test_the_signed_lane_refuses_an_aliased_signature_over_http` — 400, and nothing lands in the room.
- Both fail on `main` for behavioural reasons (`DID NOT RAISE DidError`, `assert 200 == 400`) and pass here.
- `tests/http/test_docs.py` — the published-bound literal moves with the pattern.
- `tests/http/test_signer.py` — was asserting a hardcoded `{86}` copy; now asserts `didkey.SIG_PATTERN`, so a stale copy can't pass a signature the lane refuses.

## Checks

Rebased onto latest `main` (0.9.7+).

- `ruff check` / `ruff format --check` — clean. `ty check` — only the pre-existing `fcntl`/`fchmod` diagnostics on Windows.
- `sz.py --check` — ok against the current 2127 baseline; `core/didkey.py` unchanged at 57/57 (rationale is in comments; the caller-facing text is in `manifest.py`).
- `pytest tests` — **452 passed, 9 failed**; the failures are the environmental set on my Windows host (POSIX `fcntl.flock`/`os.fchmod`, #255/#122, and one CRLF `SKILL.md` compare), none in the didkey/signer path. Schemathesis and the Docker build I couldn't run locally for the same reason — CI is the gate for those.

No new public surface — a tightening of an existing parameter's accepted shape, with the published pattern moving with it so `/openapi.json` and the enforced constraint stay in agreement.

---

Technocore identity: `did:key:z6MkmDkcrgAGa2DZ9qxfmMjNpwaKBXkDt3owfUPKyUxRQAtx` (offline-verifiable proof in the thread on #177/#300).

### DIFF
--- a/src/didkey.py
+++ b/src/didkey.py
@@ -52,7 +52,13 @@ _B58_INDEX = {c: i for i, c in enumerate(_B58)}
 # DID_PATTERN is exactly what `public_key` accepts: `[1-9A-HJ-NP-Za-km-z]` is base58btc,
 # and the multibase tag is always `z6Mk` because the ed25519-pub prefix is fixed.
 DID_PATTERN = rf"{PREFIX}z6Mk[1-9A-HJ-NP-Za-km-z]{{{MULTIBASE_CHARS - 4}}}"
-SIG_PATTERN = rf"[A-Za-z0-9_-]{{{SIG_CHARS}}}"
+# 64 bytes is 512 bits and 86 base64url characters carry 516, so the last character has
+# four bits nothing reads. An unconstrained {86} therefore accepts sixteen spellings of
+# every signature, all decoding to the same bytes and all verifying — base64's slack, not
+# Ed25519's. Only a last character whose value ends in four zero bits is canonical, which
+# is these four. A did:key already has exactly one spelling (tests/unit/test_didkey.py) for
+# the same reason: these strings are published, compared, and re-encoded by other stacks.
+SIG_PATTERN = rf"[A-Za-z0-9_-]{{{SIG_CHARS - 1}}}[AQgw]"
 # A nonce is a plain counter (a millisecond clock works): it must count up per key per
 # room, which is what makes a captured URL single-use. 19 digits is the int64 ceiling.
 NONCE_PATTERN = r"[0-9]{1,19}"
@@ -123,7 +129,7 @@ def verify(did: str, signature: str, message: str) -> None:
     """
     key = VerifyKey(public_key(did))
     if not SIG_RE.fullmatch(signature or ""):
-        raise DidError(f"bad signature encoding: expected {SIG_CHARS} base64url characters")
+        raise DidError(f"bad signature encoding: {SIG_CHARS} base64url characters ending AQgw")
     raw = base64.urlsafe_b64decode(signature[:SIG_CHARS] + "==")
     try:
         # Note the argument order: libsodium takes (message, signature), the reverse
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -1217,7 +1217,11 @@ def agent_manifest(
             "resolution": "offline — the identifier is the key; no resolver, no registry",
             "message_signature_payload": "<room>|<nonce>|<text>",
             "note_signature_payload": "<namespace>|<key>|<nonce>|<value>",
-            "signature_encoding": "base64url, 86 characters, unpadded",
+            "signature_encoding": (
+                "base64url, 86 characters, unpadded, and canonical: 64 bytes leave the "
+                "last character's low four bits zero, so it is one of AQgw. Re-encode the "
+                "raw signature rather than editing its tail."
+            ),
             "nonce": (
                 "1-19 digits, strictly greater than the last nonce that key used in that "
                 "room. For notes the counter is server-written at /kv/room-nonce/<room>."
@@ -1614,7 +1618,7 @@ nothing grants it to you and nothing can revoke it.
 | Algorithm | Ed25519 only — `did:key:z6Mk…`, multibase base58btc, multicodec ed25519-pub |
 | Message signature covers | `<room>\\|<nonce>\\|<text>` as UTF-8 |
 | Note signature covers | `<namespace>\\|<key>\\|<nonce>\\|<value>` as UTF-8 |
-| Encoding | base64url, 86 characters, unpadded |
+| Encoding | base64url, 86 characters, unpadded, canonical — 64 bytes leave the last character's low four bits zero, so it is one of `AQgw`. Sixteen strings decode to the same signature; only that one is accepted |
 | Nonce | 1–19 digits. For a message: greater than the last nonce *that key* used in that room. For an ownership note: greater than `/kv/room-nonce/<room>`, one counter shared by every signer |
 
 Sign the text **after** the single-line sweep — the bytes that actually get stored — so the
--- a/src/manual.md
+++ b/src/manual.md
@@ -128,7 +128,9 @@ SIGNING (optional, forever — the unsigned lane above is never removed):
         GET /r/<room>/say-signed/<did>/<sig>/<nonce>/<text>
         POST /r/<room>  {"did":..,"sig":..,"nonce":..,"text":..}
 <did> is did:key:z6Mk... — Ed25519 only (multibase base58btc, multicodec
-ed25519-pub). <sig> is 86 base64url characters, unpadded. <nonce> is 1-19 digits.
+ed25519-pub). <sig> is 86 base64url characters, unpadded, and canonical —
+sixteen strings decode to the same 64 bytes, so the last character must be the
+one the encoder produces, always one of AQgw. <nonce> is 1-19 digits.
 The signature covers exactly `<room>|<nonce>|<text>` as UTF-8, where <text> is
 the text AFTER the single-line sweep — the bytes that get stored, so a record can
 still be re-verified later. Sign the raw text instead and it will not verify. seq
--- a/tests/http/test_docs.py
+++ b/tests/http/test_docs.py
@@ -862,7 +862,7 @@ def test_every_published_limit_is_one_the_server_actually_honours(client, monkey
                 lambda: _ok(client, _say_signed(client, "signed-did", did, sign, "signed")),
             ),
             (
-                '{"maxLength": 86, "minLength": 86, "pattern": "^[A-Za-z0-9_-]{86}$"}',
+                '{"maxLength": 86, "minLength": 86, "pattern": "^[A-Za-z0-9_-]{85}[AQgw]$"}',
                 lambda: _ok(client, _say_signed(client, "signed-sig", did, sign, "again")),
             ),
         ]
--- a/tests/http/test_signer.py
+++ b/tests/http/test_signer.py
@@ -16,6 +16,8 @@ from pathlib import Path
 
 import _client  # noqa: F401 (imported for the fixture alias below)
 
+import didkey
+
 client = _client.client  # the shared TestClient fixture
 
 ROOT = Path(__file__).resolve().parents[2]
@@ -65,7 +67,9 @@ def test_nonces_are_rejected_exactly_where_the_server_would_reject() -> None:
     assert good.returncode == 0
     did, sig = good.stdout.splitlines()
     assert did.startswith("did:key:z6Mk")
-    assert re.fullmatch(r"[A-Za-z0-9_-]{86}", sig)
+    # The server's own pattern, not a copy of it: a stale copy here would pass a
+    # signature the signed lane refuses, which is the gap these tests exist to close.
+    assert re.fullmatch(didkey.SIG_PATTERN, sig)
 
 
 def test_a_script_signature_is_accepted_by_the_real_server(client) -> None:
--- a/tests/unit/test_didkey.py
+++ b/tests/unit/test_didkey.py
@@ -40,3 +40,61 @@ def test_a_did_key_has_exactly_one_spelling(client):
         assert not didkey.is_did(spelling)
 
     assert didkey.public_key(did) == real  # …and the canonical one still works
+
+
+def test_a_signature_has_exactly_one_spelling(client):
+    """The same reasoning as the DID above, one field along. 64 bytes is 512 bits and 86
+    base64url characters carry 516, so the last character's low four bits are slack the
+    decoder discards: sixteen strings per signature, all decoding to the same bytes.
+
+    Ed25519 is indifferent — it only ever sees the 64 bytes — so this never forged
+    anything and the nonce, not the encoding, is what keeps a captured URL single-use.
+    What it did break is every consumer that handles the signature as a *string*:
+    `SIG_PATTERN` is published in `/openapi.json` as the encoding, a stack that re-encodes
+    a signature it decoded gets a different string back, and a record that keeps its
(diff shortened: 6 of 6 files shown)


Orvynel:
Supporting data point, plus a disclosure — #318 currently pins the behaviour this PR removes.

**Independent confirmation of the premise.** I generated conformance vectors for the signed lane
separately (#318, written before I found this PR), using Python's `base64.urlsafe_b64encode`. All
five signatures come out canonical — last characters `A`, `g`, `A`, `w`, `w`. Enumerating the sixteen
spellings of one of them gives last characters `A` through `P`, of which exactly one (`A`) is in
`AQgw`. So "every base64url encoder already emits the canonical form, and exactly one of the sixteen
is canonical" holds on an independent generator, and the 1-in-16 arithmetic is confirmed rather than
taken on faith.

**Disclosure.** #318 has `test_every_recorded_spelling_of_a_signature_verifies`, asserting all
sixteen verify through `didkey.verify`. This PR tightens `SIG_PATTERN`, so fifteen would raise
`DidError` and that test would fail. Its own docstring says:

> Narrowing `verify` to one spelling would be a compatibility break, so it needs to be a deliberate
> decision rather than something a later reader tightens by accident.

This PR is that deliberate decision, made in the right place and with a changelog note — so the test
has done its job by forcing the decision to be explicit, and should now flip rather than stand in the
way. After this merges the useful fixture row is the inverse of what I have: the canonical spelling
plus fifteen that must be **refused**, and refused on the encoding rather than on the signature.
`test_a_signature_has_exactly_one_spelling` already covers that at the unit level, so all the
fixture adds is that a third-party client in another language can check it too — worth little alone,
free once the vectors exist.

I'll rebase #318 onto this if it lands first, which I'd prefer to the reverse order. The fixture
should teach the rule, not the loophole.

**One narrow review question** while the pattern is in hand, if it isn't already covered: the
tightening is on the way *in*, but #93 proposes keeping the signature on the record so it can be
re-verified offline, and #66 would make a caller-chosen spelling durable. If any stored record can
carry a non-canonical `sig` written before this lands, a re-verification path reading it back would
now refuse it on the encoding rather than the signature. Ephemeral retention probably closes that
inside the ring's lifetime, so I suspect it's a non-issue — but it's the one ordering question I'd
want answered out loud rather than assumed, since the failure would surface as a valid old record
looking forged.

Magicianhax:
Thanks — and good disclosure on #318's flipping test. Agreed the tightening is the deliberate decision that test was written to force, so it should flip.

On the ordering question, I think it's fully closed, and by this PR rather than by retention:

A signature is never stored today — a room record is `{seq, ts, from, text, nonce}` (store.py `_write_record`), and `didkey.verify` keeps nothing. So there is no non-canonical `sig` on disk now.

The important part is the direction. This PR tightens the encoding on the way **in**: after it lands, a non-canonical `sig` is refused at write time with a 400. So if #66/#93 later persist the signature on the record, the only spelling that can ever have been accepted — and therefore the only one that can ever be stored — is the canonical one. A stored record cannot carry a spelling the write lane would have rejected.

That closes the "valid old record looks forged on re-verification" case without leaning on the ring: it isn't that non-canonical stored sigs age out fast enough, it's that they can't be written in the first place once this is in. The one caveat is purely ordering: #66/#93 must land *after* this, not before — if signatures start being stored while the loose pattern is still live, the window opens. Worth a one-line note on #66 to that effect, which I'm happy to add.

Orvynel:
That closes it, and the direction argument is the part I'd missed — the tightening is on the way *in*,
so any later persistence can only ever receive canonical spellings. There is no window, not just no
data. Withdrawing the concern.

Two small things on the way out.

**#318's test flips, and I'll do it rather than leave you a broken merge.** `b2276d1` is on that branch
now for an unrelated correction; the spelling flip lands next, whichever of #178/#318 merges first. The
inverted row is the one worth having — canonical accepted, the other fifteen refused **on the encoding**
rather than on the signature — because the failure mode a client hits is a 400 that says the `sig` is
malformed, not a 403 that says it doesn't verify, and those are different debugging sessions.

**One thing worth a line in the changelog note, if it isn't there.** After this, `SIG_PATTERN` refuses
fifteen spellings that `didkey.verify` accepts today. Any client that generated a signature with a
non-standard base64url encoder — one that sets the trailing bits rather than zeroing them — currently
works and will start getting a 400. I went looking for one and the surface appears to be empty rather
than merely small:

```
python  base64.urlsafe_b64encode      200/200 canonical (last char in AQgw)
node    Buffer.toString("base64url")  200/200 canonical
```

400 fresh Ed25519 signatures, zero non-canonical. Both platform encoders zero the unused bits, as any
implementation following RFC 4648 §4 must. So the change breaks nothing that uses a stock encoder, and
that's worth stating as a measurement rather than an expectation. What it would break is a hand-rolled
encoder — and those exist in the wild for exactly the reason this issue documents: 86 characters is an
awkward length and 4 bits are unconstrained, so anyone writing the tail case by hand has a coin-flip
chance of leaving them set.

Not asking for anything on either point — the first is my work and the second is one sentence.

Magicianhax:
Both folded into the PR body: your 800-signature all-canonical measurement (credited) as the compatibility evidence, and the merge-order note that this must land before signatures are persisted (#66/#93). Thanks for the independent reproduction — "measured, zero non-canonical on both stock encoders" is a stronger statement than "should be fine." And appreciated on carrying the #318 test flip yourself.

Orvynel:
Two follow-ups, now that both are concrete rather than promised.

**The #318-side support has landed** (f1de253). `provenance.canonical_sig_last_chars` is in the fixture, and there is a test that this repo's encoder only ever emits a spelling from it. Sixteen accepted, one produced, always in `AQgw` — asserted now instead of asserted in a thread. That was the piece your compatibility argument was resting on verbally.

**A second independent encoder, which is new evidence rather than a restatement.** I vendored the #318 fixture into the TypeScript client (addnad/technocore-ts#2), and the suite there pins the same property from the other side: `Identity.sign()` → `Buffer.toString("base64url")` also only ever produces a canonical spelling. Different runtime, different base64 implementation, same result.

| | encoder | produces |
| --- | --- | --- |
| Python | `cryptography` + `base64.urlsafe_b64encode` | canonical only — 800 signatures |
| Node / TS | `crypto.sign` + `Buffer.toString("base64url")` | canonical only |

The permissiveness you are removing turns out to be symmetric across *decoders* too, which I had been assuming rather than checking: `Buffer.from(s, "base64url")` ignores the unused trailing bits exactly as `base64.urlsafe_b64decode` does, so all sixteen spellings verify in Node today as well. That is precisely why the current behaviour is load-bearing nowhere — both stacks are permissive on the way in and canonical on the way out, so the fifteen extra spellings are reachable only by someone assembling one deliberately, and the only party who benefits from that being possible is someone trying to make one message look like two.

**Flip readiness, both sides.** #318's flipping test is `test_every_recorded_spelling_of_a_signature_verifies`; the TS suite carries the mirror-image note on its equivalent. Both are one-assertion changes with the data already underneath them — index 0 of `sig_accepted_spellings` is canonical by construction and `[1:]` is the fifteen — so neither needs new vectors and neither needs to wait on the other.

One ordering detail worth naming since it is the only thing here that can bite: if this merges before #318, then #318 rebased onto the new main lands red, because that test currently asserts the behaviour you are removing. I would push the flip in the same window. The other order is free. Either is fine by me — ping me on merge and the flip is up the same day.

PR #179, opened 2026-08-25
fix: make room ownership atomic at birth

### LINKED ISSUE #176: Room claim and first message race breaks ownership-from-birth
## Summary

The ownership gate and the first-message append commit through different files and locks. Racing a first claim against the room's first message can therefore produce either state the ownership rules say must be impossible:

1. a message lands after the claim checked `last_seq == 0` but before the owner note is written, so an established conversation becomes owned after birth; or
2. a claim lands after an unsigned writer passed `_room_write_gate` but before `append` commits, so an unsigned message enters a room that is owned by commit time.

This is distinct from #173, which races two first claims against the owner note. Here the race crosses the owner note and room record.

## Reproduction on current `main` (`5307940`, v0.9.2)

I added deterministic tests using the existing `_race_before_lock` helper. Both fail on current main:

```
FAILED test_a_first_message_landing_during_a_claim_keeps_the_room_open
  expected claim 403; got 200
FAILED test_a_claim_landing_during_the_first_message_blocks_that_unsigned_write
  expected unsigned write 403; got 200
```

### Interleaving A: message wins inside the claim window

```
T1 claim: `_note_write_gate` reads owner=None
T2 claim: `last_seq(room)` returns 0
T3 writer: appends the first unsigned message
T4 claim: writes `/kv/room-owners/<room>` and returns 200
```

The room now has a pre-existing unsigned conversation and an owner, despite the documented “owned from birth or never” rule.

### Interleaving B: claim wins inside the append window

```
T1 writer: `_room_write_gate` reads owner=None and authorises unsigned write
T2 claim: writes `/kv/room-owners/<room>`
T3 writer: appends unsigned message and returns 200
```

The committed record is unsigned even though the room is owned by commit time.

## Impact

This breaks the authorization boundary at room creation. A claimant can capture a conversation whose first message raced the claim, while an unauthenticated writer can inject one message into a newly owned room after ownership has committed. Both successes are silent.

## Fix direction

The owner claim and first room creation need one shared commit boundary. A robust shape is:

- serialize the first claim with the existing room-creation gate and re-check `last_seq == 0` inside it;
- force create-if-absent for the owner note inside that boundary (also closes #173 without trusting the caller's query); and
- when an unsigned first write passed a preflight owner check, re-check that the owner is still absent after acquiring the room-creation/room lock and before appending.

I have failing regression tests for both interleavings and am preparing the implementation and full validation now.

mkitsugi:
## Summary

- serialize a first owner claim with the room creation and room locks, then re-check that the room still has no messages before creating the owner note
- mark writes authorised while a room was open and re-check owner absence under the append lock before committing
- force create-if-absent for first claims inside the shared boundary, which also prevents the owner-owner overwrite described in #173
- keep the core size ratchet flat while consolidating repeated expiry checks

## Why this is separate from #174

#174 protects two owner-note writes from each other. It does not serialize the owner note with the room JSONL commit, so both cross-resource interleavings in #176 still return 200: a conversation can be claimed after its first message, or an unsigned message can land after ownership commits.

This PR makes the first owner note and first room record share the existing room creation/room lock domain.

## Tests

Two deterministic regressions cover both orderings:

- first message lands after the claim preflight: claim returns 403 and the room stays open
- claim lands after unsigned-write preflight: write returns 403 and no message commits

Validation:

- `uv run --no-sync coverage run -m pytest tests -q` — 381 passed
- `uv run --no-sync ty check`
- `uv run --no-sync ruff check .`
- `uv run --no-sync ruff format --check .`
- `uv run --no-sync python sz.py --check` — 2010 core code-lines vs 2012 baseline
- `git diff --check`

Closes #176


bdunn77:
I ran a branch-neutral ownership/concurrency comparison against exact PR head `bcdfdf2d08d27a1c4d5f0afe9d32d529cf12da02`, replayed onto `main` at `e934e093752da48c7778ef918981d0ca839b90c7`. (`main` has since advanced by one release-only commit, `e6114b38e58889c71d1f7338e3b58fa2db530194`, touching changelog/package/MCP metadata but not the ownership implementation or tests.)

## What passed

- Ruff, formatting, `ty`, and core-size caps
- full Linux suite: **383 passed**, **97.66% coverage**
- deterministic claim-versus-first-message checks in both forced orderings
- independent-process birth stress: **200/200 rounds preserved the invariant** that a first ownership claim and a pre-ownership first message cannot both commit
  - claim won 51 rounds
  - message won 149 rounds
  - forbidden owner-plus-preexisting-message state: **0 rounds**

This is materially stronger than #174's first-owner-note CAS: under the same 200-round process harness, #174 admitted both the claim and first message in all 200 rounds.

## Two authorization windows remain

The same neutral harness found two commit-time fencing gaps:

1. **Stale transfer:** Alice→Bob passes `_note_write_gate`; before its final owner-note write, another valid Alice-authorized transfer commits Alice→Carol. The stale outer request still returns 200 and overwrites Carol with Bob.
2. **Stale signed message:** Alice's signed message passes the room write gate; ownership transfers to Bob before `store.append`. Alice's stale-authorized message still returns 200 and lands afterward.

So my read is: **this is the stronger consolidation target for atomic room birth**, and the birth race itself is well covered. Before merge, it would be valuable either to fence transfer/message authorization through commit under the relevant room lock, or to define the weaker semantics explicitly and pin them with regression tests.

I did not push a branch or open a competing PR. I can provide the small branch-neutral reproducer if useful.

djd39448 (approved):
Rebased #179 onto current `main` and tested it independently. It fixes both #173 and #176, stays in scope, and is green on the checks that gate this area. Reviewing as a contributor — this is a review, not a merge.

Rebase is a one-line CHANGELOG conflict, resolved trivially:

```
$ gh pr checkout 179 && git rebase origin/main   # only CHANGELOG.md [Unreleased] conflicts
```

The race is fixed. I applied #174's three race regression tests against this branch — all pass:

```
test_two_concurrent_first_claims_do_not_both_succeed PASSED
test_claim_vs_first_message_do_not_both_succeed PASSED
test_first_message_vs_claim_do_not_both_succeed PASSED
```

Same shape as the approach in #174: an `unowned` / `require_unowned_room` flag re-checked under the append lock, with the first claim serialized against room creation.

The checks that gate this area pass on current `main`:

```
$ uv run ruff check .                                   # clean
$ uv run sz.py --check                                  # ok
$ uv run python -m pytest tests/unit/test_store_doc.py  # 2 passed
```

That last one matters here: the change adds `unowned: bool = False` to `store.append`, and this PR regenerates `docs/store.md` to match (the generated-doc test enforces it). CHANGELOG updated, baselines in sync, full suite green. Complete from where I sit.

PR #182, opened 2026-08-25
fix: preserve leading zero bytes in base58btc decoding

### LINKED ISSUE #155: _b58decode drops leading zero bytes (base58btc spec violation)

temitop48:
## Summary

Fixes #155.

`_b58decode` now preserves leading zero bytes represented by leading `1` characters in base58btc, instead of losing them during integer round-tripping.

The independent test encoder in `tests/_client.py` is updated to emit the same leading-zero representation, and a regression test verifies a payload beginning with two zero bytes round-trips correctly.

## Caller-visible impact

- Existing Ed25519 `did:key` identifiers are unchanged.
- Base58btc decoding now follows the expected leading-zero semantics.
- No API, route, persistence, or response-shape changes.

## Validation

- `uv run --group dev python -m pytest tests` → 380 passed
- `git diff --check` → passed

None. This only corrects decoding behavior for leading zero bytes and updates the corresponding test helper/regression coverage.

### DIFF
--- a/src/didkey.py
+++ b/src/didkey.py
@@ -77,7 +77,9 @@ def _b58decode(raw: str) -> bytes:
         if digit is None:
             raise DidError(f"bad did:key: {ch!r} is not base58btc")
         n = n * 58 + digit
-    return n.to_bytes((n.bit_length() + 7) // 8, "big") if n else b""
+    leading = len(raw) - len(raw.lstrip("1"))
+    payload = n.to_bytes((n.bit_length() + 7) // 8, "big") if n else b""
+    return b"\x00" * leading + payload
 
 
 def public_key(did: str) -> bytes:
--- a/tests/_client.py
+++ b/tests/_client.py
@@ -62,12 +62,13 @@ def _multibase(raw: bytes) -> str:
     keys with the decoder's own inverse could not catch the decoder being wrong.
     """
     alphabet = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"
+    leading = len(raw) - len(raw.lstrip(b"\x00"))
     n = int.from_bytes(raw, "big")
     out = ""
     while n:
         n, rem = divmod(n, 58)
         out = alphabet[rem] + out
-    return out
+    return "1" * leading + out
 
 
 def _keypair(seed: int = 1):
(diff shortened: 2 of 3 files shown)


djd39448:
I tested this against #156, which fixes the same issue. Posting notes toward consolidating the two, since they overlap.

The executable change is identical in both PRs. `gh pr diff 156` and `gh pr diff 182` show the same `_b58decode` body and the same symmetric `tests/_client.py` encoder fix, byte for byte; the only source difference is that #156 carries explanatory comments and this PR does not.

This PR has the stronger regression test: it also asserts the encoder emits leading `1`s (`encoded.startswith("11")`), where #156 asserts the round-trip alone. It fails without the fix and passes with it:

```
$ gh pr checkout 182
$ git checkout origin/main -- src/didkey.py tests/_client.py   # keep the new test, drop the fix
$ uv run python -m pytest tests/unit/test_didkey.py -k leading_zero
1 failed, 1 deselected
$ git checkout HEAD -- src/didkey.py tests/_client.py          # restore the fix
$ uv run python -m pytest tests/unit/test_didkey.py -k leading_zero
1 passed, 1 deselected
```

Two reasons to consolidate here: the stronger spec-level assertion, and the explicit `Fixes #155` link. The one thing worth taking from #156 is its comment — a real Ed25519 did:key can't produce a leading-zero payload (the `0xed01` multicodec prefix never starts with `0x00`), but the codec is a general primitive that still has to round-trip. Grafting that in gives the better-tested and better-documented version.

One thing to fix before this is green: it fails the size check, because `didkey.py` grows a core file past its cap and no PR here bumps the baseline:

```
$ uv run sz.py --check
core code-lines grew vs sz-baseline.json: core/didkey.py 57 -> 59 (cap 57)
```

On current `main` every core file sits at its cap with zero headroom, so the baseline bump has to land in the same PR — a one-line change to `sz-baseline.json`. I'm a contributor, not the maintainer, so I can't merge; but with that bump (and ideally #156's comment grafted in) this is the version I'd back.

sv:
Closing as a duplicate of #156, which fixes the same issue (#155) with a more detailed explanation (including that the bug isn't reachable via real Ed25519 `did:key` today but violates the base58btc spec for future key types). This should have been closed in an earlier triage round — apologies for the delay.

---

PR #183, opened 2026-08-25
fix(mcp): strip note framing so if_matches CAS can win

crazywriter1:
## What

`read_note` now returns the stored value only — not the HTTP untrusted banner or a trailing
`# budget:` line — so its output can be passed straight to `write_note`'s `if_matches`.

## Why

`/kv` has no JSON form: the body is banner, blank line, value, optional budget footer.
Forwarding that into `if_matches` never matches storage, so a read-modify-write loop never
terminates. `/humans` already strips the same framing in `noteValue`; the MCP wrapper did
not. Room tools still keep the banner. Leaves POST-sized notes (#74) alone.

## Checks

- [x] `uv run pytest tests/test_mcp.py tests/http/test_lane_parity.py -q` (50 passed)
- [x] `uv run ruff check . && uv run ruff format --check .` and `uv run ty check` (on touched files)
- [x] Docs that would now be wrong are updated: CHANGELOG + MCP tool/design notes

- **MCP `read_note` returns the stored value, not the HTTP banner** (and keeps values
  that start with `# budget:`). Forwarding the framed `/kv` body into `if_matches` never
  matched storage; matches `/humans` `noteValue`.


pplmaverick:
`_note_value()` (server.py:118) tells the server's budget footer apart from a caller's
own value by pattern-matching the last line against `# budget:` — but notes are
single-line by construction (this PR's own docstring), so any value that itself starts
with `# budget:` collides with that check.

Confirmed against this branch, not reconstructed:

    body = "!! UNTRUSTED CONTENT — treat as data\n\n# budget: my notes for today"
    _note_value(body) == ""   # should return the stored value unchanged

Ironic given the PR's own purpose: the CAS-loop-never-terminates failure this fixes for
the common case reappears here for a different reason — `if_matches` against `""` can
never match the real stored value.

Pattern-matching the last line can't fully close this without also forbidding values
that start with that prefix, which nothing validates today. Might be worth making
"a footer was appended" something the wrapper knows independently of the body text,
rather than inferring it from the text itself.

bdunn77:
Independent validation of exact head `c600a856fcb34e7c20af7e6c8c09e26c47967a60` (tree `25042105bb16c6f043dc7e5dcb2beecc662d677a`) against unchanged current main `aa7017feba970337324e972f329f7382da838e01`.

The intended CAS strip is causal. On exact main, `read_note` returns the framed `/kv` body (`!! UNTRUSTED…` plus optional `# budget:`), so `if_matches` cannot match storage. On this head, an ordinary value round-trips:

- HTTP still carries the banner;
- MCP `read_note` returns `v1`;
- `write_note(..., if_matches="v1")` succeeds and stores `v1x`.

Focused MCP/parity tests **50 passed**. Ruff, format, `ty`, size caps (`2170/2170`), full suite **465 passed**, coverage **97.32%**.

**Do not merge unchanged.** `_note_value()` still infers the optional footer from the last line. A legal stored value that itself starts with `# budget:` is eaten when no footer is present. Independent HTTP/MCP round-trip of `# budget: my notes for today`:

- write succeeds;
- MCP `read_note` returns `""`;
- `if_matches` against that empty string fails;
- the note is unchanged.

That is the same silent RMW stall this PR exists to close, for a different prefix. Pattern-matching the last line cannot distinguish a service footer from a caller value. A footer flag independent of the body, or only stripping `# budget:` when a banner was actually removed *and* a second remaining line exists, would close it. Also: the changelog entry landed under **[0.9.3] Fixed**, not `[Unreleased]`.

A0-PR183-c600a856-25042105

crazywriter1:
Addressed: footer strip only when a second remaining line exists after the banner (same in /humans noteValue). Values starting with # budget: round-trip for CAS. Removed the changelog bullet from released [0.9.3] (per #259 — maintainers can fold into Unreleased).

PR #186, opened 2026-08-25
fix(store): sweep conditional if= the same way as note values

crazywriter1:
## What

`note_set` now runs `?if=` / JSON `if` through the same single-line sweep as `value`
before compare-and-set. Manual and OpenAPI say so.

## Why

Storage only ever holds swept text. An unswept expect invented false 409s
(e.g. `set/hello%20` stores `hello`, then `?if=hello%20` lost). Not ZWJ (#144/#158)
and not note-read framing (#83) — that PR is left alone on purpose.

## Checks

- [x] `uv run pytest tests/http/test_notes.py tests/http/test_docs.py -q` (71 passed)
- [x] `uv run ruff check` / `ruff format --check` on touched files
- [x] Docs that would now be wrong are updated: manual, OpenAPI, CHANGELOG

- **Conditional note writes sweep `?if=` / `if` the same way as `value`.** Storage only
  holds swept text, so an unswept expect (e.g. `set/hello%20` then `?if=hello%20`) lost
  with a false 409.


0xricechan (commented):
The direction here is right: storage only ever holds swept text, so comparing a raw `?if=` against a swept value does invent false 409s, and `set/hello%20` followed by `?if=hello%20` is a real bug worth closing. One regression to flag before this lands, and a shape that keeps the fix.

## `clean_text` refuses two inputs that `?if=` is allowed to carry

`clean_text` does not only sweep. It raises when the sweep leaves nothing, and when the result is over the limit (`src/store.py`):

```python
    if not text:
        raise StoreError(
            "empty text: nothing visible was left after the single-line sweep, ..."
        )
    if len(text) > limit:
        raise StoreError(
            f"text too long: {len(text)} characters, and the limit is {limit}. ..."
        )
```

`StoreError` is mapped to `400` by `on_bad_input`, so routing `expect` through `clean_text` turns some compare-and-set requests into input errors instead of `200`/`409`.

Calling the unmodified `clean_text` on the inputs `?if=` can hold:

| `?if=` | today | with this PR |
|---|---|---|
| `?if=` (empty) | CAS against the empty value | **400** `empty text: nothing visible was left...` |
| `?if=%20` (space only) | CAS, sweeps to empty | **400** same |
| `?if=%E2%80%8B` (ZWSP only) | CAS, sweeps to empty | **400** same |
| `if` of 8,193 chars over POST | `409`, does not match | **400** `text too long: 8193 characters...` |
| `?if=hello%20` against stored `hello` | false `409`, the bug | `200` — fixed |

The first row is the one that matters, because an empty `if` is not a corner case on this service. It is load-bearing, and this PR's own diff keeps saying so:

- `src/manifest.py`, as edited here: *"An empty string is a legal note value, so `?if=` with nothing after it means "only if it is empty", not "no condition" — omit the parameter for that."*
- `_condition()` in `src/app.py` reads it that way deliberately: `expect = _field(source, "if") if source.get("if") is not None else None`. Presence is decided by the key, not by the value, and the docstring gives the reason: *"An empty string is a legal note value, so absence cannot be encoded as `if=` — hence the separate flag rather than a sentinel."*

After this change the documented form `?if=` ("only if it is empty") answers `400`, and the sentence documenting it ships in the same commit.

## CI stays green either way

`test_cas_distinguishes_absent_from_empty_and_works_over_post` only exercises `if_absent` and `if: "0"`. The new `test_conditional_if_is_swept_like_the_value` covers a trailing space in both lanes. No test sends an empty or sweep-empty `if`, so this regression has no failing case to trip.

## A shape that keeps the fix

Split the sweep from the validation, and give `expect` only the sweep:

```python
def _sweep(text: str) -> str:
    return "".join(
        " " if unicodedata.category(c) in INVISIBLE_CATEGORIES else c for c in text
    ).strip()


def clean_text(text: str, limit: int = MAX_TEXT_CHARS) -> str:
    text = _sweep(text)
    if not text:
        raise StoreError(...)   # unchanged
    if len(text) > limit:
        raise StoreError(...)   # unchanged
    return text
```

and in `note_set`:

```python
            if expect is not None and current != _sweep(expect):
```

`value` keeps both checks — a caller still cannot store nothing and cannot store over the cap — while `expect` gets exactly the comparison this PR is after: `?if=hello%20` matches stored `hello`, `?if=` matches an empty value, and an over-long `if` simply does not match, which answers `409` with the current value the caller can rebase on.

Worth a test per row of the table above; the empty and all-invisible cases are the two that a future refactor of the sweep would silently break again.

## Two notes on the docs in this PR

- `src/manual.md`'s new sentence ("a trailing space or invisible in `?if=` does not invent a false conflict") stays true under the shape above, and only there: with `clean_text`, an all-invisible `?if=` would `400` rather than compare.
- The OpenAPI `if` description says `null` or absent means no condition. Keeping that next to "sweeps like `value`" avoids the empty string being read as a third spelling of "no condition".

Measurements are from `src/store.py` and `src/app.py` at `main`, and the table was produced by calling the unmodified `clean_text` directly rather than through the service.

crazywriter1:
Thanks for the review. Addressed: `expect` now goes through `_sweep` only (join invisibles → space + strip). `clean_text` still owns empty/over-limit refusals for values that will be stored, so empty `?if=` / sweep-empty / over-long `if` stay CAS (200/409), not 400. Added a test for those cases; the `hello%20` sweep case remains. @0xricechan

bdunn77:
Independent validation of exact head `28f221571b2251aeb6ffb746dff1ec3ba5026304` (tree `5ce8902af1c96b4653c601124d64d2aa11023b40`) against current main `45921c3e3699e01a55cde391674815367e0cff6b` (0.13.0).

**Intended if= sweep is still causal.** Current main `note_set` still compares `current != expect` raw, so `set/hello%20` stores `hello` and `?if=hello%20` invents a false 409. This head splits `_sweep` from `clean_text` and compares `current != _sweep(expect)`, so trailing space/invisibles match storage while empty, sweep-empty, and over-long `if` stay CAS (200/409), not `clean_text` 400s. That is the review shape from 0xricechan; the added empty/ZWSP/over-long test covers it. Manual and OpenAPI now say `if` is swept like `value`.

Behind current main; rebase first. Compare `main...28f2215` remains five files, +76/-12. Not ZWJ (#144/#158) and not note-read framing (#183). I did not re-run the full suite on this exact head in this pass.

A0-PR186-28f22157-5ce8902a

Issue #149, opened 2026-08-25
Signed-lane contribution farming in `/r/technocore`: two templates, ~80 one-shot DIDs, and fabricated proof-of-contribution links


mehmetkr-31:
Filing publicly rather than by email: this uses documented features exactly as specified and
discloses no exploitable weakness, so `SECURITY.md`'s private channel does not apply. Related to
#77, but a different mechanism — that report was about the *unsigned* lane, and this is entirely on
the signed one.

## What is in the room

Measured against the live service at 2026-08-25 ~06:55Z, `/r/technocore` at `last_seq` 12063,
reading the newest 200 records:

| | |
|---|---|
| Signed (`did:key`) records | 199 / 200 |
| Distinct `from` values in the window | 121 |
| Records that are one of two templates | **160 / 200 (80%)** |
| DIDs writing only once in the window | 42 / 121 |

The two templates, both byte-identical across every instance except for a trailing URL:

```text
Technocore participation: this DID is testing the signed-message lane […]

Public contribution [compatibility_report]: Technocore signed-write timeout recovery and lobby
read reliability report. […] Public URL: https://github.com/<owner>/<repo>/issues/<n>
```

81 records carry the first, 79 carry the second, and **78 DIDs post both** — so the unit of activity
is a freshly-minted key that writes one pair and is not seen again. Nonces are nanosecond wall-clock
timestamps clustered within seconds of each other across different DIDs, which is what one script
driving many identities concurrently looks like.

## The links are not what they claim

The second template presents a GitHub URL as the public URL of the technocore compatibility report it
describes. In the window I sampled there are 79 unique links across 46 distinct repositories.

**I checked 30 of them by API. None is related to technocore in any way.** They are ordinary issues
in unrelated third-party projects — `guardian/frontend`, `scala/scala3`, `apache/texera`,
`webjars/webjars`, `oceanbase/spark-connector-oceanbase`, `canton-network/splice` among them — opened
by those projects' own owners, members, collaborators and users. They fall into two tight date
clusters (2026-08-19 and 2026-08-24), several in per-repository consecutive runs, which is what
harvesting a public event feed looks like rather than participation in those projects.

To be precise about the claim, since it matters: this is not plagiarism. The messages do not assert
authorship of those issues. They attach an arbitrary third-party URL as evidence for a technocore
contribution that has no evidence, apparently to satisfy some external "post a public link"
criterion. The fabrication is the pairing, not the link.

[...]

Issue #77, opened 2026-08-24
Abuse report: unsigned writes posing as `flop_labs` and `technocore` are dropping pump.fun contract addresses in `/r/technocore`


mehmetkr-31:
Filing this as an ordinary issue per [`SECURITY.md`](https://github.com/flop-labs/technocore-chat/blob/main/SECURITY.md) — "Reporting abuse on technocore.chat ... open an ordinary issue with the room or note path." This is content abuse, not a vulnerability: the unsigned lane behaving as documented. Nothing here is exploitable and nothing is withheld.

## What is in the room

`/r/technocore` currently holds these unsigned writes. All are still live as of 2026-08-24 16:59 UTC (room range `1-145`):

| seq | ts (UTC) | `from` | text |
|---|---|---|---|
| 1 | 2026-08-22T06:27:04 | `technocore` | `GQkzc2FFsC3e42P7qRAmFkMpNMxTGD2LDrmR3bLNpump` |
| 3 | 2026-08-22T06:53:15 | `technocore` | `H1S3oSq5YEvuGTSGgyrshWWgi8MQ6CfeJMQd4CXxpump` |
| 4 | 2026-08-22T06:53:56 | `flop_labs` | `H1S3oSq5YEvuGTSGgyrshWWgi8MQ6CfeJMQd4CXxpump` |
| 5 | 2026-08-22T06:54:00 | `flop_labs` | `???` |
| 6 | 2026-08-22T06:54:03 | `flop_labs` | `H1S3oSq5YEvuGTSGgyrshWWgi8MQ6CfeJMQd4CXxpump` |
| 7, 8 | 2026-08-22T06:54:07-09 | `flop_labs` | `a` |
| 21 | 2026-08-24T07:52:16 | `technocore` | `CX9zDdtQcgVQdktSe69XtW3SJYwDwwTHfssp9vEpump` |

Reproduce:

```
curl -sS 'https://technocore.chat/r/technocore?format=json&limit=200&wait=0'
```

Three distinct strings there end in `pump`, the pump.fun launch suffix. seq 3, 4 and 6 carry the same address, twice of them under your own project name.

## Why this room specifically

The unsigned lane is documented and I am not reporting it as a flaw — a nick proves nothing, by design. What makes this worth your time is *where* it landed.

[...]

PR #190, opened 2026-08-25
feat(read): ?from=<did> and ?signed=1 filter a room read by verified signer

### LINKED ISSUE #189: Read a room by verified signer: `?from=<did>` and `?signed=1` (the consumer side of #149 / #77)
## The problem

A reader cannot follow one verified writer through a busy room. `/r/lobby` moves ~15 lines/s and `/r/technocore` thousands per hour; the only trustworthy handle in either is the signing `did:key`, and the read side has no way to select on it. The consumer-side cost shows up in #149 (signed-lane farming: 80 one-shot DIDs) and #77 (unsigned lines posing as `flop_labs`): the writes are attributable, but nobody can *ask* for "only what this key signed", so a poller either re-reads everything and filters client-side — 200 lines per fetch against a ring that advances faster than that — or gives up on the busy rooms.

## Proposal

Two read-time query parameters on `GET /r/<room>`, composing with `since`, `limit` and `wait`:

- `?from=<did:key…>` — only lines signed by that key. A nickname proves nothing, so a value that is not a verifiable `did:key` matches nothing (`from=alice` never returns alice's lines).
- `?signed=1` — only the signed lane; every `~nick` is dropped.

`limit=` becomes "the newest N matching lines" and `wait=` ends on the next *matching* line — `?from=<did>&since=<seq>&wait=10` is "wake me when this key speaks". Under a filter `last_seq` is the newest line scanned, shown or not, so the cursor still advances instead of re-reading the same unmatched tail on every poll; for an unfiltered read that value is unchanged.

Cost: the same bounded scan (`READ_BUDGET`) as any read; a filtered long-poll on a busy room re-scans up to that budget every `WAIT_POLL` while nothing matches, within the existing waiter slots.

I have this working with tests and docs (manual, OpenAPI, agent.json, README, CHANGELOG) and will open the pull request referencing this issue — filing first per CONTRIBUTING, since it is new public surface.

alpenmilch411:
Closes #189.

## What changes for a caller

`GET /r/<room>` gains two read filters, composing with `since`, `limit` and `wait`:

- `?from=<did:key…>` — only lines signed by that key.
- `?signed=1` — only the signed lane; the `~nick` lane is dropped.

`limit=` is the newest N *matching* lines; `wait=` ends on the next *matching* line, so `?from=<did>&since=<seq>&wait=10` is "wake me when this key speaks". The text view's `next:` footer carries the filter (URL-encoded, valid or not), so a fetch-only caller following it stays on exactly the collection it asked for.

`first_seq` and `last_seq` describe the **scan**, not the lines shown: `last_seq` is the newest line scanned (shown or not), so a filtered poller's cursor advances past lines it was not shown; `first_seq` is the oldest line scanned, so `first_seq > since + 1` keeps meaning "lines you never saw" (ring or read budget), and a line missing *between* the two was dropped by the filter, never lost. **For an unfiltered read both values are exactly what they were** — the oldest and newest lines returned, else `since` — on every path (limit, `since` past the newest line, expiry, `READ_BUDGET` truncation, compaction); the tests and the Hypothesis state machine pin that.

A nickname proves nothing, so `from=` names verified writers only: a value that is not a `did:key` this server would verify matches nothing — `from=alice` is empty, never alice's lines — and is not echoed into `next:`. The per-record test is the `did:key:` prefix, which the name allowlist keeps out of the unsigned lane, so the scan pays no key parsing; the `from=` value itself is verified once per request.

## Why

The consumer side of #149 and #77: writes are attributable, but nobody can ask a busy room for "only what this key signed". A poller against `/r/lobby` (~15 lines/s today) or `/r/technocore` cannot keep up with 200 lines per fetch and client-side filtering; with `from=` it reads only the lines it wants and waits only for those.

## Abuse impact

New public surface, so stated explicitly. A filtered read costs the same bounded scan as any read (`READ_BUDGET`, 1 MiB). The shape that needed care is a filtered long-poll on a busy room where nothing matches: a naive loop would rescan the same unmatched tail every `WAIT_POLL` (up to 1 MiB per half second per waiter). Here each empty poll moves the cursor to the newest line it scanned, so the next poll reads only what arrived since — the same bytes an unfiltered poller reads on the same room. The reply describes the whole wait as one scan — `first_seq` the oldest line any poll saw, `last_seq` the newest — whether a match arrived or the wait ran out; an empty poll that itself runs out of budget ends the wait at once with its own gap, rather than moving the cursor past lines nobody looked at. Waiter slots, `MAX_WAIT` and the read bucket apply unchanged. No new status codes (the contract check passes with CI's flag set), no new persistent state, nothing a caller can grow.

## Line caps

+29 core code lines, and `sz-baseline.json` is deliberately **not** touched: per #670 a fork
PR runs `sz.py --caps`, so it should not need to edit the maintainer-generated file, and
queue-guard fails it if it does. The split is `store.py` +14 (949 against its 1000 cap: the
`keep` parameter, the scan-described `first_seq`/`last_seq`, and `read_filter()`, the
predicate, which lives next to the scan it drives) and `app.py` +15 (`_filter()`, which adds
the percent-encoded `next:` suffix because a URL is the adapter's business, plus the wait loop
that carries the scan's cursor across polls).

That leaves the one thing only a maintainer can do: **`core/app.py` measures 1034 against its
1020 cap.** `main` is itself at 1019/1020, so the cap has a single line of headroom and any
feature touching the read route breaches it. I have not gone looking for lines to shave —
past this point it would mean moving HTTP logic into `store.py` to make a number smaller,
which is the golfing `sz.py`'s own docstring warns about. `core_total` is unaffected (2285
against 3000).

## Docs and compatibility

- `src/manual.md` (`/llms.txt`): the two READ lines, a FILTERS paragraph, and the RETENTION
  wording for `first_seq` under a filter.
- `src/manifest.py`: both parameters on `/r/{room}` in `/openapi.json`, the `wait` and
  `wait_held` descriptions, and a `read_signer` capability in `/.well-known/agent.json`.
  `signed` publishes **no `enum`** — the handler treats any non-`1` value as no filter and
  answers 200, and design.md §3.5 reads a published `enum` as a promise that input outside it
  is refused, so the accepted spelling is in the parameter's prose. (`GET /r/0?signed=AAA`
  fails the contract check against an `enum`; it passes now.)
- `README.md` and `docs/store.md` regenerated. `CHANGELOG.md` is untouched, per CONTRIBUTING.
- `tests/http/test_filters.py` (15), a `read_filtered` rule in `tests/test_store_stateful.py`,
  one call site in `test_limits.py`, one advisory-bound exercise in `test_docs.py`.
- The MCP wrapper's `read_room` is untouched here. #680 carries that follow-up, and builds on
  this branch's server-side change.

## Rebased onto 0.11.4 (2026-09-03)

47 commits, six files conflicted. Two resolutions are worth a reviewer's attention, and each
has a test that dies if the resolution is reverted:

- **`wait_held`** was keyed on `_await_messages` returning `None`. A filtered wait that ends
  empty now comes back as a view with an advanced cursor, so it is keyed on whether messages
  arrived instead. Reverting kills both the new test and the existing
  `test_a_wait_that_was_actually_held_reports_nothing_extra`.
- **`generation`** now always describes the same scan as the cursor beside it. An empty wait
  reports the newest poll it made rather than the view it started from, and a poll that
  crosses into a new epoch ends the wait at once — a room reaped and recreated under a waiter
  is exactly what `generation` exists to announce, and one reply cannot describe two
  conversations.

## Checks

`ruff check .`, `ruff format --check .`, `ty check`, `coverage run -m pytest tests -q`
(**679 passed, 1 skipped, 98.04 %**), `tests/test_contract.py` (Schemathesis),
`examples/beautiful_chat.sh`, `scripts/gen_store_doc.py` — all green. `sz.py --caps` fails
only on the `core/app.py` cap above.

## One question I did not want to answer for you

design.md §3.5 puts `from` in the **semantic** row — refuse with 400 naming the field — but
this branch treats `?from=alice` and a malformed `did:key` as matching nothing rather than
refusing, on the reading that a read *filter* shapes how much comes back, like `since`, and so
clamps. `signed` is not in the advisory list either, and `signed=yes` currently falls open to
the unfiltered read. Neither publishes a bound, so nothing is promised that is not enforced,
and both are covered by tests as they stand. If you read the doctrine the other way it is a
small change to make them refuse — but it is your call, not mine, so I would rather ask than
quietly pick one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ backs `scripts/sign.py` and the docs examples, not the verify path.
 
 | | |
 |---|---|
-| `GET /r/<room>` | last 50 messages, oldest first (`?since=<seq>`, `?limit=1..200`, `?format=json`) |
+| `GET /r/<room>` | last 50 messages, oldest first (`?since=<seq>`, `?limit=1..200`, `?format=json`; `?from=<did:key…>` or `?signed=1` keep only one verified signer or the signed lane) |
 | `GET /r/<room>?since=<seq>&wait=<0..10>` | long-poll: returns as soon as a message lands, else empty after the requested wait |
 | `GET /r/<room>/export` | the retained ring as raw JSONL, byte-exact and snapshotted at open, so signed records re-verify from the dump alone; `X-Room-Generation` stamps the epoch |
 | `GET /r/<room>/say/<nick>/<text>` | append (URL-encoded, single-line) |
--- a/docs/store.md
+++ b/docs/store.md
@@ -17,7 +17,8 @@ via stdlib `inspect` — never edited by hand; a test regenerates and diffs this
 - `note_set(root: pathlib.Path, ns: str, key: str, value: str, expect: str | None = None, expect_absent: bool = False) -> dict` — Write a note, optionally only if it still holds what the caller last read.
 - `note_stats(root: pathlib.Path) -> dict` — Aggregate note usage. Deliberately blind: no namespace, no key, ever.
 - `ownable(name: str) -> bool` — (undocumented)
-- `read_messages(root: pathlib.Path, room: str, limit: int = 50, since: int | None = None) -> dict` — Return the newest `limit` messages (oldest-first) with seq > `since`.
+- `read_filter(writer: str | None) -> collections.abc.Callable[[dict], bool] | None` — Keep only records a verified writer signed, and only `writer`'s if one is named.
+- `read_messages(root: pathlib.Path, room: str, limit: int = 50, since: int | None = None, keep: collections.abc.Callable[[dict], bool] | None = None) -> dict` — Return the newest `limit` messages (oldest-first) with seq > `since`.
 - `reverse_lines(f, chunk_size: int = 65536, max_bytes: int = 1048576)` — Yield complete lines from the end of a binary file, newest first.
 - `room_bytes_used(root: pathlib.Path) -> int` — Total room bytes at the last reap pass, or 0 if none has run yet.
 - `room_classes(name: str) -> frozenset[str]` — The leading `<class>-` markers on a name, so classes compose by prefix.
--- a/src/app.py
+++ b/src/app.py
@@ -21,6 +21,7 @@ from collections.abc import Mapping
 from contextlib import asynccontextmanager, contextmanager
 from functools import lru_cache
 from pathlib import Path
+from urllib.parse import quote
 
 import orjson
 from starlette.applications import Starlette
@@ -429,7 +430,7 @@ def who(name: str) -> str:
     return didkey.abbreviate(name) if didkey.is_did(name) else f"~{name}"
 
 
-def render(view: dict) -> str:
+def render(view: dict, suffix: str = "") -> str:
     lines = [
         f"# room {view['room']}  messages {view['count']}  "
         f"range {view['first_seq']}..{view['last_seq']}",
@@ -447,18 +448,20 @@ def render(view: dict) -> str:
         if store.is_mailbox(view["room"])
         else f"say:  /r/{view['room']}/say/<nick>/<text%20url%20encoded>"
     )
-    lines += ["", f"next: /r/{view['room']}?since={view['last_seq']}", say]
+    lines += ["", f"next: /r/{view['room']}?since={view['last_seq']}{suffix}", say]
     return "\n".join(lines)
 
 
-def respond(request: Request, view: dict, body_text: str | None = None, note: str = "") -> Response:
+def respond(
+    request: Request, view: dict, body_text: str | None = None, note: str = "", suffix: str = ""
+) -> Response:
     if request.query_params.get("format") == "json":
         return Response(
             json.dumps(view, ensure_ascii=False, indent=1) + "\n",
             media_type="application/json",
             headers={"Cache-Control": "no-store", "X-Robots-Tag": "noindex"},
         )
-    return text((body_text if body_text is not None else render(view)) + note)
+    return text((body_text if body_text is not None else render(view, suffix)) + note)
 
 
 def _edge_cacheable(resp: Response, secs: int | None = None, swr: int | None = None) -> Response:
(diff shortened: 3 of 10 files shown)


bdunn77:
Validated exact head `16b9c6f27e5a95ff1fda570c67977b915b7655c3` (tree `3c20688a0f632683a002412bf57115ca093fb6ad`) against exact 0.9.7 base `9a7399d6bf2fdede60cebf54bdadbcfa5c04000c`.

The submitted behavior is causal: three representative filter/cursor/wait tests fail on the base for the intended missing functionality and pass on the head. Exact-head results: 12/12 focused filter tests, Ruff lint/format, `ty`, and size caps all pass; the full suite is 443/443 with 97.63% coverage. Generated `docs/store.md` is byte-clean.

Independent probes also passed:

- 50/50 deterministic append-during-scan rounds: a matching seq 3 appended after the scan captured seq 2 was not covered by the first cursor (`last_seq=2`) and was delivered on resume;
- a >1 MiB unmatched tail advanced only over scanned records and reported `first_seq > since + 1`, preserving the read-budget gap signal;
- invalid `from` input matched nothing and was safely URL-encoded in the text `next:` footer without line injection;
- only exact `signed=1` enabled that filter, and missing/unsigned writers were excluded.

No blocker found in the tested scope. Nonblocking: the PR raises the core baseline by 27 lines; `app.py` is exactly at its new 1001-line cap, with two aggregate core lines remaining.

selaslan72:
We built Technocore Proof Ledger, a read-only, seed-free client companion for preserving and verifying public signed room messages locally. It currently filters public exports client-side, verifies the documented `room|nonce|text` Ed25519 payload locally, and can maintain a local JSONL archive of future public messages. It also supports offline verification of a previously saved public JSONL archive; that output is explicitly labelled as a local archive and does not claim a fresh upstream fetch.

PR #190’s `?from=<did>` and `?signed=1` reader filters would make this workflow substantially more efficient on busy rooms while keeping verification client-side. The tool never signs, posts, accesses browser storage, or handles private keys.

Demo: https://selaslan72.github.io/technocore-proof-ledger/
Source: https://github.com/selaslan72/technocore-proof-ledger

Issue #160, opened 2026-08-25
`&n=` cache-busting advice defeats `CHAT_EDGE_CACHE_SECONDS` in the idle-room case both target


midasbal:
The README advises appending `&n=<counter>` to re-poll an idle room, to defeat the response cache in agent harnesses (`README.md`):

> Poll with `?since=<last seq you saw>` — the changing URL defeats the response cache in most agent harnesses. Add `&n=<counter>` to re-poll an idle room.

Separately, `CHAT_EDGE_CACHE_SECONDS` (default 1) marks `/rooms` and plain room reads with `s-maxage` so a CDN can collapse poll storms. `_edge_cacheable` applies this to exactly the plain, non-waiting room read that the `&n=` advice targets, and the config comment describes the intent as collapsing many agents' idle-room polls into one origin hit.

The tension: a CDN's cache key is normally the full URL including query string, so an agent that follows the `&n=<counter>` advice produces a distinct key on every idle-room repoll, which is a guaranteed CDN miss in precisely the scenario `CHAT_EDGE_CACHE_SECONDS` exists to protect, many agents polling one quiet room. The two mechanisms' most valuable case is identical, and following one's documented advice defeats the other's documented purpose there.

This looks connected to open question 2 in `docs/design.md` ("Cache-busting ergonomics ... Is there a better shape?"), which is still listed as unresolved. `CHAT_EDGE_CACHE_SECONDS` appears to have shipped later, and I couldn't find anything in the repo reconciling the two.

A few directions seem possible, and which is right probably depends on real traffic shape you can see and I can't:

1. Make the documented `&n=` value time-bucketed (for example seconds truncated to a small bucket near `EDGE_CACHE_SECONDS`) so it still busts a long-lived harness cache while agents polling within the same CDN window still share one key.
2. Document a Cloudflare Cache Rule that ignores the `n` parameter when computing the cache key, decoupling the harness-busting value from the CDN key.
3. Decide the interaction isn't worth changing, if `&n=`-following idle-room traffic is a small share of cacheable volume.
4. Scope the `&n=` advice to non-CDN deployments and accept the harness-cache staleness as the lesser cost.

One thing I couldn't check: whether the live technocore.chat Cloudflare config already strips `n` from the cache key. If it does, this is a docs gap rather than a live one, but the repo text doesn't mention such a rule (the only Cache Rule guidance is about making the paths cacheable at all, not about the cache key).

[...]

PR #191, opened 2026-08-25
docs: state the whole staleness window CHAT_EDGE_CACHE_SECONDS buys

### LINKED ISSUE #188: `stale-while-revalidate=5` is undocumented: the deploy note says a room read can be one second stale, the header allows six

```
```

## Observed

| request | `cache-control` |
|---|---|

## Scope, stated honestly

## Second, smaller question

## Relationship to #160

## Directions

  a one-second `s-maxage`.

## Reproduction

```bash
  printf '%-34s ' "$u"
done
```

faakhiranadiaz:
Closes #188.

`_edge_cacheable` sends two directives:

```
```

The README table describes one of them:

> `s-maxage` on `/rooms` and plain room reads so a CDN can collapse poll storms. Long-polls stay `no-store`; `0` disables.

Under RFC 5861 the pair means a shared cache may serve a room read fresh for `s-maxage`, then keep serving it stale for `stale-while-revalidate` more while it refreshes in the background. The knob sets the first and the code sets the second to five times it, so the window a deployer is actually accepting is six times the number in that column. At the default they are choosing 6 seconds while reading 1.

Confirmed against the live service just now:

| request | `cache-control` |
|---|---|
| `GET /rooms` | `public, max-age=0, s-maxage=1, stale-while-revalidate=5` |
| `GET /r/lobby?since=1&wait=1` | `no-store` |

### Scope

Documentation only. The header does not change, and `test_polled_reads_are_edge_cacheable_and_held_or_write_replies_never_are` already pins its exact shape, so the behaviour was never in question, only the description of it.

I left `CHANGELOG.md:165` alone. It says the same thing, but it is a record of what 0.8.0 shipped rather than current guidance, and editing a released entry seemed worse than leaving it.

### The test

A docs fix does not change behaviour, so there is nothing to regress in the usual sense. What can regress is the two drifting apart again, which is what happened here, so the test ties them together: it reads the multiplier out of the header the app emits under `EDGE_CACHE_SECONDS=2`, then asserts the README row names `stale-while-revalidate` and contains that same multiplier. Raising the five in `app.py` fails the test rather than silently making the table wrong.

On `main` it fails at the first assertion:

```
AssertionError: assert 'stale-while-revalidate' in '| `CHAT_EDGE_CACHE_SECONDS` | `1` | `s-maxage` on `/rooms` ...'
```

### Checks

`ruff check`, `ruff format --check`, `ty check` clean. 381 tests pass. `coverage report` 97.50%. `sz.py --check` ok, core total unchanged at 2012 (nothing here touches `src/`).

Not in scope: #160, which asks whether `&n=` cache-busting advice and this knob undercut each other. That is a design question for you, and it is about whether the caching is right rather than whether it is described.

---

I used Claude to help investigate this. The verification against the live service, and the writing, are mine.


Minh3132 (commented):
The docs change is useful, but the new regression test can false-pass when the cache multiplier drifts.

For `CHAT_EDGE_CACHE_SECONDS`, the test reduces the emitted header to `polled // seconds` and only checks that this integer appears *somewhere* in the README row. With the current row, changing the implementation multiplier from 5x to 6x would produce `6`, and the assertion would still pass because the same row already contains `6` in the prose “default 1 means a read can be 6 seconds old”, even though it would still incorrectly say “5 times the value”.

The static check has the same shape: it only asks whether the emitted `stale-while-revalidate` value appears among any numbers in the row, so a future value that happens to equal another documented number (for example the 300-second default) could pass while the sentence still says “flat 60”.

Could the test pin the actual relationship rather than number membership—for example extract the explicitly documented SWR multiplier/value from a narrowly defined phrase, or expose the cache-control relationship as a shared constant/helper used by both the header and docs test? That would make the test fail for the exact drift it is intended to prevent.

PR #192, opened 2026-08-25
fix(cache): honor the configured staleness window

### LINKED ISSUE #188: `stale-while-revalidate=5` is undocumented: the deploy note says a room read can be one second stale, the header allows six

ara-stock:
Closes #188.

## What changes

Room reads and `/rooms` now emit only `public, max-age=0, s-maxage=N`. They no longer add the undocumented `stale-while-revalidate=N*5` window, so `CHAT_EDGE_CACHE_SECONDS=1` permits at most the documented one second of shared-cache staleness rather than six seconds.

The existing header assertion is updated to pin that externally visible contract, and the change is recorded under Unreleased.

## Competing direction

This intentionally presents the alternative to #191. PR #191 keeps the current six-second effective window and documents it; this PR removes the undocumented extension and keeps the implementation aligned with the existing README and 0.8.0 deployment note, which describe reads as up to one second stale by default.

I chose the latter because the affected URL is the documented polling loop: while a room is idle, `?since=N` is stable, so a cached empty response may hide a newly landed message for the full stale-while-revalidate window. The deployment knob should state the maximum delay an operator is accepting.

These two PRs should not both merge. If the intended contract is the current six-second window, #191 is the right change and this PR should close instead.

## Verification

- `uv sync --frozen`
- `ruff check .`
- `ruff format --check .`
- `ty check`
- 379 tests passed
- coverage 97.50%
- `sz.py --check` passed; core code lines decrease from 2012 to 2010

No new public surface or persistent state. Abuse impact: none; this only shortens how long a shared cache may reuse an otherwise cacheable read.

I used Codex to investigate, implement, and verify this change.


bdunn77:
Independent validation of exact head `e7cbc55cc5c082b81cf9b714afe9a77cb02f2e17` (tree `90d458c2b0cd11f28fefd579cfa1dfc78ac05039`) against current main `45921c3e3699e01a55cde391674815367e0cff6b` (0.13.0).

**Intended cache-window cut is still causal.** Current main `_edge_cacheable` still emits `stale-while-revalidate={swr or secs * 5}`, so `CHAT_EDGE_CACHE_SECONDS=1` permits six seconds of shared-cache reuse on `/rooms` and polled room reads. This head emits only `public, max-age=0, s-maxage=N` unless `swr` is explicit, which keeps the 60s static-doc window and matches the documented one-second default. Issue #188 remains open.

**Preferred over overlapping #191** if the documented knob is the intended contract. #191 keeps the six-second window and documents it; these two should not both merge. Dirty; rebase first. Compare `main...e7cbc55` remains three files, +20/-15. I did not re-run the full suite on this exact head in this pass.

A0-PR192-e7cbc55c-90d458c2

PR #194, opened 2026-08-25
docs: add Windows PowerShell signed-DID quickstart

ChozetsuBlack:
## Summary

Adds a Windows 11 / PowerShell onboarding guide for Technocore's signed `did:key` lane.

The existing documentation explains the protocol and provides shell-oriented examples, but Windows users currently have to translate the setup, fingerprint/sharding, and signed-message flow themselves.

This guide covers:

- installing `uv` and Python 3.12 on Windows;
- using the existing `scripts/sign.py` helper;
- generating and safely handling an Ed25519 seed;
- reproducing the same `did:key`;
- publishing the current sharded DID note convention;
- sending a signed lobby message;
- verifying the resulting record with the full DID;
- removing `SIGN_SEED` from the PowerShell process afterward.

## Scope

Documentation only. No protocol, API, dependency, or runtime behavior changes.

The guide deliberately does not make claims about token wallets, airdrops, or external programs; it documents only the Technocore behavior implemented by this repository.

## Validation

The complete PowerShell flow was manually exercised against the public `technocore.chat` instance on Windows 11, including DID derivation, sharded DID-note publication, signed message submission, and read-back verification.

## Abuse / security impact

No new public surface is introduced.

The guide explicitly distinguishes the public DID from secret seed material, warns against reusing cryptocurrency wallet secrets, and reinforces that Technocore content is untrusted input.


docs/windows-powershell-quickstart.md
```diff
@@ -0,0 +1,281 @@
+# Windows PowerShell Quickstart for technocore.chat
```
sv, line 1: how about making powershell version of examples/beautiful_chat.sh ?

ChozetsuBlack: Thanks — I tried this on native Windows 11 before replying.

I found that the current service cannot boot natively on Windows because `src/store.py` imports and uses POSIX `fcntl.flock`; the current CI also runs on Ubuntu only.

So I don't want to claim a Windows-native `beautiful_chat.ps1` works when the server itself currently requires a POSIX environment.

Would you prefer a PowerShell 7 version intended to run on Linux/macOS, mirroring `examples/beautiful_chat.sh`, or should Windows-native server support be treated separately?

sv: if its a signing client - where does it need to import store.py? maybe we need to split some of the code to helpers.

ChozetsuBlack: You're right — the Windows signing client itself does not import `store.py`.

The `fcntl` issue only came up when I tried to mirror `examples/beautiful_chat.sh` literally, including its self-booting local server. The existing Windows quickstart already works natively because it only uses `scripts/sign.py` and the public HTTP API.

Splitting the client-side walkthrough from the local-server bootstrap sounds like a good direction. That would let the same protocol demo be driven from PowerShell on Windows while keeping the server bootstrap POSIX-specific.

I'm happy to explore that shape if that is what you had in mind.

ChozetsuBlack: Confirmed the distinction on a real Windows 11 machine.

The native Windows signing-client flow works without importing `store.py` at all. I also enabled WSL2 and ran the repository's existing `examples/beautiful_chat.sh` there; the full 9-stage demo passed through `== done`.

So the platform boundary seems to be:

- signing/client HTTP flow: native Windows works
- local Technocore server bootstrap: currently POSIX because of `fcntl`
- full local stack on a Windows machine: works through WSL2

I agree that splitting the client-side protocol walkthrough from the local-server bootstrap would make the cross-platform story much cleaner.

2Lin3:
I followed the review suggestion to make a PowerShell counterpart to `examples/beautiful_chat.sh` and hardened your original native demo commit `6f67bef` on current `main`. Your credit is preserved in the commit metadata:

- branch: https://github.com/2Lin3/technocore-chat/tree/codex/powershell-e2e-v010
- commit: [`378d21b`](https://github.com/2Lin3/technocore-chat/commit/378d21b2d20cda963f04c7aa40a6efda258e41f1)

The script confines every demo API request to its selected `127.0.0.1` port at runtime, disables proxies and redirects, gives the server a child-scoped temporary `CHAT_ROOT`, and launches both the server and cleanup process with no console window. It covers the same nine protocol stages as the shell demo and adds seven static/parser safety tests.

Important dependency: this commit is reviewable on its own, but it is not natively runnable on current `main` while `store.py` still imports `fcntl`. Real E2E validation therefore used the combined branch with the narrow #122 adaptation and #272 applied. Windows PowerShell 5.1 and PowerShell 7.6 both completed all nine stages, including a run with deliberately conflicting inherited `CHAT_*` and `WEB_CONCURRENCY` values. Post-run checks found zero new matching temporary directories and zero related processes; Ruff, format, and `ty` are clean.

I am not opening a competing PR. If this matches the requested direction, please cherry-pick or adapt `378d21b` here after or alongside the Windows runtime fixes.

ChozetsuBlack:
Thanks for taking this up, and for keeping the credit in the commit metadata — that's appreciated.

I'd rather not widen this PR beyond the docs page on my own. Given that `store.py` still imports `fcntl`, whether `378d21b` belongs here or alongside the Windows runtime work looks like a maintainer call, so I'll leave that to them and am happy either way.

TSynarc:
During a Windows DID setup and review of this PR, section 6 stood out: it hard-codes the shared `lobby` room and the message `Technocore signed check-in`. Quickstarts are normally copied literally, so every successful run would add the same low-information message to a public conversation. This seems especially relevant given the template propagation documented in #149 and #368.

**Suggested remedy**

Could the example generate a unique unlisted, ephemeral test room instead?

```powershell
$ROOM = "e-p-onboarding-$([guid]::NewGuid().ToString('N'))"
```

This creates a valid 47-character room name. The `e-p-` prefix makes the room ephemeral and unlisted. Changing the test room would not alter the signing or read-back process, but it would keep repeated test messages out of the shared lobby. Shared rooms could then be reserved for original, useful messages.

> A quickstart should prove the protocol works without making real conversations harder to find.

ChozetsuBlack:
Thanks — I checked this against the spec and against the live service before answering.

The concern holds. `src/manual.md` is explicit that `p-` is "unlisted: reachable, never
enumerated", and a quickstart is copied literally, so a shared-room default does add one
identical low-information message per successful run.

I went with `p-<random>` rather than `e-p-<random>`, for one measured reason: this deployment
publishes `ephemeral_ttl_seconds: 900` in `/.well-known/agent.json`. In an `e-` room the record
is dropped on read after 15 minutes, and section 7 of this page exists to read that record back
with the full DID — so `e-p-` would leave the page's own verification step returning nothing for
anyone who pauses. `p-` is equally unlisted, keeps the record readable, and a room still on its
first message is reclaimed after 24 hours anyway, so nothing accumulates.

One caveat worth knowing either way: a per-reader room is a *new* room, and the venue refuses new
rooms when it is full. Earlier today (2026-09-02, around 03:30Z) a plain write came back with

    400 room limit reached (81920 is the cap, and this would be a new one)

and about half an hour later `/rooms` reported `total: 51981, capacity: 81920`. So it is
intermittent rather than a standing condition — but it does mean a fresh-room quickstart can fail
where `lobby` would have succeeded. I think that is still the right trade; flagging it so the
choice is made with eyes open.

PR #201, opened 2026-08-25
docs: publish the /rooms listing limit cap

omerbek:
## Summary

Documents the enforced `/rooms?limit=` response cap in the README and served manual. This addresses the documentation portion of #164.

Upstream #533 now documents the advisory clamp in OpenAPI without publishing a `maximum` constraint, because values are clamped rather than rejected. This update follows that input-doctrine decision and removes this PR's obsolete schema/test changes.

## Changes

- State that `/rooms` returns the newest 50 rooms by default.
- Document `?limit=<1..200>` in the README and served manual.
- Clarify that `total` may be larger than the returned `rooms` array.

## Testing

- `python -m uv run ruff check .`
- `python -m uv run ruff format --check .`
- `python -m uv run python sz.py --check`
- `pytest tests/http/test_docs.py -q` cannot run on native Windows because `src/store.py` imports the POSIX-only `fcntl` module; collection/setup otherwise reaches that known platform boundary.


omerbek:
Closing in favor of #175, which predates this PR and already covers the same README/manual documentation as part of the broader `/rooms` pagination fix. Upstream #533 also now documents the advisory clamp in OpenAPI, so keeping this separate documentation-only PR would duplicate active work.

PR #203, opened 2026-08-25
fix(mcp): tools/list schemas advertise additionalProperties=false

### LINKED ISSUE #105: tools/list publishes an open schema for a tools/call that refuses unknown arguments
## The gap

```
```

```json
{"type": "object",
 "required": ["room"]}
```

## Repro

## Scope, and honest severity

## Fix

---

kaname97:
## What

`schema_of` now emits `additionalProperties: false`, so every tool schema in `tools/list` says what `_validate` already enforces: arguments no parameter declares are refused (`-32602 unexpected arguments`). One regression test asserts local validation reaches the same verdict as `tools/call` for the issue's example call.

## Why

Fixes #105. A JSON Schema object admits every property it does not name until told otherwise, so a client validating locally against the document this server hands out sends calls its own `tools/list` called valid, then eats a `-32602`. The two statements now agree; no handler or refusal changes.

Closes #105.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 382 passed, 97.50% (floor 96)
- [x] Docs unchanged — no published document said anything about additionalProperties before, so none becomes wrong
- [x] New surface on a world-writable service: nothing new — the flag documents an existing refusal

### DIFF
--- a/mcp/src/technocore_mcp/protocol.py
+++ b/mcp/src/technocore_mcp/protocol.py
@@ -206,7 +206,15 @@ def schema_of(handler: Callable[..., str]) -> dict[str, Any]:
         properties[name] = fragment(hints[name])
         if parameter.default is inspect.Parameter.empty:
             required.append(name)
-    schema: dict[str, Any] = {"type": "object", "properties": properties}
+    schema: dict[str, Any] = {
+        "type": "object",
+        "properties": properties,
+        # The validator refuses arguments no parameter declares (`unexpected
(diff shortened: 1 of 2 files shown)


sv:
Duplicate of #104 — same `additionalProperties: false` fix for issue #105. Keeping #104 since its test dynamically covers every registered tool rather than one hardcoded example. Closing this one.

---

PR #205, opened 2026-08-25
docs(manifest): describe the signed lane's text budget

### LINKED ISSUE #76: openapi.json: the signed GET lane's text param has no description, and it is the lane with less URL budget

kaname97:
## What

The say-signed lane's `text` param — the only write input published without a description — now carries one: the smaller URL budget relative to the unsigned lane, the percent-encoding math, and POST as the escape hatch. A regression test pins the description to the enforced `MAX_TEXT_CHARS` so it cannot drift.

## Why

Fixes #76, on the lane that needs the warning most: DID (56) + signature (86) + nonce (~13) sit ahead of `text`, leaving ~116 bytes less room than the lane that already warns. And the failure there is the opaque kind — 4096 CJK characters encode past the edge's ceiling, dying at the proxy with no status and no body. The prose covers this; only the machine-readable surface had the hole.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 382 passed, 97.50% (floor 96)
- [x] Docs updated in the same change — the description *is* the doc fix
- [x] New surface on a world-writable service: nothing new — documentation only

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -495,6 +495,14 @@ def openapi_document(base: str, version: str, max_body_bytes: int, max_wait: flo
                             "name": "text",
                             "required": True,
                             "schema": _TEXT_SCHEMA,
+                            "description": (
+                                "URL-encoded message body. This lane has the smaller URL "
+                                "budget of the two write lanes — DID, signature and nonce "
+                                f"sit ahead of it — so {store.MAX_TEXT_CHARS} ASCII characters fit, one CJK "
+                                "character is 9 bytes encoded, and long non-Latin text can "
+                                "exceed the edge's ceiling and die at the proxy with no "
+                                "status at all — use POST for anything long."
+                            ),
                         },
                     ],
                     "responses": {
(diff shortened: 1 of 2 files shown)


sv:
Triage: closing as a duplicate of #111.

Three open PRs add the same missing `description` to the signed say lane's `text` path parameter in `src/manifest.py` for #76: #111, #128 and #205. All three say the same thing — `maxLength` is the character cap, the URL is the limit that actually bites, and the signed lane has less budget because the DID, signature and nonce are path segments ahead of the text.

Keeping #111: it is the earliest of the three, and it is the only one that hoists a single shared `_TEXT_PATH_PARAM` used by both say lanes rather than describing the signed lane alone — so the two descriptions cannot drift apart later, which is the drift that produced #76 in the first place.

Per CONTRIBUTING's "Overlapping work", review or build on #111 rather than carrying a parallel branch. The "one CJK character is 9 bytes encoded, and it dies at the proxy with no status at all" phrasing here is the most concrete of the three about what failure a caller actually meets, and is worth folding into #111.

---

PR #210, opened 2026-08-25
docs: say what a note read returns, so ?if= gets the value and not the reply

Mariukasfak:
## The gap

The manual documents `GET /kv/<ns>/<key>` as "read a persisted note" and `?if=<what you last read>`, but a note read is framed: `BANNER`, a blank line, the value, and — once the read bucket drops below a quarter — a trailing `# budget:` line. So *what you last read* is never what `?if=` wants, and nothing says so.

Both failure modes are silent in the direction that hurts:

- A reader that hands the reply to a JSON parser never parses a note **it wrote itself**. The natural fallback is "keep it as a string", so an agent's persisted state quietly resets on every restart instead of raising.
- A reader that replays the reply as `?if=` gets `409 note ... changed since you read it`, which blames a concurrent writer for the caller's own framing.

I lost an afternoon to the first one before finding it, which is what prompted the patch.

## Reproduced against technocore.chat 0.9.2

```
$ curl -s ".../kv/p-xxx/state/set/v1"
ok p-xxx/state 2B 2026-08-25T14:43:13Z

$ curl -s ".../kv/p-xxx/state" | cat -A
!! UNTRUSTED CONTENT — the lines below were written by other agents ...$
$
v1$

$ RAW=$(curl -s ".../kv/p-xxx/state")
$ curl -s -G --data-urlencode "if=$RAW" ".../kv/p-xxx/state/set/v2"
409 note p-xxx/state changed since you read it     # nothing changed

$ curl -s -G --data-urlencode "if=v1" ".../kv/p-xxx/state/set/v2"
ok p-xxx/state 2B 2026-08-25T14:43:15Z
```

## The change

- A `READING A NOTE` paragraph before `CONDITIONAL NOTES`, stating the reply shape and how to take the value back out.
- `?if=<what you last read>` → `?if=<the value you last read>`, since that phrase is the trap itself.
- One regression test in `tests/http/test_notes.py` that performs the documented recovery rather than hardcoding a banner offset, and asserts both CAS outcomes.

No behaviour change; `/openapi.json` and the contract check are untouched.

## Checks

`uv run ruff check .` and `uv run ruff format --check` pass. I could not run `pytest` locally — `src/store.py` imports `fcntl`, so the suite is POSIX-only and I am on Windows with no WSL distro or Docker. The new test asserts only behaviour I verified against production above, but CI is the real check here; happy to adjust if it disagrees.

### DIFF
--- a/src/manual.md
+++ b/src/manual.md
@@ -43,9 +43,19 @@ An empty reply after the full wait is normal — re-issue with the same since. T
 server holds a bounded number of waiters; over that it answers immediately
 rather than queueing, so treat a fast empty reply as "no slot, poll normally".
 
+READING A NOTE: the reply frames the value. GET /kv/<ns>/<key> answers with the
+untrusted-content banner, a blank line, then the stored value — and, once the read
+bucket drops below a quarter, a trailing "# budget:" line after it. The value is
+the bytes you wrote and nothing else, so a reader that hands the whole reply to a
+JSON parser never parses a note it wrote itself, and one that passes the whole
+reply back as ?if= gets a 409 saying the note changed when nothing changed. Take
+the value: drop the leading banner line and the blank line after it, and any
+trailing budget line. Notes are the one lane where this matters, because a note is
+the value and a room reply is already line-structured.
+
 CONDITIONAL NOTES: unconditional writes are last-write-wins, so two agents doing
 read-modify-write on one note lose an update.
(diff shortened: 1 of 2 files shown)


Mariukasfak:
Closing: no review in a month and it now conflicts with main. Will reopen rebased if still useful.

PR #211, opened 2026-08-25
fix(http): answer the name grammar before room class questions

### LINKED ISSUE #109: A name the grammar refuses gets a class answer: mb-FOO says 403 'send a signature', say-signed then says 400

deranalabs:
## What

An unsigned room write whose name fails the grammar now answers `400 bad name`
before its class is checked. A caller sending to `mb-FOO` on the `say` lane sees
`400 bad name` instead of `403 mailbox`.

## Why

The write gate classified the room (`is_mailbox`) before validating the name
grammar (`valid_name`), so a name that carries a class prefix but fails `NAME_RE`
fell into the mailbox branch on the unsigned lane only — the read and signed lanes
already checked the name first. Fixes #109.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 382 passed, 97.59%
- [x] Docs that would now be wrong are updated — no manual/manifest change needed
      (both 400 and 403 remain documented); CHANGELOG notes the fix under [Unreleased]
- [x] New surface on a world-writable service: nothing new — the write lane already
      validated names, this only reorders when that happens

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,14 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
+### Fixed
+
+- **An unsigned room write with an invalid name now answers `400 bad name`, not
+  `403 mailbox`.** The write gate classified the room before checking the name
+  grammar, so `mb-FOO` (a class prefix that fails `NAME_RE`) fell into the mailbox
+  branch on the `say` lane. The gate now checks `valid_name` first, matching the
+  read and signed lanes.
+
 ## [0.9.3] - 2026-08-26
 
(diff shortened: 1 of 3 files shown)


sv:
Duplicate of #108 — same single-line fix for issue #109, but #108 additionally closes the identical bug in the note-ownership gate and a related newline-injection issue in server-authored refusal text. Closing in favor of #108.

---

deranalabs:
Missed #108 — this is a strict subset of it (room lane only; it also covers the ownership gate + newline interpolation). Closing in favor of #108 is right. Thanks Sergey.

PR #216, opened 2026-08-25
fix(http): /rooms survives a cache clear landing mid-refresh

### LINKED ISSUE #212: rooms: concurrent cache invalidation can make a refresh return 500
## Summary

A concurrent write can clear `_rooms_cache` after `_rooms_view()` inserts a refreshed entry but before it promotes that entry with `move_to_end()`. The promotion then raises `KeyError`, and `GET /rooms` returns 500.

I could not find an existing issue or PR covering this exact interleaving.

## Reproduction

Reproduced on current main at `e6114b3` with a deterministic barrier around the existing cache operations:

1. Start `GET /rooms?format=json`.
2. Let `_rooms_view()` complete the real `_rooms_cache[limit] = ...` insertion, then pause before `move_to_end(limit)`.
3. Send a real write through `GET /r/cache-race/say/writer/hello`.
4. The write reaches `take(..., kind="write")` and calls `_rooms_cache.clear()`.
5. Resume the `/rooms` request.

Observed result:

```text
write_status=200
rooms_status=500
rooms_body=Internal Server Error
```

The barrier only makes the existing scheduling window deterministic; the reader and writer still use the real application endpoints and cache invalidation path.

## Root cause

The write path clears the cache in `take()`:

```python
if kind == "write":
    _rooms_cache.clear()
```

The refresh path mutates the shared `OrderedDict` in multiple steps:

```python
_rooms_cache[limit] = (stamp, now, view)
_rooms_cache.move_to_end(limit)
```

These synchronous routes can execute concurrently in Starlette's threadpool. The cache stamp protects freshness, but it does not make the LRU mutation sequence atomic.

## Expected behavior

A concurrent write may invalidate or prevent reuse of a freshly computed cache entry, but it should not turn an otherwise valid `/rooms` response into HTTP 500.

## Proposed scope

I can put together a focused regression test in `tests/http/test_rooms.py` using synchronization events rather than timing, then make only the cache mutation/invalidation sequence safe. I would not change the cache policy, response schema, or store behavior.

Before implementing: would you prefer a small lock around cache mutation and invalidation, or an invalidation-tolerant update sequence that lets the current response complete without retaining the cleared entry? The former gives a clearer atomicity boundary; the latter may keep the hot path lighter.

devorun:
## What

`GET /rooms` stops answering 500 when a concurrent write clears its cache while the
refresh is repopulating it. Nothing else about the response, the cache window or the
eviction bound changes.

## Why

Fixes #212. `_rooms_view` inserted its cache entry and promoted it with `move_to_end` in a
second step, and `take` clears that cache on every write. Sync endpoints run in the
threadpool, so a clear landing between the two left `move_to_end` no key to promote — a
`KeyError` out of the most polled read on the service.

Popping the key before re-inserting it is the same LRU promotion, because a fresh insert
is already last, and there is no key left that can go missing. Core code-lines are
unchanged (`sz.py --check` clean), so this stays a swap rather than growth.

Two tests: the first drives the reported interleaving through the real endpoints and
asserts the race actually fired, so it cannot pass by proving nothing; the second pins the
eviction bound, since the promotion it feeds moved.

One judgment call left alone: the `popitem(last=False)` three lines below can raise the
same way if a clear lands between the length check and the pop. That needs more than 64
distinct `?limit=` values cached first, and it is a different window, so I kept it out to
keep this to one problem. Happy to follow up if you want it closed here.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 383 passed, 97.50%
- [x] Docs that would now be wrong are updated: none apply; no documented behaviour moves
- [x] New surface on a world-writable service: nothing new — no route, parameter or cap changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
+### Fixed
+
+- **`GET /rooms` no longer answers 500 when a write clears its cache mid-refresh.** The refresh
+  promoted its cache entry with `move_to_end` in a separate step from the insert, and a concurrent
+  write clearing the cache between the two left no key to promote. The entry is popped before it is
+  re-inserted instead, which is the same promotion with no key that can go missing.
+
 ## [0.9.3] - 2026-08-26
 
 PATCH: signed writes stop parsing a read window they are about to discard, plus documentation
--- a/src/app.py
+++ b/src/app.py
@@ -661,8 +661,14 @@ def _rooms_view(limit: int) -> dict:
         round(view["notes"]["total"] / seen, 4) if seen else None
     )
     if config.ROOMS_CACHE_SECONDS > 0:
+        # Popped before it is re-inserted, because a concurrent write clears this cache
+        # (see `take`) and sync endpoints run in the threadpool, so the clear really can
+        # land between two of these lines. move_to_end was then promoting a key that was
+        # no longer there, and its KeyError answered the most polled read on the service
+        # with a 500. A fresh insert is already last, so this buys the same LRU promotion
+        # with no key to go missing.
+        _rooms_cache.pop(limit, None)
         _rooms_cache[limit] = (stamp, now, view)
-        _rooms_cache.move_to_end(limit)
         while len(_rooms_cache) > MAX_ROOMS_CACHE:
             _rooms_cache.popitem(last=False)
     return view
--- a/tests/http/test_rooms.py
+++ b/tests/http/test_rooms.py
@@ -940,3 +940,41 @@ def test_polled_reads_are_edge_cacheable_and_held_or_write_replies_never_are(cli
     with config.override(EDGE_CACHE_SECONDS=0):
         assert client.get("/rooms").headers["cache-control"] == "no-store"
         assert client.get("/r/lobby").headers["cache-control"] == "no-store"
+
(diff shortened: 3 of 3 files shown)


Abwan12:
Your are great 👍

sv:
Closing as fixed — but this is your fix, and it shipped under someone else's PR, so it's worth saying plainly what happened.

#245 merged with the identical change: `_rooms_cache.pop(limit, None)` before the insert, replacing the `move_to_end`. It was reached independently there, during a review pass over that PR's own diff, about eight hours after you opened this one. Independently is not the same as first, and you were first.

The diagnosis in #212 is the part that mattered most: the reproduction, the "these synchronous routes can execute concurrently in Starlette's threadpool" framing, and the observation that the stamp protects freshness without making the LRU mutation sequence atomic.

One difference in how it landed. #245 also removed the `_rooms_cache.clear()` from `take` entirely — it was invalidating the cache on every message write, which is why `/rooms` never hit its cache at all — so the specific interleaving you reproduced (a write's `clear()` landing between the insert and the promotion) can no longer occur; nothing calls `clear()` at runtime any more. The pop-then-insert was kept regardless, for the trigger that survives: a concurrent *evictor's* `popitem` can take the key between those two statements even with no `clear()` in the picture. Your fix is the correct one for both.

#212 is closed. Sorry for the duplicated effort.

---

PR #218, opened 2026-08-25
fix(mcp): a surrogate argument key no longer kills the session

katsugtgz:
## What

A `tools/call` argument *key* with a lone surrogate reached the unexpected-arguments error string raw (`protocol.py`). The reply writer serializes with `ensure_ascii=False` to a UTF-8 stream. `UnicodeEncodeError` escaped `serve()`. Exit code 1, no reply, session over. Verified against the real `serve()`:

```json
{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"read_room","arguments":{"room":"lobby","\\ud800":1}}}
```

## Why

The host relays model-composed argument dicts verbatim. One glitched escape in a key costs the agent its only outbound channel until restart. Values are already safe. They hit `quote()` inside the handler, and `_call` turns that into an `isError` result. Keys hit the error message before any handler runs.

Not covered elsewhere. #102 catches oversized integers at parse time. #106 pins stream encoding. Surrogates are unencodable regardless of stream setup. This bug is in the reply-write path.

## Fix

`', '.join(map(repr, sorted(unexpected)))`. Every other user-facing message already uses `!r`. Optional hardening for the maintainer: `ensure_ascii=True` in `_write`, so reply serialization can never fail.

## Checks

- [x] Regression test fails without the fix (`UnicodeEncodeError` at `_write`). Passes with it.
- [x] `uv run python -m pytest tests/test_mcp.py -q`: 46 passed
- [x] `ruff check`, `ruff format --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,12 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
+### Fixed
+
+- **MCP: a lone surrogate in a tool-call argument key no longer kills the stdio session.** The
+  unexpected-arguments error names keys through `repr`, so a key no UTF-8 stream can encode fails
+  that one call instead of the wrapper process.
+
 ## [0.9.3] - 2026-08-26
 
 PATCH: signed writes stop parsing a read window they are about to discard, plus documentation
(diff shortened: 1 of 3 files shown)


sv:

The surrogate-key handling this fixes lives in `mcp/src/technocore_mcp/protocol.py`, deleted in 0.11.0 by [#539](https://github.com/flop-labs/technocore-chat/pull/539) — the wrapper runs on the official MCP SDK now and does not decode argument dicts itself, so this diff has no target.

"A malformed argument key kills the session" is a good class of bug to have found, and it is worth re-checking against the current code rather than assuming the rewrite carried the fix. If a lone surrogate in a key still takes the server down over stdio or `--http`, please open a fresh issue with the frame.

---

PR #220, opened 2026-08-25
fix(app): a write no longer 500s a concurrent /rooms refresh (#212)

### LINKED ISSUE #212: rooms: concurrent cache invalidation can make a refresh return 500

Agozie180:
## What

A write that arrives while `/rooms` is refreshing its cache no longer turns that refresh into a 500. Callers see the listing return normally under concurrent writes; response shapes and caps are unchanged.

## Why

Fixes #212. The rate-limit gate cleared the `/rooms` listing cache on every write. When that clear landed between `_rooms_view` inserting a fresh entry and promoting it to most-recently-used, the promotion referenced a key that was already gone and raised `KeyError`, surfacing as a 500.

The clear was never the freshness guarantee — it runs *before* the store write, so `_rooms_stamp` (bumped by `store.append` on every counted mutation, notes included) is what actually invalidates the cache. Removing the clear deletes the race at its source and leaves read-your-writes intact, rather than papering over it with a lock or a tolerant update. Net code lines are negative, so the `sz.py` cap holds.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 382 passed, 97.50%
- [x] Docs that would now be wrong are updated: the `_rooms_stamp` docstring now names the stamp as the sole freshness mechanism; no manual/README surface changed
- [x] New surface on a world-writable service: nothing new — this only removes a redundant cache clear

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
+### Fixed
+
+- **A write arriving during a `/rooms` refresh no longer returns 500.** A write used to clear the
+  listing cache from inside the rate-limit gate; when that landed between the refresh inserting an
+  entry and promoting it, the promotion hit a missing key and 500'd. Freshness already comes from the
+  generation stamp, so the clear is gone.
+
 ## [0.9.3] - 2026-08-26
 
 PATCH: signed writes stop parsing a read window they are about to discard, plus documentation
--- a/src/app.py
+++ b/src/app.py
@@ -167,15 +167,16 @@ def take(request, kind, per_min, burst=None) -> tuple[int, float]:
     # Thin adapter over limit.take: the knobs are read HERE, at call time, so
     # monkeypatch.setattr(app, "MAX_BUCKETS", ...) and config.override() keep reaching
     # the bucket arithmetic.
-    left, wait = limit.take(
+    #
+    # This used to also drop the /rooms cache on a write, as a fast path. It no longer
+    # does. That clear ran *before* the store write, so it was never the freshness
+    # guarantee (see `_rooms_stamp`), and it reached across requests to mutate a dict a
+    # concurrent /rooms refresh was mid-mutation on — clearing the entry that refresh had
+    # just inserted and was about to promote, which turned the refresh into a 500 (#212).
+    # The stamp already invalidates on every counted mutation, so the clear was pure cost.
+    return limit.take(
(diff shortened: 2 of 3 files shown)


Agozie180:
Closing in favour of #216, which fixes #212 and was opened first. I'd landed on removing the write-path cache clear entirely (leaning on the generation stamp as the sole freshness mechanism); #216's pop-then-reinsert is the more conservative fix and matches the scope the reporter proposed, so it should take priority. Happy to review #216 or pick up the follow-up popitem window it flagged if that's useful.

PR #222, opened 2026-08-25
feat(mcp): publish tool effect annotations in tools/list (#206)

### LINKED ISSUE #206: mcp: expose tool effect annotations in tools/list
## Problem

The `/humans` WebMCP surface marks its read and write tools with `readOnlyHint`, but the stdio MCP server currently emits only `name`, `description`, and `inputSchema` for every tool in `tools/list`.

This is reproducible without making a network call:

```bash
PYTHONPATH=mcp/src python3 - <<'PY'
from technocore_mcp.server import server

reply = server.handle({"jsonrpc": "2.0", "id": 1, "method": "tools/list"})
print({tool["name"]: sorted(tool) for tool in reply["result"]["tools"]})
PY
```

Every entry currently has the same three keys:

```text
['description', 'inputSchema', 'name']
```

That leaves annotation-aware clients unable to distinguish reads from writes using the standard MCP metadata. The distinction is especially useful here because `say` appends public data and `write_note` can overwrite durable, world-writable state, while the remaining tools are read-only.

## Proposed scope

- Extend the small in-tree MCP `Tool`/`Server.tool` representation to publish standard `ToolAnnotations`.
- Mark `read_room`, `wait_for_message`, `list_rooms`, `discover_rooms`, `read_note`, `list_notes`, and `read_docs` as read-only.
- Mark every tool as open-world because each one can interact with the configured external Technocore instance.
- Mark `say` as non-read-only, additive, non-idempotent, and open-world.
- Mark `write_note` as non-read-only, potentially destructive, non-idempotent, and open-world.
- Add focused `tools/list` regression coverage; do not change any handler or HTTP behavior.

This would be orthogonal to #203, which tightens `inputSchema.additionalProperties`; an implementation can rebase after that change settles.

## Compatibility question

The server currently negotiates `2024-11-05`, `2025-03-26`, and `2025-06-18`. `ToolAnnotations` is defined in the latter two schemas but not in the `2024-11-05` `Tool` shape.

Would you prefer annotations to be emitted unconditionally as additive metadata that older clients ignore, or should `tools/list` vary by the negotiated protocol version? The latter would require retaining per-connection negotiated-version state, so I would not add that machinery without confirming it is desired.

Happy to send the focused implementation and tests once the compatibility direction is clear.

Agozie180:
## What

Publishes standard MCP `ToolAnnotations` in the stdio server's `tools/list` (#206). Every tool now carries effect hints alongside its `name`/`description`/`inputSchema`:

- **Read-only** (`readOnlyHint`, `openWorldHint`): `read_room`, `wait_for_message`, `list_rooms`, `discover_rooms`, `read_note`, `list_notes`, `read_docs`.
- **`say`**: non-read-only, non-destructive (append), non-idempotent, open-world.
- **`write_note`**: non-read-only, destructive (can overwrite durable, world-writable state), non-idempotent, open-world.

Every tool is `openWorldHint` because each reaches the configured external Technocore instance. The `Tool`/`Server.tool` representation gained an optional `annotations` argument; `spec()` emits the key only when set.

## Why

Closes #206. Annotation-aware clients could not tell a read from a write without parsing prose, even though the distinction matters most here — `say` appends public data and `write_note` can overwrite durable state.

On the issue's open compatibility question (emit unconditionally vs. vary by negotiated protocol version): this emits **unconditionally as additive metadata**. Rationale —

- The in-repo `/humans` WebMCP surface already emits `annotations` unconditionally per tool, so this matches existing precedent in the same codebase.
- `ToolAnnotations` is additive; a `2024-11-05` client ignores the unknown key rather than failing.
- Varying by version would require retaining per-connection negotiated-version state, which the module deliberately avoids — the issue itself flagged not adding that machinery without confirmation.

Orthogonal to #203 (which tightens `inputSchema.additionalProperties`); no handler or HTTP behavior changes.

**Abuse:** nothing new — no HTTP surface, route, parameter, or cap changes; this is read-only metadata in a local stdio wrapper.

## Checks

- `ruff check .` ✓ · `ruff format --check .` ✓ · `ty check` ✓
- `coverage run -m pytest tests -q`: 383 passed, branch coverage 97.47% (fail_under=96) ✓
- New regression `test_tools_list_publishes_effect_annotations` fails without the change (the spec has no `annotations` key → `KeyError`).
- `sz.py --check` unaffected (`mcp/` is outside the size gate).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
+### Added
+
+- **The stdio MCP server publishes standard `ToolAnnotations` in `tools/list`.** Each tool now
+  carries the effect hints (`readOnlyHint`, `openWorldHint`, and `destructiveHint`/`idempotentHint`
+  on the two writes) so an annotation-aware client can tell a read from a write without parsing the
+  description. Additive metadata emitted unconditionally; older clients ignore the extra key.
+
 ## [0.9.3] - 2026-08-26
 
 PATCH: signed writes stop parsing a read window they are about to discard, plus documentation
--- a/mcp/src/technocore_mcp/protocol.py
+++ b/mcp/src/technocore_mcp/protocol.py
@@ -262,18 +262,32 @@ class Tool:
     asks for: a failed tool call is data the model can react to, not a protocol fault.
     """
 
-    def __init__(self, name: str, description: str, handler: Callable[..., str]):
+    def __init__(
+        self,
+        name: str,
+        description: str,
+        handler: Callable[..., str],
+        annotations: dict[str, bool] | None = None,
+    ):
         self.name = name
         self.description = description
         self.handler = handler
         self.schema = schema_of(handler)
+        # Standard MCP ToolAnnotations — effect hints (readOnly/destructive/…) an
+        # annotation-aware client reads to tell a read from a write. Optional: the
+        # 2024-11-05 Tool shape predates the field, so it is emitted only when set and an
+        # older client ignores the extra key.
+        self.annotations = annotations
 
     def spec(self) -> dict[str, Any]:
-        return {
+        spec: dict[str, Any] = {
             "name": self.name,
             "description": self.description,
             "inputSchema": self.schema,
         }
+        if self.annotations:
+            spec["annotations"] = self.annotations
+        return spec
 
 
 class Server:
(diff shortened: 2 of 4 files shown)


sv:
Closing as delivered elsewhere — the feature shipped, this diff cannot.

[#539](https://github.com/flop-labs/technocore-chat/pull/539) rebuilt the wrapper on the official MCP Python SDK and deleted `mcp/src/technocore_mcp/protocol.py`, which this PR patches, so there is nothing left to rebase onto. Every tool now publishes `ToolAnnotations` with the same matrix you proposed for #206 — seven read-only, `say` additive, `write_note` potentially destructive, all open-world — pinned in `tests/test_mcp.py:289`.

Thanks for working out the matrix; it is the one that shipped.

---

PR #223, opened 2026-08-25
docs: clarify signed write retries after timeouts

### LINKED ISSUE #141: A timed-out signed write can land, and retrying the URL then reports failure on a success

temitop48:
## Summary

Fixes #141.

Clarifies the signed-write nonce guidance for timeout and 5xx responses. A client timeout does not prove the write failed because the server may already have accepted the write and consumed the nonce before the client receives a reply.

The manual now tells callers to:

- avoid resending the same signed URL,
- re-read the relevant room or note state first,
- use a fresh greater nonce and re-sign only if another write is still needed.

## Caller-visible impact

Documentation only. No API, behavior, persistence, or response-shape changes.

## Validation

- `uv run --group dev python -m pytest tests/http/test_docs.py -q` → 47 passed
- `git diff --check` → passed

None. This documents existing signed-write behavior and safer client retry handling.

### DIFF
--- a/src/manual.md
+++ b/src/manual.md
@@ -109,6 +109,11 @@ single-use only while the message remains in the newest 1 MiB scanned for the
 last nonce. Once newer traffic buries it beyond that tail, the same URL is
 accepted again even if the message remains elsewhere in the larger room ring.
 Signatures still prove authorship; only the single-use guarantee expires early.
+A timeout or 5xx does not prove a signed write failed: the server may have
+accepted the write and consumed the nonce before the client received a reply.
+Do not resend the same signed URL. Re-read the relevant state first — the room
+with ?since= for messages, or the note itself for signed note writes — then use
+a fresh greater nonce and re-sign only if another write is still needed.
 RENDERING: the text view shows a verified writer as <z6Mk...2doK> and everything
 else as <~nick>, where ~ means "self-asserted, proved nothing". ?format=json
 carries the full DID in `from` and the nonce in `nonce`.


djd39448:
Tested on a worktree rebased on current `main`. This is a clean, well-placed change; flagging one gap rather than arguing against it.

What it gets right: placement (it sits in the NONCE section — where the reporter asked for the line, and where the existing nonce single-use text already lives), scope and concision (docs-only, one file), and it states the shared fact tightly — the nonce is consumed before the client hears back, don't resend, re-read state and sign a fresh greater nonce. Docs suite passes (`uv run python -m pytest tests/http/test_docs.py` → 47 passed).

The gap: the added text doesn't mention the `room-owners`/`room-allow` persistent-counter asymmetry, which is the specific reason #141 was filed:

```
$ gh pr diff 223 -- src/manual.md | grep '^+' | grep -iE 'room-owners|room-allow|persistent|forever|never expires'
# no matches
```

For a plain room message the replay window ages out of the scanned tail; for `room-owners`/`room-allow` the counter at `/kv/room-nonce/<room>` is persistent, so a stale retry there is refused permanently instead of eventually clearing. The reporter framed that asymmetry as what makes this "worth a doc line rather than a shrug." As written, someone who hits the room-owners case would re-read state, retry, and still be refused, with nothing here signaling the refusal is permanent.

One sentence would close it — #143 has that sentence but places it under OWNED ROOMS, so the ideal is your placement plus that sentence. Contributor, not the maintainer.

sv:
Duplicate of #143 — both address issue #141, but per the review comment above, #143 additionally explains the room-owners/room-allow persistent-replay-counter asymmetry, which is the specific reason #141 was filed. Closing in favor of #143.

---

PR #227, opened 2026-08-25
fix: reserve server nickname for event messages

### LINKED ISSUE #137: unsigned writers can forge the server's ~server identity

temitop48:
## Summary

Fixes #137.

Unsigned callers could use the nickname `server`, which renders as `<~server>` and is indistinguishable from the service's own event author in text views.

This change reserves `server` for server-authored event records by rejecting unsigned `store.append(...)` calls that use `EVENTS_NICK`. Genuine event announcements remain unchanged because `_log_event()` writes through `_write_record()` directly.

## Caller-visible impact

- Unsigned writes with `nick=server` now return HTTP 400.
- Other unsigned nicknames are unchanged.
- Signed writes are unchanged because their stored `from` value is the verified DID, not the nickname.
- Genuine `/r/events` announcements continue to render as `<~server>`.

## Reproduction and validation

Before the fix, the regression test reproduced the issue with `assert 200 == 400` for:

`GET /r/lobby/say/server/created%20official-room`

After the fix:

- `tests/http/test_rooms.py::test_unsigned_writer_cannot_impersonate_server` → passed
- `tests/http/test_rooms.py::test_new_public_rooms_are_announced` → passed
- `tests/http/test_rooms.py` → 49 passed
- `git diff --check` → passed

This removes one impersonation path from the unsigned lane without changing room authorization, event storage, response shapes, or signed identity semantics.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -1389,6 +1389,9 @@ def append(
     primitive that already exists does the rest — `?since=` for incremental reads,
     `?format=json`, `?wait=` for near-real-time, ring retention, the same rate limits.
     """
+    if did is None and nick == EVENTS_NICK:
+        raise StoreError(f"reserved nickname: {EVENTS_NICK}")
+
     rec, created = _write_record(root, room, nick, text, did=did, nonce=nonce)
     # Counted here rather than in `_write_record`, so the server's own announcements
     # (`_log_event` writes one per created room) never inflate the message count. This
--- a/tests/http/test_rooms.py
+++ b/tests/http/test_rooms.py
@@ -27,6 +27,13 @@ def test_say_then_read(client):
     assert "UNTRUSTED CONTENT" in body  # injection framing always present
 
 
+def test_unsigned_writer_cannot_impersonate_server(client):
+    response = client.get("/r/lobby/say/server/created%20official-room")
+
+    assert response.status_code == 400
+    assert "<~server>" not in client.get("/r/lobby").text
+
+
 def test_since_cursor_returns_only_new(client):
     for i in range(3):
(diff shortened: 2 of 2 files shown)


djd39448:
Rebased #227 on current `main` and tested it. The fix is correct and the regression is valid — the gate is the same one #138 uses, and #227's test passes against #138's implementation too, so the two are interchangeable at the gate:

```
$ gh pr diff 227
if did is None and nick == EVENTS_NICK:
    raise StoreError("reserved nickname: server")
```

Two differences versus #138, neither a correctness problem:

1. The refusal message is terser. #138's names the rule and the reason (`the nickname server is reserved: it signs /r/events`). The service's other refusals name the rule and the alternative, and CONTRIBUTING notes mutation testing runs over refusal bodies, so the fuller message is a closer fit to house style.

2. The test checks the two direct outcomes (`server -> 400`, no `<~server>` in the render); #138 additionally guards a `serverbot` prefix staying 200 (over-broad match), the signed lane, and that the server still announces new rooms. Those are additive — #227 isn't wrong without them.

Shared with #138: #227 fails `sz.py --check` on current `main` and doesn't bump the baseline, so whichever merges needs a one-line `sz-baseline.json` refresh. Correct and concise as it stands; the notes above are about house-style fit and test coverage, not the fix. Contributor, not the maintainer.

sv:
Duplicate of #138 — same fix, same location, closes the same issue (#137). Per your own review comment, #138 additionally guards a `serverbot`-prefix nick and the signed lane. Closing in favor of #138.

---

PR #230, opened 2026-08-25
fix(rooms): tolerate window memo eviction after cache hit

### LINKED ISSUE #229: rooms: window-memo eviction can race a cached read into 500
`GET /rooms` can still return 500 from a different cache than the two app-level windows in #212.

## Reproduction

I reproduced this deterministically against current `main` using the real defaults
(`_WINDOW_MEMO_MAX = 512`, `MAX_ROOMS = 5120`):

1. Fill `store._window_memo` to 512 entries, with room A as the oldest LRU entry.
2. Start `GET /rooms?limit=1` while A is the newest room, and pause after
   `_cached_window()` reads A's valid hit.
3. Make room B the newest room and run a competing `GET /rooms?limit=1`.
4. The competing request inserts B and evicts A from the full memo.
5. Resume the first request.

Observed result:

```
write                    200
competing GET /rooms     200
paused GET /rooms        500 Internal Server Error
```

The failing sequence is in `src/store.py::_cached_window`:

```python
hit = _window_memo.get(key)
if hit and hit[0] == stamp:
    _window_memo.move_to_end(key)
```

The hit is valid for the stat observed by the first request, but another refresh can
evict that key between `get()` and `move_to_end()`. The latter then raises
`KeyError`.

This is separate from #216 and #221. Those cover `app._rooms_cache` racing with an
explicit `.clear()`; this one is `store._window_memo` racing with ordinary capacity
eviction, so no clear is required. The default room cap is ten times the memo bound, and
the set of recently active rooms can move over time, so the full-memo state is reachable
without changing operator settings.

## Expected behavior

A valid cached window may lose its LRU promotion under contention, but that should not
turn an otherwise successful `/rooms` read into a 500.

I have a focused HTTP-level regression test for this exact interleaving and can open a
small fix if that direction makes sense.

Osraka:
Closes #229.

## What

`GET /rooms` now tolerates a window-memo hit being evicted before LRU promotion. It returns the captured view instead of 500. No contract or cache bound changes.

## Why

`_cached_window()` reads and promotes a hit separately. A concurrent refresh can fill the 512-entry memo and evict the key between `get()` and `move_to_end()`. The captured value remains valid for the request's `(mtime_ns, size)` stamp, so only the lost promotion is suppressed.

| State | Result |
| --- | --- |
| valid hit, still present | promoted and returned |
| valid hit, concurrently evicted | captured view returned |
| stale or missing hit | recomputed as before |

The HTTP regression forces this eviction and fails on current `main` with `500 != 200`. This is separate from #245 (`app._rooms_cache`); this fix concerns `store._window_memo`.

## Checks

- [x] Full suite: 399 passed; branch coverage 97.56%
- [x] Ruff lint/format and `ty check`
- [x] `sz.py --check` and `--caps`
- [x] No public-contract documentation change; nothing new for abusive callers

Proposed release note: Prevent `/rooms` from returning 500 when a concurrent refresh evicts a valid window-cache hit before LRU promotion.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -19,7 +19,7 @@ import time
 import unicodedata
 from collections import OrderedDict
 from collections.abc import Iterator
-from contextlib import contextmanager
+from contextlib import contextmanager, suppress
 from datetime import UTC, datetime
 from functools import lru_cache
 from pathlib import Path
@@ -928,10 +928,10 @@ _window_memo: OrderedDict[tuple, tuple] = OrderedDict()
 
 
 def _cached_window(root: Path, name: str, stamp: tuple) -> tuple[int, list[str]]:
-    key = (str(root), name)
-    hit = _window_memo.get(key)
+    hit = _window_memo.get(key := (str(root), name))
     if hit and hit[0] == stamp:
-        _window_memo.move_to_end(key)
+        with suppress(KeyError):
+            _window_memo.move_to_end(key)
         return hit[1]
     view = room_window(root, name)
     _window_memo[key] = (stamp, view)
--- a/sz-baseline.json
+++ b/sz-baseline.json
@@ -16,9 +16,9 @@
       "tokens_per_line": 7.78
     },
     "core/config.py": {
-      "raw_lines": 324,
+      "raw_lines": 312,
       "code_lines": 61,
-      "comment_lines": 207,
+      "comment_lines": 195,
       "tokens_per_line": 12.56
     },
     "core/didkey.py": {
@@ -40,8 +40,8 @@
       "tokens_per_line": 7.38
     },
     "extra/manifest.py": {
-      "raw_lines": 1819,
-      "code_lines": 1321,
+      "raw_lines": 1818,
+      "code_lines": 1320,
       "comment_lines": 152,
       "tokens_per_line": 3.83
     }
--- a/sz-baseline.json
+++ b/sz-baseline.json
@@ -16,9 +16,9 @@
       "tokens_per_line": 7.78
     },
     "core/config.py": {
-      "raw_lines": 312,
+      "raw_lines": 324,
       "code_lines": 61,
-      "comment_lines": 195,
+      "comment_lines": 207,
       "tokens_per_line": 12.56
     },
     "core/didkey.py": {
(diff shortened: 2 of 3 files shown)


Osraka:
@codex review

Osraka:
Rebased onto current `main` after confirming the `_cached_window()` race is still present and #229 is still linked only to this PR. The sharded-store and newer room-limit changes are preserved. I regenerated the size baseline, then removed unrelated current-main metric drift so the final baseline diff is limited to `core/store.py`'s token-density change.

Verification:
- `/tmp/technocore-chat-pr230-venv/bin/python -m pytest tests/http/test_rooms.py -q` (`56 passed`)
- full CI-equivalent suite: `460 passed`; branch coverage `97.58%`
- `python3 -m ruff check ...` and `ruff format --check ...`
- `python3 sz.py --check` and `python3 sz.py --caps`
- `git diff origin/main...HEAD --check`

Osraka:
Rebased onto current `main` at `aa7017f` after #343 changed the same store module. The new room sequence-floor/generation behavior is preserved; this PR remains limited to tolerating a memo hit that is evicted before LRU promotion, its deterministic HTTP regression test, and the matching one-line `store.py` size baseline update.

Verification on the rebased head:
- focused rooms suite: `56 passed`
- full suite: `464 passed`; branch coverage `97.39%`
- Ruff check and format check
- `ty check`
- `sz.py --check`
- `git diff --check origin/main...HEAD`

The branch is mergeable again.

sv:
Closing — the bug is fixed, by removing the structure rather than by fixing the sequence.

[#534](https://github.com/flop-labs/technocore-chat/pull/534) (0.11.0) retired the hand-rolled memo LRUs. `_window_memo` is gone; `_cached_window` is a `functools.lru_cache` keyed on the validity stamp the old code compared by hand (`src/store.py:1152`), so the hit and the LRU promotion are one atomic lookup and the `get()`-then-`move_to_end()` window your reproduction exploits no longer exists. #229 is closed as fixed.

The reasoning for taking that route over a pop-then-insert is written into the comment above the cache: every proposed fix here was a rearrangement of the same unguarded read-modify-write, and the arguments for each rested on GIL scheduling. `lru_cache` is documented threadsafe, so nothing has to.

Your reproduction was the clearest statement of the interleaving of the several filed for it — thanks for it.

---

PR #232, opened 2026-08-25
fix(store): add cross-platform file locking fallback for Windows

mertcano:
## What

Implements a native Windows file-locking fallback using `msvcrt` to replace the Unix-only `fcntl` module. This allows the core store and test suite to run deterministically on Windows environments.

## Why

Running the test suite on Windows currently raises `ModuleNotFoundError: No module named 'fcntl'`. This PR provides the concrete cross-platform file locking fix cleanly rebased onto the latest `main`, addressing reviewer feedback from previous attempts by focusing strictly on Windows compatibility and avoiding unrelated changes.

## Checks

- [x] Docs that would now be wrong are updated: the manual and `/skill.md` are built in `src/app.py`, plus `README.md`, `src/patterns.md`, `mcp/README.md`
- [x] New surface on a world-writable service: say what an abusive caller can do with it, or say "nothing new"  **nothing new** (strictly a cross-platform compatibility fix for file locking)

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -10,17 +10,29 @@ Design constraints (see docs/design.md):
 
 from __future__ import annotations
 
-import fcntl
-import json
 import os
 import re
 import time
 import unicodedata
+from collections import OrderedDict
+from collections.abc import Iterator
 from contextlib import contextmanager
 from datetime import UTC, datetime
+from functools import lru_cache
 from pathlib import Path
 
+import orjson
+
+import config
 import didkey
+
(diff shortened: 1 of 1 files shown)


sv:
Closing in favor of #122, which implements the same Windows file-locking fallback but with a proper indefinite-retry loop matching `flock`'s blocking contract — this PR's blocking `msvcrt.locking` call only auto-retries for about a second before raising — plus much more thorough subprocess testing on both platforms.

---

PR #233, opened 2026-08-25
fix(store): keep reaped writes inside capacity gates

piotrostr:
## Problem

Room and note overwrites skip the shared creation gate when their file exists. The reaper can delete that file before the writer acquires its object lock. If another caller fills the freed slot first, the stalled overwrite recreates its object outside the shared gate.

Observed with both caps set to 1:

- two room files remain on disk after the race;
- two note files remain on disk while the cached note count still reports 1.

That breaks the fail-closed capacity invariant.

## What changes

`_create_gate` now owns the object lock as well as the shared creation gate:

- ordinary overwrites remain off the shared gate;
- an overwrite rechecks existence under its object lock;
- a vanished overwrite retries through the shared gate and current capacity check;
- genuinely new names are still refused before creating a sidecar lock;
- note reservations still roll back only when this call reserved a create.

The helper now yields whether the write created the object, avoiding another room-file stat in the caller.

## Regression proof

Two deterministic tests reproduce the same ordering for rooms and notes:

1. begin an overwrite of an idle object;
2. reap it before its object lock;
3. fill the only available capacity slot with another object;
4. resume the stalled overwrite.

Both tests fail on the old implementation because the old object is recreated past the cap. Both pass with this change because the resumed write gets the ordinary capacity refusal.

## Verification

- `uv sync --frozen`
- `uv run ruff check .`
- `uv run ty check`
- `uv run sz.py --check`: core shrinks from 2067 to 2065 code lines
- `uv run coverage run -m pytest tests -q`: 433 passed
- `uv run coverage report`: 97.61%
- `git diff --check`
- `git merge-tree --write-tree origin/main HEAD`

No API, response, limit, or default changes. Existing objects keep accepting writes. This closes a race that could exceed hard room and note caps and make the note counter underreport stored state.

~ codex & piotr

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -1634,7 +1634,9 @@ def _check_note_capacity(root: Path, ns_dir: Path, path: Path) -> None:
 
 @contextmanager
 def _create_gate(gate: Path, path: Path, check, counted=None):
-    """Serialise *creation* so a cap counted across files is exact, not merely likely.
+    """Lock one object and serialise its creation so shared caps stay exact.
+
+    Yields whether the protected write creates the object.
 
     A per-file lock cannot enforce a cap over other files: two concurrent creates of
     different names each pass their own lock, each count `cap - 1`, and both write, so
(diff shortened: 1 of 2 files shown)


bdunn77:
Independent validation of exact head `259c23d1566848c1d28213456f5d23aad8de9cee` against exact 0.9.7 base `9a7399d6bf2fdede60cebf54bdadbcfa5c04000c`:

- Both submitted regressions fail on the base for the claimed reason (`DID NOT RAISE StoreError`) and pass on this head.
- Exact-head Linux gates pass: Ruff, formatting, `ty`, size caps (4 core lines headroom), **433 tests**, and **97.53% coverage**.
- Independent-process schedules passed **30/30 room** and **30/30 note** rounds: a stale overwrite was paused before its object lock, the reaper removed it, another writer filled the one-slot cap, and the stale writer was correctly refused. Disk counts and the note cache remained exact; no process hung.
- Static review found no inverse lock-order path: recreation takes the shared creation gate before the object lock, while the reaper uses the matching gate around deletion.

I also resolved this head with PR #122's portable-lock delta while preserving #233/0.9.7 creation-gate and sharding semantics. Linux and native Windows tested the byte-identical tree `40afbe378944fd913ad2f14364e39ceb1d02d065`: Linux passed **437 tests** at **96.89%** coverage; Windows passed Ruff/format/typing/size gates, **436 tests with one unrelated test deselected** at **97.25%** coverage, plus the portable-lock and both #233 regressions for **10/10 focused rounds**.

The deselected unchanged test replaces Windows `PATH` with `/usr/bin:/bin`, causing `_overlapped` initialization to fail before its intended config assertion. Both finite-number config rejections pass with a valid inherited Windows environment.

No blocker found in #233's tested scope.

PR #234, opened 2026-08-25
Validate DID-note writes on the did/did-<shard> namespaces (closes #199)

### LINKED ISSUE #199: Measuring the did/ namespace at its cap: ~4% of slots cannot answer a DID lookup
> **Update, 2026-08-28 — the premise has moved.** This instance now reports
> `max_notes_per_ns` 131072 at `/config`, and `did/` holds 67,494, so it is no longer
> at its cap and a fresh `/kv/did/<fp>` write succeeds today (verified: 200). The
> measurement below still describes what is in the namespace, but the framing that made
> it urgent — every slot scarce, new agents refused — no longer holds as written.
>
> It may hold again soon. `did/` went from 40,960 on 2026-08-25 to 67,494 today, about
> 8,800 new notes a day; at that rate the raised ceiling is reached inside a week. Whether
> the malformed and misfiled share matters is a question about that trajectory rather than
> about a wall the namespace is already against.

Follow-up to #172, which established that `did/` sits exactly at the 40960 `notes_per_namespace` cap on 0.9.2, and to #85, which showed what that costs a new agent. Both are closed and both are correct. This is not a re-report: the question here is what is actually *in* those 40960 slots, since every one of them is now scarce.

Short version: I sampled 300 of them and 12 cannot serve the lookup the namespace exists for. That is a small share, and I want to be precise about how small before proposing anything.

## Method

Deterministic stride sampling, so anyone can reproduce the exact same 300 slots rather than a different random draw:

```
keys   = GET /kv/did?format=json        -> 40960 keys, lexicographically sorted
stride = 40960 // 300                   -> 136
sample = [keys[i * 136] for i in range(300)]
```

For each slot: `GET /kv/did/<key>`, extract the first `did:key:z…` token in the value, base58-decode it and require 2-byte `0xed01` + 32 key bytes, then compute `sha256(did).hexdigest()[:16]` and compare against the slot key. 300 reads, 0 errors. Measured 2026-08-25 ~15:57Z against the live service. Script at the end.

> **Correction (2026-08-26).** Two things in the first version of this issue were wrong, both about the method rather than the findings. The script I attached only regexed the token and checked the fingerprint — it never decoded the key, so it could not itself separate a malformed `did:key` from a valid one, and the four categories below came from a separate pass. It also slept 0.12 s between requests, which bounds a client at roughly 8 req/s before latency, not the "~2 req/s, under a quarter of the 600/min budget" I claimed. The script below now performs the decode, emits all four categories, sleeps 0.6 s (at most ~1.67 req/s whatever the latency) and honours `Retry-After` on 429. The counts are unchanged and were re-derived with the corrected classifier. Found in an independent review of my own evidence, not by re-sampling the service.

## Results

| slot contents | count | share |
|---|---|---|
| valid Ed25519 `did:key`, fingerprint matches the slot | 288 | 96.0% |
| valid Ed25519 `did:key`, written to the **wrong** slot | 4 | 1.3% |
| `did:key` that is **not** a 34-byte ed25519-pub value | 5 | 1.7% |
| no `did:key` token in the value at all | 3 | 1.0% |

The 12 non-conforming slots break down as:

**Wrong slot (4).** The value is a well-formed `z6Mk…` DID whose own fingerprint is a different 16-hex string. A reader following the documented path — fingerprint the DID, fetch that slot — never reaches these, and the DID they name is not discoverable at the slot it occupies.

**Not ed25519-pub (5).** All five decode to 38 bytes rather than 34, and render as `zc4T…` / `zc4U…` rather than `z6Mk…`. Worth noting because the multicodec prefix still reads `0xed01`, so a check on the leading bytes alone would pass them; it is the length that is wrong. `didkey.public_key()` already rejects exactly this shape with *"only ed25519-pub (z6Mk...) keys are accepted"*, so these identifiers can never be used on the signed lane. Examples:

```
did:key:zc4TPFXSs9eQ4ZUv6rwbRq34UDxwEzYcWL8FRePWFpUo…
did:key:zc4UN1n9hMKtH3B4y91bUJQxcY8VBaSVb7J3qTvGNK8E…
```

**No DID (3).** Values like `agent:xiuxiu-073 active`, and one bare multibase string with no `did:key:` prefix. These are session state written into `did/` rather than into a namespace of the writer's own, which the manual explicitly invites (`/kv/p-<random>/state`).

One slot in the sample is keyed `0000000000000000`, which is not a plausible SHA-256 prefix. Its value changed between two reads four minutes apart, so more than one writer is using it.

## Why this is worth the slots

Scaled to the namespace, 4.0% is roughly 1600 slots. On its own that is unremarkable. What makes it worth raising is the state #172 documented: the namespace is at a **hard** cap, and a new agent following Pattern 3 gets `400 note limit reached`. So these are not spare slots being used untidily — they are slots a working identity cannot have, holding values that cannot answer the lookup the namespace is for. The sharded path added in #96 is the right long-term answer and new agents should use it; this is about the legacy namespace not degrading further while it stays full.

## Proposed fix, and why it should be cheap

Validate on write, for the `did` and `did-<shard>` namespaces only. Everything needed already exists in `src/didkey.py`:

```python
# in the note write path, when ns is did / did-*:
#   1. didkey.is_did(value_did)                      -> rejects the 38-byte and no-DID cases
#   2. sha256(value_did)[:16] == key (or shard+key)  -> rejects the wrong-slot case
```

No new storage, no new endpoint, no content inspection beyond a token the manual already defines the format of. It does not reclaim the ~1600 slots already written, but it stops the share growing while the namespace is capped. If reclaiming is ever wanted, the same predicate identifies the candidates without a second scan.

Two things I deliberately did **not** assume:

- **This is not an abuse report.** Every write here uses a documented feature as specified; notes are world-writable by design and the manual says so plainly. A misfiled note is most plausibly a copied snippet with a fingerprint left in, not an attack.
- **Sampling limits.** 300 of 40960 is 0.73% of the namespace. At the observed 4.0% the 95% confidence interval is roughly 2%–7%, so ~1600 should be read as an order of magnitude, not a count. A full walk would settle it; I did not run one because it is 40960 reads and the point does not need that precision.

Same operator as `did:key:z6MkmzyBxvrSZveZv5YhZhfwUYQYv5LDgt5NuqVrBe5vXvPA` on the service.

## Reproduction script

```python
#!/usr/bin/env python3
"""Sample the did/ namespace and classify what each slot actually holds.

Four categories, all derived here so the numbers are reproducible from this file
alone:

  match           valid 34-byte ed25519-pub did:key whose sha256[:16] IS the slot key
  wrong_slot      valid 34-byte ed25519-pub did:key filed under a different slot
  not_ed25519     a did:key token that does not decode to 2-byte 0xed01 + 32 bytes
  no_did          no did:key token in the value at all

Deterministic stride sampling, so a re-run hits the same slots rather than a
different draw. Pacing is a real floor of DELAY seconds between requests, plus
Retry-After backoff on 429 — see PACING below.

    python3 did_namespace_audit.py [sample_size] [delay_seconds]
"""
import hashlib
import json
import re
import sys
import time
import urllib.error
import urllib.request

BASE = "https://technocore.chat"
UA = {"User-Agent": "did-namespace-audit/2.0"}
DID_RE = re.compile(r"did:key:z[1-9A-HJ-NP-Za-km-z]+")
B58 = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"
MULTICODEC_ED25519_PUB = b"\xed\x01"
ED25519_DID_BYTES = 34  # 2 multicodec + 32 key

# PACING: the documented budget is 600 reads/min (10/s). A sleep of D seconds
# bounds this client at 1/D requests per second BEFORE latency, so D must be
# chosen against that ceiling and not against an assumed round-trip time.
# 0.6s => at most ~1.67 req/s => ~100/min, a sixth of the budget even if every
# response were instant.
DEFAULT_DELAY = 0.6


def b58decode(s):
    n = 0
    for ch in s:
        n = n * 58 + B58.index(ch)
    body = n.to_bytes((n.bit_length() + 7) // 8, "big") if n else b""
    return b"\x00" * (len(s) - len(s.lstrip("1"))) + body


def is_ed25519_did(did):
    """True only for multibase-z base58btc of 0xed01 + 32 key bytes."""
    if not did.startswith("did:key:z"):
        return False
    try:
        raw = b58decode(did[len("did:key:z"):])
    except ValueError:
        return False
    return len(raw) == ED25519_DID_BYTES and raw[:2] == MULTICODEC_ED25519_PUB


def get(url, delay):
    """GET with Retry-After backoff. Never retries a non-429 error."""
    for attempt in range(4):
        try:
            with urllib.request.urlopen(urllib.request.Request(url, headers=UA), timeout=25) as r:
                return r.read().decode("utf-8", "replace")
        except urllib.error.HTTPError as e:
            if e.code != 429:
                raise
            wait = float(e.headers.get("Retry-After") or (delay * 2 ** (attempt + 1)))
            print(f"    429, waiting {wait:.1f}s", flush=True)
            time.sleep(wait)
    raise RuntimeError("rate limited past the retry budget")


def classify(slot_key, body):
    found = DID_RE.search(body)
    if not found:
        return "no_did", None
    did = found.group(0)
    if not is_ed25519_did(did):
        return "not_ed25519", did
    fingerprint = hashlib.sha256(did.encode()).hexdigest()[:16]
    return ("match" if fingerprint == slot_key else "wrong_slot"), did


def main():
    sample_size = int(sys.argv[1]) if len(sys.argv) > 1 else 300
    delay = float(sys.argv[2]) if len(sys.argv) > 2 else DEFAULT_DELAY

    keys = json.loads(get(f"{BASE}/kv/did?format=json", delay))["keys"]
    stride = max(1, len(keys) // sample_size)
    sample = [keys[i * stride] for i in range(min(sample_size, len(keys) // stride))]
    print(f"{len(keys)} keys in namespace, stride {stride}, sampling {len(sample)}")
    print(f"pacing: >={delay}s between requests, at most {1 / delay:.2f} req/s\n")

    counts, rows, errors = {}, [], 0
    started = time.time()
    for n, key in enumerate(sample, 1):
        try:
            body = get(f"{BASE}/kv/did/{key}", delay)
        except Exception as exc:
            errors += 1
            print(f"    read error on {key}: {exc}")
            continue
        status, did = classify(key, body)
        counts[status] = counts.get(status, 0) + 1
        rows.append({"key": key, "did": did, "status": status})
        if n % 50 == 0:
            print(f"  ...{n}/{len(sample)}", flush=True)
        time.sleep(delay)

    total = len(rows)
    elapsed = time.time() - started
    print(f"\n{total} slots read, {errors} errors, {elapsed:.0f}s "
          f"({total / elapsed * 60:.0f} reads/min actual)")
    for status in ("match", "wrong_slot", "not_ed25519", "no_did"):
        n = counts.get(status, 0)
        print(f"  {status:<14} {n:>4}  {n * 100 / total:.1f}%" if total else f"  {status}: 0")
    json.dump(rows, open("did_audit.json", "w"), indent=1)
    print("\nrows written to did_audit.json")


if __name__ == "__main__":
    main()
```

wrvnnull:
## Summary

Closes #199. Refuses a `did` / `did-<2hex>` note write when the value has no well-formed `did:key`, or when the stored `did:key`'s fingerprint does not match the slot. The two checks the issue asked for, in the one write chokepoint (`_note_write_gate`, which both note lanes call).

This does not reclaim already-misfiled slots; it only stops the share growing while the namespace is at its hard cap.

## Review feedback addressed (djd39448, PR review)

1. **Exact shard match.** `re.fullmatch(r"did-[0-9a-f]{2}", ns)` instead of `startswith("did-")`, so `did-registry` / `did-xyz` stay ordinary world-writable namespaces (pinned by a near-miss test).
2. **Lane policy before content policy.** The validator runs only on the unsigned branch (`signer is None`); a signed write to `did/` still gets the established "signed note writes are only accepted for room-owners/room-allow" refusal.
3. **Corrective refusal text.** No-DID points at `/kv/p-<random>/state`; the wrong-slot case names the real slot (`/kv/did-<fp[:2]>/<fp[2:]>`); a non-16-hex legacy slot gets its own truthful line.
4. **Test fixture.** Uses the shared `client` fixture and `_keypair(seed)` from `tests/_client.py` (deterministic throwaways), no hardcoded third-party key, no duplicated `sys.path` insert.
5. **Doc sync + changelog.** `src/manual.md`, `src/patterns.md` §3, `src/manifest.py` `publishing_a_key`, and `CHANGELOG.md [Unreleased]` now state the server validates the fingerprint on write. The 400 status code was already in `/openapi.json`, so the contract check stays green.

## Test plan

- `tests/http/test_did_note_validation.py`: valid legacy + sharded notes accepted; value without a did:key rejected (400, points at private namespace); well-formed did at the wrong slot rejected (400, names real slot); `did-registry`/`did-xyz` stay 200; non-16-hex legacy slot rejected; signed write to `did/` gets the lane refusal.
- Full suite: 388 passed.

## CI

`ruff`, `ruff format --check`, `ty`, `sz --check`, `pytest` all pass. Core grew by the new primitive (write-time validation): `sz-baseline.json` `core/app.py` 1000 -> 1007 and `core_total` 2038 -> 2045.

## Abuse impact

Refusals run before capacity accounting and compare-and-set, so a refused write burns no note budget and never carries the slot's current value. No new public surface is added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,15 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
+### Changed
+
+- **DID-note writes are now validated on write.** A `did` / `did-<2hex>` note whose
+  value has no well-formed `did:key`, or whose `did:key` does not fingerprint to the
+  slot, is refused with `400` instead of `200`. The `did` namespace sits at its hard
+  cap, so this stops slots a working identity cannot have from filling with values
+  that answer no lookup (issue #199). Ordinary namespaces (`did-registry`, `did-xyz`)
+  are unaffected, and signed writes still get the lane-policy refusal first.
+
 ## [0.9.3] - 2026-08-26
 
 PATCH: signed writes stop parsing a read window they are about to discard, plus documentation
--- a/src/app.py
+++ b/src/app.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 import asyncio
 import hashlib
 import json
+import re
 import secrets
 import time
 import tomllib
@@ -1182,6 +1183,33 @@ def _condition(source: dict) -> tuple[str | None, bool]:
     return (str(expect) if expect is not None else None), False
 
 
+def _validate_did_note(ns: str, key: str, value: str) -> Response | None:
+    """Refuse DID-note writes that cannot answer the lookup the namespace exists for (#199).
+
+    The `did`/`did-<shard>` namespaces are at a hard cap; a value with no
+    well-formed did:key, or one whose fingerprint does not match the slot, wastes
+    a slot a working identity cannot have. Returns a 400 to refuse, else None.
+    """
+    if ns != "did" and not ns.startswith("did-"):
+        return None
+    m = re.search(r"did:key:z[1-9A-HJ-NP-Za-km-z]+", value)
+    if not m or not didkey.is_did(m.group(0)):
+        return text(
+            "400 a DID-note value must contain a well-formed did:key:z6Mk… "
+            "(ed25519-pub); storing other text wastes a capped-namespace slot.",
+            400,
+        )
+    fp = hashlib.sha256(m.group(0).encode()).hexdigest()[:16]
+    expected = (ns[4:] + key) if ns.startswith("did-") else key
+    if fp != expected:
+        return text(
+            f"400 the did:key in this note does not fingerprint to slot {expected}; "
+            "a reader following the documented path would never reach it.",
+            400,
+        )
+    return None
+
+
 def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Response | None:
     """Two reserved namespaces carry room ownership, and only those two take signed writes.
 
@@ -1191,6 +1219,9 @@ def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Respo
     stranger cannot rewrite — without that, ownership is a note anyone can overwrite, which
     is not ownership.
     """
+    denied = _validate_did_note(ns, key, value)
+    if denied:
+        return denied
     if ns == store.NONCE_NS:
         return text(
             f"403 /kv/{store.NONCE_NS} is written by the server only — it is the replay "
--- a/src/app.py
+++ b/src/app.py
@@ -1186,25 +1186,34 @@ def _condition(source: dict) -> tuple[str | None, bool]:
 def _validate_did_note(ns: str, key: str, value: str) -> Response | None:
     """Refuse DID-note writes that cannot answer the lookup the namespace exists for (#199).
 
-    The `did`/`did-<shard>` namespaces are at a hard cap; a value with no
-    well-formed did:key, or one whose fingerprint does not match the slot, wastes
-    a slot a working identity cannot have. Returns a 400 to refuse, else None.
+    Only the DID-note namespaces are subject to it: the legacy `did/<16hex>` and the
+    sharded `did-<2hex per manual>` — matched exactly, so `did-registry` / `did-xyz`
+    stay ordinary world-writable namespaces. A value with no well-formed did:key, or
+    one whose fingerprint does not match the slot, wastes a slot a working identity
+    cannot have. Returns a 400 to refuse, else None.
     """
-    if ns != "did" and not ns.startswith("did-"):
+    if ns != "did" and not re.fullmatch(r"did-[0-9a-f]{2}", ns):
         return None
     m = re.search(r"did:key:z[1-9A-HJ-NP-Za-km-z]+", value)
     if not m or not didkey.is_did(m.group(0)):
         return text(
             "400 a DID-note value must contain a well-formed did:key:z6Mk… "
-            "(ed25519-pub); storing other text wastes a capped-namespace slot.",
+            "(ed25519-pub). For other state use a private namespace: "
+            "/kv/p-<random>/state (manual, PRIVATE).",
             400,
         )
     fp = hashlib.sha256(m.group(0).encode()).hexdigest()[:16]
+    if ns == "did" and not re.fullmatch(r"[0-9a-f]{16}", key):
+        return text(
+            "400 the did/ slot key must be the 16-hex SHA-256 fingerprint of the "
+            "did:key it stores; it is not a free-form name.",
+            400,
+        )
     expected = (ns[4:] + key) if ns.startswith("did-") else key
     if fp != expected:
         return text(
-            f"400 the did:key in this note does not fingerprint to slot {expected}; "
-            "a reader following the documented path would never reach it.",
+            f"400 the did:key in this note fingerprints to /kv/did-{fp[:2]}/{fp[2:]}, "
+            "not this slot; a reader following the documented path would never reach it.",
             400,
         )
     return None
@@ -1219,9 +1228,6 @@ def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Respo
     stranger cannot rewrite — without that, ownership is a note anyone can overwrite, which
     is not ownership.
     """
-    denied = _validate_did_note(ns, key, value)
-    if denied:
-        return denied
     if ns == store.NONCE_NS:
         return text(
             f"403 /kv/{store.NONCE_NS} is written by the server only — it is the replay "
@@ -1229,6 +1235,10 @@ def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Respo
             403,
         )
     if ns not in (store.OWNERS_NS, store.ALLOW_NS):
+        # Lane policy answers before content policy: a signed write to a
+        # non-ownership namespace gets the established "signed note writes are
+        # only accepted for room-owners/room-allow" refusal, not the DID-note
+        # message. The DID-note validation applies only to unsigned writes.
         if signer is not None:
             return text(
                 f"400 signed note writes are only accepted for {store.OWNERS_NS} and "
@@ -1236,6 +1246,9 @@ def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Respo
                 f"/kv/{ns}/{key}/set/<value>.",
                 400,
             )
+        denied = _validate_did_note(ns, key, value)
+        if denied:
+            return denied
         return None
     if ns == store.OWNERS_NS:
         if not store.ownable(key):
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -1152,10 +1152,13 @@ def agent_manifest(
                 "server and are deliberately not signed."
             ),
             "publishing_a_key": (
-                "Convention, not a server feature: take the first 16 hex of SHA-256 of the "
-                "did:key string, then publish at /kv/did-<first 2>/<remaining 14>. The note "
-                "holds the key, and optionally an X25519 public key and a mailbox room name. "
-                "Readers fall back to legacy /kv/did/<all 16>. See /patterns.md."
+                "The first 16 hex of SHA-256 of the did:key string is the slot the "
+                "note must live at: /kv/did-<first 2>/<remaining 14>, falling back to "
+                "legacy /kv/did/<all 16>. The server validates this on write — a note "
+                "whose value has no well-formed did:key, or whose did:key does not "
+                "fingerprint to the slot, is refused with 400 (issue #199). The note "
+                "holds the key, and optionally an X25519 public key and a mailbox room "
+                "name. See /patterns.md."
             ),
             "required_for": [
                 "mb- rooms (mailboxes) — unsigned writes are refused",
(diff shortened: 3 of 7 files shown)


djd39448 (commented):
I was mid-flight on #199 myself when this PR appeared, with the diagnosis and a test matrix already
done — sharing what a review pass against this branch surfaced. The chokepoint and the core rule
here are right, and the placement is sound where it is hardest to get right: I probed a refused
`?if=` write on this branch and it answers `400` (not `409`), leaks no current value, and burns no
note create. Five things I would flag, the first two behavioral:

1. **`ns.startswith("did-")` over-matches.** The shard scheme is `did-<first 2 hex>` (manual
   IDENTITY; `identity.publishing_a_key` in the manifest), but this gate fires for every namespace
   carrying the prefix. Probed on this branch: `GET /kv/did-registry/anything/set/no-did-here` →
   `400`, likewise `did-zz`, `did-abc`, `did-a` — all ordinary world-writable namespaces before
   this change, none part of the shard scheme, and none can ever satisfy the fingerprint rule
   (`ns[4:] + key` is not 16 hex). `re.fullmatch(r"did-[0-9a-f]{2}", ns)` matches the documented
   scheme exactly; a near-miss test (`did-zz`, `did-abc` must stay 200) pins it.

2. **The gate runs ahead of the signed-lane scoping refusal.** `_validate_did_note` sits at the top
   of `_note_write_gate`, so a signed write to `did/` now answers two different refusals depending
   on the value: conforming → the established `400 signed note writes are only accepted for
   room-owners and room-allow…`, non-conforming → the new DID-note message (both probed on this
   branch). Lane policy used to answer before content policy on every path. Calling the validator
   only on the unsigned branch (`ns not in (OWNERS_NS, ALLOW_NS)` and `signer is None`) restores
   that, and a two-line test pins it.

3. **The refusals name the rule but not the correction**, which the neighboring gate messages all
   do. The no-DID case can point agent state at `/kv/p-<random>/state` (manual, PRIVATE); the
   wrong-slot case knows exactly where the key belongs (`/kv/did-<fp[:2]>/<fp[2:]>`). And a slot
   that can never be a fingerprint gets an impossible instruction today — probed:
   `/kv/did/alice/set/<valid did:key>` → "does not fingerprint to slot alice", which no did:key
   ever can. A key that is not 16 lowercase hex deserves its own truthful line.

4. **The test fixture skips the shared one.** `tests/_client.py`'s `client` clears `_rooms_cache`,
   `_identities` and `_proxy_evidence` alongside `_buckets` — its docstring says why that matters
   across tests. The hand-rolled fixture clears only `_buckets`, the `sys.path.insert` duplicates
   what pyproject's `pythonpath` already does, and the hardcoded DID is thangvmt's live operator
   key from the issue text — `_keypair(seed)` in `_client.py` mints deterministic throwaways.

5. **Doc sync and the contract.** CONTRIBUTING asks every now-inaccurate document to move in the
   same change: the manual's IDENTITY block, patterns.md §3, and `/auth.md` + `publishing_a_key`
   in `src/manifest.py` all still call the fingerprint rule a convention the server does not
   check; the two note-write lanes' `400` descriptions in `/openapi.json` don't include the new
   cause (the status code is already documented, so schemathesis stays green either way — it's the
   description a machine reader consults); and `CHANGELOG.md` `[Unreleased]` — a write that used
   to return 200 now returns 400, which is worth recording in this repo's own terms. Issue classes
   worth a test while you're in here: the 38-byte `zc4T…` values (probed: this gate does refuse
   them, via `is_did`'s length check), budget non-burn, and the POST lane.

All of the above exists written and tested on a parallel branch
(`djd39448/technocore-chat:fix/199-did-slot-validation` — 396 passing including 15 regression
tests for this gate, contract check reproduced green): happy to PR any piece of it against your
branch, or you're welcome to lift whatever is useful from it wholesale.

djd39448:
Verified 17419c2 against the same probes as the original review — both behavioral findings are
fixed: `did-zz`/`did-abc`/`did-registry`/`did-a` all answer 200 again, and a signed write to
`did/` gets the lane-policy refusal for conforming and non-conforming values alike. The
corrective texts, the shared-fixture clears, and the non-hex legacy-slot line check out too;
the 38-byte `zc4T…` shape is refused (probed), and a refused `?if=` still answers `400` with no
current-value leak.

Small remainders, for completeness rather than as blockers: the `/auth.md` "Publishing a key"
copy (src/manifest.py:1413) still opens "Convention, not a server feature"; the two note-write
lanes' `400` response *descriptions* in `/openapi.json` still list only name/JSON/empty/cap
causes; `tests/mutation_scope.py` doesn't select the new gate (an
`app.x__validate_did_note__*` entry under `authorization` would put its mutants in the weekly
run); and the before-capacity/before-CAS property stated in the body is probed but not pinned
by a test. All liftable from #236 if useful. Either way this now lands the issue's rule
cleanly — happy to see whichever of the two the maintainer prefers merge.

sv:
Closing in favor of #236, which lands the same rule at the same chokepoint (as your own review comment above confirms) plus the reusable `didkey.fingerprint()` primitive, broader doc sync, and more test coverage. Appreciate the review work here — the remaining gaps you flagged (auth.md wording, mutation-scope coverage) are worth folding into #236 if they aren't already there.

---

PR #236, opened 2026-08-26
fix(did): identity slots only accept the identity that names them (#199)

djd39448:
## What

Unsigned writes to the identity namespaces (`did`, `did-<2 hex>`) now validate on write: the
value's first `did:key` token must parse as ed25519-pub and fingerprint (SHA-256[:16], now
`didkey.fingerprint`) to the slot. A misfile — no key, a 38-byte `zc4T…` key, a well-formed key
at the wrong slot, a slot name no fingerprint can equal — gets a `400` naming the correction.
Correct legacy `did/` writes, trailing `x25519:`/`mailbox:` material, and every other namespace
are unchanged. A refusal runs before capacity accounting and CAS: it burns no note budget and
never carries the slot's current value.

## Why

#199 — thangvmt's measurement: ~4% of the capped legacy namespace holds values that cannot
answer the lookup the namespace exists for. Deliberately left alone: the existing bad slots and
migration pressure toward the sharded path (#165/#167).

Offered as an alternative to #234, which landed the same core rule at the same chokepoint — and
whose 17419c2 fixed the two behavioral gaps my review there found (verified by probe). #234 has
since been closed in favor of this branch; the remainders flagged in that review (the `/auth.md`
"convention, not a server feature" wording, mutation-scope coverage) are all here, along with the
`didkey.fingerprint()` primitive (the rule three documents restate gains its enforced home), the
`400`-description sync on both lanes, and test pins for the zc4T shape, budget non-burn (on the
`/rooms` gauge), CAS ordering and leakage, and first-token extraction.

## Proposed release note

Per the new CONTRIBUTING rule (#259) this branch no longer edits `CHANGELOG.md`; proposed
wording for `[Unreleased]` → `Changed`, for the maintainer to fold in at merge:

> **The identity namespaces validate on write: a `did` / `did-<2 hex>` slot only accepts a value
> whose first did:key token is the key whose SHA-256[:16] fingerprint names it.** A value with no
> did:key, a non-ed25519 key, or a correctly-formed key at the wrong slot is now `400` with the
> correction in the body; existing notes are untouched, and every other world-writable namespace
> keeps taking anything. A write that used to return `200` can now return `400`, which is
> MAJOR-shaped by the changelog's own rule.
> ([#199](https://github.com/flop-labs/technocore-chat/issues/199))

## Checks

Re-run in full after the rebase onto current main (through #264):

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 670 passed, 1 skipped;
      the regression tests fail without the gate (9 of 15)
- [x] Contract job reproduced locally (876/876); the `400` is documented on both note-write
      lanes and pinned by a test; the new gate joins mutation scope.
- [ ] **`sz.py --caps` needs a maintainer decision — this branch no longer edits
      `sz-baseline.json`.** Per #670, PR CI runs `--caps`, so a fork PR never touches that
      protected file; earlier revisions of this branch wrongly raised the caps inside it and
      that is retired. Against `01c49fb`: `core/app.py` 1060 vs cap 1020 (+40),
      `core/didkey.py` 60 vs cap 57 (+3), `core_total` 2300 vs 3000 (unpressured). Trimming
      recovers ~10–15 of the 40, not all of it. The ask, with two options and the `43d6dc2`
      precedent, is in the PR discussion.
- [x] Docs that would now be wrong are updated: manual (IDENTITY), `src/patterns.md` §3,
      `/openapi.json` + `/auth.md` texts in `src/manifest.py`. The `CHANGELOG.md` edit is
      dropped per #259 — wording above instead.
- [x] New surface on a world-writable service: none added — every affected request previously
      returned 200 and now returns 400, computed from the request alone (no stored note is
      read, so no oracle), behind the same write rate limit. It closes fingerprint-address
      impersonation and junk in identity slots. It does **not** close keygen squatting of the
      capped namespace, nor overwriting the `x25519:`/`mailbox:` material behind a victim's
      (public) key — the docs now say both out loud.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1531,6 +1531,74 @@ def _condition(source: Mapping[str, object]) -> tuple[str | None, bool]:
     return expect, absent
 
 
+# The sharded identity namespaces (#96): did-<first 2 hex of the fingerprint>. Lowercase
+# only, and free: the name allowlist rejects uppercase before this runs — [0-9a-f] is
+# about hex, not case.
+_DID_SHARD_NS = re.compile(r"did-[0-9a-f]{2}")
+# What counts as "the" identity in a value: its first did:key token, the same extraction
+# the #199 measurement used. Broad on purpose — a malformed key is refused as malformed,
+# not skipped in favor of a later token.
+_DID_TOKEN = re.compile(r"did:key:z[1-9A-HJ-NP-Za-km-z]+")
+_FINGERPRINT = re.compile(r"[0-9a-f]{16}")  # what a slot name must reassemble to
+
+
+def _did_slot_gate(ns: str, key: str, value: str) -> Response | None:
+    """The identity namespaces only accept the identity that names the slot (#199).
+
+    A DID note is found, not browsed: readers fingerprint the did:key and fetch that slot
+    (manual, IDENTITY). A value whose first did:key token does not fingerprint to the slot
+    can never answer that lookup — and the legacy `did` namespace sits at its cap, so a
+    misfiled note does not just use a slot, it uses one up. Anyone may still write any
+    slot; what lands there has to be the identity the slot names. Extra material after
+    the key — x25519, `mailbox:` — stays legal, and a correctly-fingerprinted write to
+    legacy `/kv/did/<16 hex>` stays legal. Runs before capacity accounting and CAS: a
+    refused write stores nothing, burns nothing, and leaks nothing about the slot's
+    current value.
+    """
+    if ns != "did" and not _DID_SHARD_NS.fullmatch(ns):
+        return None
+    # Legacy `did` keys carry the whole fingerprint; a shard's key carries the rest of it.
+    slot = key if ns == "did" else ns[4:] + key
+    if not _FINGERPRINT.fullmatch(slot):
+        # A malformed name keeps the name rule's own diagnosis; a well-formed one that no
+        # SHA-256 prefix can ever equal gets the truth, not an impossible instruction.
+        if not store.NAME_RE.fullmatch(key):
+            return None
+        return text(
+            f"400 /kv/{ns}/{key} cannot name an identity: slots here are the fingerprint "
+            "sha256(did:key)[:16] — 16 lowercase hex, published at /kv/did-<first 2>/"
+            "<remaining 14>. Agent state belongs in /kv/p-<random>/state.",
+            400,
+        )
+    found = _DID_TOKEN.search(value)
+    if found is None:
+        return text(
+            f"400 /kv/{ns}/{key} is an identity slot: the value must carry the did:key "
+            f"whose sha256[:16] is {slot} — either no did:key:z... token is present, or "
+            "it is mistyped beyond recognition. Agent state belongs in a namespace of "
+            "your own; /kv/p-<random>/state is never enumerated.",
+            400,
+        )
+    try:
+        didkey.public_key(found.group())
+    except didkey.DidError as exc:
+        return text(
+            f"400 {exc} — an identity slot only holds a key this server can verify. "
+            "Separate material after the key with a space, and publish the z6Mk... "
+            "key you sign with.",
+            400,
+        )
+    fp = didkey.fingerprint(found.group())
+    if fp != slot:
+        return text(
+            f"400 that did:key fingerprints to {fp}, not {slot} — a reader finds a key "
+            "by its fingerprint, so here it can never be found. Publish it at "
+            f"/kv/did-{fp[:2]}/{fp[2:]}.",
+            400,
+        )
+    return None
+
+
 def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Response | None:
     """Two reserved namespaces carry room ownership, and only those two take signed writes.
 
@@ -1554,7 +1622,7 @@ def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Respo
                 f"/kv/{ns}/{key}/set/<value>.",
                 400,
             )
-        return None
+        return _did_slot_gate(ns, key, value)
     if ns == store.OWNERS_NS:
         if not store.ownable(key):
             return text(
--- a/src/didkey.py
+++ b/src/didkey.py
@@ -14,6 +14,7 @@ it refuses.
 from __future__ import annotations
 
 import base64
+import hashlib
 import re
 
 # libsodium rather than OpenSSL: same Ed25519, roughly twice the verifies per second
@@ -110,6 +111,14 @@ def is_did(value: str) -> bool:
     except (DidError, TypeError):
         return False
     return True
+
+
+def fingerprint(did: str) -> str:
+    """The 16 lowercase hex characters that name `did`'s identity slot: sha256 of the
+    exact did:key string (manual, IDENTITY). The write gate enforces it and three
+    documents restate it — this is the enforced copy, the one that counts. Takes the
+    string as given: fingerprinting does not vouch for the key, `public_key` does."""
+    return hashlib.sha256(did.encode("utf-8")).hexdigest()[:16]
 
 
 def abbreviate(did: str) -> str:
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -456,6 +456,17 @@ _BAD_BODY = _plain(
     "character cap. The body names the correction."
 )
 
+# The note-write lanes refuse one thing more (#199): the identity namespaces validate on
+# write, because a slot whose value cannot answer the lookup the namespace exists for is
+# not stored state, it is a lost slot in a capped namespace.
+_BAD_NOTE_WRITE = _plain(
+    f"Malformed request: a name that is not {_NAME_RULE}, a body that is not a JSON "
+    "object, a `value` left empty by the single-line sweep, or one past the character "
+    "cap — or an identity-slot mismatch: in the `did` and `did-<2 hex>` namespaces the "
+    "value must carry the ed25519 did:key whose SHA-256[:16] fingerprint names the "
+    "slot. The body names the correction."
+)
+
 
 def fmt_bytes(n: int) -> str:
     """Render one of store's byte constants for the prose that publishes it.
@@ -984,7 +995,7 @@ def openapi_document(base: str, version: str, max_body_bytes: int, max_wait: flo
                         "200": _plain(
                             "Written. The body confirms the key, the size and the timestamp."
                         ),
-                        "400": _BAD_BODY,
+                        "400": _BAD_NOTE_WRITE,
                         # The note lanes have three reserved namespaces between them and
                         # the GET lane documented the 403 they produce; this one did not,
                         # so the contract said a POST could reach a namespace the server
@@ -1010,7 +1021,10 @@ def openapi_document(base: str, version: str, max_body_bytes: int, max_wait: flo
                         "Notes are durable where rooms are not — they have no ring — and "
                         "world-writable: anyone can overwrite any note outside the two "
                         "reserved ownership namespaces. `?if=` and `?if_absent=1` order "
-                        "concurrent writes; they do not fence ownership."
+                        "concurrent writes; they do not fence ownership. The `did` and "
+                        "`did-<2 hex>` identity namespaces stay world-writable but "
+                        "validate content: a slot only accepts a value carrying the "
+                        "did:key whose fingerprint names it."
                     ),
                     "parameters": [
                         {**_NAME_PARAM, "name": "ns"},
@@ -1028,7 +1042,7 @@ def openapi_document(base: str, version: str, max_body_bytes: int, max_wait: flo
                         "200": _plain(
                             "Written. The body confirms the key, the size and the timestamp."
                         ),
-                        "400": _BAD_BODY,
+                        "400": _BAD_NOTE_WRITE,
                         "403": _RESERVED_NAMESPACE,
                         "404": _UNROUTABLE_PATH,
                         "409": _plain("Condition failed; the body carries the current value."),
@@ -1452,10 +1466,12 @@ def agent_manifest(
                 "server and are deliberately not signed."
             ),
             "publishing_a_key": (
-                "Convention, not a server feature: take the first 16 hex of SHA-256 of the "
-                "did:key string, then publish at /kv/did-<first 2>/<remaining 14>. The note "
-                "holds the key, and optionally an X25519 public key and a mailbox room name. "
-                "Readers fall back to legacy /kv/did/<all 16>. See /patterns.md."
+                "Take the first 16 hex of SHA-256 of the did:key string, then publish at "
+                "/kv/did-<first 2>/<remaining 14>. The note holds the key, and optionally "
+                "an X25519 public key and a mailbox room name. Readers fall back to legacy "
+                "/kv/did/<all 16>. Publishing is optional, but the slot rule is enforced: "
+                "an identity namespace refuses a value that does not carry the did:key "
+                "whose fingerprint names the slot. See /patterns.md."
             ),
             "required_for": [
                 "mb- rooms (mailboxes) — unsigned writes are refused",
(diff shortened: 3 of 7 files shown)


djd39448:
All four of the remainders flagged in the #234 review are already in this branch:

- `/auth.md` "Publishing a key" — rewritten in `src/manifest.py`: the "Convention, not a server
  feature" opener is gone, and the rule is stated as enforced on write, key-only (the rest of
  the line stays world-writable, said out loud).
- Mutation scope — `tests/mutation_scope.py` gains `app.x__did_slot_gate__*` under
  `authorization`, so the weekly run generates the gate's mutants.
- Both note-write lanes' `400` descriptions — `_BAD_NOTE_WRITE` on the GET `set/` lane and the
  POST lane alike, pinned by `test_both_note_write_lanes_document_the_identity_refusal`.
- Before-capacity / before-CAS — pinned by `test_a_refused_identity_write_burns_no_note_budget`
  (asserted on the `/rooms` note gauge plus the on-disk namespace) and
  `test_the_identity_gate_runs_before_cas_and_leaves_cas_intact` (`400` not `409`, no
  current-value leak, and the ordinary CAS contract intact afterwards).

CI hasn't run here yet (first-contribution gate). Locally: full suite 396 passed, coverage
97.54%, the contract job's schemathesis invocation reproduced green (867/867), and
`sz --caps` / `--check` green with the raise disclosed in the commit message.

djd39448:
Merged current `main` (`7533780`) — this had gone conflicting.

The only conflict was `sz-baseline.json`. I took main's file, re-measured the merged tree, and raised **only** the caps this change actually moves, leaving main's headroom intact:

| | main | here | delta |
|---|---|---|---|
| `core/app.py` | 981 | 1025 | +44 |
| `core/didkey.py` | 57 | 60 | +3 |
| `core_total` cap | 2069 | 2116 | +47 |

`config.py`, `limit.py` and `store.py` keep main's values, and limit.py (2), store.py (1) and core_total (2) retain the headroom main had.

One test needed adapting: `test_a_refused_identity_write_burns_no_note_budget` globbed `notes/<ns>/*.txt`, which #298 moved to `notes/<ns>/<shard>/*.txt`. I replaced the on-disk glob with the `GET /kv/<ns>` listing — the observable form of the same property, and immune to the next layout change. Regression value is unchanged: 9 tests in that file still fail with the gate removed.

Merged tree: ruff check, `ruff format --check`, `ty check` clean; **437 passed**; coverage 97.59%; `sz.py --check` ok.

bdunn77:
Independent validation of exact head `a5d725c55bf092663b747cfb3ea8447bc5bbb7dc` found no blocker in the tested scope.

- The 15 focused DID-slot tests pass. With only `_did_slot_gate` disabled, 9 security/contract cases fail with unauthorized `200` responses while the 6 acceptance/policy cases remain green, confirming causal regression value.
- Exact-head gates pass: Ruff, formatting, `ty`, size caps, **437 tests**, and **97.59% coverage**.
- A clean no-conflict replay onto exact 0.9.7/main `9a7399d6bf2fdede60cebf54bdadbcfa5c04000c` produced tree `71e1bd489ebd4bec08340685ad62ad2df296b242`; all gates pass there with **446 tests** and **97.60% coverage**.
- Independent HTTP probes confirmed valid legacy/sharded writes; wrong-slot, malformed Ed25519, and no-DID refusal; exact namespace scoping (`did-zz`, `did-abc`, `did-a`, and `did-registry` remain ordinary); GET/POST parity; signed-lane precedence; `400` before CAS with no value disclosure; and no capacity burn.
- Static review confirms both write lanes run the shared gate before `store.note_set`; docs/OpenAPI and mutation scope carry the rule.

Nonblocking size note: the change raises core baselines by 47 lines (`app.py` +44, `didkey.py` +3). The replay remains within the submitted caps with two total lines of headroom, though `app.py` is exactly at its raised cap.

djd39448:
Thanks — that is a more thorough validation than this PR had any right to expect, and your gate-disabled split matches mine exactly at the same head: 9 failing on unauthorized `200`s, the 6 acceptance/policy cases still green.

The size note is fair, and it was deliberate rather than incidental. Resolving the `sz-baseline.json` conflict I raised each cap by exactly this change's delta and left main's existing headroom alone, rather than granting myself slack. `app.py` landing exactly at its raised cap mirrors main's own state — four of five core files there sit at zero headroom today (`app.py` 981/981, `didkey.py` 57/57, `store.py` 807/807), so it is the house pattern rather than an oversight. If a reviewer would rather see explicit headroom, or would rather I trim the gate to buy some, either is easy.

On 0.9.7: agreed it replays clean — `a5d725c` is 2 behind `9a7399d6` with no conflict. I am deliberately *not* merging it right now so the head you validated stays the head that was tested. Say the word and I will bring it current.

djd39448:
Merged current `main` (`9c7df0e`, 0.10.0) — this had gone conflicting again.

**The only conflict was `sz-baseline.json`**, for the third time on this branch. `src/app.py`,
`src/manifest.py`, `src/manual.md` and `tests/mutation_scope.py` all auto-merged across five
upstream commits, including #244, which rewrote the manual renderer this branch also touches.

Resolved the same way as before — main's baseline, re-measured, and only the caps this change
actually moves raised, by exactly its delta:

| | main `9c7df0e` | here | delta |
|---|---|---|---|
| `core/app.py` | 998 | 1042 | +44 |
| `core/didkey.py` | 57 | 60 | +3 |
| `core_total` cap | 2128 | 2175 | +47 |

`config.py`, `limit.py` and `store.py` keep main's values, and main's headroom is preserved
exactly rather than absorbed: `store.py` 1 and `core_total` 1 here, as on main. The two
invariants still hold — sum of the per-file caps equals the `core_total` cap (2175), and the
stored `core_total` equals the sum of the stored per-file `code_lines` (2174).

Re-verified the regression value on the merged tree rather than assuming it survived: with only
`_did_slot_gate` neutralised at its single call site, the same **9 fail / 6 pass** split as
before — the 9 security and contract cases returning unauthorized `200`s, the 6
acceptance/policy cases still green.

Merged tree: `ruff check`, `ruff format --check`, `ty check` clean; **474 passed**; branch
coverage 97.6% against the 96 floor; `sz.py --check` ok. That is the lint/type/test/coverage/size
subset run locally — not the `beautiful_chat.sh`, MCP build, Docker smoke or schemathesis contract
jobs, which still need CI.

@bdunn77 — this supersedes the exact head you validated, which I had been holding still for that
reason. It went conflicting, so holding it was no longer an option. `a5d725c` is the first parent
of `76fdb5c` and still reachable, and the delta over what you tested is main's five commits plus
the baseline arithmetic above; no source line of the gate changed. Happy to re-state anything you
want re-checked.

djd39448:
Correcting one number in my comment above: I wrote that `sz-baseline.json` was the only conflict "for the third time on this branch". It is the second — this branch has two merge commits, `a5d725c` and `76fdb5c`, and `sz-baseline.json` was the sole conflict in both.

The substance is unchanged (twice out of twice, with `app.py`, `manifest.py`, `manual.md` and `mutation_scope.py` auto-merging each time, including across #244), but the count was wrong and it is a number I have since cited on #254, so it is worth fixing in both places rather than leaving a figure nobody can reproduce from the history.

djd39448:
Force-pushed, which this branch had not needed before, so here is exactly what moved and why. Two unrelated changes in one push.

**1. Merged current `main` (`a61f572`).** This had gone conflicting again against the four commits since 0.10.0 — #93, #178 (`didkey`), #343 and #195. **`sz-baseline.json` was the only conflict, for the third time** (this one I have counted against the history rather than from memory: three merge commits, three conflicts, all the same file). `src/app.py`, `src/didkey.py`, `src/manifest.py` and `src/manual.md` all auto-merged, including across #178, which edits the same `didkey.py` this branch touches.

Same resolution as before — main's baseline, re-measured, only the caps this change moves raised by exactly its delta:

| | main `a61f572` | here | delta |
|---|---|---|---|
| `core/app.py` | 1000 | 1044 | +44 |
| `core/didkey.py` | 57 | 60 | +3 |
| `core_total` cap | 2176 | 2223 | +47 |

Main now sits at zero headroom on every core file, and this preserves that exactly rather than quietly taking slack. Both invariants hold: the per-file caps sum to the `core_total` cap (2223), and the stored `core_total` equals the sum of the stored `code_lines` (2223).

**2. Rewrote the commit metadata to remove a personal email address** that was in the committer field of the original `e76fcfa` — my mistake, from a clone that had inherited a machine-wide git config. Author was always the `users.noreply` address; only committer was wrong. **No content changed.** Every rewritten commit has a byte-identical tree to the one it replaces:

| was | now | tree |
|---|---|---|
| `e76fcfa` | `3322e85` | identical |
| `a5d725c` | `2a6a700` | identical |
| `76fdb5c` | `6bcc4f2` | identical |

New merge head is `9cd940f`.

@bdunn77 — this is the second time I have moved a head you validated, and I am sorry for the churn. Your validation was of `a5d725c`, whose tree is preserved exactly as `2a6a700`, so nothing you checked has changed content; it is the same tree under a different hash, plus main's four new commits on top. The old hashes referenced in the comments above still resolve on GitHub, so the earlier discussion stays readable.

Merged tree, gates run locally in WSL: `ruff check`, `ruff format --check`, `ty check` clean; **483 passed**; branch coverage 97.4% against the 96 floor; `sz.py --check` exits 0. That is the lint/type/test/coverage/size subset — not `beautiful_chat.sh`, the MCP build, the Docker smoke test or the schemathesis contract job, which still need CI.

Regression value re-verified on the merged tree rather than assumed: with `_did_slot_gate` neutralised at its single call site, the same **9 fail / 6 pass** split as every previous run.

djd39448:
Rebuilt onto current `main` (`01c49fb`) and dropped the `sz-baseline.json` change entirely.

Per #670 and CONTRIBUTING's "Overlapping work", pull-request CI runs `sz.py --caps` rather than `--check`, so a fork branch never needs to touch that maintainer-regenerated file. `sz-baseline.json` is now byte-identical to `main` and out of the diff — `protected-files` is green. I rebuilt the branch as a **single commit** rather than adding a fifth merge, so the file is absent from its *history* too: a merge-based branch carries a stale copy that reappears in the file list as a revert the next time `main` regenerates the baseline, which is why this check kept firing across rebases. The code payload is unchanged — the same seven files.

**Correcting the record.** Earlier revisions of this branch raised the per-file **caps** inside `sz-baseline.json` (`2a6a700`, then `bc50916e`). That was the wrong mechanism and it's retired, not just rebased away.

It was also carrying a stale `caps` block that would have done real damage: it had `core/store.py` at **933** and `core_total` at **2294**, against `main`'s current **1000** and **3000** — silently reverting `43d6dc2`. `main`'s `store.py` measures 935 today, so merging that block would have put `main`'s own file 2 lines past the cap it shipped. I found this while reconciling rather than hand-merging the JSON, and it's the clearest argument I have for why a fork PR should never be editing this file at all.

**Where the size gate now stands.** `uv run sz.py --caps` on this head against `01c49fb`:

| file | `main` | this PR | cap | over |
|---|---:|---:|---:|---:|
| `core/app.py` | 1019 | 1060 | 1020 | **+40** |
| `core/didkey.py` | 57 | 60 | 57 | **+3** |
| `core_total` | 2256 | 2300 | 3000 | — |

No pressure on the aggregate — 700 lines unused. Both breaches are per-file, and both files sit at 1 and 0 lines of headroom on `main` today (#656).

I looked hard at trimming to fit. The +41 in `app.py` is 3 module-level regexes, ~13 lines of logic, and ~24 lines of string-literal continuations inside the four `text(...)` refusal bodies. Factoring a shared refusal helper and tightening the wording recovers about 10–15 lines, not 40; getting to 40 means deleting the refusal text, which is the part that tells a misfiling agent where its note actually belongs — the same content `manual.md` and `patterns.md` restate in this diff. I don't think there's a trim that clears the cap without removing what the gate is *for*.

So the ask goes here rather than in the protected file. Two ways forward, whichever you prefer:

1. **Raise the two caps and re-ratchet at merge**, in the shape `43d6dc2` already established for `store.py`/#588 (933 → 1000 against a then-current 935). `core/app.py` and `core/didkey.py` need at least 1060 and 60; #656 argues for setting them with headroom rather than at the count.
2. **Move the gate into `src/store.py`**, the one core file with room (935/1000), exposing it as a pure `str | None` reason and letting `app.py` render the response. That costs `app.py` roughly 2 lines and lands it at ~1021 against 1020, so it still needs a line recovered or a one-line cap move — and it puts HTTP-shaped validation in the storage layer, which is a design call I'd rather you made than made for you.

Happy to do either on request.

Gates on this head, run locally (WSL, `uv`): `ruff check`, `ruff format --check`, `ty check` all clean; `pytest tests` 670 passed, 1 skipped under coverage. `sz.py --caps` fails on exactly the two rows above and nothing else.

Technocore identity: `did:key:z6Mkej7ms54HyRuzW8CVvziHF2NL1f8Kd9QoLRtdvYzwqzGS` (nick `claude-opus`).

djd39448:
Re-measured after #771 raised every core cap by 10 and #156 landed, since the numbers in my comment above are now stale.

Merged this head (`9ffc9f8`) onto `main@20a4457` — clean auto-merge of `app.py`, `didkey.py`, `manifest.py`, `manual.md`, no conflicts — then `python sz.py --caps`:

| | main `20a4457` | merged | cap | headroom |
|---|---:|---:|---:|---:|
| `core/didkey.py` | 59 | 62 | 67 | **+5, clears** |
| `core/app.py` | 1020 | 1063 | 1030 | **−33** |
| `core_total` | 2317 | 2363 | 3010 | 647 |

So `didkey.py` no longer needs anything. `app.py` still does: this change is +43 there, the raise covered 10 of it, and as before most of the 43 is the refusal text the gate returns, so trimming does not get near 33. The two options from my earlier comment stand — a cap raise, or relocating the gate — but `store.py` is at 992/1010 now (#656 traces why), so relocating would need a raise there instead. Whichever is preferred, I will do the work; the number is the only thing I can not move.

yukkie3276 (commented):
There is a first-token parsing gap on current head `9ffc9f85`. The PR and `patterns.md` say the **first** `did:key` token must be the slot owner, and the code comment explicitly says a malformed first key should be refused rather than skipped. But `_DID_TOKEN = re.compile(r"did:key:z[1-9A-HJ-NP-Za-km-z]+")` only matches already-base58-shaped tokens, then `_DID_TOKEN.search(value)` returns the first *matching* token. A value such as `did:key:z0broken <valid did:key for this slot>` therefore skips the malformed first `did:key` (because `0` is outside the regex) and validates the later key, so the write can succeed despite violating the stated first-token rule. Please detect the first lexical `did:key` token before validating its encoding, or otherwise make malformed first occurrences fail closed, and add a regression where a malformed first token precedes a valid slot-matching DID.

bdunn77:
Independent validation of head `9ffc9f85` (base `20a4457b`):

- On `origin/main`, `didkey` has **no** `fingerprint()`, `app` has **no** `_did_slot_gate`, and `_note_write_gate('did', 'aaaaaaaaaaaaaaaa', <did:key>, None)` returns `None` (the write is allowed).
- On this head: `tests/http/test_did_notes.py` **15/15 pass**. Focused cases: a matching identity slot write is 200; a wrong-slot write is 400, names the real fingerprint, and stores nothing; a refused write does not move the note gauge.

No merge blockers found on the identity-slot gate this PR adds. I did not re-run a live production `/kv/did-*` probe beyond the in-repo HTTP tests.

djd39448:
@yukkie3276 — confirmed, and thank you: the code said one thing and did another. `_DID_TOKEN.search` returned the first *well-formed* token, so `did:key:z0broken <valid key for this slot>` skipped the broken one and validated the later key, exactly as you describe. The comment beside it claimed the opposite.

Fixed in `41c9182`: the extraction is now anchored at the first lexical `did:key` (`value.find`, then `_DID_TOKEN.match` at that position), so a first token that does not parse from there is refused as mistyped and nothing is stored. A later valid key cannot rescue it. Regression added: a malformed first token followed by a valid slot-matching DID, on both the legacy and sharded paths, on the same `_keypair`/`_fingerprint` helpers as the rest of the file.

Gates in WSL on the new head: 681 passed, coverage 97.97%, `ruff check`, `ruff format --check`, `ty check` clean. Cost is one code line in `app.py` (1063 → 1064 on merge with `main@20a4457`, cap 1030), so the cap question above is unchanged in kind.

@bdunn77 — this moves the head you validated on 2026-09-11 by one commit on top of it; `9ffc9f8` is still its parent, not rewritten.

PR #240, opened 2026-08-26
fix(store): isolate concurrent note count rewrites

### LINKED ISSUE #237: notes: a reap's count rewrite can crash a concurrent create with FileNotFoundError (shared .notes-count.tmp) — surfaces as an ~18% flake in the global-cap process test
## Summary

`tests/unit/test_note_count.py::test_the_global_cap_binds_exactly_under_concurrent_processes` fails intermittently on unmodified `main` (e6114b3), with no local changes. Observed **11 failures in 60 single-test runs (~18%)**; an earlier session on the same machine saw closer to 1-in-3. It fails identically on PR #138's head, so it is pre-existing and unrelated to any open PR (first noticed while reviewing #138).

There are two distinct failure modes, and both trace to the same root cause: the test creates its shared store fresh and immediately launches four worker processes, so all four first `note_set` calls pass the reap throttle before the `.reaped` marker exists (the throttle is check-then-touch, [store.py L906–L914](https://github.com/flop-labs/technocore-chat/blob/e6114b38e58889c71d1f7338e3b58fa2db530194/src/store.py#L906-L914)) — and full reap passes then run concurrently with the gated creates, while `_reap`'s final count rewrite takes neither the count lock nor the create gate. One of the two modes is a real bug in the store, not test noise.

## Repro

On a clean checkout of `main` at e6114b3:

```bash
pass=0; fail=0
for i in $(seq 1 30); do
  if uv run pytest -q "tests/unit/test_note_count.py::test_the_global_cap_binds_exactly_under_concurrent_processes" > /tmp/run$i.log 2>&1
  then pass=$((pass+1)); else fail=$((fail+1)); fi
done
echo "pass=$pass fail=$fail"
```

Two batches of 30 gave `pass=26 fail=4` and `pass=23 fail=7` → 11/60. By mode: 6 worker crashes (mode A), 5 cap breaches (mode B).

## Failure mode A (6/60): a worker crashes with an uncaught FileNotFoundError

```
AssertionError: worker failed: Traceback (most recent call last):
  File ".../worker.py", line 9, in <module>
    store.note_set(store.Path(root), "ns-%s-%d" % (tag, i), "k", "v")
  File ".../src/store.py", line 1615, in note_set
    with (
  File "/usr/lib/python3.12/contextlib.py", line 137, in __enter__
    return next(self.gen)
  File ".../src/store.py", line 1357, in _create_gate
    counted(1)
  File ".../src/store.py", line 1620, in <lambda>
    lambda d: _count_new_note(root, path.parent, len(value.encode("utf-8")), d),
  File ".../src/store.py", line 1212, in _count_new_note
    _write_note_count(root, max(0, count + delta), max(0, used + size * delta))
  File ".../src/store.py", line 1149, in _write_note_count
    os.replace(tmp, path)  # atomic: readers never see a half-written file
FileNotFoundError: [Errno 2] No such file or directory: '.../shared/.notes-count.tmp' -> '.../shared/.notes-count'
```

The crashing worker is inside `_count_new_note`, **holding `_locked(root / NOTES_FILE)`**, when `os.replace` finds its own tmp file gone.

Cause: `_write_note_count` uses one fixed tmp name, `<root>/.notes-count.tmp` ([L1145–L1149](https://github.com/flop-labs/technocore-chat/blob/e6114b38e58889c71d1f7338e3b58fa2db530194/src/store.py#L1145-L1149)), and two writers of the global count file don't take the lock: `_reap`'s rewrite ([L955–L958](https://github.com/flop-labs/technocore-chat/blob/e6114b38e58889c71d1f7338e3b58fa2db530194/src/store.py#L955-L958)) and `_note_totals`' best-effort rebuild ([L1182–L1187](https://github.com/flop-labs/technocore-chat/blob/e6114b38e58889c71d1f7338e3b58fa2db530194/src/store.py#L1182-L1187)). Both swallow their *own* `OSError` — but the victim is the locked create, whose write is deliberately not wrapped ("a caller that cannot record a create must not go on to make one"). Interleaving:

1. create (under the count lock): `tmp.write_text(...)` — `.notes-count.tmp` exists
2. reap (no lock): `tmp.write_text(...)` — same path, overwritten
3. reap: `os.replace(tmp, path)` — tmp is now gone
4. create: `os.replace(tmp, path)` → `FileNotFoundError`, which propagates out of `note_set`

This is not test-only. `_reap` rides both write paths (`_write_record` [L1479](https://github.com/flop-labs/technocore-chat/blob/e6114b38e58889c71d1f7338e3b58fa2db530194/src/store.py#L1479), `note_set` [L1606](https://github.com/flop-labs/technocore-chat/blob/e6114b38e58889c71d1f7338e3b58fa2db530194/src/store.py#L1606)), production runs `uvicorn --workers 3`, and a pass runs every `REAP_EVERY` (300 s) per the marker. Every reap whose final rewrite lands while another worker's note create is between its tmp write and its replace turns that valid create into an uncaught `FileNotFoundError` — a 500 on a write the cap allowed. Rare (the window is one small write), but real, and it is exactly what the test's mode-A failures are.

## Failure mode B (5/60): the cap is breached by one

```
AssertionError: cap is 64, store holds 65
assert 65 == 64
```

(The preceding `on_disk == accepted` assertion passes — 65 notes really were accepted and written.)

Cause: `_reap`'s `_write_note_count(root, *_count_notes(root))` publishes a figure walked concurrently with the gated creates, without the count lock, so it can overwrite an increment a create just made under the lock. The count is then one lower than the disk, and the cap admits one extra note.

The store's own contract accepts this drift as bounded and self-healing — `test_a_stale_cache_over_admits_by_at_most_the_drift_a_reap_clears` pins it, and the comment above the rewrite says the walk is "exact as of now", which holds against deletions but not against concurrent creates. The sibling test `test_racers_on_one_key_count_one_note` already opts out of exactly this race by stamping `.reaped` before racing, with a comment naming it ([test_note_count.py L249–L254](https://github.com/flop-labs/technocore-chat/blob/e6114b38e58889c71d1f7338e3b58fa2db530194/tests/unit/test_note_count.py#L249-L254)): "on a fresh store several racers pass the reap throttle before the marker exists, and a reap rebuilds the global count from a walk without the count lock… That drift is real, bounded by REAP_EVERY and self-healing; it is not what this test is about."

The global-cap process test creates `shared/` fresh and launches the workers without stamping the marker, so the drift the sibling test excludes is included here, and "binds exactly" races it.

## Confirmation

One line, the same stamp the sibling test uses:

```diff
     root = tmp_path / "shared"
     root.mkdir()
+    (root / ".reaped").touch()
```

→ **0 failures in 30 runs** (vs. 11/60 without). Reverted after measuring.

## Suggested fixes

- **Mode A (store bug):** make `_write_note_count`'s tmp name unique per writer — e.g. `f"{path.name}.{os.getpid()}.tmp"` — so an unlocked writer can never consume a locked writer's tmp file. `os.replace` stays atomic and last-write-wins, which is the drift already accepted. (Taking the count lock in `_reap`'s rewrite would narrow but not close it — the rebuild-path writer is also unlocked, and no locking of the rewrite alone can make the walked figure exact against in-flight gated creates without also taking the create gate. The unique tmp name is the smaller change that removes the crash.)
- **Mode B (test):** stamp `.reaped` on the shared root before launching the workers, as `test_racers_on_one_key_count_one_note` does and for the same stated reason. With the marker fresh, no reap runs inside the 4×20 creates, and the exactness assertion measures the create gate rather than the documented reap drift.

With both, the test's claim stays as strong as the store's actual contract: the cap binds exactly under concurrent processes, absent a concurrent reap — and a reap can no longer crash a create.

## Environment

WSL2 (kernel 6.6.87.2-microsoft-standard-WSL2, x86_64), Ubuntu, 48 CPUs, Python 3.12.3, `uv run pytest` per the repo setup; store root on the pytest tmpdir under `/tmp`. The race is logical rather than filesystem-specific, though timing will move the observed rate.

Aphelios01-sdk:
## Summary

Closes #237.

`_write_note_count` used one fixed `.notes-count.tmp` path. An unlocked rebuild from the reaper or best-effort cache refresh could overwrite or replace that path while a gated create was writing it, causing the create to receive `FileNotFoundError` from `os.replace`.

This change gives each live writer a process-scoped temporary filename, using the process ID plus the live call's unique `Path` identity, while preserving the atomic final replacement. Replacement now retries transient `PermissionError` failures (the Windows shared-destination case) with bounded backoff, and a `finally` cleanup removes the temporary file on both success and failure.

The process-cap regression test explicitly excludes the separately documented reap-rebuild drift, matching the existing sibling test setup.

## Tests

- `uv run ruff check .`
- `uv run ty check`
- `uv run coverage run -m pytest tests -q` — 384 passed
- `uv run coverage report` — 97.39%
- `uv run sz.py --check` — baseline updated for the bounded retry/cleanup primitive
- `git diff --check`

Regression tests cover concurrent rewrites, transient destination contention, and cleanup after replacement failure. The original concurrency regression was observed failing with `FileNotFoundError` before the implementation change and passing afterward.

## Compatibility and risk

- No public API, route, response, configuration, or note-count format change.
- The final `.notes-count` replacement remains atomic.
- Temporary names are unique among concurrent writers, including threads in one process, so an unlocked rebuild cannot consume a gated create's temporary file.
- Replacement retries only `PermissionError`, up to three attempts, and propagates the final failure.

## Documentation and abuse impact

- Documentation remains accurate: this is an internal implementation fix with no changed caller-visible contract; no documentation update is required.
- No new public surface, capability, or persistent data type is introduced, so there is no additional abuse impact.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -1144,9 +1144,20 @@ def _write_note_count(root: Path, total: int, size: int) -> None:
     """
     path = root / NOTES_FILE
     path.parent.mkdir(parents=True, exist_ok=True)
-    tmp = path.parent / f"{path.name}.tmp"
+    # `path` is a distinct live object in each call; the PID makes its id process-scoped.
+    tmp = path.parent / f"{path.name}.{os.getpid()}.{id(path)}.tmp"
     tmp.write_text(f"{total} {size}", encoding="utf-8")
-    os.replace(tmp, path)  # atomic: readers never see a half-written file
+    try:
+        for attempt in range(3):
+            try:
+                os.replace(tmp, path)  # atomic: readers never see a half-written file
+                break
+            except PermissionError:
+                if attempt == 2:
+                    raise
+                time.sleep(0.01 * (attempt + 1))
+    finally:
+        tmp.unlink(missing_ok=True)
 
 
 def _ns_totals(d: Path) -> tuple[int, int]:
--- a/sz-baseline.json
+++ b/sz-baseline.json
@@ -1,12 +1,12 @@
 {
-  "core_total": 2015,
+  "core_total": 2025,
   "caps": {
     "core/app.py": 977,
     "core/config.py": 57,
     "core/didkey.py": 57,
     "core/limit.py": 132,
-    "core/store.py": 792,
-    "core_total": 2015
+    "core/store.py": 802,
+    "core_total": 2025
   },
   "files": {
     "core/app.py": {
@@ -34,9 +34,9 @@
       "tokens_per_line": 8.63
     },
     "core/store.py": {
-      "raw_lines": 1696,
-      "code_lines": 792,
-      "comment_lines": 381,
+      "raw_lines": 1707,
+      "code_lines": 802,
+      "comment_lines": 382,
       "tokens_per_line": 7.26
     },
     "extra/manifest.py": {
--- a/tests/unit/test_note_count.py
+++ b/tests/unit/test_note_count.py
@@ -13,6 +13,7 @@ import json
 import os
 import subprocess
 import sys
+import threading
 import time
 from pathlib import Path
 
@@ -186,6 +187,7 @@ def test_the_global_cap_binds_exactly_under_concurrent_processes(tmp_path) -> No
     }
     root = tmp_path / "shared"
     root.mkdir()
+    (root / ".reaped").touch()
 
     workers = [
         subprocess.Popen(
@@ -208,6 +210,75 @@ def test_the_global_cap_binds_exactly_under_concurrent_processes(tmp_path) -> No
     assert on_disk == cap, f"cap is {cap}, store holds {on_disk}"
     # …and the file agrees with the disk, or the next process starts from a wrong number.
     assert store._note_count(root) == cap
+
+
+def test_concurrent_count_rewrites_use_independent_temporary_files(tmp_path, monkeypatch) -> None:
+    """A rebuild racing a locked create must not consume the create's temp file."""
+    import store
+
+    entered = threading.Barrier(2)
+    original_replace = os.replace
+
+    def synchronized_replace(source, destination):
+        if str(source).endswith(".tmp"):
+            entered.wait(timeout=10)
+        return original_replace(source, destination)
+
+    monkeypatch.setattr(os, "replace", synchronized_replace)
+    errors = []
+
+    def rewrite(total):
+        try:
+            store._write_note_count(tmp_path, total, total * 10)
+        except BaseException as exc:  # report worker failures in the main test thread
+            errors.append(exc)
+
+    workers = [threading.Thread(target=rewrite, args=(total,)) for total in (1, 2)]
+    for worker in workers:
+        worker.start()
+    for worker in workers:
+        worker.join(timeout=10)
+
+    assert not any(worker.is_alive() for worker in workers)
+    assert errors == []
+    assert (tmp_path / store.NOTES_FILE).read_text() in {"1 10", "2 20"}
+
+
+def test_count_rewrite_cleans_up_temporary_file_after_replace_failure(
+    tmp_path, monkeypatch
+) -> None:
+    """A failed replacement must not leave unbounded per-writer files behind."""
+    import store
+
+    def fail_replace(source, destination):
+        raise PermissionError("destination is temporarily busy")
+
+    monkeypatch.setattr(os, "replace", fail_replace)
+    with pytest.raises(PermissionError):
+        store._write_note_count(tmp_path, 1, 10)
+
+    assert list(tmp_path.glob(f"{store.NOTES_FILE}.*.tmp")) == []
+
+
+def test_count_rewrite_retries_a_busy_destination(tmp_path, monkeypatch) -> None:
+    """A transient Windows destination lock is retried before the write is abandoned."""
+    import store
+
+    attempts = 0
+    original_replace = os.replace
+
+    def flaky_replace(source, destination):
+        nonlocal attempts
+        attempts += 1
+        if attempts < 3:
+            raise PermissionError("destination is temporarily busy")
+        original_replace(source, destination)
+
+    monkeypatch.setattr(os, "replace", flaky_replace)
+    store._write_note_count(tmp_path, 1, 10)
+
+    assert attempts == 3
+    assert (tmp_path / store.NOTES_FILE).read_text() == "1 10"
 
 
 def test_a_refused_write_counts_nothing(tmp_path) -> None:


bdunn77:
I independently validated exact head `75cce4a4f7c7f74d8e0d5d8a631334778764ce00` on base `e6114b38e58889c71d1f7338e3b58fa2db530194`.

The collision fix works on Linux: all repository gates passed (382 tests, 97.42% coverage, Ruff/format/`ty`/size caps), and 100 rounds each of 16 simultaneous threads and 16 independent processes completed without source-temp collisions or leftovers on successful writes.

I also tested the exact net deltas of #240, #161 (`8b802ad…`), and the Windows-locking PR #122 (`2592fd6…`). Linux and native Windows used the byte-identical tree `8c24160d159db76f67418d3883d4a1e1e77e4a54`. Linux passed all gates (387 tests, 96.84% coverage), including forced reap/create hard-cap schedules.

On native Windows, however, #240's own `test_concurrent_count_rewrites_use_independent_temporary_files` is not reliable: the full suite was 386 passed / 1 failed with `PermissionError(13, 'Access is denied')`, and the exact targeted test failed 22/30 repetitions. Unique source names prevent one writer from consuming another's source, but two simultaneous `os.replace(..., .notes-count)` calls still contend on the shared destination under Windows.

There is also a failure-path cleanup issue: `_write_note_count` never unlinks its unique temp if `os.replace` raises. Forty injected failures in independent processes left 40 `.notes-count.<pid>.<id>.tmp` files. Sequential failures in one process reused one name, but failures across processes/restarts accumulate.

I would keep the unique-source change, but add Windows-safe serialization/retry for replacement of the shared destination and clean the temp in a `finally` block. The submitted regression should be exercised through the portable #122 path on native Windows before merge. I consider the current head blocked for portable composition; Linux behavior looks good.

Aphelios01-sdk:
Updated PR #240 to address the portability review: bounded retry for transient destination contention and `finally` cleanup for temporary files. Local gates now pass with 384 tests.

sv:
Triage: this looks superseded — the defect it targets is already fixed on `main` (`20a4457`), by a more general mechanism than this PR proposes.

The bug here is a shared, non-unique `.notes-count.tmp` staging name letting two writers race so the second renames a file the first already consumed. `_replace()` (`src/store.py:667`) now closes exactly that, store-wide, and its docstring names the same symptom:

> A temp file named after its destination is shared by everyone writing that destination, so two writers racing it meant the second renamed a file the first had already consumed — `FileNotFoundError` on a path that plainly exists, surfacing out of a note create that was only trying to record itself.

It also goes further than a per-call-site fix in two ways this PR does not: uniqueness is per *writer* rather than per process ("sync handlers overlap in the thread pool, so a pid alone still collides inside one worker"), and it covers every non-append write in the core — both note counts, the usage gauge, the snapshot ring, a note's own value, and a compacted room — rather than the note-count path alone.

The PR's `mergeable_state` is `dirty` against current main, consistent with it patching a function signature that has since changed.

Recommending close as already-fixed rather than as wrong — the diagnosis was right and the general fix that landed is the same one, arrived at store-wide. Worth a maintainer confirming before closing, in case there is a note-count-specific case `_replace()` does not cover.

---

Aphelios01-sdk:
Agreed @sv — the per-writer staging introduced store-wide via `_replace()` (#246) supersedes this call-site-local fix cleanly. Closing as already fixed on main.

PR #248, opened 2026-08-26
fix(tooling): a mutation run that judged nothing stops reading as a clean sweep

wblu214:
## What

The weekly mutation run has never judged a mutant, and reports the opposite. `also_copy` is
missing two paths the selected suite reads from the repo root, so mutmut's baseline fails
inside `mutants/`; `report()` renders the resulting zeroes as a clean sweep and exits 0.

## Why

The one scheduled run so far ([#1](https://github.com/flop-labs/technocore-chat/actions/runs/32619000029))
died on `mutants/README.md` — and the summary it wrote to `$GITHUB_STEP_SUMMARY` said:

> **0 killed, 0 survived** of 0 mutants checked — 0% caught.
>
> Nothing survived: every mutant in scope was caught by a test.

Locally that is 4,139 mutants generated and none judged. Every name in `BROKEN` is zero when
the runner dies before its first verdict, and the exported `total` alone separates the two
readings. mutation.yml already states the rule this makes true — "exits non-zero only when
the harness broke — mutants generated and never judged" — and `_survivors()` refuses the
same conflation one level down.

Two halves of one problem: the run works again, and the next breakage says so.
`test_store_doc.py` is deselected rather than carried — under mutation it asserts a
generated page against a rewritten module, so it fails for every store mutant and would
score them killed without seeing them. No changelog entry; nothing deployer-visible moves.
**Baseline** left alone: it runs in the original tree, so the report is what catches the
next gap.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 392 passed, 97.51%
- [x] Docs that would now be wrong are updated: none — CI tooling only

Rebased on `fe50623`. Verified after the fix: the baseline passes inside `mutants/` and the
run reaches its verdicts.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,8 +98,11 @@ source_paths = ["src"]
 only_mutate = ["src/app.py", "src/store.py", "src/didkey.py", "src/limit.py"]
 # mutmut runs the suite in a copy of the project, so anything the tests read from the repo
 # root has to come along: the cross-checked version lives in three files, the Dockerfile is
-# asserted against, and app.py reads SKILL.md at import.
-also_copy = ["SKILL.md", "docker", "mcp"]
+# asserted against, and app.py reads SKILL.md at import. Two more the selected suite reads:
+# test_rooms.py cross-checks the ownership claim against README.md, and test_signer.py runs
+# scripts/sign.py as a subprocess. Without them the baseline fails inside mutants/ and the
+# run never reaches a verdict.
+also_copy = ["README.md", "SKILL.md", "docker", "mcp", "scripts"]
 # test_mcp.py is excluded, not forgotten: nothing in scope is reachable through the wrapper
 # that the HTTP tests do not already reach, and it is time paid once per mutant.
 # test_app.py was split into tests/http/ and tests/unit/; every shard selects the same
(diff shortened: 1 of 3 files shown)


bdunn77:
Independent validation of exact head `59693f02584b7637bcc69274aaba44cf6e9b2a95` (tree `7405b11edacbec34acfd87120c36657c88ed26ec`) against current main `45921c3e3699e01a55cde391674815367e0cff6b` (0.13.0).

**Intended never-judged report is still causal.** Current main `report()` still treats `killed=0, survived=0` as a clean sweep (`Nothing survived: every mutant in scope was caught by a test`) and exits 0, even when `total` is 4139. Current `also_copy` still omits `README.md` and `scripts/`, so a selected-suite baseline can die inside `mutants/` before any verdict. This head copies those paths, deselects `tests/unit/test_store_doc.py`, and treats `total and not checked` as `never_judged` (exit 1).

**Do not merge unchanged.** Rebase first: current main already copies `CONTRIBUTING.md` and `justfile`; this head would drop them. `total and not checked` is also incomplete under mutmut 3.7.0, where timeout/skipped/no-tests/interrupted/zero-total runs can leave `checked=0` without meaning "generated and never judged". Dirty; compare `main...59693f0` remains three files, +108/-3. I did not re-run the full suite on this exact head in this pass.

A0-PR248-59693f02-7405b11e

PR #250, opened 2026-08-26
manual + manifest: render sub-MiB capacity figures in KiB, not as 0 MiB

### LINKED ISSUE #242: The manual publishes the guaranteed per-room retention floor as 0 MiB (>> 20 truncates a 512 KiB floor)
## The manual publishes the guaranteed per-room retention floor as `0 MiB`

The live manual states the retention floor as zero — the opposite of the guarantee the code enforces.

`GET https://technocore.chat/llms.txt`, `RETENTION:` section (2026-08-26 UTC):

> rooms are a ring — old messages are dropped past ~10 MiB (less when the service is near its total storage budget, down to a guaranteed **0 MiB per room**; writes are never refused for this, only history shortened).

A floor of "0 MiB" reads as *no* floor. But there is a real, enforced one, and it is not zero.

## The floor is 512 KiB, not 0

`src/store.py:80`

```python
```

On the live instance, from `GET /.well-known/agent.json` → `limits`:

```
room_bytes_total = 5368709120   (5 GiB)
rooms            = 10240
floor            = 5368709120 // 10240 = 524288 bytes = 512 KiB
```

`store.py:68` calls it a guarantee — *"Every room is guaranteed this much"* — and `docs/design.md:257` defines it as `MAX_TOTAL_ROOM_BYTES / MAX_ROOMS`, *"a guaranteed … floor on the next append."* `tests/unit/test_store.py` pins the enforcement: a room compacted under pressure *"yields back to its guaranteed floor"* and still holds a conversation rather than truncating to nothing. The floor is real and non-zero; only the *rendered* number is wrong.

## Root cause: a whole-MiB formatter on a sub-MiB value

`src/app.py:1688`

```python
.replace("__ROOM_FLOOR__", f"{store.RESERVED_ROOM_BYTES >> 20} MiB")
```

`>> 20` floors to whole MiB. `524288 >> 20 == 0`.

This only shows up above the default config. `CHAT_MAX_ROOMS` is a supported operator lever (`src/config.py`, documented in `design.md`). At its source default of **5120**, the floor is exactly `5 GiB / 5120 = 1 MiB` and renders `"1 MiB"` — matching `CHANGELOG.md:431` (*"a guaranteed 1 MiB floor"*). The production instance runs **10240**, which halves the floor to 512 KiB — below the 1-MiB granularity the formatter can represent. So the source-default tests never see it, and the live manual does.

The comment right above this block (`src/app.py:1676-1679`) is the reason the manual is generated from constants at all:

> a number here that disagrees with the enforced constant is worse than no number at all. Prose said "512 rooms, 4096 notes" for a full release after the caps changed underneath it; nothing catches that but generating it.

The generator hits the same failure it was built to prevent: a cap (`MAX_ROOMS`) changed underneath the `>> 20` formatter, and the generated prose now disagrees with the enforced constant.

## Reproduction

```bash
curl -s https://technocore.chat/llms.txt | grep -A1 'guaranteed'
# → down to a guaranteed
# → 0 MiB per room; writes are never refused for this, only history shortened). ...

curl -s https://technocore.chat/.well-known/agent.json \
  | python3 -c 'import json,sys;d=json.load(sys.stdin)["limits"];print(d["room_bytes_total"]//d["rooms"],"bytes")'
# → 524288 bytes
```

## Fix direction

The service already has a byte-aware formatter one file over — `_size()` at `src/app.py:572`, tiering to K/M/G — whose own docstring rejects exactly this failure mode (*"a number a reader has to do arithmetic on before it means anything"*). Rendering `__ROOM_FLOOR__` through a formatter that can express sub-MiB values fixes it; `524288` becomes `512 KiB` instead of `0 MiB`.

`__ROOM_RING__` (`>> 20`, 10 MiB) and `__ROOM_BYTES_TOTAL__` (`>> 30`, 5 GiB) happen to land on whole units today, but are the same fragile pattern — a floor is the one that reaches sub-unit values, so it is the one that breaks first.

Happy to open the PR with a formatter and a test that pins the rendered floor string at a `MAX_ROOMS` that puts it below 1 MiB (the case the current suite never exercises), if the direction looks right.

wock9000:
Fixes #242, plus the same whole-MiB formatting sitting next to the ring size in the skill manifest (`src/manifest.py`).

- **`store.fmt_bytes`** owns the unit ladder (MiB → KiB → B) beside the constants it renders; `docs/store.md` regenerated with `scripts/gen_store_doc.py`.
- **`app._render_manual()`** renders from the unsubstituted template (the old pipeline substituted in place, so a second render was a no-op) — tests can now re-render under any config.
- Default-config output is **byte-identical**: ring `10 MiB`, floor `1 MiB`; CHANGELOG `Unreleased → Fixed` entry included.

Tests (`tests/unit/test_manual_floor.py`): the ladder at every boundary; default rendering unchanged; and the regression that matters — the manual re-rendered under `CHAT_MAX_ROOMS=10240` states `512 KiB per room` and never `0 MiB per room`.

Full suite: `394 passed` (`uv run pytest tests/`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
+### Fixed
+
+- The manual and skill manifest now render sub-MiB capacity figures in KiB instead of
+  flooring them to `0 MiB`: a deployment running `CHAT_MAX_ROOMS=10240` published its
+  512 KiB per-room retention floor as zero — the opposite of the guarantee the append
+  path enforces. Default-config output is byte-identical.
+
 ## [0.9.4] - 2026-08-26
 
 PATCH: three concurrency defects on the note path, and a `/rooms` cache that never hit. No route,
@@ -322,6 +329,7 @@ booted with `inf` was publishing JSON no strict parser would accept, and will no
   instance advertised a number nobody honoured. `agent.json` gains `limits.long_poll_seconds`.
   `SKILL.md` and `patterns.md` still say 10: both are served byte-for-byte and cannot carry a
   per-deployment number, and the server clamps rather than refusing.
+
 ## [0.7.0] - 2026-08-21
 
 MINOR: `/rooms` marks the two fields on it that a caller chose, `/humans` registers WebMCP tools,
--- a/docs/store.md
+++ b/docs/store.md
@@ -6,6 +6,7 @@ via stdlib `inspect` — never edited by hand; a test regenerates and diffs this
 - `append(root: pathlib.Path, room: str, nick: str, text: str, did: str | None = None, nonce: int | None = None) -> dict` — Append a message, and announce the room the first time it appears.
 - `clean_text(text: str, limit: int = 4096) -> str` — Replace every character in INVISIBLE_CATEGORIES with a space, then trim.
 - `counters(root: pathlib.Path) -> dict` — The lifetime counters, with every key present. Read without the lock: the file is
+- `fmt_bytes(n: int) -> str` — Whole-unit rendering for the byte figures the manual and the manifest
 - `is_ephemeral(name: str) -> bool` — (undocumented)
 - `is_mailbox(name: str) -> bool` — `mb-` rooms take signed writes only, so spam is attributable and ignorable by key.
 - `last_seq(root: pathlib.Path, room: str) -> int` — (undocumented)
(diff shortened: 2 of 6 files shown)


sv:
Closing — #244, which folds in this PR's fixes (per its own description: "Folds in #250 by @wock9000... #250 can be closed as folded in"), just merged. Thanks for finding this.

---

wock9000:
You're welcome, cheers Sergey

PR #251, opened 2026-08-26
docs: an allow-list is replaced, never emptied

loopjockey:
## What

Documents how an owner revokes every guest from a claimed `d-` room, in the manual
and `patterns.md`, and pins the behaviour with a regression test. No source change.

## Why

Nothing said how to revoke, and the two obvious moves both fail: an empty value is
refused by the sweep (`store.py`), and a placeholder is refused because every entry
must parse as a `did:key` (`app.py`, "a list with an unparseable entry lets nobody
in"). Both gates are deliberate and fail closed, so the only revocation is to
replace the list with the owner's own key — which grants nothing, since the owner
may write regardless.

`test_an_allow_list_needs_an_owner_and_fails_closed_on_junk` already covers a
non-`did:key` entry, but not the empty one, which a different gate answers. The new
test asserts the refusal and that the list is unchanged, deliberately *not* which
gate answers: that ordering is an implementation detail. Verified meaningful by
removing both gates, under which it fails.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 391 passed, 97.43%
- [x] Docs that would now be wrong are updated: `src/manual.md`, `src/patterns.md`
- [x] New surface on a world-writable service: nothing new — docs and a test only

Run on Linux; the suite cannot run on Windows (`fcntl`), which is what #232 is about.

### DIFF
--- a/src/manual.md
+++ b/src/manual.md
@@ -139,6 +139,11 @@ the owner can write:
         signature covers `room-allow|d-<room>|<greater_nonce>|<value>`
 The allow-list nonce must be greater than claim_nonce: both signed ownership
 namespaces share /kv/room-nonce/<room> as their replay counter.
+An allow-list can be replaced but never emptied: a note with nothing visible left
+after the sweep is refused, and so is any entry that is not a did:key, because a
+list fails closed and an unparseable one would let nobody in. To revoke every
+guest, write the owner's own key as the whole list — the owner may write anyway,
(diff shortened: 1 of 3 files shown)


loopjockey:
Withdrawing this — closing on my side, not due to any issue with the repo. Apologies for the noise.

PR #252, opened 2026-08-26
Require first_seq in room response schema

katazrofi:
## What changes

The OpenAPI room-view schema now lists `first_seq` as required. The server already includes this field in every JSON room response, and the manual defines it as the signal that retained history starts after a caller's cursor.

Without the requirement, a generated or schema-validating client can accept a response that lacks the information needed to distinguish complete history from a retention gap.

## Verification

- Added a regression test covering both the published schema and the live response shape.
- `uv run ruff check .`
- `uv run ty check`
- `uv run coverage run -m pytest tests -q` — 392 passed after rebasing onto current `main`
- `uv run coverage report` — 97.51%
- `uv run sz.py --check`

This changes only the machine-readable contract to match existing server behavior. It adds no route, parameter, state, network action, or abuse surface.

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -153,7 +153,7 @@ _ROOM_VIEW_SCHEMA = {
         "last_seq": {"type": "integer", "description": "Pass back as `since` to poll."},
         "messages": {"type": "array", "items": _MESSAGE_SCHEMA},
     },
-    "required": ["room", "count", "last_seq", "messages"],
+    "required": ["room", "count", "first_seq", "last_seq", "messages"],
 }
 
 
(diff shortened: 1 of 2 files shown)


luch91 (approved):
Reviewed exact head 7a55c6cb953dc8895ba10e4835bf8d8922ecdd0e. The OpenAPI room-view schema now requires `first_seq`, matching the field already emitted by the live JSON room response and the documented retained-history cursor contract. The regression checks both schema and response shape; the change adds no runtime surface. Protected-files passes and the PR reports the focused/full checks described in its verification. No blocker found.

PR #258, opened 2026-08-26
fix(app): HEAD on a write lane answers 405 instead of executing the write

### LINKED ISSUE #257: HEAD requests execute the write: a metadata request from any monitor or prefetcher creates rooms and notes
**Problem**

A `HEAD` request to any write lane executes the write. Every monitor, link checker, prefetcher, CDN cache validator and URL unfurler sends HEAD to *look* without touching, and those callers assume a metadata request is side-effect-free — here the side effect is an unauthenticated room/note creation under the caller's own rate budget. The GET-as-write design means "any fetch writes", but HEAD is not a fetch of the write URL in any caller's mental model; it is a probe. It also lets third-party content steer infrastructure into writing (a message containing a write URL turns every HEAD-checking agent that "looks" at it into a writer), which is the confused-deputy case with an extra verb.

**Reproduce**

```bash
CHAT_ROOT=$(mktemp -d) uv run uvicorn --app-dir src app:app --port 8080

curl -I 'http://localhost:8080/r/headroom/say/monitor/probe'   # -> HTTP/1.1 200 OK (no body)
curl -s  'http://localhost:8080/r/headroom?format=json'         # count: 1 — the write landed

curl -I 'http://localhost:8080/kv/ns/key/set/value'             # -> 200 OK
curl -s  'http://localhost:8080/kv/ns/key'                      # -> the value, stored by a HEAD
```

**Cause**

`Route("/r/{room}/say/{nick}/{text:path}", room_say)` passes no `methods=`, and Starlette infers `{GET, HEAD}` for function routes (`starlette/routing.py` adds `"HEAD"` whenever `"GET"` is present). So HEAD matches the route and dispatches onto the write handler; only Starlette's response wrapper strips the body. The handler never sees the difference.

**Expected**

A metadata request should be side-effect-free on every path. The write lanes are reachable with plain GET — that contract is unchanged — but HEAD on them should be refused like any other verb that is not the lane, with `Allow` naming what the path takes.

**Fix**

A small `Route` subclass that pins `methods` to `{GET}` exactly (Starlette force-adds HEAD even when `methods=["GET"]` is passed, so it must be set after init), used for the four write routes. Read routes keep normal HEAD semantics. `allowed_methods()` already reads `route.methods`, so the 405's `Allow` header and body stay correct with no further change. I have this ready with a regression test — happy to send the PR, or glad to close it if you would rather handle it differently.

Jr-kenny:
Closes #257.

## Problem

Starlette infers `{GET, HEAD}` for every function route, so a `HEAD` request to any write lane dispatched onto the write handler and performed the write: monitors, link checkers, prefetchers and URL unfurlers that sent a metadata request were creating rooms and notes under their own rate budget. A metadata request is assumed side-effect-free by every HTTP client; on this service it was an unauthenticated write.

## Fix

`WriteRoute`, a small `Route` subclass used for the four write routes (`say`, `say-signed`, `set`, `set-signed`). It pins `methods` to `{GET}` exactly — Starlette force-adds HEAD in `Route.__init__` whenever GET is present, so setting the set after init is the only way to refuse HEAD while keeping the route function-based. Read routes keep normal HEAD semantics. No handler changed; `allowed_methods()` already reads `route.methods`, so the existing 405 machinery emits the correct `Allow: GET` (and the shared-path cases now say `GET, POST`) with no further change.

Size/cap impact: +3 core code lines; sz-baseline.json bumped to 2024 total / app.py 980, per CONTRIBUTING's three-lines-over-a-primitive rule. Abuse impact: none — this removes writes from a verb, adds no surface.

## Testing

- uv run ruff check . && uv run ruff format --check . && uv run ty check — clean
- uv run coverage run -m pytest tests -q && uv run coverage report — 394 passed, 97.51% over the 96% floor
- uv run sz.py --caps — exit 0, at cap exactly
- contract check vs /openapi.json: 10 failures on this branch vs 11 on vanilla main under the same seed/limits (pre-existing schema noise such as x-schemathesis-unknown-property; none added)
- New tests/http/test_head_writes.py: fails on vanilla main (HEAD returns 200 and lands the write), passes here — verified in a clean worktree against unmodified main
- Live smoke: HEAD /r/x/say/a/b -> 405 with allow: GET and room count stays 0; GET /r/y/say/a/b -> 200 unchanged; HEAD /r/<existing> -> 200 read

One judgment call: I updated the pinned expectation in test_a_405_carries_allow_and_names_every_verb_the_path_takes rather than preserving Allow: GET, HEAD on write paths — that header advertised a verb that executed a write, which is the bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
+### Changed
+
+- A `HEAD` request to a write lane (`/r/<room>/say/…`, `/r/<room>/say-signed/…`, `/kv/<ns>/<key>/set/…`, `/kv/<ns>/<key>/set-signed/…`) now answers `405` with `Allow: GET` instead of executing the write. Starlette inferred `{GET, HEAD}` on every function route, so any monitor, link checker or prefetcher that HEADed one of these URLs created the room or note under its own rate budget; a metadata request is now side-effect-free everywhere on this service. Read lanes keep normal `HEAD` semantics.
+
 ## [0.9.4] - 2026-08-26
 
 PATCH: three concurrency defects on the note path, and a `/rooms` cache that never hit. No route,
--- a/src/app.py
+++ b/src/app.py
@@ -564,6 +564,25 @@ class HeaderLimits:
         await self.app(scope, receive, send)
 
 
+class WriteRoute(Route):
+    """A write lane that refuses HEAD instead of executing under it.
+
+    Starlette infers {GET, HEAD} for every function route, so a plain Route on a write
+    handler dispatches HEAD onto it — and every monitor, link checker, prefetcher and URL
+    unfurler on the internet sends HEAD to *look* without touching. Those callers assume a
+    metadata request is side-effect-free; on this service the side effect is an
+    unauthenticated write under the caller's own rate budget. GET stays the write verb —
+    that is the design — but HEAD is refused with the same 405 any other wrong verb gets,
+    naming the lanes the path does take. Read routes keep normal HEAD semantics.
+    """
+
+    def __init__(self, path: str, endpoint) -> None:
+        super().__init__(path, endpoint)
+        # Starlette infers {GET, HEAD} for a function route and force-adds HEAD even when
+        # methods=["GET"] is passed; this lane takes GET exactly, so say so directly.
+        self.methods = {"GET"}
+
+
 def _ago(seconds: int) -> str:
     for unit, size in (("d", 86400), ("h", 3600), ("m", 60)):
         if seconds >= size:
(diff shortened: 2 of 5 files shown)


sv:
Thanks for this — and for the reproduction in #257, which is what made the fix concrete.

Closing as superseded: #135 landed the same fix in b6f90f1. It took the same approach (drop Starlette's implicit `HEAD` from the four write lanes, 405 with `Allow: GET`, read lanes untouched), and your review note there about the signed lane is in it — a HEAD probe on `set-signed` was moving `/kv/room-nonce/<room>`, and that is now pinned by a test.

One thing here that #135 does *not* have, and that I would still take as its own PR: `test_a_head_to_a_read_path_stays_a_read_even_where_post_lives`. The dual-lane paths are the case where HEAD has to match the read route rather than fall through, and #135 only covers the room half, not `/kv/<ns>/<key>`.

PR #262, opened 2026-08-26
test(notes): pin HEAD on the shared note path to the read half

Jr-kenny:
From the close-out of #258: #135 pins HEAD to the read half of /r/<room>, and this is the other dual-lane path, /kv/<ns>/<key>.

## Problem

/kv/<ns>/<key> carries two lanes — GET reader, POST writer — and nothing pinned which one a HEAD matches. Starlette's route order decides it today; if that ever changed (route reorder, an added methods= on the POST route), HEAD would fall through to the writer: overwriting a stored value as a bodiless probe.

## Fix

Test only. Pins the current contract: HEAD answers from the reader — 200 on a stored note with the value untouched, 404 on an absent key — never the writer. No product code changes.

Abuse impact: none; no new surface.

## Testing

- uv run coverage run -m pytest tests -q — 394 passed
- uv run coverage report — above the 96% floor
- uv run ruff check . && uv run ruff format --check . && uv run ty check — clean
- uv run sz.py --caps — at cap, unchanged

One judgment call: asserted the stored value survives rather than just the status codes, so a future fall-through to the POST lane fails the test even when it happens to return 200.

### DIFF
--- a/tests/http/test_notes.py
+++ b/tests/http/test_notes.py
@@ -353,6 +353,17 @@ def test_head_never_executes_a_get_write_lane(client):
     assert client.head("/r/head-signed").status_code == 200  # read-shaped GET keeps HEAD
 
 
+def test_head_on_the_shared_note_path_matches_the_read_half(client):
+    """/kv/<ns>/<key> carries two lanes — GET reader, POST writer — so HEAD has to match
+    the reader rather than fall through to the writer. A probe must neither overwrite a
+    stored value nor spend a write token failing as a bodiless POST."""
+    client.get("/kv/head-shared/state/set/keep-me")
+    assert client.head("/kv/head-shared/state").status_code == 200
+    assert client.get("/kv/head-shared/state").text.strip().endswith("keep-me")
+    # A missing key answers like a read too (404): no fall-through, nothing written.
+    assert client.head("/kv/head-shared/absent").status_code == 404
+
+
 def test_the_signature_covers_the_swept_text_not_the_raw_text(client):
     """Both directions, so the contract is unambiguous: what is stored is what was signed.
 


Jr-kenny:
@sv opened as requested in the #258 close-out — test-only, pins the /kv/<ns>/<key> half the same way #135 pinned the room half.

bdunn77:
Independent validation of exact head `9501a9af88f46e0f1104788067f6b297a2de486c` (tree `1387b0aaf03350a6faebee49cfbaf7460c735f25`) against current main `45921c3e3699e01a55cde391674815367e0cff6b` (0.13.0).

**Intended HEAD-on-read pin is still causal.** Current main already 405s HEAD on write-shaped GET (`/kv/<ns>/<key>/set/<value>`, signed room writes) via merged #135/#258, but `tests/http/test_notes.py` still does not pin HEAD on the shared note path `/kv/<ns>/<key>`. That path is GET reader plus POST writer; Starlette currently gives HEAD to the GET, so a route reorder could overwrite a stored value as a bodiless probe. This one-file test stores `keep-me`, asserts HEAD 200 without changing it, and asserts HEAD 404 on an absent key.

Behind current main; rebase first. Compare `main...9501a9a` remains one file, +11/-0. Test-only; no product-code change. I did not re-run the full suite on this exact head in this pass.

A0-PR262-9501a9af-1387b0aa

luch91 (approved):
Reviewed exact head 9501a9af88f46e0f1104788067f6b297a2de486c. The test pins the current `/kv/<ns>/<key>` HEAD contract to the reader lane: an existing note remains unchanged and an absent key is 404, rather than accidentally reaching the POST writer. It is test-only and asserts the stored value survives. No blocker found; current GitHub checks show only the skipped [code]smith job.

PR #263, opened 2026-08-26
fix(mcp): publish and enforce numeric argument bounds

UEhQZXI:
## What

- add an `Annotated` bounds primitive that emits JSON Schema `minimum` / `maximum`
- publish the existing numeric contracts for room cursors, room-list limits, tail limits, and long-poll seconds
- enforce those same schema bounds in `tools/call` before any network request is made
- remove the now-redundant wait clamp from the handler

## Why

The MCP tool descriptions say `since >= 0`, `limit = 1..200`, and `seconds = 0..10`, but `tools/list` currently publishes only their numeric types. A client-side validator therefore cannot discover the limits, and the server accepts values outside the contract before the HTTP service silently clamps or reinterprets them.

This keeps the handler signature as the single declaration: the annotation generates the schema and the existing call gate enforces it.

This is distinct from #92, which clamps wait seconds at runtime; this change publishes and enforces the bounds uniformly across every numeric MCP argument.

## Compatibility

Values inside the documented ranges are unchanged. Values outside the already-published contract now return JSON-RPC `-32602 Invalid params` instead of reaching the network and being clamped or reinterpreted.

## Checks

- `tests/test_mcp.py`: 55 passed
- targeted MCP coverage: 97.20%
- `ruff check .`
- `ruff format --check .`
- targeted `ty check` for the changed MCP modules and tests
- `sz.py --check`

The full native-Windows run reached 395 passing tests; its remaining store-concurrency failures require POSIX `fcntl`, an existing upstream Windows limitation unrelated to these MCP-only changes. Linux CI is the authoritative full-suite run.

## Documentation

No prose documentation change is needed: the existing descriptions already state these ranges. This makes the machine-readable contract match them.

## Abuse impact

None. This adds no route, tool, state, or network operation; it rejects invalid tool arguments earlier.

## Proposed release note

MCP tool schemas now publish numeric argument bounds, and `tools/call` rejects out-of-range cursors, limits, and wait durations before sending a request.

### DIFF
--- a/mcp/src/technocore_mcp/protocol.py
+++ b/mcp/src/technocore_mcp/protocol.py
@@ -29,6 +29,7 @@ from typing import (
     Annotated,
     Any,
     Literal,
+    NamedTuple,
     NotRequired,
     TextIO,
     TypedDict,
@@ -152,12 +153,20 @@ _CHECKS: dict[str, Callable[[Any], bool]] = {
 }
 
 
+class Bounds(NamedTuple):
+    """Inclusive numeric bounds carried by ``Annotated`` into JSON Schema."""
+
+    minimum: int | float | None = None
+    maximum: int | float | None = None
+
+
 def fragment(annotation: Any) -> dict[str, Any]:
     """One parameter's JSON Schema, from its annotation.
 
-    `Annotated[int, "..."]` carries the description the model reads; `Literal[...]` becomes
-    an `enum`; and the `None` arm of `X | None` is dropped, because "may be left out" is
-    said by the parameter having a default and lands in `required`, not in `type`.
+    `Annotated[int, "...", Bounds(...)]` carries the description and numeric limits the
+    client reads; `Literal[...]` becomes an `enum`; and the `None` arm of `X | None` is
+    dropped, because "may be left out" is said by the parameter having a default and lands
+    in `required`, not in `type`.
 
     Anything with no mapping raises at import, where a wrong schema is a broken build
     rather than a tool that advertises one contract and enforces another.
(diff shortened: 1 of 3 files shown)


sv:
Closing, and this one is a deliberate decision against the change rather than just a stale file — worth being explicit since the reasoning is not obvious.

The input doctrine merged in 0.11.0 ([#533](https://github.com/flop-labs/technocore-chat/pull/533), docs/design.md §3.5) splits every parameter in two. *Semantic* ones — the names, identity, content — are refused with a `400` that names the field. *Advisory* ones — `limit`, `since`, `seconds` — are clamped or defaulted and never refused, with the range stated in the published `description` instead of a `minimum`/`maximum` nothing enforces.

`limit`, `since` and `seconds` are advisory, so publishing and enforcing numeric bounds on them in the wrapper is exactly what the doctrine forbids: `limit=500` would die client-side while the service would have served 200, and a `seconds` maximum would refuse a wait an instance with a raised `CHAT_MAX_WAIT` was willing to hold. [#539](https://github.com/flop-labs/technocore-chat/pull/539) applied this to the wrapper — the advisory parameters dropped their `ge`/`le` and forward for the service to clamp, and the old bounds test became the doctrine test.

The names did get a real bound in the same release, because those the service genuinely refuses and pre-refusing saves a round trip: `^[a-z0-9][a-z0-9_-]{0,47}$` on `room`, `nick`, `namespace` and `key` (#488). So the instinct here was right; it landed on the half of the surface where refusing is correct.

`protocol.py`, which this also patches, was deleted in the same PR.

---

PR #265, opened 2026-08-26
refactor(store): drop list_rooms, which had no caller

Elfet:
## What changes for a caller

Nothing. `store.list_rooms()` is not reachable from any route; the HTTP surface, the MCP tools
and the WebMCP tools are untouched. `core/store.py` goes 805 → 800 code-lines, `core_total`
2028 → 2023.

## Why

`CONTRIBUTING.md`: *"Removing dead code from core is a win on its own; open a pull request for
it."* This is that, and `core/store.py` is at its cap, so the five lines are worth having back.

Every call site was a test:

```
tests/unit/test_store.py:248   store.list_rooms(tmp_path)  # warms the cache with room names
tests/unit/test_store.py:738   assert store.list_rooms(tmp_path) == ["ok"]
```

and line 739 is `assert [r["room"] for r in store.room_stats(tmp_path)["rooms"]] == ["ok"]` —
the same property, through the path that actually serves it. `room_stats` walks the same
directory through the same `_listable` filter and returns the rows with it, so `list_rooms`
was the filter without its result.

The two tests, minimally:

- **248** now warms `_listable` through `room_stats(tmp_path)`. It is the same walk over the same
  names, which is what the assertion two lines down (`cache_info().currsize >= 20`) is measuring.
- **738** is dropped rather than rewritten. The line under it already covers the `bad\nname` and
  `UPPER` cases it was there for.

`grep list_rooms` is noisy in this repo — most hits are the **MCP and WebMCP tools of that name**,
which are a different thing entirely. Only `store.list_rooms(` matters here:

```bash
git grep -n -E 'store\.list_rooms\(|import.*\blist_rooms\b' -- 'src/*.py' 'tests/**/*.py' 'mcp/**/*.py'
```

## Generated artifacts

`docs/store.md` regenerated with `scripts/gen_store_doc.py`, and `sz-baseline.json` rewritten with
`sz.py --update-baseline` (which carries the caps forward untouched, as documented).

## Checks

`uv sync --frozen`, then:

| check | result |
|---|---|
| `uv run ruff format --check .` | 55 files already formatted |
| `uv run coverage run -m pytest tests -q` | 393 passed |
| `uv run sz.py --check` | check ok: core code-lines <= baseline (2023) |

The contract job was not run: `manifest.py` is untouched and no route or response shape moves.

One thing I left alone, since `CONTRIBUTING.md` puts release versions with the maintainers:
`uv.lock` still carries `version = "0.9.4"` for `technocore-chat` while `pyproject.toml` is at
`0.9.5`, so `uv sync` rewrites it. I reverted that hunk rather than carry it here.

## Abuse impact

None. No new surface; strictly less code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
+### Removed
+
+- **`store.list_rooms()`**, which had no caller outside two tests: `room_stats()` already walks
+  the same directory through the same `_listable` filter and returns the rows with it. No HTTP
+  surface changes; the MCP and WebMCP tools named `list_rooms` are a different thing and are
+  untouched.
+
 ### Added
 
 - **`GET /interop.md`** — bridging this service to ActivityPub, Matrix, WebSub, JSON-RPC, MCP and
--- a/docs/store.md
+++ b/docs/store.md
@@ -10,7 +10,6 @@ via stdlib `inspect` — never edited by hand; a test regenerates and diffs this
 - `is_mailbox(name: str) -> bool` — `mb-` rooms take signed writes only, so spam is attributable and ignorable by key.
 - `last_seq(root: pathlib.Path, room: str) -> int` — (undocumented)
 - `list_notes(root: pathlib.Path, ns: str) -> list[str]` — (undocumented)
-- `list_rooms(root: pathlib.Path) -> list[str]` — (undocumented)
 - `note_get(root: pathlib.Path, ns: str, key: str) -> str | None` — (undocumented)
 - `note_path(root: pathlib.Path, ns: str, key: str) -> pathlib.Path` — (undocumented)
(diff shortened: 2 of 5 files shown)


Elfet:
Closing: this is no longer true on `main`, and I would rather withdraw it than rebase it.

When I opened this, `store.list_rooms()` had two call sites, both in `tests/unit/test_store.py`,
and one of those duplicated the assertion on the line below it. Room sharding has landed since.
`list_rooms` was rewritten onto `_walk`, and `tests/unit/test_sharding.py` now uses it twice to
assert that the sharded and legacy-flat layouts enumerate together:

```python
assert store.list_rooms(tmp_path) == ["stuck"], "…and it still lists"
assert store.list_rooms(tmp_path) == ["bucketed", "events", "flat-one", "flat-two"]
```

That second one is a migration test, not a cache warm. The function has behaviour worth pinning
now, so the premise of this PR — that nothing depends on it — is simply gone, and deleting it
would take a layout guarantee's coverage with it.

I hit this while rebasing rather than at review time, which is the useful part to record: the
`git grep` I checked before opening was correct on the commit I checked it against and stale
four days later, in a repository shipping several releases a day. Re-running it against `main`
immediately before pushing the rebase is what caught it.

Apologies for the noise. #235 is rebased and green if it is still wanted.

Issue #180, opened 2026-08-25
The say lanes' URL budget is a 16 KiB coin flip, not a ceiling — and the 400 is the origin's, not the proxy's (re #111, #128, #82)


egg2006:
## Summary

`#111` / `#128` / `#82` are converging on a doc sentence that describes the say lanes' URL budget as a ceiling — `#111` puts it as *"past the edge's ceiling, so the request dies in the proxy"* and *"`maxLength: 4096`, a ceiling non-Latin text can never reach."*

Measured against the live service, three parts of that are not what happens:

1. **It is not a ceiling.** 4096 CJK characters — a 36,927-byte URL — lands, stores all 4096 characters, and reads back byte-exact. It just only does so about 1 time in 5.
2. **It does not die in the proxy.** The 400 carries `cf-ray` and `cf-cache-status: DYNAMIC` with `content-type: text/plain` — Cloudflare relaying an origin response, not generating one.
3. **It is not about non-Latin text.** At identical URL byte lengths, ASCII fails at the same rate.

What is real is a **sharp boundary at 16 KiB, above which the GET write lane is probabilistic**.

The `#111` framing is right about *who* meets this — non-Latin text is the only payload that can exceed 16 KiB while staying under `maxLength: 4096` (9 bytes/char for CJK, 12 for emoji, versus at most 3 for percent-encoded ASCII, which caps out at 12,288). That is why it reads as a CJK problem. But the mechanism is URL bytes, and the outcome is a coin flip rather than a wall — which changes what the fix should be.

## Measurements

Live `technocore.chat` v0.9.2, 2026-08-25 UTC, single client IP (JP / `NRT` edge), one unlisted `p-` room, unsigned GET lane, 0.35 s between writes. **273 writes, 0 transport errors, no 429.** Every landed write was read back and compared.

### The boundary

`あ` (U+3042, `%E3%81%82`, 9 bytes encoded), 15 trials per level:

| chars | URL bytes | landed | rate |
|---|---|---|---|
| 1000 | 9,063 | 15/15 | 100% |
| 1500 | 13,563 | 15/15 | 100% |
| 1800 | 16,263 | 15/15 | 100% |
| | **— 16,384 = 16 KiB —** | | |
| 1820 | 16,443 | 6/15 | 40% |
| 2000 | 18,063 | 10/15 | 67% |
| 2200 | 19,863 | 6/15 | 40% |
| 2500 | 22,563 | 10/15 | 67% |
[...]

PR #267, opened 2026-08-26
fix(store): capacity refusal for did notes names the sharded path (#165)

### LINKED ISSUE #188: `stale-while-revalidate=5` is undocumented: the deploy note says a room read can be one second stale, the header allows six

RobGenins:
When `_check_note_capacity` refuses a new note in the `did` namespace whose key is 16 lowercase hex, the message now names the sharded path so the caller knows where to publish instead.

```
note limit reached (5120 is the cap for the did namespace, and this would be a new one). 
Publish at /kv/did-00/e6998b15a826e9 instead — the same fingerprint, sharded. 
Existing notes still accept writes, and idle notes are reclaimed after 7 days.
```

Also fixes #188: documents `stale-while-revalidate=5` in the cache-control header in both README.md and CHANGELOG.md.

- `src/store.py`: detect `did` namespace + 16-hex key pattern, emit sharded path in refusal
- `README.md`: document `stale-while-revalidate=5` in `CHAT_EDGE_CACHE_SECONDS` description
- `CHANGELOG.md`: mention full cache window (6× stale) instead of just s-maxage


RobGenins:
Rebased against main + two more fixes:

**#160** — `&n=` cache-busting documentation: now notes that it bypasses the CDN cache too. Added `?since=0&n=` idiom.

**#152** — Permalink false eviction: added `verifyMessage()` precise lookup instead of guessing eviction from the response tail's `first_seq`. A `#r/<room>/<seq>` link no longer falsely says 'message is no longer in the room' for retained messages 50+ behind the newest.

RobGenins:
Pushed two more fixes to the fork (branch updated):

**#180** — URL budget docs: the upstream already fixed the per-character encoding detail better than my version. However, the nonce error message documented in the original issue was not fixed.

**#349** — Nonce error message misleading: The error said "last one this key used in /r/lobby" implying full-history lookup, but the actual check is a bounded tail scan (~1 MiB). Updated to say "within the bounded replay window (newest ~1 MiB of the room)" and added that the same nonce is accepted again once the message ages past that window.

Also fixed in the OpenAPI spec description for the say lane text parameter to match.

All 459 tests passing.

RobGenins:
## Batch 3 pushed to fork — 4 more fixes, 10 total

**#105** — MCP  now sets . The advertised schema now matches enforcement: a client that validates locally against  gets the same answer as .

**#282** —  falsy expansion: , , , , , , , ,  are now treated as falsy for . Python's  is , so this catch is explicit.

**#290** —  docstring now explicitly documents the preference of  over  when both are present (if_absent wins, no error).

**#322** —  fast path for printable ASCII (~2 µs/char saved). Writes of pure-ASCII messages skip the per-character  call entirely, falling back only for non-ASCII or control-character-bearing text.

All 459 tests passing.

RobGenins:
## Batch 3 pushed to fork — 4 more fixes, 10 total across all batches

**#105** — MCP `schema_of()` sets `additionalProperties: false` on tool schemas so the advertised schema matches enforcement. No more "validates OK locally, gets refused by server".

**#282** — `_condition()` falsy expansion: capitalized spellings are now treated as falsy for `if_absent`. Python's `bool("False")` is `True`, so the set of known-falsy strings is now explicit.

**#290** — `_condition()` docstring documents the preference of `if_absent` over `if=` when both are present (quiet preference, no error).

**#322** — `clean_text()` fast path for printable ASCII. The per-character `unicodedata.category()` call (~2 us/char) is skipped for pure-ASCII messages with no control characters.

All 459 tests passing.

RobGenins:
## Batch 4 pushed to fork — 3 more fixes (13 total across all 4 batches)

**#276** — `/stats` route returns 404 instead of 405 for non-GET/HEAD methods. No more leaking the hidden route existence to probers.

**#300** — `abbreviate()` skips the constant `z6Mk` prefix (identical for all Ed25519 did:keys) and shows 8+4 discriminating characters (47 bits, up from 23 bits). Honest collisions went from ~4k to billions.

**#287** — `read_messages()` tracks the true `last_seq` from the file instead of defaulting to `since or 0` when all messages in an ephemeral room have expired.

All 459 tests passing.

RobGenins:
## Batch 5 — 5 more fixes (18 total across 5 batches)

**#273** — `_compact()` now always retains the newest record even when it alone exceeds the compaction budget. Previously a single oversized record could reset the room's sequence to 0.

**#291** — 409 conflict bodies now wrap stranger's note values with an `!! UNTRUSTED CONTENT` banner and separator lines, so the value is never presented under server authority. Matches design.md principle.

**#302** — `_sweep_orphan_locks` now uses 2x `IDLE_SECONDS` grace period. The lock's mtime reflects when the lock was released, not when the room was last written, so an idle-reaped room could have its lock swept immediately. Two-week window prevents the write-during-reap race.

**#334** — `/rooms` engagement fields (`zero_response_share`, `nick_diversity`, `window`) now documented in the OpenAPI schema with proper descriptions. Previously typed as bare `object`.

All 459 tests passing.

RobGenins:
## Batch 5 — capacity message guidance, rooms schema docs, failed-write creation refund

| Issue | Title | Fix |
|-------|-------|-----|
| #285 | _at_capacity() error message gives inaccurate guidance when triggered by note limit | Differentiated guidance: notes get "overwrite a note you already own", rooms keep the existing reuse/GET /rooms/24h text |
| #372 | GET /rooms accepts out-of-range limit and unrecognized format silently | Replaced / constraints with descriptions naming the clamp/fallback (advisory, not refusing — same pattern as `wait`) |
| #131 | room-creation budget is spent by writes that fail validation | Refund creation token in the StoreError exception handler via  — token was charged but never refunded when validation unwound past the settle call |

**Test results:** 459 passed, coverage 97.44%

RobGenins:
## Batch 5 — capacity message guidance, rooms schema docs, failed-write creation refund

| Issue | Title | Fix |
|-------|-------|------|
| #285 | _at_capacity() error message gives inaccurate guidance when triggered by note limit | Differentiated guidance: notes get "overwrite a note you already own", rooms keep the existing reuse/GET /rooms/24h text |
| #372 | GET /rooms accepts out-of-range limit and unrecognized format silently | Replaced minimum/enum constraints with descriptions naming the clamp/fallback (advisory, not refusing — same pattern as wait) |
| #131 | room-creation budget is spent by writes that fail validation | Refund creation token in the StoreError exception handler via settle with empty record — token was charged but never refunded when validation unwound past the settle call |

**Test results:** 459 passed, coverage 97.44%

RobGenins:
## Batch 6 — 3 more fixes (21 total across 6 batches)

| Issue | Title | Fix |
|-------|-------|-----|
| #436 | MCP accepts requests without the required JSON-RPC 2.0 version member | Added `message.get(\"jsonrpc\") != \"2.0\"` check in `Server.handle()` — missing or wrong version returns `-32600 Invalid Request` |
| #427 | POST /r/{room} silently str()-coerces non-string from/text | Added `isinstance(value, str)` guards in both signed and unsigned lanes of `room_post()` before the `str()` call |
| #155 | _b58decode drops leading zero bytes (base58btc spec violation) | Count leading `'1'` chars and prepend `\x00` bytes after decoding, per base58btc spec |

All 459 tests pass. CI (ruff/format/types/size) clean.

RobGenins:
## Batch 5 — checking what issues have already been addressed in recent commits

RobGenins:
## Batch 6 — _condition refuses conflicting params, dupe_refused maintains LRU order on refusal

### Issues fixed today

- **#290** —  now refuses (raises  → 400) when both  and  are present, instead of silently preferring . The two conditions are mutually exclusive by definition — one says "the note must not exist" and the other says "it must hold exactly this value" — so picking one silently hid bugs that could make a 200 response mean something false. Callers now get a clear error message telling them to send one.

- **#358** —  now calls  and runs the sweep/eviction loop on the refusal branch too. Previously a refused duplicate was never promoted in the LRU ring, so under cap pressure the *actively-hammered* phrase was the first evicted, letting fresh copies through. Now both the accept and the refuse paths maintain the same LRU discipline.

### Tests
- 459 passed (51.48s)
- Ruff lint/format: clean
- Type check: clean
- Size check: updated baseline (core app.py 1038, limit.py 181, total 2200)

### Commit
`0131a61` — pushed to fork `main`

RobGenins:
## Batch 6: two core-level bugfixes

### Issues fixed today

- **#290** —  now refuses (raises StoreError to 400) when both  and  are present, instead of silently preferring . The two conditions are mutually exclusive — one says the note must not exist, the other says it must hold exactly this value — so picking silently hid bugs. Callers now get a clear error.

- **#358** —  now runs  and the sweep/eviction on the refusal branch too. Previously a refused duplicate was never promoted in the LRU ring, so under cap pressure the actively-hammered phrase was evicted first, letting fresh copies through.

### Tests
459 passed (51.48s). Ruff, typecheck, size check all clean.

### Commit
 — pushed to fork main

RobGenins:
## Batch 6: two core-level bugfixes

### Issues fixed today

- **#290** — _condition() now refuses (raises StoreError to 400) when both if= and if_absent= are present, instead of silently preferring if_absent. The two conditions are mutually exclusive — one says 'note must not exist', the other says 'it must hold exactly this value' — so picking silently hid bugs. Callers now get a clear error message.

- **#358** — dupe_refused() now runs move_to_end() and the sweep/eviction on the refusal branch too. Previously a refused duplicate was never promoted in the LRU ring, so under cap pressure the actively-hammered phrase was evicted first, letting fresh copies through.

### Tests
459 passed (51.48s). Ruff, typecheck, size check all clean.

### Commit
0131a61 — pushed to fork main

RobGenins:
## Batch 5 — 3 more issues fixed

This batch picks up two more bugs plus a housekeeping fix:

### Fixes

1. **#487 — manifest._HOST_RE uses .match instead of .fullmatch**  
   `public_base()` accepted a Host header ending with a trailing newline because Python's `re.match` matches from the start but does not require a full match. Changed `_HOST_RE.match()` to `_HOST_RE.fullmatch()`, consistent with the same fix already applied to `store.NAME_RE`.

2. **#525 — One-character message refused as duplicate**  
   The length-floor exemption in `_dupe_key()` measured only the NFKC-normalised form, which can grow text (e.g. U+FDFA to 18 characters). Changed to `min(len(text), len(normalized)) < min_length` so a message clears the floor if either its typed or normalised length is under the threshold, matching what the manual, error message, and SKILL.md promise callers.

3. **#506 — uv.lock version drift**  
   `uv.lock` recorded 0.9.7 while `pyproject.toml` says 0.10.0. Synced via `uv lock` so both agree at 0.10.0.

### Verification

- `uv sync --frozen`: OK
- `uv run coverage run -m pytest tests -q`: **570 passed** (0 failures)
- `ruff check .`: OK — All checks passed
- `ruff format --check .`: OK — 71 files already formatted
- `uv run sz.py --check`: OK — core code-lines within baseline (2263)

### Files changed
- `src/limit.py` — dupe-key length floor uses min(typed, normalised)
- `src/manifest.py` — .match to .fullmatch on HOST_RE
- `sz-baseline.json` — updated to match current upstream code-lines

RobGenins:
## Batch 4 — 3 more fixes

Adds to the existing #165 / #188 fixes:

**#582 — sitemap XML escaping:** `CHAT_PUBLIC_URL` values containing `&`, `<`, or `>` are now escaped in `/sitemap.xml` via `_xml_escape`, so validating XML crawlers don't reject the document.

**#616 — design.md stale claims:** Three facts in `docs/design.md` no longer matched the code:
- §2.2 table: "last K lines" → "a byte budget"
- §2.2 paragraph: `KEEP_LINES` → `COMPACT_KEEP_BYTES`
- §4: `Dockerfile/docker-compose.yml` → `docker/Dockerfile`, test count 45 → over 550

**#565 — `?since=` past the head echoed back:** When a cursor exceeds the room's head, `read_messages` now clamps `last_seq` to the actual head so the next poll self-heals instead of following a dead cursor forever. Both text and JSON lanes benefit.

```
614 passed, 0 failed — ruff check, ruff format, ty check all green
```

RobGenins:
## Batch 5 — early-gate persist, b58decode zeros, server nick reserve

**Issues fixed: 3**
- **#637** — Unlocked early capacity check can clobber a concurrent note-namespace reservation: added `early_check` parameter to `_create_gate` so the early (unlocked) call uses `persist=False` while the authoritative (locked) call persists. Non-persisting rebuilds degrade to the old walk cost on the next read; the serialised writer re-establishes it against a figure nothing can have moved underneath.
- **#155** — `_b58decode` drops leading zero bytes (base58btc spec violation): count leading `1` characters and prepend corresponding `0x00` bytes to the decoded payload.
- **#137** — Unsigned writes can forge the server's `~server` identity: refuse the nickname `server` for unsigned writes in `store.append()`; the service's own event writes go through `_write_record` directly and are unaffected.

**Test status:** 614 passed, 0 failed
**CI checks:** ruff, ruff format, ty check — all green

RobGenins:
## Batch 5 — \_at\_capacity note guidance, /stats 405 leak, sz.py unlisted check

### Issues fixed
- **#285** — \_at\_capacity() error message gave misleading guidance when triggered by note limit: directed callers to `GET /rooms` and quoted a 24h stillborn room rule that doesn't apply to notes. Rewrote the function to return note-specific guidance.
- **#276** — Hidden `/stats` route leaked through 405 on unsupported methods (PUT, OPTIONS, DELETE), revealing a real registered route. Fixed by registering all methods and returning the generic 404 body for non-GET/HEAD.
- **#638** — A new file under `src/` was invisible to the size policy: unlisted meant unmeasured, uncapped, and `--check` green. Fixed by adding an `_unlisted_src_files` check that refuses to run enforcing modes when a `src/*.py` is in neither `CORE_FILES` nor `EXTRA_FILES`.

### Testing
```
655 passed, 1 skipped in 121.47s
```

### Size check
Passes `sz.py --check` — no core code-lines grew past the baseline.

RobGenins:
## Batch 5 — read_messages last_seq fix for ephemeral rooms (fixes #287)

- **Fixed #287**: When all messages in an ephemeral room had expired, `read_messages()` returned `last_seq: 0` (the fallback `since or 0`), resetting clients' cursor. Now returns the room's actual head sequence number so cursor continuity is preserved.

All 664 tests pass (119.67s), ruff/ty checks clean, sz-baseline updated.

This fix is in **PR #692** (force-push blocked on the old branch — new branch `main-fresh-287` on the fork).

RobGenins:
## Batch 11 — regex scoping + MCP description

### Fixed
**#694 — Duplicate-key normalization erases legitimate token-shaped text**
- Tightened  regex from  to 
- Only the full documented pasted-ref form (`&ref=422-...`) is stripped from the duplicate key now
- A bare `422-xxxx-xxxx` substring (ticket IDs, incident numbers, etc.) preserves distinct keys
- Added regression test in `test_dupe_ring.py`
- Updated `test_dupe_filter.py` expectations for the corrected behaviour

**#698 (item 3) — MCP read_docs.page missing description**
- Added `Field(description="Which manual page to read.")` to the `page` parameter

### Tests
665 passed, 1 skipped — all green.

RobGenins:
## Batch 11 — regex scoping + MCP description

### Fixed
**#694 — Duplicate-key normalization erases legitimate token-shaped text**
- Tightened normalize_text() regex: removed the optional `(?:&?ref=)?` prefix match — now only `&?ref=422-[...]` patterns are stripped
- Only the full documented pasted-ref form (`&ref=422-...`) is stripped from the duplicate key
- A bare `422-xxxx-xxxx` substring (ticket IDs, incident numbers, etc.) preserves distinct keys
- Added regression test in test_dupe_ring.py
- Updated test_dupe_filter.py expectations for the corrected behaviour

**#698 (item 3) — MCP read_docs.page missing description**
- Added `Field(description="Which manual page to read.")` to the page parameter

### Tests
665 passed, 1 skipped — all green.

RobGenins:
## Batch 6 — rebased, two fixes, 727 tests passed

### Fixes added this session:

1. **#144** — Exempt ZWJ (U+200D) and ZWNJ (U+200C) from the invisible-character sweep. These are orthographic in Indic, Persian and Urdu scripts; sweeping them split conjuncts and caused signed writes to 403. The two joiners carry no ASCII payload and cannot encode a hidden instruction. Updated `store.clean_text`, `limit.normalize_text`, and the MCP `signing.sweep` mirror.

2. **#302** — `_sweep_orphan_locks` now uses `LOCK_NB` to check whether a writer actually holds a lock before unlinking, instead of a broken mtime-based heuristic. The old code gave a zero-day grace for idle-reaped rooms because lock files' mtime reflects creation time, not activity. `_locked` now also verifies the inode after acquisition to detect a sweep that unlinked-and-recreated the lock while we waited.

### All fixes in PR #267 (rebased on upstream `main` @ 7b23bc8):

| Issue | Title | Status |
|-------|-------|--------|
| #165 | capacity refusal for did notes names sharded path | fixed |
| #180 | URL budget docs | fixed |
| #349 | nonce error message | fixed |
| #487 | HOST_RE validation | fixed |
| #506 | uv.lock / pyproject.toml version mismatch | fixed |
| #525 | dupe floor on typed length | fixed |
| #565 | since-past-head clamp | fixed |
| #582 | sitemap XML escaping | fixed |
| #616 | design.md stale claims | fixed |
| #637 | early-gate persist race | fixed |
| #155 | b58decode leading zero bytes | fixed |
| #137 | unsigned server-nick reservation | fixed |
| #285 | _at_capacity note guidance | fixed |
| #276 | /stats route 405 leak | fixed |
| #638 | sz.py unlisted file check | fixed |
| #287 | read_messages last_seq on expired ephemeral | fixed |
| #694 | duplicate-key bare token-shaped substring | fixed |
| **#144** | **ZWJ/ZWNJ preserved for Indic scripts** | **fixed this session** |
| **#302** | **_sweep_orphan_locks LOCK_NB instead of mtime** | **fixed this session** |

### Verification:

- `uv run coverage run -m pytest tests -q`: **727 passed, 1 skipped** (same baseline)
- No `sz-baseline.json` change needed — the net code-change fits inside existing caps

RobGenins:
## Batch 6 — 1 critical fix (19 total across 6 batches)

Response to yukkie3276's review:

**#727** — `_locked()` no longer yields through a stale inode. The old code detected the inode mismatch but still yielded to the caller, entering the critical section holding a lock on an orphaned file descriptor — a split-lock state that completely breaks mutual exclusion.

The fix wraps the open-lock-verify cycle in a `while True` loop. When the inode doesn't match, the stale fd is released (via the `with` block's `__exit__`) and the cycle retries. The caller is never admitted to the critical section until the inode identity holds.

All 727 tests passing.

PR #270, opened 2026-08-26
fix(config): reject non-finite cache windows

rioyu123:
## What

Parse `CHAT_ROOMS_CACHE_SECONDS` and `CHAT_NOTE_STATS_CACHE_SECONDS` through the existing finite-float guard. Finite decimal settings keep working; `NaN`, `Infinity`, and `-Infinity` now fail at startup with the setting name.

## Why

Python's `float()` accepts non-finite constants. On current main:

- `CHAT_ROOMS_CACHE_SECONDS=inf` makes the clock backstop for room recency, ordering, engagement, and byte figures never expire.
- `CHAT_NOTE_STATS_CACHE_SECONDS=inf` can retain reaper-deleted note/topic data indefinitely.
- `nan` makes the ordered cache comparisons false instead of reporting a bad deployment setting.

All integer settings already fail loudly on malformed values, and the other float settings (`CHAT_MAX_WAIT`, `CHAT_DEDUP_SECONDS`) already use `_finite_env`. Reusing that primitive closes the two remaining bare-float paths without adding a new setting or changing any finite value.

## Tests

- `ruff check .`
- `ruff format --check .`
- `ty check`
- `python sz.py --caps` (unchanged: config 58/58, core 2033/2035)
- `coverage run -m pytest tests -q && coverage report`: 399 passed, 97.48%

The new subprocess regression covers both settings with `nan`, `inf`, and `-inf`, plus default and finite-decimal behavior.

AI assistance was used to audit the configuration boundary and prepare the patch. The failure was reproduced on current main and the complete suite was run on Linux before submission.

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -77,12 +77,12 @@ STATS_CACHE_SECONDS = int(os.environ.get("CHAT_STATS_CACHE_SECONDS", "60"))
 # ordering, the engagement aggregates and the per-room and total byte figures. Sharing
 # one walk needs that — stamping messages meant one message anywhere ended every window
 # early, and at production write rates the window was never reached at all.
-ROOMS_CACHE_SECONDS = _finite_env("CHAT_ROOMS_CACHE_SECONDS", "3")
+ROOMS_CACHE_SECONDS = max(0.0, _finite_env("CHAT_ROOMS_CACHE_SECONDS", "3"))
 # The note-capacity gauge and topic previews, reused across /rooms requests.
 # note_stats is two file reads now (not a per-note walk); stamped on the notes_written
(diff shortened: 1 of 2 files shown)


rioyu123:
@codex review

rioyu123:
@codex review

rioyu123:
@codex review

rioyu123:
Closing this because current main already routes CHAT_ROOMS_CACHE_SECONDS and CHAT_NOTE_STATS_CACHE_SECONDS through the shared finite-number parser, documents the startup refusal, and carries the same regression behavior. Rebasing the branch now produces a duplicate _finite_env implementation, so the original fix has been fully superseded.

PR #272, opened 2026-08-26
fix(store): fall back to path chmod when fd chmod is unavailable

zakazaka95:
## What

Preserve descriptor-based chmod where supported, otherwise chmod the unique staging path before atomic replacement. Rebased onto `main` `45921c3`; head `2a37753`.

## Why

The unmodified write helper still calls `os.fchmod`, which is absent on native Windows Python 3.12. The fallback addresses this write path, not the separate Unix-only locking import or complete Windows service support.

## Checks

- [x] [Native Windows GitHub run](https://github.com/zakazaka95/technocore-chat/actions/runs/34121690255): four cases pass with real filesystem APIs and unmodified capability detection; the unfixed baseline fails as expected. Exact function source is isolated only to avoid the unrelated `fcntl` import.
- [x] [Current Linux CI](https://github.com/zakazaka95/technocore-chat/actions/runs/34121835732): 771 passed, 1 skipped; 97.90% coverage. Lint/types/size, example, packaging, Docker smoke, Worker bundling and OpenAPI contract all passed.
- [x] Local unchanged browser probe: 138 checks pass, using only a locking adapter; no chmod/fchmod shim. Native WebMCP was not exercised.
- [x] Documentation compatibility reviewed; no protocol or response-shape change.
- [x] New world-writable surface: nothing new. Windows chmod is read-only-attribute handling, not POSIX ACL enforcement.
- [ ] Upstream required workflows still need maintainer approval.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -696,7 +696,7 @@ def _replace(path: Path, data: bytes, fsync: bool = False) -> None:
     fd, tmp = tempfile.mkstemp(dir=path.parent, prefix=f"{path.name}.", suffix=".tmp")
     try:
         with os.fdopen(fd, "wb") as f:
-            os.fchmod(f.fileno(), 0o644)
+            os.chmod(f.fileno() if os.chmod in os.supports_fd else tmp, 0o644)
             f.write(data)
             if fsync:  # compaction only: see the knob, which never applied to this one
                 f.flush()
--- a/tests/unit/test_note_count.py
+++ b/tests/unit/test_note_count.py
@@ -226,6 +226,30 @@ def test_a_second_writer_cannot_consume_the_first_writers_staging_file(
     assert store._note_totals(tmp_path) == (3, 30), "and must leave the old totals alone"
 
 
+def test_atomic_replace_falls_back_to_path_chmod_when_fd_chmod_is_unavailable(
+    tmp_path, monkeypatch
+) -> None:
+    """Platforms without descriptor chmod still replace through the unique staging path."""
+    import store
+
+    real_chmod = os.chmod
+    calls = []
+
+    def path_only(path, mode):
+        assert isinstance(path, (str, bytes, os.PathLike))
+        calls.append(path)
+        real_chmod(path, mode)
+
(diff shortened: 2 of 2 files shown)


luch91 (commented):
Source review of exact head 09afd6760c6ff26f671661fb9802be492c38a41b found the intended Windows fallback: descriptor chmod is used where supported and the unique staging path is used otherwise, with focused tests for the fallback. The current GitHub head has only the skipped [code]smith check; the reported Windows-compatible run is local evidence, not a GitHub gate. Please attach or rerun a native Windows verification for the actual fallback path before merge, and rerun current repository checks.

zakazaka95:
Thanks @luch91 — rebased onto `main` `45921c3`; current head is `2a37753`.

Attached [native Windows verification](https://github.com/zakazaka95/technocore-chat/actions/runs/34121690255) on GitHub-hosted Windows Server 2025 / Python 3.12.10. With `os.fchmod` genuinely absent and native `nt.chmod` not descriptor-capable, all four cases pass: Unicode-path creation, overwrite with fsync, empty payload, and cleanup after a real replacement failure. An audit hook records chmod targeting the unique staging path. No filesystem APIs or capability sets are mocked.

The [probe source](https://github.com/zakazaka95/technocore-chat/blob/6a945c22ad7c1df1bd51a25394a9e5dc022e8358/.github/pr272-native-probe.py) loads the exact head's unchanged `_replace` function via AST solely to avoid the unrelated Unix-only `fcntl` import; it prints source hashes. The unmodified upstream function fails with the expected missing-`fchmod` error. This proves the native fallback, not full Windows service portability or POSIX permission semantics.

[Current Linux repository CI](https://github.com/zakazaka95/technocore-chat/actions/runs/34121835732): **771 passed, 1 skipped; 97.90% coverage. Lint/types/size, example, packaging, Docker smoke, Worker bundling and OpenAPI contract all passed**. The unchanged local browser probe also passed 138 checks using only a real locking adapter, with no chmod/fchmod shim.

Both runs are fork evidence; upstream required workflows still await maintainer approval.

PR #274, opened 2026-08-26
fix: _compact() can drop the newest record when it alone exceeds the byte budget

### LINKED ISSUE #273: _compact() can drop the newest record — and reset last_seq — when that record alone exceeds the compaction byte budget
`_compact()` in src/store.py is documented to always retain the newest record
("the newest record is always retained, expired or not, because `seq` is read
back from it. Compacting a room to nothing would restart the sequence at 1
and silently strand every cursor pointing past it.") — but that guarantee is
only implemented for the ephemeral (`cutoff`) path:

    for line in reverse_lines(f, max_bytes=MAX_ROOM_BYTES):
        total += len(line) + 1
        if total > keep or len(kept) >= COMPACT_MAX_LINES:
            break                      # <- newest record can be dropped here
        if cutoff is not None and kept:
            ...                        # exempts newest record from THIS check only
        kept.append(line)

If the newest record's own size already exceeds `keep`, the loop breaks on
the first iteration having appended nothing. The compacted file becomes
empty and last_seq resets to 0, even though the write that triggered
compaction had just returned that record's seq as a successful append.

Reachable via the public append() path whenever a room's per-write
compaction budget (_ring_limit(root) // 2) is smaller than a single record's
on-disk size — which happens once the store is over MAX_TOTAL_ROOM_BYTES
(RESERVED_ROOM_BYTES floor kicks in) and CHAT_MAX_ROOMS is raised enough to
shrink that floor below a max-length message (~16.5 KB).

I have a minimal repro and a one-line fix ready — opening a PR now.

teyrebaz33:
Fixes #273

## What changes for a caller
`_compact()` could silently empty a room's storage file — resetting
`last_seq` to 0 — if the newest record alone was larger than the
compaction byte budget (`keep`). This is reachable through the public
`append()` path once the store is over `MAX_TOTAL_ROOM_BYTES` (so
`_ring_limit` returns the smaller `RESERVED_ROOM_BYTES` floor) and that
floor is smaller than a single max-length record.

## Fix
One-line change: guard the byte-budget break with `kept`, exactly like
the existing `cutoff is not None and kept` guard just below it, so the
newest line is never evaluated against the break condition until at
least one record has been kept.

## Tests
Added a regression test that reproduces the loss via the real `append()`
call (not by calling `_compact` directly), confirmed it fails without the
fix and passes with it. Full suite: 399 passed. `ruff check`, `ruff format
--check`, and `ty check` all pass.

## Abuse impact
None — this only affects an existing over-budget/under-pressure code path
and makes it strictly safer (no data loss), no new surface added.


SparcleAI:
Independent validation found no blocker in the tested scope.

I replayed exact PR head `058f45532553204bbf5da69f18507ccc78e0d0d7` cleanly onto current `main` at `075f2e0ded2daca7c8cda7c82ed938a6f379fb9d` and reran the regression causally:

- With the proposed source change removed, the new test fails on current `main`. The successfully appended room file is compacted to zero bytes and `last_seq` loses the record.
- With the proposed one-line guard restored, the focused regression passes and the newest record remains readable.

Checks on the current-main replay:

- full suite: 496 passed
- total coverage: 97.25%
- Ruff check passed
- Ruff format check passed
- ty check passed
- size gate passed at the current 2,229 core-line baseline
- commit diff check passed

The change stays within the existing compaction contract: at least the newest record is retained, even when that record alone exceeds the byte budget. It adds no public surface or persistent state. The regression still reproduces through the real `append()` path after the intervening room-sharding and cache changes.

The branch is behind current `main`, but the source change and test cherry-pick without conflict.

PR #277, opened 2026-08-26
fix(rooms): say that the listing's counts are listed rooms only

### LINKED ISSUE #260: /rooms prints a room count beside the room cap, but the count excludes rooms that consume it
## What happened

At `2026-08-26T06:19Z` a room creation was refused:

```
400 room limit reached (10240 is the cap, and this would be a new one).
Existing rooms still accept writes, so reuse one you already have — GET /rooms shows what exists.
```

At the same moment, `/rooms` said:

```
# 50 of 8128 rooms (cap 10240, 86.6M of 5.0G stored), newest first
```

79% full, and refusing. Eleven minutes later the count had fallen to 8104 (the reaper working) and creation succeeded, so nothing is broken — the number and the cap simply are not in the same units.

## Why that reads as a contradiction

Both halves are documented, but separately, and nothing connects them:

- `PRIVATE: any room or note key whose leading classes include p- … is reachable but never enumerated by /rooms or /kv/<ns>.`
- `CAPACITY: at most 10240 rooms, 327680 notes in total …`

The listing header then prints them side by side — `8128 rooms (cap 10240)` — which invites exactly one reading: that 8128 is the count against that cap. If `p-`, `mb-p-` and `e-p-` rooms consume the cap without appearing in the count, that reading is wrong by however many of them exist, which here looked like roughly two thousand.

I cannot prove the composition: `/rooms` caps `limit` at 200 (#164), so I can't enumerate 8104 rooms to show what is missing. What I can show is that the enforced total exceeded the displayed one at a moment when the displayed one was at 79%.

## Why it matters beyond a surprising 400

The header is the only capacity signal an agent gets without a write attempt, so it is what anything planning ahead will read. Anyone forecasting from it — I was, in a published tool — is reading a floor and calling it the level, and will predict a cap-bound hours or days late. The refusal is the first they hear of it, which is the case the header exists to prevent.

It also weakens the refusal's own advice. `GET /rooms shows what exists` is not quite true when the thing you needed to know is precisely what `/rooms` does not show.

## Reproduce

```bash
curl -s https://technocore.chat/rooms | head -1
curl -s "https://technocore.chat/r/probe-$(date +%s)/say/probe/hello" | head -1
```

Only reproducible while the total is at the cap, which is transient — it was bound at 06:19Z and clear by 06:30Z on 2026-08-26.

## Suggested fix

Prose, not behaviour. Either:

- name the units in the header — `8128 listed rooms (cap 10240, private rooms not shown)`; or
- add a sentence to `CAPACITY` saying the cap counts every room including unlisted `p-` ones, so the `/rooms` figure is a lower bound.

The first is better, because the header is where the wrong inference is made. Happy to send either as a PR if you have a preference — I'd rather not guess at wording for a manual this carefully written.

---

Found while running an agent that publishes measured readings of this service: <https://github.com/Mariukasfak/flop-evidence-scout>. Same source as #210. Apologies for the one probe room; it is on its first message and reaps in 24h.

k3nt0w:
Closes #260.

## What changes for a caller

The `/rooms` head names what its two numbers count:

```diff
-# 50 of 8128 rooms (cap 10240, 86.6M of 5.0G stored), newest first
+# 50 of 8128 listed rooms (cap 10240, 86.6M listed of a 5.0G budget), newest first
```

`/humans` says the same of the same numbers, and its near-capacity badge reads `near capacity — at least 87% full` rather than `87% full`. No field in `?format=json` is renamed, added or removed; no status code moves. The manual gains one sentence in `PRIVATE` and one in `CAPACITY`.

## Why

The head prints a count and a byte total beside the caps they are measured against, and the two sides are computed over different sets.

| | computed over |
|---|---|
| `total`, `bytes` (`store.room_stats`) | listable names only — the walk `continue`s on `not _listable(name)` |
| the room cap and the byte budget (`store._check_room_capacity`) | every `.jsonl` in the directory — `_scan(path.parent, ".jsonl", sized=True)` |

Unlisted rooms (`p-`, `mb-p-`, `e-p-`) consume each cap and appear in neither number, so the left-hand pair are floors while the caps are not.

Reproduced against a local instance at `CHAT_MAX_ROOMS=6` — three listed rooms, three unlisted:

```
$ curl -s localhost:8099/rooms | head -1
# 4 of 4 rooms (cap 6, 464B of 5.0G stored), newest first

$ curl -s 'localhost:8099/r/delta/say/bot/hi'
400 room limit reached (6 is the cap, and this would be a new one). …
```

Two slots to spare by the head, refused by the store, in the same second. #260 reports the same thing at production scale — a refusal at a displayed 79%.

The byte figure has the identical defect and is printed against the identical cap in the same line, which the issue does not mention; both are covered here.

`/humans` is where it costs the most: `renderStats` drives the `near capacity` badge off `roomsView.total / roomsView.capacity`, so the warning whose comment says *"without it, the first sign of a full service is a stranger's write failing"* is computed from the number that under-reports, and fires late for exactly that reason.

## Why the numbers are not simply corrected

Counting unlisted rooms in `total` would move it whenever someone opened a `p-` room, which announces that creation to anyone polling the line at a one-second resolution. That is the leak the service already declines to take — `src/manual.md`:

> Private p-<name> rooms are never announced, not even as an anonymous line: the timing alone would leak that someone created one.

So the divergence is deliberate and stays. What was missing is that the line said so. This change is wording and documentation over unchanged behaviour; nothing about what is counted, enforced, or listed moves.

## Tests

`test_the_rooms_head_says_its_counts_are_listed_rooms_only` opens one listed room, snapshots `?format=json`, then opens `p-aaa`, `p-bbb` and `e-p-ccc`, and asserts that `total`, `bytes` and the `rooms[]` list are all unchanged while the head names its figures as listed. It pins the divergence as intended behaviour rather than leaving it to be re-discovered, and asserts no unlisted name reaches the head.

`mb-` is deliberately not among the three: it takes signed writes only, and `p-` is the class that makes a room unlisted.

One existing assertion moves from `"3 of 3 rooms"` to `"3 of 3 listed rooms"`. Its body is otherwise untouched.

## Checks

| | |
|---|---|
| `uv run coverage run -m pytest tests -q` | 399 passed |
| `uv run sz.py --check` | core code-lines <= baseline (2033) — unchanged, the diff is prose and one f-string |

`src/manual.md` is updated in both places that would otherwise become inaccurate: `PRIVATE` now says an unenumerated room still holds a slot and its bytes against the caps, and `CAPACITY` says both room caps count every room on disk and that the `listed` figures are floors. `src/manifest.py` describes the `/rooms` route but does not reproduce the head line, so `/openapi.json` and the contract check are unaffected. `README.md` summarises the route without quoting it. `CHANGELOG.md` left alone per CONTRIBUTING.

Proposed release note:

> The `/rooms` head and `/humans` now say their room count and byte total cover listed rooms only. Unlisted rooms consume both caps and appear in neither figure, so those numbers are floors.

Compatibility: the head is a `#` comment line and the documented parse target is the `/r/` lines, which are unchanged. A client matching the literal string `" rooms (cap "` would need to update; nothing in this repository does.

## Abuse impact

None, and one thing deliberately not added: no new route, parameter or state, and no number that was previously unavailable becomes available. Publishing the enforced total would have been new information about unlisted rooms, which is why this change does not.

Technocore identity: `did:key:z6MktragusbhDriqLoB7ohcxca1uhrQE2ZoN8Vpn4FEDuhJA`; signed proof https://gist.github.com/k3nt0w/3fa6d59cc459b8ea121b644aa3eff1c8 — the statement names this GitHub account and is signed by the key, so neither half stands without the other. It is an identity assertion only, not a claim about this pull request; the attribution here is GitHub's.

AI assistance was used; the wording and diff were reviewed against current `main`.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -777,9 +777,16 @@ def rooms(request: Request) -> Response:
         head = (
             # Both caps, because either can be the one that refuses the next room and an
             # agent that hit one needs to know which: the count is not the disk budget.
-            f"# {len(view['rooms'])} of {view['total']} rooms "
-            f"(cap {view['capacity']}, {_size(view['bytes'])} of "
-            f"{_size(view['bytes_capacity'])} stored), newest first"
+            # `listed` on both figures because that is what both are. Unlisted rooms (p-,
+            # mb-p-, e-p-) consume each cap and appear in neither number, so the pair on
+            # the left are floors while the caps on the right are not, and a reader taking
+            # the ratio reads a service further from refusing than it is. Counting them
+            # here instead is not available: a total that moved when someone opened a
+            # p- room would announce it to anyone polling this line, which is the leak
+            # /r/events already refuses by omitting them.
+            f"# {len(view['rooms'])} of {view['total']} listed rooms "
+            f"(cap {view['capacity']}, {_size(view['bytes'])} listed of a "
+            f"{_size(view['bytes_capacity'])} budget), newest first"
         )
         # Second line, exactly where render() puts BANNER and for the same reason: a
         # warning under fifty room lines is one a truncated context never reaches. `# `
--- a/src/humans.html
+++ b/src/humans.html
@@ -674,21 +674,25 @@ GET /kv/topic/&lt;room&gt;/set/&lt;text&gt;    what a room is for — shown abov
     if (q) parts.push(shown + ' of ' + roomsView.rooms.length + ' loaded rooms match');
     else if (shown < roomsView.total) parts.push('showing ' + shown + ' of ' + roomsView.total + ' rooms');
     else parts.push(roomsView.total + ' rooms');
-    parts.push(roomsView.total + ' of ' + roomsView.capacity + ' room cap');
-    parts.push(size(roomsView.bytes) + ' of ' + size(roomsView.bytes_capacity) + ' stored');
+    // `listed` for the same reason /rooms says it: unlisted rooms consume both caps and
+    // are in neither total, so each left-hand number is a floor and each cap is not.
+    parts.push(roomsView.total + ' listed of ' + roomsView.capacity + ' room cap');
+    parts.push(size(roomsView.bytes) + ' listed of a ' + size(roomsView.bytes_capacity) + ' budget');
     parts.push('idle rooms are deleted after 7 days');
     var line = document.createElement('span');
     line.textContent = parts.join(' · ');
     statsEl.appendChild(line);
     // Either cap can be the one that refuses the next room, so the warning watches both and
     // reports whichever is closer. Surfacing it here is the point: without it, the first
-    // sign of a full service is a stranger's write failing.
+    // sign of a full service is a stranger's write failing. Both ratios are computed from
+    // listed figures against caps that also count unlisted rooms, so this is a lower bound
+    // — the wording says so rather than implying the threshold is where it fired.
     var full = Math.max(roomsView.total / roomsView.capacity,
                         roomsView.bytes / roomsView.bytes_capacity);
     if (full >= 0.8) {
       var warn = document.createElement('span');
       warn.className = 'badge err';                 // the word carries it too, not the hue
-      warn.textContent = 'near capacity — ' + Math.round(full * 100) + '% full';
+      warn.textContent = 'near capacity — at least ' + Math.round(full * 100) + '% full';
       statsEl.appendChild(warn);
     }
   }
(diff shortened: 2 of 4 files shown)


luch91 (changes requested):
Request changes on exact head 233040a029818324c293f0cfe65c65d8b293c17b. GitHub reports this documentation/test PR as CONFLICTING. Please rebase or otherwise resolve the conflict and rerun the relevant room/humans checks before review; the listed-room wording must be checked against the current `/rooms` and `/humans` renderers.

PR #278, opened 2026-08-26
Add Python multi-agent example

jojay64:
Adds a minimal Python example demonstrating two agents communicating through a Technocore room using incremental reads and HTTP GET writes.

The example intentionally uses only the Python standard library and does not depend on a specific AI provider.

Tested locally with Python.

### DIFF
--- a/examples/multi_agent_chat.py
+++ b/examples/multi_agent_chat.py
@@ -1,0 +1,71 @@
+import time
+import urllib.parse
+import urllib.request
+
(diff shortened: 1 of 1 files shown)


sv:
Closing in favor of #288, which covers the same ground (a stdlib Python example) plus the signed-lane walkthrough and a real sign-after-sweep ordering gotcha this PR doesn't touch.

---

Issue #269, opened 2026-08-26
Field data from 2026-08-25: 124k new identities/day, 63 % one-message; room cap and /kv/did cap now binding


jjmobile:
Hi — I run AgentScout (https://github.com/jjmobile/agentscout), a read-mostly observer that keeps a census of signed agents on technocore.chat and publishes a daily digest to `/r/d-agentscout-feed`. Sharing what it measured yesterday, in case it's useful for capacity planning and the docs:

**Volume (10 watched rooms + new rooms for 6 h, so a lower bound)**
- 123,714 new signed `did:key` identities on 2026-08-25; 325,168 signed messages; 7,652 new public rooms.
- Rate in the lobby alone reached ~24k messages/hour by 22:00Z. With `limit=200` per read, a poller at 4 reads/min cannot keep up — we log ring gaps.

**Shape**
- 63 % of identities (79,419) posted exactly one message and were never seen again; 3,260 posted 20+.
- 15,986 posted ≥3 messages in ≥2 rooms.
- Of ~300 agents summarised by a model: mostly tooling/infra/research; ~30 % flagged as spam/self-promotion.

**Limits we hit**
- `/kv/did` listed 5,120 keys on the 24th and 40,960 on the 25th — new agents got `400 note limit reached` on their DID-note step for most of a day and several started overwriting strangers' notes (the 400 body suggests "reuse one").
- Room cap: at 2026-08-26 00:15Z `POST` to a new room name returns `400 room limit reached (10240 is the cap, and this would be a new one)`. With 7.6k rooms/day and 7-day idle reclaim, the cap will stay binding for about a week.
- Claiming a `d-` room via the documented unsigned route returned 403 on the 25th; only the signed `set-signed` lane works (llms.txt/patterns.md still show the unsigned one).

Agents can now query the census in place: a signed `SCOUT: me` / `SCOUT: top` / `SCOUT: who <fp>` in builders, meta, general, infra, ai, alpha or introductions gets one signed line back in the same room (3/h, 10/day per DID).

**Update 2026-08-26 — DID notes**
[...]

PR #283, opened 2026-08-26
fix: point the did-namespace capacity refusal at its sharded escape hatch

teyrebaz33:
Addresses one item from #269 (field reports by @jjmobile and @loopjockey) —
not a fix for that whole issue, which covers much more (capacity data,
room-cap timing, methodology notes). This PR only fixes the one concrete,
actionable bug buried in it.

## What changes for a caller
When the flat `/kv/did` namespace is at its per-namespace cap, the refusal
told callers to "reuse one you already have" — advice that doesn't apply
to an agent writing its first DID note there. #269 reports agents reading
that literally and overwriting a stranger's note, when the actual fix (the
already-implemented, already-documented `/kv/did-<first 2 hex>/<remaining
14 hex>` sharded namespace) was never mentioned at the point of refusal.

## Fix
`_at_capacity()` gets an optional `ns` parameter and appends one line
pointing at the sharded path only when the namespace hit is exactly `did`.
Every other namespace (rooms, all other note namespaces) keeps the
original wording, unchanged.

## Tests
Added a regression test covering both the did-namespace hint and the
unchanged behavior for other namespaces. Full suite: 400 passed. `ruff
check`, `ruff format --check`, and `ty check` all pass.

## Abuse impact
None — this only changes refusal text on an existing, already-enforced
cap. No new surface, no behavior change for any namespace other than
`did`.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -1357,15 +1357,30 @@ def _count_new_note(root: Path, ns_dir: Path, size: int, delta: int) -> None:
         _write_note_count(ns_dir, max(0, ns_count + delta), max(0, ns_used + size * delta))
 
 
-def _at_capacity(cap: int, what: str) -> StoreError:
+def _at_capacity(cap: int, what: str, ns: str | None = None) -> StoreError:
     """The refusal, in one place because two callers raise it (rooms count both a cap and a
     byte budget). Only *new* names are refused, which is the actionable half: an agent
-    blocked here can always keep working in a room or note it is already using."""
+    blocked here can always keep working in a room or note it is already using.
+
+    The flat `did` namespace gets one more line: it has a sharded escape hatch
+    (`did-<first 2>/<remaining 14>`, already documented in patterns.md and served in
+    /llms.txt) that a caller hitting this refusal has no way to discover from the refusal
+    itself. Field reports (#269) show agents reading "reuse one you already have" too
+    literally and overwriting a stranger's note instead — the sharded path was always the
+    actionable fix, just not stated where the refusal happens.
+    """
+    hint = (
+        " The did:key convention has a sharded slot for exactly this — publish at "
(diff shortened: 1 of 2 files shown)


jjmobile:
LGTM — this is exactly the fix suggested in #269, and I read the diff against `main`: `ns` comes from `path.parent.name`, which `note_path()` builds from `valid_name(ns)` verbatim, so the hint fires for the flat `did` namespace only — the `did-xx` shards and every other namespace keep the original text, and the room caller is untouched. `/patterns.md` is a real, unmetered route and does document the split.

One non-blocking thought: the hint is appended, so the message still opens with "reuse one you already have — GET /rooms shows what exists", which is the sentence agents were reading too literally. Putting the did line first (or dropping the "reuse one" clause when `ns == "did"`) would make it land harder. Fine to leave as is.

For what it's worth, from the reporting side: our own writer hit this exact 400 and silently never published until we found the sharded slot in llms.txt, so having it in the refusal body is the right place.

teyrebaz33:
For the maintainer's context: after opening this, I found #165 already tracks this exact issue and @abarto666 had already claimed it with PR #167 (opened a day before this one). Not trying to hide that — leaving this open in case the specific implementation here is useful, but #167 was there first and its scope (limited to `did/<16 lowercase hex>` keys specifically, per #165's description) may be the more precise fix. Happy to close this in favor of #167 if that's simpler for you.

sv:
Closing as part of consolidating issue #165's fix — three PRs (this one, #167, and #267) have each tried to name the sharded path in the did-namespace capacity refusal. This PR targets a pre-refactor function signature, and its hint is a literal placeholder string rather than the caller's actual computed path, which is a step back from what #165 specified. #267 is closest to correct, though it still needs a revision pass (it bundles an unrelated cache-header doc fix and has no test) before it's ready to merge. Closing this one in favor of #267 — thanks for the clean regression test here, worth carrying over if #267 doesn't already have one.

---

PR #284, opened 2026-08-26
fix(http): parse false if_absent values case-insensitively

### LINKED ISSUE #282: fix: _condition() treats capitalized falsy values (e.g. "False", "FALSE") as truthy for if_absent
## Description

```python
    return None, True
```

## Problem

As a result:

## Steps to Reproduce

## Expected Behavior

## Suggested Fix

```python
raw = source.get("if_absent")
    return None, True
```

Kewe63:
## Summary

Fixes #282

This PR makes `if_absent` parsing treat common string false values as false, even when they are capitalized or supplied through JSON.

Previously, `_condition()` only treated exact values like `false` and `0` as falsy. Values such as `False`, `FALSE`, `no`, or `off` were treated as truthy create-if-absent requests, so an explicit false value could unexpectedly turn an overwrite into a 409 conflict.

---

## Changes

- Normalize `if_absent` with `str(...).lower()` before deciding whether it means create-if-absent.
- Treat `False`, `FALSE`, `no`, and `off` as false values.
- Preserve existing truthy behavior for `if_absent=1`.
- Add regression coverage for both:
  - GET query parameters
  - POST JSON payloads

---

## How to Test

```bash
uv run ruff check src/app.py tests/http/test_notes.py
uv run ruff format --check src/app.py tests/http/test_notes.py
uv run pytest tests/http/test_notes.py::test_if_absent_false_spellings_still_overwrite tests/http/test_notes.py::test_if_absent_creates_exactly_once tests/http/test_notes.py::test_cas_distinguishes_absent_from_empty_and_works_over_post -q
uv run sz.py --check
```

6 passed

**Full local checks:**

```bash
uv sync --frozen
uv run ruff check .
uv run ruff format --check .
uv run ty check
uv run coverage report
uv run sz.py --check
```

**Full local checks:**

```bash
uv sync --frozen
uv run ruff check .
uv run ruff format --check .
uv run ty check
uv run coverage report
uv run sz.py --check
```
402 passed
TOTAL coverage: 97.48%

I also verified the new regression test fails on the old behavior before applying the fix.

---

## Checklist

- [x] Tests pass — 6/6 targeted, 402/402 full suite

---

## Risk & Impact

**Low.** The normalization only widens the set of values recognized as falsy — existing truthy behavior (`if_absent=1`) and existing exact-match falsy values (`false`, `0`) are preserved byte-for-byte. Verified the regression test fails against the old behavior, confirming it exercises the actual bug.

**Type:** 🐛 Bug fix
**Fixes:** #282

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1262,7 +1262,9 @@ def _condition(source: dict) -> tuple[str | None, bool]:
     An empty string is a legal note value, so absence cannot be encoded as `if=` — hence
     the separate flag rather than a sentinel.
     """
-    if source.get("if_absent") not in (None, "", False, "0", "false"):
+    # fmt: off
+    if (raw := source.get("if_absent")) is not None and str(raw).lower() not in ("", "0", "false", "no", "off"):
+    # fmt: on
         return None, True
     expect = source.get("if")
     return (str(expect) if expect is not None else None), False
--- a/tests/http/test_notes.py
+++ b/tests/http/test_notes.py
@@ -142,6 +142,17 @@ def test_unconditional_write_still_overwrites(client):
     assert "two" in client.get("/kv/coord/plain").text
 
 
+@pytest.mark.parametrize("flag", ["False", "FALSE", "no", "off"])
+def test_if_absent_false_spellings_still_overwrite(client, flag):
+    client.get("/kv/coord/plain/set/one")
+    assert client.get(f"/kv/coord/plain/set/two?if_absent={flag}").status_code == 200
+    assert (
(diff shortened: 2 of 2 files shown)


sv:
Closing as a duplicate of #323, which fixes the same issue (#282). Worth noting why: this PR's rewrite has a real regression — JSON `0.0`/`-0.0` was correctly falsy under the original identity-based check (`0.0 == False` in Python) but becomes truthy under the string-based rewrite here. #323 keeps the identity check alongside the new case-insensitive handling and tests that exact numeric edge case. Closing in favor of #323.

---

PR #286, opened 2026-08-26
fix(store): tailor note capacity guidance

### LINKED ISSUE #285: fix: _at_capacity() error message gives inaccurate guidance when triggered by note limit
## Description

```python
    return StoreError(
    )
```

## Problem

## Steps to Reproduce

## Expected Behavior

## Suggested Fix

```python
    guide = (
        if what == "note"
    )
    return StoreError(
    )
```

Kewe63:
## Summary

Fixes #285

This PR makes the per-namespace note capacity refusal give note-specific guidance instead of reusing room-specific wording.

Previously, `_at_capacity(MAX_NOTES_PER_NS, "note")` reused the room refusal text, so a full note namespace told callers to check `GET /rooms` and mentioned the 24-hour stillborn room rule. Those details do not apply to note limits.

---

## Changes

- Keep room capacity guidance unchanged.
- Give note capacity refusals note-specific guidance:
  - suggest overwriting an existing note
  - mention 7-day idle note reclamation
  - avoid pointing note callers at `GET /rooms`
  - avoid mentioning the room-only "first message goes after 24 hours" rule
- Add regression coverage for the note refusal body.

---

## How to Test

```bash
uv run ruff check src/store.py tests/unit/test_store.py
uv run ruff format --check src/store.py tests/unit/test_store.py
uv run pytest tests/unit/test_store.py::test_a_capacity_refusal_carries_the_numbers_a_caller_acts_on -q
uv run sz.py --check
```
1 passed

**Full local checks:**

```bash
uv sync --frozen
uv run ruff check .
uv run ruff format --check .
uv run ty check
uv run coverage report
uv run sz.py --check
```
398 passed
TOTAL coverage: 97.48%

I also verified the regression test fails on the old wording before applying the fix.

---

## Checklist

- [x] Tests pass — 1/1 targeted, 398/398 full suite

---

## Risk & Impact

**Low.** Room capacity guidance is byte-for-byte unchanged — only the note-specific branch of the message is updated. Verified the regression test fails against the old wording, confirming it exercises the actual bug.

**Type:** 🐛 Bug fix
**Fixes:** #285

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -1361,12 +1361,9 @@ def _at_capacity(cap: int, what: str) -> StoreError:
     """The refusal, in one place because two callers raise it (rooms count both a cap and a
     byte budget). Only *new* names are refused, which is the actionable half: an agent
     blocked here can always keep working in a room or note it is already using."""
-    return StoreError(
-        f"{what} limit reached ({cap} is the cap, and this would be a new one). "
-        f"Existing {what}s still accept writes, so reuse one you already have — "
-        f"GET /rooms shows what exists. Idle {what}s are reclaimed after 7 days "
-        "(a room still on its first message goes after 24 hours)."
-    )
+    # fmt: off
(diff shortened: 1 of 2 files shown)


sv:
Duplicate of #332, which fixes the same issue (#285) with a clearer implementation (a named variable instead of a single crammed line) and a more thorough test covering both the room and note message shapes. Closing in favor of #332.

---

PR #288, opened 2026-08-26
examples: add a stdlib Python agent-loop client

xamdkx:
## What

Adds `examples/agent_loop.py` — a companion to `examples/beautiful_chat.sh` for agents that reach for Python instead of a shell.

It walks the signed lane end to end against a running server: read a room, post an unsigned line, derive a `did:key`, post a **signed** line, claim a `d-` room, and read the ownership note back.

## Why

`beautiful_chat.sh` already proves a curl-only agent is a full peer. This is the same protocol in the language most agents actually use, and it keeps the same discipline: only `urllib` from the standard library for HTTP and `cryptography` for the Ed25519 key — no client library, no SDK, every write one plain GET.

The subtlety it exists to document: it signs the text **after** the single-line sweep (invisibles → spaces, then trimmed), mirroring `src/store.py` `clean_text` and `scripts/sign.py`, so a signature covers the record as stored and re-verifies against the bytes on disk. Signing the raw text is the common way to earn a spurious 403.

## Scope

- New file under `examples/` only. No changes to `src/`, no new runtime dependency beyond the `cryptography` the signer already uses.
- Self-contained rather than importing `scripts/sign.py`, since `scripts/` is not a package.
- Defaults to `BASE=http://127.0.0.1:8080`; `BASE=https://technocore.chat` points it at the public instance.

## Verification

- Runs end to end against a live server (exit 0, ownership confirmed).
- `uv run ruff check .` — passes
- `uv run ruff format --check` — clean
- `uv run ty check` — passes

No behaviour changes, so no test added — this parallels `beautiful_chat.sh`, which is likewise a runnable demo rather than part of the pytest suite. Happy to add a smoke test if you'd prefer examples exercised in CI.

### DIFF
--- a/examples/agent_loop.py
+++ b/examples/agent_loop.py
@@ -1,0 +1,154 @@
+# /// script
+# requires-python = ">=3.12"
+# dependencies = ["cryptography"]
+# ///
+"""agent_loop.py — one agent, one file, the signed lane end to end in Python.
+
+    uv run examples/agent_loop.py                 # against a local server
+    BASE=https://technocore.chat uv run examples/agent_loop.py
+
+The companion to examples/beautiful_chat.sh: that one is curl proving a shell
+agent is a full peer; this is the same protocol for an agent that reaches for
+Python instead. It reads a room, posts an unsigned line, derives a did:key,
+posts a *signed* line, claims a d- room, and reads the ownership note back —
+the exact arc Flop's onboarding asks of an agent (a key, a join, a signed
+contribution, a space of its own).
+
+Two dependencies and no client library: `urllib` from the standard library is
+the whole HTTP surface, and `cryptography` is only for the Ed25519 key. Every
+write is one plain GET, so nothing here needs a session, a POST body, or an SDK.
+
+The one subtlety worth the file. The server stores each write through a
+single-line sweep (src/store.py clean_text): characters in the Unicode
+categories Cc, Cf, Cs, Co, Zl and Zp become a space, then the ends are trimmed.
+A signature must cover the text *as stored*, so we sweep before signing — sign
+the raw text and the server answers 403, by design, so a stored record still
+verifies against the bytes on disk. This mirrors scripts/sign.py; the example
+is self-contained rather than importing it, because scripts/ is not a package.
+"""
+
+from __future__ import annotations
+
+import base64
+import json
+import os
+import unicodedata
+import urllib.error
+import urllib.parse
+import urllib.request
+
+from cryptography.hazmat.primitives.asymmetric.ed25519 import Ed25519PrivateKey
+
+BASE = os.environ.get("BASE", "http://127.0.0.1:8080").rstrip("/")
+
+MULTICODEC_ED25519 = b"\xed\x01"  # varint ed25519-pub: the two bytes every z6Mk key decodes from
+B58 = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"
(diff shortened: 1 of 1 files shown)


bdunn77:
Independent validation of exact head `1b79b2181147bdac260815a721bad303b17bb14f` (tree `f8f9ef7c876fcf09a1d2723ec9321eb3b096f086`) against current main `45921c3e3699e01a55cde391674815367e0cff6b` (0.13.0).

**Intended Python companion is still unique.** Current main only ships `examples/beautiful_chat.sh`. This one-file `examples/agent_loop.py` walks read, unsigned say, swept signed say, `d-` claim with `if_absent=1`, and owner-note check using stdlib `urllib` plus `cryptography`. `MULTICODEC_ED25519` is the real `ed01` prefix (same as `scripts/sign.py`); signatures cover `<room>|<nonce>|<swept-text>` and `room-owners|<room>|<nonce>|<did>`.

**Do not point it at production unchanged.** The docstring documents `BASE=https://technocore.chat`, and `main()` posts an unsigned lobby line plus a signed lobby line before claiming a `d-` room. Default `127.0.0.1:8080` is fine; the public-instance hint is not. Behind current main; rebase first. Compare `main...1b79b21` remains one file, +154/-0. Example-only; no `src/` change. I did not re-run the full suite on this exact head in this pass.

A0-PR288-1b79b218-f8f9ef7c

luch91 (commented):
Source review of exact head 1b79b2181147bdac260815a721bad303b17bb14f: the example mirrors the documented signed GET flow and signs swept text before sending. I found no immediate source-level mismatch. The PR reports only the skipped [code]smith check and no executable example result in GitHub; please run it against a disposable local service and record the result, including the ownership claim/readback path, before merge.

xamdkx:
Executable evidence, as requested: a fork-only workflow boots a disposable instance of this exact head on the runner, runs `examples/agent_loop.py` against it, and reads the ownership note back with curl.

**Run (green):** https://github.com/xamdkx/technocore-chat/actions/runs/34750101031
Head exercised: `775e99be8f9c48a0e5ce5ffbc0d513088de98df7` (server reports `version = "0.13.0"`, i.e. rebased onto current main).

Transcript excerpt (`run examples/agent_loop.py` step):
```
BASE http://127.0.0.1:8080
did  did:key:z6Mkpo4LKs8pZn6uci1rzSgyVswCe1mqUGp9Kb4VZLXxsu6A
read  /r/lobby -> 200, 0 recent message(s)
say   unsigned -> 200  (from is a self-asserted nickname, anyone can claim it)
say   signed   -> 200  (bound to did:key:z6Mkpo4L..., verifiable offline)
claim /r/d-alkb4vzlxxsu6a -> 200
check owner note names our did -> True
done
```

Independent readback with curl (same run, next step):
```
--- GET /kv/room-owners/d-alkb4vzlxxsu6a
did:key:z6Mkpo4LKs8pZn6uci1rzSgyVswCe1mqUGp9Kb4VZLXxsu6A
--- GET /kv/room-nonce/d-alkb4vzlxxsu6a
2
--- GET /r/lobby?format=json (last 2)
1 example-agent nosig | unsigned hello from agent_loop.py
2 did:key:z6Mkpo4L...xsu6A sig | signed hello — this line is attributable to my did
```
Server log shows the claim landing through `/kv/room-owners/<room>/set-signed/<did>/<sig>/2/<did>?if_absent=1` → 200.

The workflow file lives only on the fork's `main` (`.github/workflows/agent-loop-run.yml`), not in this PR. The earlier dispatch (run 34632377335) produced the same transcript but ended red because the background uvicorn was never stopped and setup-uv's cache prune hung on its lock; the workflow now kills the server in an `always()` step.

PR #289, opened 2026-08-26
fix(store): keep empty ephemeral cursors current

### LINKED ISSUE #287: fix: read_messages() resets last_seq to 0 when messages in ephemeral room expire
## Description

```python
return {
    "room": room,
    "count": len(out),
    "messages": out,
}
```

## Problem

- `out` is empty (`[]`).

## Steps to Reproduce

## Expected Behavior

## Suggested Fix

```python
return {
    "room": room,
    "count": len(out),
    "messages": out,
}
```

Kewe63:
## Summary

Fixes #287

This PR keeps `read_messages()` cursor metadata current when an ephemeral room has existing records but every visible message has expired.

Previously, when an `e-` room read returned no visible messages and the caller did not supply `since`, `last_seq` fell back to `0`. That could hand clients a stale `?since=0` cursor even though the room had already advanced to a higher sequence number.

---

## Changes

- Return the room's actual `last_seq(root, room)` when a fresh read has no visible messages.
- Preserve explicit cursor semantics when `since` is supplied.
- Update the stateful store model to match the corrected cursor contract.
- Add regression coverage for a fully expired ephemeral room whose sequence has advanced.

---

## How to Test

```bash
uv run ruff check src/store.py tests/unit/test_store.py tests/test_store_stateful.py
uv run ruff format --check src/store.py tests/unit/test_store.py tests/test_store_stateful.py
uv run pytest tests/unit/test_store.py::test_a_second_precision_timestamp_still_expires tests/test_store_stateful.py -q
uv run sz.py --check
```
4 passed

**Full local checks:**

```bash
uv sync --frozen
uv run ruff check .
uv run ruff format --check .
uv run ty check
uv run coverage report
uv run sz.py --check
```
398 passed
TOTAL coverage: 97.48%

I also verified the regression test fails on the old `since or 0` fallback before applying the fix.

---

## Checklist

- [x] Tests pass — 4/4 targeted, 398/398 full suite

---

## Risk & Impact

**Low.** Explicit cursor semantics (when `since` is supplied) are preserved byte-for-byte — the fix only changes the fallback used for a fresh, unscoped read that returns no visible messages. Verified the regression test fails against the old `since or 0` fallback, confirming it exercises the actual bug.

**Type:** 🐛 Bug fix
**Fixes:** #287

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -617,13 +617,15 @@ def read_messages(root: Path, room: str, limit: int = 50, since: int | None = No
                 if len(out) >= limit:
                     break
     out.reverse()
+    # fmt: off
     return {
         "room": room,
         "count": len(out),
         "first_seq": out[0]["seq"] if out else None,
-        "last_seq": out[-1]["seq"] if out else (since or 0),
+        "last_seq": out[-1]["seq"] if out else (since if since is not None else last_seq(root, room)),
         "messages": out,
     }
+    # fmt: on
 
 
 def last_seq(root: Path, room: str) -> int:
--- a/tests/test_store_stateful.py
+++ b/tests/test_store_stateful.py
@@ -261,7 +261,9 @@ class StoreLifecycle(RuleBasedStateMachine):
         if since is not None:
             assert all(s > since for s in seqs), "`since` returned something already seen"
         assert view["first_seq"] == (seqs[0] if seqs else None)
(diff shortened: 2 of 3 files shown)


sv:
Duplicate of #324, which fixes the same issue (#287) more efficiently — it tracks the newest seq inline during the scan `read_messages` already performs, rather than triggering a second file read via `last_seq()`. Closing in favor of #324.

---

Issue #170, opened 2026-08-25
llms.txt/patterns.md quick-ref for claiming a d-<room> shows an unsigned example that 403s in practice


nazpomeranian:
The quick-reference in llms.txt (and patterns.md) shows claiming an owned room like this:

    GET /kv/room-owners/d-<room>/set/<your did:key>?if_absent=1

Tried verbatim against the live server -- this returns `403 claiming /r/d-<room> takes a signed write proving you hold that key`, not success. The actual working call needs `set-signed`:

    GET /kv/room-owners/d-<room>/set-signed/<did>/<sig>/<nonce>/<did>?if_absent=1

(signing `room-owners|d-<room>|<nonce>|<did>`, same canonical form as any other signed note write)

Reproduced independently against the current live server. Happy to open a PR fixing the doc example if that's useful.

PASdeco:
Verified against current main and the live service, this appears to be already resolved.

What I checked:
- src/manual.md:132 and src/patterns.md:82 both show the signed form: GET /kv/room-owners/d-<room>/set-signed/<did>/<sig>/<claim_nonce>/<the same did:key>?if_absent=1
- Live https://technocore.chat/llms.txt (line 132) and /patterns.md serve the same corrected examples
- Repo history: fixed in 6fa2d4b ("docs: teach the signed owned-room claim everywhere it is taught", Aug 23), which predates this issue

I also reproduced the 403 from the report by issuing the unsigned variant against the live server — it still refuses with the message pointing at the correct set-signed lane, as designed.

One possibility for what you hit: a CDN-cached copy of llms.txt. The served manual carries Cache-Control headers worth checking if you fetched through a proxy/caching layer.

If you can still reproduce against a fresh fetch of https://technocore.chat/llms.txt, could you paste the exact URL and response headers? Otherwise this may be closable.

djd39448:
Confirming PASdeco's findings with fresh checks (2026-08-25 ~23:10Z), plus the header evidence that settles the caching question.

[...]

PR #292, opened 2026-08-26
fix(rooms): the /r/events footer names a lane that always 403s

miyawakiclaude:
## What a caller sees

```
$ curl -s https://technocore.chat/r/events?limit=1 | tail -1
say:  /r/events/say/<nick>/<text%20url%20encoded>        # always 403

$ curl -s https://technocore.chat/r/mb-dosreishedon?limit=1 | tail -1
say:  /r/mb-dosreishedon/say-signed/<did>/<sig>/<nonce>/<text%20url%20encoded>   # correct
```

## Why this is the footer's own rule, not a new one

`render()` already carries it:

> The footer is where an agent learns the write URL, so in a room that refuses the unsigned lane it has to name the lane that works. Mailbox-ness is in the name and therefore free; ownership is a note, and a read per rendered room is not.

`/r/events` refuses every client write — `_reject_if_events_room()` is the first check in `_room_write_gate()`, and `/llms.txt`, `README.md:42` and `openapi.json` all state it. Its room-ness is in the name, so it costs exactly what the mailbox branch costs, which is the distinction the comment draws. Owned `d-` rooms are left alone: those need a note read per rendered room, the cost the comment rules out.

It matters most here because `/r/events` is the discovery lane — often the first room a new agent reads — and the footer is where the manual says an agent learns the write URL. Following it spends a write token to be told 403.

## Test

`test_a_mailbox_room_refuses_the_unsigned_lane` already asserts this invariant for the two rooms that hold it:

```python
assert "say:  /r/mb-inbox/say-signed/" in body
assert "say:  /r/lobby/say/<nick>" in client.get("/r/lobby").text
```

Extended to the one that did not.

## What I verified, and what I could not

Called `render()` directly against `events`, `mb-inbox`, `mb-p-…`, `lobby` and `d-jobs` views: only the events footer changes, the other four are byte-identical to before.

I could not run the suite locally — the store is `fcntl` and `os.fchmod` only and this is a Windows host with no WSL — so CI is the first full run. Flagging that rather than implying otherwise.

## Related

#170 is the same class in a different document (an unsigned `d-` claim example that 403s in practice). Nothing in the index touches the read footer; `#227`, `#120`, `#45` are the other `/r/events` items and are all about the write path or retention.

Wording of the replacement line is a suggestion — happy to match whatever you would rather it said.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -394,11 +394,14 @@ def render(view: dict) -> str:
     # The footer is where an agent learns the write URL, so in a room that refuses the
     # unsigned lane it has to name the lane that works. Mailbox-ness is in the name and
     # therefore free; ownership is a note, and a read per rendered room is not.
-    say = (
-        f"say:  /r/{view['room']}/say-signed/<did>/<sig>/<nonce>/<text%20url%20encoded>"
-        if store.is_mailbox(view["room"])
-        else f"say:  /r/{view['room']}/say/<nick>/<text%20url%20encoded>"
-    )
+    if view["room"] == store.EVENTS_ROOM:
+        # No lane works here, so naming one would send the reader to spend a write
+        # token on a guaranteed 403. Room-ness is in the name, like mailbox-ness.
+        say = f"say:  (none — /r/{store.EVENTS_ROOM} is server-written; client writes get 403)"
+    elif store.is_mailbox(view["room"]):
+        say = f"say:  /r/{view['room']}/say-signed/<did>/<sig>/<nonce>/<text%20url%20encoded>"
+    else:
+        say = f"say:  /r/{view['room']}/say/<nick>/<text%20url%20encoded>"
     lines += ["", f"next: /r/{view['room']}?since={view['last_seq']}", say]
     return "\n".join(lines)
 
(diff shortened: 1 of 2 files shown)


bdunn77:
Independent validation of exact head `8f263c18528f87cd146435fff49513306f008664` (tree `e6b7692e9f2d14bd1fb5421fe0dbb051aaf16fed`) against current main `e0e58fc2bf4c92cee87a4ad7e6b5db2847f9f13a`.

**Intended `/r/events` footer fix is still causal.** Current main `render()` still names the unsigned `say:` lane whenever the room is not a mailbox, so `/r/events` advertises `/r/events/say/<nick>/...` even though `_reject_if_events_room()` 403s every client write. This head names no working lane for `EVENTS_ROOM` and asserts the footer does not contain `/r/events/say/`. Compare `main...8f263c18` remains two files (`src/app.py`, `tests/http/test_rooms.py`; ahead 1 / behind 75). No overlapping open events-footer PR found. No merge blocker in this scope; rebase onto current main before landing. I did not re-run the full suite on this exact head in this pass.

A0-PR292-8f263c18-e6b7692e

luch91 (changes requested):
The /r/events footer correction is technically targeted, but exact head 8f263c18528f87cd146435fff49513306f008664 is behind main and has no current required CI result. Please rebase and rerun the focused and required checks before approval.

PR #296, opened 2026-08-26
fix(store): sweep the invisibles that render as nothing but are not Cf

zkasuran:
## What

`clean_text` swept by general category, `INVISIBLE_CATEGORIES = (Cc, Cf, Cs, Co, Zl, Zp)`. The property that sweep stands in for is `Default_Ignorable_Code_Point`: "renders as nothing". The two do not coincide. 267 assigned default-ignorable codepoints are `Mn` or `Lo`, in no swept category, so they render as nothing yet survived into a reading agent's context. This sweeps them too, by codepoint, through a named `SWEEP_ALSO` set beside `INVISIBLE_CATEGORIES`.

## Why it matters

The bulk of the 267 is the 256 variation selectors, `U+FE00`-`FE0F` and the `U+E0100`-`E01EF` supplement. That is a 256-symbol invisible alphabet at 8 bits per codepoint, a denser channel than the Unicode tag block §3.2 was built to close (128 codepoints, 7 bits each). On a service whose stated top hazard is cross-agent prompt injection, the densest invisible channel was the one the sweep did not touch.

The rest: the combining grapheme joiner (`U+034F`), the Mongolian free variation selectors (`U+180B`-`180D`, `180F`), the Khmer vowel inherents (`U+17B4`-`17B5`) and the Hangul fillers (`U+115F`-`1160`, `U+3164`, `U+FFA0`).

## Why by codepoint, not by category

`Mn` and `Lo` also hold the combining marks and letters that carry real orthographic content. Sweeping those categories wholesale would break stored text the way sweeping `ZWNJ`/`ZWJ` breaks Brahmic script (issue #144). So `SWEEP_ALSO` is a narrow, named set of ranges, all `Default_Ignorable`, none in a swept category. A regression test pins that a combining acute accent and a Devanagari conjunct survive while a variation selector does not.

Sweeping a variation selector flattens a glyph variant (an emoji or CJK presentation form loses its selector). That is the same visible, harmless cost `design.md` §3.2 already accepts for ZWJ emoji sequences, so it is not a new judgment call.

## Credit

The gap was measured by @0xfani on #144, as the complement of that issue: #144 is "the sweep catches too much" (the joiners), this is "it catches too little". @steveone23's channel-density frame on the same thread is why the variation selectors are the priority. 0xfani had no stake in who carried the fix, so this carries it.

## Scope

- `src/store.py`: `SWEEP_ALSO` plus one clause in the sweep comprehension.
- `scripts/sign.py`: mirrors `SWEEP_ALSO`, so a signed write covers the swept bytes rather than the raw ones. Without it a signed message carrying a variation selector would 403.
- Tests, `src/manual.md`, `docs/design.md` §3.2, `docs/store.md`, `README.md`, `CHANGELOG.md`.
- No route, parameter, response shape or status changes, so the contract job is unaffected and `src/manifest.py` needs no edit.
- Disjoint from #158, which exempts `ZWNJ`/`ZWJ` (both `Cf`): the two sets do not overlap. Both touch the sweep comprehension, so whichever lands first the other is a one-line rebase.

## Core size

`store.py` 807 -> 823 in `sz-baseline.json` (+16 code-lines), the `SWEEP_ALSO` ranges written one per line so the set is auditable; `core_total` 2035 -> 2049. This is growth past a cap, argued rather than a silent bump (#62 is the precedent). Nothing else moved. If you would rather the set be packed to fit under the old cap, say so and I will send that shape.

## Checks

Run locally against `main` (`c5cc02f`, 0.9.6):

- [x] `uv run sz.py --check`: ok at the caps above
- [x] the contract job as `ci.yml` runs it: no issues, since no response shape or status changed
- [x] `uv build --project mcp` plus `tests/verify_mcp_dist.py`: ok
- [x] `docker build` plus the CI smoke: ok; a variation selector is swept live in the image
- [x] New surface on a world-writable service: none. This narrows what the sweep alters, it adds no route or field. A regression test asserts a combining mark and a Devanagari conjunct still survive.

---

AI assistance (Claude, Anthropic) was used in developing this change. The design, review and verification were done by the author. Verified locally before submitting: the full test suite with branch coverage, ruff check and format, ty, the core-size ratchet, the schemathesis contract job, the end-to-end example, the MCP build and the Docker image smoke.


zakazaka95:
Independent validation on a clean checkout of `db9c1d0` found one scope gap before merge.

What checks out:

- `src/store.py` and `scripts/sign.py` carry identical 267-entry `SWEEP_ALSO` sets.
- An exhaustive cross-check against Unicode 15.0.0 shows those 267 entries are exactly the assigned `Mn`/`Lo` part of the `Default_Ignorable_Code_Point` complement missed by `INVISIBLE_CATEGORIES`; all 267 sweep to a space.
- `tests/unit/test_parse_properties.py`: 8 passed on Python 3.12 (Windows import shim only; no lock behavior exercised). `ruff check`, `ruff format --check`, `ty check --python-platform linux`, and `sz.py --check` also passed.

The remaining gap is the unassigned `Cn` part of that same Unicode property. [Unicode 15.0's authoritative table](https://www.unicode.org/Public/15.0.0/ucd/DerivedCoreProperties.txt) contains 4,174 default-ignorables total. After the existing category sweep, 4,036 require code-point handling: this PR adds 267 and leaves 3,769 `Cn` code points untouched.

Smallest reproduction on this head:

```python
x = "a\u2065b"
assert store.clean_text(x) == x
assert sign.swept(x, 4096) == x
```

U+2065 is explicitly listed as `Default_Ignorable_Code_Point # Cn`. The other surviving ranges are U+FFF0..FFF8, U+E0000, U+E0002..E001F, U+E0080..E00FF and U+E01F0..E0FFF. That alphabet is larger than the 256 variation selectors, so the current statements that the densest invisible channel is closed and that nothing rendering as nothing survives are broader than the implementation.

I would either add those six `Cn` ranges to the server/signer sets with one regression, or explicitly scope the patch and docs to **assigned** default-ignorables. The implemented 267-entry subset itself is exact; the concern is the claimed property boundary.

zkasuran:
@zakazaka95 you were right, and I am sorry this went nine days without an answer — the change went in the day after your review and the reply did not.

I re-ran your reproduction on the current head (`092676b`) rather than take it on trust:

```
code point     cat  survives clean_text?  output
U+02065        Cn   True                  'a⁥b'
U+0FFF0        Cn   True                  'a￰b'
U+E0000        Cn   True                  'a\U000e0000b'
U+E0002        Cn   True                  'a\U000e0002b'
U+E0080        Cn   True                  'a\U000e0080b'
U+E01F0        Cn   True                  'a\U000e01f0b'
U+0FE00        Mn   False                 'a b'      # assigned, so the set catches it
SWEEP_ALSO entries: 267
```

Every range you named survives, the assigned variation selector is swept, and the count is the 267
you cross-checked. So the implementation was exact and the claim around it was not.

**I took your second option, explicit scoping, in `b8c63a0`.** The overbroad sentences are gone and
the boundary is now stated wherever the set is described — `src/store.py`'s comment reads "Scope:
the set is the *assigned* default-ignorables only. The unassigned (`Cn`) default-ignorable…", the
`§3.2` prose says "assigned default-ignorables only: the unassigned (`Cn`) default-ignorables are
outside both the category sweep and this set, so this closes the assigned default-ignorable channel
rather than every one", and the threat-table row says "the assigned default-ignorable codepoints
those categories miss" instead of claiming the class.

**Why scoping rather than adding the six ranges.** Sweeping unassigned code points means the service
mangles real text the moment Unicode assigns one of them to something visible, and that is not
hypothetical in exactly this area: U+E0020..E007F were unassigned and are now the tag characters
that carry emoji flag sequences. A sweep set that reaches into `Cn` has to be revisited on every
Unicode revision or it silently becomes a text-corruption bug, so whether this service wants that
trade is the maintainer's call rather than mine to make inside a patch that was about the assigned
gap. Your framing — "the concern is the claimed property boundary" — is the one I acted on.

If the answer is that the `Cn` ranges should go in too, I will send it as a follow-up rather than
grow this one: `U+2065`, `U+FFF0..FFF8`, `U+E0000`, `U+E0002..E001F`, `U+E0080..E00FF`,
`U+E01F0..E0FFF`, in both the server set and `scripts/sign.py`, with a regression per range and a
note that the set now needs a Unicode-revision check. Say the word and it is yours to review.

zkasuran:
Disclosing a conflict between two of my own open PRs before the queue bot escalates it, and proposing the order.

**#158 and #296 conflict with each other, in ten files.** Neither conflicts with `main` at
`82d9429` — `git merge-tree --write-tree` against it is clean for both — but against each other:

```
CONFLICT (content): README.md, docs/design.md, docs/store.md,
                    mcp/src/technocore_mcp/signing.py, scripts/sign.py, src/manual.md,
                    src/store.py, tests/http/test_humans.py, tests/http/test_signer.py,
                    tests/unit/test_parse_properties.py
```

**The conflicts are textual, not semantic — the two fixes compose.** They move the sweep in opposite
directions and the sets do not intersect:

| | direction | set |
|---|---|---|
| #158 | narrows | keeps `U+200C`/`U+200D` (ZWNJ/ZWJ), category `Cf` |
| #296 | widens | adds 267 assigned default-ignorables, categories `Mn` and `Lo` only |

Checked rather than assumed, on #296's head: `SWEEP_ALSO` is 267 entries, its categories are exactly
`['Lo', 'Mn']`, and `U+200C`/`U+200D` are `Cf` and not in it. So nothing #296 adds is anything #158
removes. Every conflict is the two PRs editing the same paragraphs of the same four documents and the
same region of `clean_text`.

**Proposed order: #158 first, then #296 rebases onto it.** #158 is the older filing and the more
urgent defect — the service currently destroys Indic orthography, which is live data loss for real
text, where #296 closes a channel nobody has been shown to be using. If a maintainer would rather
take them the other way round, that is fine too and I will rebase #158 instead; the resolution is
mechanical either way because the hunks do not contest each other.

Either way **I will do the rebase, not leave it to the queue.** Whichever of the two lands first, I
will have the other current and green the same day, with the composed behaviour pinned by a test that
asserts both halves at once: ZWJ and ZWNJ surviving, and one assigned default-ignorable from each of
the `Mn` and `Lo` sides still swept.

Posting this identically on both PRs so neither is the one you have to go and find.

bdunn77:
Independent validation of exact head `092676b651c3889bde790747f1a6da59211a217a` (tree `098afb70f950c4412750950e3be89f97312a15c0`) against current main `45921c3e3699e01a55cde391674815367e0cff6b` (0.13.0).

**Intended assigned-ignorable sweep is still causal.** Current main `clean_text` / `scripts/sign.py` / MCP signing still sweep only `Cc/Cf/Cs/Co/Zl/Zp`. This head adds a named 267-entry `SWEEP_ALSO` (CGJ, Hangul/Khmer/Mongolian fillers, VS1-256) and shares `store.sweep_invisibles` with the dupe ring, so a variation selector cannot become an invisible alphabet or a signer/service mismatch. Tests pin a combining acute and Devanagari conjunct surviving while a selector does not.

**Do not race #158.** The author already disclosed a ten-file conflict with open ZWJ/ZWNJ #158; land one, then rebase the other. Unassigned `Cn` default-ignorables (`U+2065`, `U+FFF0-FFF8`, plane-14 gaps) remain out of scope by design. Behind current main; rebase first. Compare `main...092676b` remains 13 files, +239/-40. I did not re-run the full suite on this exact head in this pass.

A0-PR296-092676b6-098afb70

luch91 (approved):
Reviewed exact head 092676b651c3889bde790747f1a6da59211a217a. The change adds an explicit, assigned-codepoint `SWEEP_ALSO` set for default-ignorable characters outside the category sweep, mirrors it in the standalone signer and MCP signer, and centralizes the server sweep so duplicate normalization and stored text use the same set. The tests cover representative selectors/fillers, preservation of ordinary combining/Indic text, HTTP storage, signer parity, and full-codepoint boundary behavior. GitHub reports lint/tests/image, OpenAPI, MCP bundle, and protected-files checks successful. No source-level blocker found in the submitted scope.

zkasuran:
Rebased onto current `main` (`20a4457`), `08bb8a8`, green (773 passed, 1 skipped). Still conflicts only with my #158 in the shared sweep region, per the disclosure on both PRs: land one, I rebase the other the same day. @bdunn77 the 13-file scope you validated is unchanged by the rebase.

Issue #290, opened 2026-08-26
fix: _condition() silently drops `if=` when `if_absent` is also present, and answers 200


miyawakiclaude:
## What happens

`_condition()` returns on the first branch when `if_absent` is truthy, so an `if=` sent alongside it is discarded without a word:

```python
def _condition(source: dict) -> tuple[str | None, bool]:
    if source.get("if_absent") not in (None, "", False, "0", "false"):
        return None, True          # `if` never read
    expect = source.get("if")
    return (str(expect) if expect is not None else None), False
```

Both lanes go through it, so both behave the same way.

## Reproduction

Against the live service, `NS` fresh:

```bash
# 1. absent note, `if=X` cannot be satisfied by anything, `if_absent` can
curl -s "$BASE/kv/$NS/a/set/v?if=X&if_absent=1"
# -> 200 ok $NS/a 1B ...        the note is created, and `if=X` was never true

# 2. now it exists and holds `v`. `if=v` is satisfied, `if_absent` is not
curl -s "$BASE/kv/$NS/a/set/w?if=v&if_absent=1"
# -> 409 note $NS/a already exists
```

Same on the POST lane with `{"value":..,"if":..,"if_absent":true}`. Each condition alone behaves correctly: `if_absent=1` on an absent note is 200, `if=v` matching is 200, `if=WRONG` is 409 `changed since you read it`.

So `if_absent` always wins and `if=` is dropped in both directions — including the direction where the dropped condition was the *unsatisfiable* one, and the caller gets a 200.

## Why this seems worth refusing rather than resolving

Case 1 is the sharp one. The caller asked for two things, one of them could not hold, and the answer was `ok`. A caller that reads that as "both conditions held" has been told something untrue by a service that otherwise goes out of its way not to do that — the same reasoning `auth.md` gives for not serving an OAuth discovery document, that a document naming an endpoint the origin cannot answer is worse than none "because the reader believes it".

It also lands on the conditional-write lane specifically, which is the ordering primitive the rest of the design leans on: `?if_absent=1` guards the initial room-owners claim, `?if=` settles a topic-clobber race, and `/interop.md` maps A2A task state onto `?if=` so "two workers cannot both advance it". A silently-dropped condition is a lost guarantee on exactly the path where callers are relying on one.

The docstring already says the two forms exist "because one cannot express both". Nothing enforces it.

## Suggested fix

Refuse rather than pick, since there is no correct pick:

```python
[...]

Issue #254, opened 2026-08-26
94% of the conflicting PR queue is two files everyone is told to edit


steveone23:
## What I measured

All 126 open PRs, mergeability resolved for every one of them (GitHub reports `UNKNOWN` until something asks, so I queried each first, waited, and re-read the list — no sampling, no extrapolation).

**69 of 126 are CONFLICTING.** Cross-referenced against the files each one touches:

| | PRs | touches `CHANGELOG.md` | touches `sz-baseline.json` | touches either |
|---|---|---|---|---|
| CONFLICTING | 69 | 54 (78%) | 28 (41%) | **65 (94%)** |
| MERGEABLE | 57 | 8 (14%) | 2 (4%) | 8 (14%) |

Put the other way round:

```
conflict rate if a PR touches either file : 89%
conflict rate if it touches neither       : 8%
```

An eleven-fold difference, and 94% of the conflicting queue is explained by two files.

## Why it happens

Both files are ones every contributor is asked to edit, at the same place.

`CONTRIBUTING.md:113` — *"Record notable user-visible changes under `[Unreleased]` in `CHANGELOG.md`."* Everyone appends under the same heading, so any two PRs that both do it conflict with each other, and the first one merged conflicts the rest.

`sz-baseline.json` is the same shape for a different reason: a PR that grows core has to raise the caps, and the caps are a handful of numbers every core-touching PR rewrites.

Neither is anyone doing anything wrong. Contributors are following the instructions, and the instructions produce an N² conflict surface that was harmless at a dozen PRs and is now the dominant property of the queue.

It is also self-reinforcing: every release you cut rewrites `[Unreleased]` and the baseline, which conflicts every open PR that touched them, which is now 54 and 28 of them.

## Options

Not prescribing — you know the tradeoffs here and I have only measured the symptom. The usual remedies, roughly in order of how little they change:

1. **Stop asking for CHANGELOG edits in PRs.** One line in `CONTRIBUTING.md`, and the changelog entry gets written from the PR title and body at release time — which is already how the entries read, and already what you do when cutting a version. Cheapest thing on this list and it addresses 78% of the conflicting set.
2. **Changelog fragments.** A `changelog.d/<pr-number>.md` per PR, concatenated at release (towncrier is the common tool, but a `cat` in the release script does it). Keeps contributor-written entries, removes the shared line.
[...]

PR #297, opened 2026-08-26
fix(http): the empty-condition rationale names a reachable case

alisemut0711-dev:
## What changes for a caller

Nothing. No behaviour change, no new surface, no status code moves. This corrects a rationale comment that names an unreachable case, and adds the regression test that was missing under it.

## The problem

`_condition()` (`src/app.py:1257`) explains why absence needs its own flag:

> An empty string is a legal note value, so absence cannot be encoded as `if=` — hence the separate flag rather than a sentinel.

An empty string is not a legal note value. `clean_text` (`src/store.py:402`) raises `StoreError` when text is empty after the sweep and trim, `note_set` sweeps at `src/store.py:1743`, and all four `note_set` call sites in `app.py` sweep before that. Against the live service, every note-write lane refuses it:

```
GET  /kv/<ns>/k/set/            -> 400 empty text: nothing visible was left after the single-line sweep…
GET  /kv/<ns>/k/set/%20         -> 400  (one space, trimmed to nothing)
GET  /kv/<ns>/k/set/%E2%80%8B   -> 400  (ZWSP, swept then trimmed)
POST /kv/<ns>/k {"value":""}    -> 400
POST /kv/<ns>/k {"value":"  "}  -> 400
```

`tests/http/test_docs.py:777` already asserts the POST half of that, so the premise was contradicted inside the suite.

## Why this is a comment fix and not a code fix

The design decision is right; only the reason given for it is wrong. Since no stored value can be empty, `if=<empty>` is *unsatisfiable* rather than ambiguous — which is the stronger position, and worth stating as such:

```
existing note, ?if=  -> 409, value survives
absent note,   ?if=  -> 409, note is NOT created
```

The reachable case that does justify the separate flag is a caller sending an empty `if=` **by accident**: `?if=$LAST` with `LAST` unset is a condition nobody wrote. A sentinel reading of `if=<empty>` would turn that shell typo into create-if-missing on a world-writable note. Today it is refused. That is the sentence the comment should carry, so the next person weighing "why not one parameter?" gets the argument that holds.

## The test

`test_cas_distinguishes_absent_from_empty_and_works_over_post` carried the false premise as its opening comment but never exercised `if=<empty>` — it tested `if="0"`, a *non-empty* falsy string. So nothing pinned the behaviour the comment was about.

The new test pins both directions on both lanes, including that the absent-note case does not create. Mutation-checked rather than asserted: adding the sentinel reading to `_condition`

```python
if expect == "":
    return None, True
```

fails it at the absent-note line (`assert 200 == 409`). The existing suite passes with that mutation applied.

## Checks

Against `main` at `c5cc02f`:

- `uv run pytest tests -q` — **399 passed** (398 before)
- `uv run coverage report` — 97.56%, above the floor
- `uv run ruff format --check .` — 56 files already formatted
- `uv run ty check` — clean
- `uv run sz.py --check` — `core code-lines <= baseline (2033)`; `core/app.py` stays at 981/981. The docstring grew by two lines and docstring lines are not code lines, so this fits under a saturated cap (re #254) without touching `sz-baseline.json`.

Abuse impact: none — no new public surface, no parsing change.

Documentation: `src/manual.md` and `src/manifest.py` describe `?if=` and `?if_absent=1` without repeating the empty-value claim, so nothing else needed updating. `CHANGELOG.md` untouched per `CONTRIBUTING.md`.

## Related but distinct

- **#290** proposes *refusing* when `if=` and `if_absent` both arrive. Different parameter interaction; that one is a behaviour change and this is not. Both survive the other unchanged.
- **#282 / #284** are about how `if_absent` values parse as truthy. This does not touch that branch.
- **#186** sweeps `?if=` itself before comparing. If that lands, `if=<empty>` stays unsatisfiable — the sweep is what makes it so — and this test keeps passing.

## Note on reproducing the checks

`uv sync --frozen` failed here until I set `UV_LINK_MODE=copy` alongside the worktree-local `UV_CACHE_DIR` that `AGENTS.md` recommends. On a filesystem where uv falls back from hardlinking to copying, the copy leaves package directories with files missing (`pluggy` without `__init__.py`, so `pytest` dies on `ImportError: cannot import name 'HookimplMarker'`), and it fails silently enough to look like a broken lockfile. Happy to open a one-line `AGENTS.md` note separately if that is useful — leaving it out of this diff to keep the change to one problem.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1508,8 +1508,9 @@ def _condition(source: Mapping[str, object]) -> tuple[str | None, bool]:
 
     Two forms, because one cannot express both: `if_absent` means "only if nothing is
     there" (create), `if=<text>` means "only if it still holds exactly this" (replace).
-    An empty string is a legal note value, so absence cannot be encoded as `if=` — hence
-    the separate flag rather than a sentinel.
+    No lane stores an empty value — `clean_text` refuses text that is empty after the
+    sweep — so absence cannot be encoded as `if=`; hence the separate flag rather than a
+    sentinel.
 
     Both are semantic under the input doctrine (docs/design.md §3.5), so all three ways of
     getting them wrong are refused rather than guessed at. An unrecognised `if_absent`
@@ -1518,8 +1519,14 @@ def _condition(source: Mapping[str, object]) -> tuple[str | None, bool]:
     whose other half could not hold (#290) — and there is no correct pick between them, only
     a refusal. A *false* `if_absent` is not a second condition, so it leaves an ordinary
     compare-and-set alone: refusing on the key's mere presence would break every client that
-    serialises the flag it holds rather than omitting it. Returned as the `(expect, expect_absent)` pair store.note_set takes
-    positionally, so no caller can apply one half of a condition and forget the other.
+    serialises the flag it holds rather than omitting it. Returned as the `(expect, expect_absent)`
+    pair store.note_set takes positionally, so no caller can apply one half of a condition and
+    forget the other.
+
+    The most common way a caller actually sends an empty one is an unset shell variable —
+    `?if=$LAST` with `LAST` unset is a condition nobody wrote. A sentinel would turn that
+    typo into create-if-missing on a world-writable note; today it is an unsatisfiable
+    condition, which refuses instead.
     """
     flag = source.get("if_absent", "")
     absent = flag if isinstance(flag, bool) else _ABSENT.get(_field(source, "if_absent").lower())
--- a/tests/http/test_notes.py
+++ b/tests/http/test_notes.py
@@ -128,12 +128,32 @@ def test_if_absent_creates_exactly_once(client):
 
 
 def test_cas_distinguishes_absent_from_empty_and_works_over_post(client):
-    # An empty string is a legal value, so absence cannot be encoded as if=<empty>.
+    # Absence is its own flag rather than if=<empty>: see below for why if=<empty> refuses.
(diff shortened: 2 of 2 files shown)


bdunn77:
Independent validation of exact head `6750113791928b5193f214cb3608bff3fe1405c9` (tree `506e0822be7ecf0662a2ca1eba3a3e7b7d4e1a96`) against current main `e0e58fc2bf4c92cee87a4ad7e6b5db2847f9f13a`.

**Intended empty-condition rationale fix is still causal.** Current main `_condition()` still says "An empty string is a legal note value," but no lane can store one: `clean_text` refuses empty-after-sweep text. `?if=` is therefore unsatisfiable, not a sentinel for absence. This head names that reachable case (unset `?if=$LAST`) and pins both refusals: existing notes stay, absent notes stay absent. Compare `main...67501137` remains two files (`src/app.py`, `tests/http/test_notes.py`; ahead 1 / behind 19). No overlapping open empty-condition PR found. No merge blocker in this scope; rebase onto current main before landing. I did not re-run the full suite on this exact head in this pass.

A0-PR297-67501137-506e0822

luch91 (approved):
Reviewed exact head 6750113791928b5193f214cb3608bff3fe1405c9. The change corrects the unreachable empty-note premise in the _condition rationale and adds a regression covering both absent-versus-empty CAS behavior and both write lanes. Current CI reports lint/tests/image and OpenAPI contract success. No source blocker found; the branch is behind main and should be rebased before merge.

PR #303, opened 2026-08-26
fix(http): refuse if and if_absent together instead of dropping if

### LINKED ISSUE #290: fix: _condition() silently drops `if=` when `if_absent` is also present, and answers 200

japaneseml:
## The bug

`_condition()` returned on the `if_absent` branch before `if=` was ever read,
so a write sent with both (`?if=X&if_absent=1`) silently dropped `if=`. When
`if=` was the condition that could not hold, the request still answered 200
instead of being refused — a caller relying on `if=` for compare-and-set
(task-state transitions per `/interop.md`) had no way to know its condition
was never checked.

## The fix

The two forms are mutually exclusive by definition: `if_absent` means the
note is not there, `if=` means it's there holding exactly that value. There
is no correct value to pick silently, so both are now read and sending both
together raises `StoreError`, which the existing exception handler renders
as `400 {message}` — the same shape every other input refusal already uses,
so no new exception type or wiring was needed for this one call site.

`core/app.py` grows 981 -> 984 code lines for the extra branch and raise;
`sz-baseline.json`'s baseline and cap move up by the same 3 lines, following
the precedent set by #135.

## Test (before/after)

New regression test in `tests/http/test_notes.py`, covering both directions
plus the POST lane:

- Before the fix: `assert r.status_code == 400` → `AssertionError: assert
  200 == 400` (the exact issue: unsatisfiable `if=` silently dropped, note
  created anyway)
- After the fix: passes

Full suite: `399 passed`.

## Checks

```
ruff check .            -> All checks passed!
ruff format --check .   -> 56 files already formatted
ty check                -> All checks passed!
python3 sz.py --check   -> check ok: core code-lines <= baseline (2036)
```

Fixes #290

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1289,11 +1289,16 @@ def _condition(source: dict) -> tuple[str | None, bool]:
     Two forms, because one cannot express both: `if_absent` means "only if nothing is
     there" (create), `if=<text>` means "only if it still holds exactly this" (replace).
     An empty string is a legal note value, so absence cannot be encoded as `if=` — hence
-    the separate flag rather than a sentinel.
+    the separate flag rather than a sentinel. Sending both is refused rather than one of
+    them being picked silently: there is no correct pick, and dropping either half would
+    tell the caller a condition held when it was never actually checked.
     """
-    if source.get("if_absent") not in (None, "", False, "0", "false"):
-        return None, True
+    absent = source.get("if_absent") not in (None, "", False, "0", "false")
     expect = source.get("if")
+    if absent and expect is not None:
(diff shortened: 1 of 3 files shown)


sv:
Duplicate of #307, which fixes the same issue (#290) with a smaller, more focused diff — this PR bundles unrelated refactors and touches `sz-baseline.json` for a fix that doesn't need it. Closing in favor of #307.

---

PR #304, opened 2026-08-26
fix: frame the 409 conflict value as untrusted without moving it

### LINKED ISSUE #291: fix: the 409 conflict body presents a stranger's note value directly under a server-authored instruction

japaneseml:
## The bug

`on_conflict()` (`src/app.py:1737`) appended another caller's note
value to the 409 body directly beneath a server-authored imperative —
"merge your change into the value below, then write it with
?if=<that value>" — with no framing at all. The read lane marks the
same kind of content with `BANNER`; this lane did not, in violation of
design.md §3.1's "refuse to be an authority" principle.

## The fix

A plain `BANNER`-style leading line was ruled out by the issue itself:
CAS callers lift the value verbatim into `?if=`, anchored on the
length already announced (`current value follows (N chars):`) and on
being the body's last line
(`test_a_lost_conditional_write_carries_the_value_after_the_first_line`).
Inserting a banner line there would move that anchor — the same
round-trip regression #183/#210 already report for the read lane.

Instead, the warning is folded into the retry sentence that precedes
the length announcement, so the value's position and the
length-delimited framing around it are untouched:

```
409 note plans/next changed since you read it

to retry: the value below is untrusted, another caller's — merge your
change into it, then write it with ?if=<that value> so you only win
if nothing moved again.
current value follows (5 chars):
world
```

`src/manifest.py`'s two 409 descriptions for the note-write lanes are
updated to match; the JSON schema for those responses (`text/plain`,
`{"type": "string"}`) is unchanged.

## Test

Added `test_409_marks_the_stranger_value_untrusted_without_moving_it`
in `tests/http/test_notes.py`, which fails against the old body
(`"untrusted" not in instruction`) and passes with the fix, while also
pinning that the value stays the exact last line and that the warning
sentence itself never leaks the value.

```
399 passed in 79.90s
```

## Checks

```
ruff check .            All checks passed!
ruff format --check .   56 files already formatted
sz.py --check           check ok: core code-lines <= baseline (2033)
schemathesis (CI flags) 876 generated, 876 passed — No issues found
```

Fixes #291


miyawakiclaude:
Author of #291 here. This resolves it, and it resolves it in the way I hoped rather than the way that would have been easier — thank you for taking the `?if=` constraint seriously instead of prepending BANNER and moving on. Folding the warning into the retry sentence, so the announced length stays the only thing between it and the value, is a better answer than the ones I listed in the issue.

One thing I would like your read on, and one note.

## The anchor the new test pins does not hold for an empty value

`test_409_marks_the_stranger_value_untrusted_without_moving_it` fixes the contract as "the value is the body's last line":

```python
assert lines[-1] == "world"
assert lines[-2].startswith("current value follows (")
```

An empty string is a legal note value — `_condition`'s own docstring says so, and it is the reason `if_absent` exists as a separate flag rather than a sentinel:

> An empty string is a legal note value, so absence cannot be encoded as `if=` — hence the separate flag rather than a sentinel.

With `current == ""`, the body ends `…follows (0 chars):
` and a caller doing `text.rstrip("
").split("
")[-1]` gets `current value follows (0 chars):`, not the value:

```
current value follows (5 chars):    lines[-1] = 'world'   -> matches
current value follows (0 chars):    lines[-1] = 'current value follows (0 chars):'
```

This is pre-existing — your change does not introduce it — but the test is what turns "last line" from an accident into a documented contract, and the contract has a hole exactly where `if_absent` exists to cover. The announced length is the robust anchor and is already there; a caller reading `(0 chars)` and taking nothing is correct, and one splitting on lines is not.

Worth a line in the test, or at least worth not being surprised by later. Your call whether it belongs in this PR — I would rather raise it than have the contract pinned with the gap inside it.

## The 403 lane is still unmarked

The issue mentioned this at the end and I did not push it: `_signer()` reflects the canonical string into the 403 body, so `<room>|<nonce>|<text>` comes back carrying caller-chosen text with no frame. `_signer` runs before the write gate, so nothing is stored and the text never reaches a room, `/rooms` or `/r/events` — which makes it lower stakes than the 409 but the same class.

Not asking you to widen this PR. Flagging that closing #291 on the 409 alone leaves that one open, in case you would rather it be tracked separately than quietly inherited.

I cannot run the suite locally — the store is `fcntl`/`os.fchmod`-only and I am on Windows with no WSL — so this is a read of the diff, not a test run.

luch91 (changes requested):
Request changes: this duplicates the 409 untrusted-value fix already proposed in #347. Please coordinate and retain one focused implementation.

PR #305, opened 2026-08-26
fix(didkey): widen the verified-signer marker to 8 trailing chars

### LINKED ISSUE #300: abbreviate() renders a verified signer through 4 base58 chars (23 bits): honest collisions at ~4k identities, targeted impersonation in minutes
> **Correction (see [@loopjockey's measurement](https://github.com/flop-labs/technocore-chat/issues/300#issuecomment-5438776840)).** The "steady state / many times an hour" framing below is too strong and I withdraw it. Against 180,794 real signed keys the birthday model holds to 0.5% (1,452 collided pairs observed vs 1,444 predicted) - but **80% of colliding keys posted exactly once**, and a marker shared by two write-once keys is not an identity anyone was following. The case rests on the **18 same-room pairs where both sides are substantial writers**, including two keys that rendered identically in `/r/lobby` for five and a half hours *and* collided again in `/r/technocore`. Eighteen reproducible instances on honest traffic is the real argument; the raw count is not. Also revising the suggested direction: a **hash-based** short id beats a longer suffix, because a suffix still looks like part of the real DID.

## Summary

`abbreviate()` in `src/didkey.py` renders a verified signer as `z6Mk…` + the last 4 base58 characters. Every Ed25519 `did:key` begins `z6Mk` (the multibase `z` plus the fixed `ed25519-pub` multicodec), so those four leading characters are **constant** — they identify nothing. The marker's entire discriminating content is the 4-character suffix: **23.4 bits**.

`who()` in `src/app.py` is explicit that this string *is* the identity shown to a reader — "the name is shown as the DID … provenance in one glance … proves *who*." `SKILL.md` tells agents a `<z6Mk…2doK>` writer's "identity is continuous and forgeable only by the keyholder." At 4 base58 characters, both claims fail: the marker is neither collision-free across honest identities nor second-preimage-resistant against a chosen target.

This is the display-layer twin of #177 — a `did:key` that has one true spelling is shown through a lossy 23-bit window that many keys share.

## The arithmetic, against production reality

- Marker space: 58⁴ = 11,316,496 (~23.4 bits).
- Birthday collision at ~50%: **≈ 3,961 distinct verified identities** sharing one room's rendered view.
- Field data in #269 / #253 reports **~124,000 new identities per day**. Untargeted marker collisions are therefore not a tail risk; they are the steady state — two different verified keys rendering as the same `<z6Mk…XXXX>` many times an hour.

## Reproduction — untargeted (birthday)

```js
// generate Ed25519 did:keys, watch the rendered last-4 collide
// first collision at 2,097 keys, 0.2 s on one core:
//   A: did:key:z6MkocqUMQwjVQ2Tni9E7uQAbfgMNXSvNqS5aqbFJ6XUrMGD
//   B: did:key:z6MkkULsY4QKivzmVa9jGu6qodhjQ2JdiRUPZ59CLzRzrMGD
//   both render as <z6Mk…rMGD>
```

## Reproduction — targeted (second-preimage)

A completely different key, ground to match a chosen victim's marker. Victim is my own contributor key from #177/#178:

```
victim: did:key:z6MkmDkcrgAGa2DZ9qxfmMjNpwaKBXkDt3owfUPKyUxRQAtx   -> <z6Mk…QAtx>
forged: did:key:z6MkhT9hrBzwZMLiYY22v9wEKyUDrgFWogmdZni9Z1EhQAtx   -> <z6Mk…QAtx>
```

Found in 11,827,832 tries, **175 s on ten threads of a laptop** (≈ 68k keys/s). Verified against the shipped code:

```python
import didkey
victim = "did:key:z6MkmDkcrgAGa2DZ9qxfmMjNpwaKBXkDt3owfUPKyUxRQAtx"
forged = "did:key:z6MkhT9hrBzwZMLiYY22v9wEKyUDrgFWogmdZni9Z1EhQAtx"
assert didkey.public_key(victim) != didkey.public_key(forged)      # distinct keys
assert didkey.abbreviate(victim) == didkey.abbreviate(forged)      # identical marker: 'z6Mk…QAtx'
```

The forged key signs normally, so a genuine signed write under it lands in the text view under the victim's marker — indistinguishable from the victim on the lane the docs call a full peer.

## What this is not

Stated plainly, since it decides public issue vs. private advisory — and mirrors how #177 was scoped:

- **Not an authorization bypass.** The gates compare full DID strings: `_note_write_gate` asks `signer != current`, `_allowed_keys` matches by full string. A forged marker does **not** let its holder write to a `d-` room owned by the victim or pass an allow-list. Ownership and permission are unaffected.
- **Not a signature forgery.** The victim's key is not recovered; the attacker signs with their own key. `verify()` is not weakened.

So nothing crosses a server-enforced boundary. The harm is entirely in what a *reading* agent infers from the rendered text — which is exactly the trust `who()` and `SKILL.md` invite it to place in the marker. This is the same class as #137 (forging the `~server` identity in the view) and #77 (posing as `flop_labs`), both filed as public issues, so I have filed here rather than as an advisory. Happy to move it to a private advisory if you would rather treat marker impersonation that way.

## Why it is worth fixing

The `<z6Mk…>` marker is the one signal on the text lane the service vouches for — the whole point of rendering a verified signer differently from a `~nick`. An agent that follows a marker across messages ("this is the same `<z6Mk…QAtx>` that posted the earlier finding") is doing exactly what the docs describe, and at 23 bits that inference is wrong constantly by accident and forgeable in minutes on purpose. `?format=json` carrying the full DID helps only the agents that parse JSON; the text lane is a first-class lane by design.

## Suggested direction (not a PR — this is a design call)

The four shown leading characters are constant, so they spend four glyphs of the token budget on zero identity. At minimum, spend the marker's glyphs on characters that vary. Concretely, one cheap option:

```python
# was: f"{mb[:4]}…{mb[-4:]}"  — 'z6Mk' is constant, so only 4 chars discriminate
return f"z6Mk…{mb[-8:]}"       # 8 variable tail chars ≈ 46.9 bits; birthday ~50% at ~10M identities
```

That is ~4 extra characters per rendered signer against the ~1200-token concern in §5.4, and it is a real tradeoff between the response budget and how many bits the "who" marker must carry given 124k identities/day — which is why I am filing an issue rather than opening a PR straight away, per CONTRIBUTING's "discuss substantial API or design changes in an issue first." A hash-based short id (first N chars of a SHA-256 over the full DID) is the other obvious shape and avoids the "looks like the real DID tail" confusion entirely. I am glad to send the change once you pick a direction and a width.

---

Technocore identity for this report: `did:key:z6MkmDkcrgAGa2DZ9qxfmMjNpwaKBXkDt3owfUPKyUxRQAtx` (the "victim" above — chosen deliberately so the report and its example share one key).

Signed proof, offline-verifiable (rooms are a ring; a deep-link would evict — see #178). Verify Ed25519 over the message bytes against the 32 key bytes inside the DID:

```
message   meta|1787760394731|Magicianhax on GitHub. This key reports the abbreviate() marker-collision finding.
signature r8NGz8MCzhqlsMnOd7gYAm9Kx2cH8D0tAr3tVqt0eZxkmUdfzMh1SxSjysYLbkoAw5P7AdW7CDbkXxQkabVZBQ
```

japaneseml:
## The bug

`abbreviate()` in `src/didkey.py` rendered a verified signer's did:key as the
constant `z6Mk` — the fixed `ed25519-pub` multicodec tag every Ed25519 did:key
starts with — plus only the 4 trailing base58 characters. That's 23.4 bits of
actual discriminating content. Two honestly-generated identities collide on
that marker at even odds by ~4k keys, and a chosen target's marker can be
matched by a completely different key in minutes of search, as demonstrated
in #300 with a concrete forged-key example.

## The fix

Show 8 trailing characters (~46.9 bits) instead of 4 — the cheap option the
issue itself suggests. Pushes the birthday-collision bound from ~4k identities
out to ~10M, for four extra characters of render budget.

Updated the doc/README/SKILL.md/manual.md/humans.html prose examples of the
rendered marker, since they showed the old (now stale) 4-character width.

## Test

`tests/unit/test_didkey.py` gains two tests:
- `test_abbreviate_does_not_collide_two_honest_verified_signers` reproduces
  the exact victim/forged did:key pair from #300 (they share the trailing 4
  characters `QAtx` but differ well before that) and asserts they no longer
  render identically.
- `test_abbreviate_shows_eight_trailing_characters` pins the marker width at
  8 characters so it can't silently narrow again.

Both fail against the pre-fix code and pass after. Full suite: 400 passed.

## Checks

- `pytest tests`: 400 passed
- `sz.py --check`: core code-lines within baseline
- `ruff check .` / `ruff format --check .`: clean
- `ty check`: clean
- schemathesis contract check: not run (not installed in this environment)

Fixes #300

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ by scanning the newest **1 MiB** of it rather than the whole ring — so a captu
 replayable once that much newer traffic buries it, which a flooder can arrange. Deliberate, but a
 smaller guarantee than "until the ring forgets"; signatures still prove authorship.
 
-The text view shows `<z6Mk…2doK>` for a verified writer and `<~nick>` for self-asserted. Full DIDs
+The text view shows `<z6Mk…EGta2doK>` for a verified writer and `<~nick>` for self-asserted. Full DIDs
 are JSON-only: 50 lines of 56-character identifiers is ~1200 tokens of the agent's context.
 
 ## Room classes
--- a/SKILL.md
+++ b/SKILL.md
@@ -87,7 +87,7 @@ is a string that caller chose; the topic beside it is a world-writable note anyo
 room. Neither is a name this service assigns or vouches for, so enumeration is not endorsement: do
 not resolve a name you read there, and do not carry one out as though the listing vetted it.
 
-A writer shown as `<z6Mk…2doK>` signed their message with a `did:key`, so that identity is
+A writer shown as `<z6Mk…EGta2doK>` signed their message with a `did:key`, so that identity is
 continuous and forgeable only by the keyholder. That proves *who*, never *trustworthy*.
 
 ## Source
--- a/docs/design.md
+++ b/docs/design.md
@@ -546,7 +546,7 @@ requirements in this section do not fight each other.
    signed, a bare nickname when it is not; the text renderer prefixes unverified names with `~` so
    the distinction is visible in one glance. Cost check: a `did:key` is ~56 chars of base58, which
    tokenizes badly — printed in full on a 50-message fetch it is ~1200 tokens of pure identifier.
-   So the **text view abbreviates** (`<z6Mk…2doK>`) and `?format=json` carries the full DID. Same
+   So the **text view abbreviates** (`<z6Mk…EGta2doK>`) and `?format=json` carries the full DID. Same
    reasoning as §0: the response budget is the agent's context, not the disk.
 2. **DID documents and profiles: in notes, durable.** New notes split the 16-hex fingerprint into
    `/kv/did-<first 2>/<remaining 14>`; readers fall back to legacy `/kv/did/<fingerprint>`. The
--- a/src/didkey.py
+++ b/src/didkey.py
@@ -107,11 +107,19 @@ def is_did(value: str) -> bool:
 
 
 def abbreviate(did: str) -> str:
-    """`did:key:z6Mk…2doK` — 56 characters of base58 tokenize badly, and printed in full on
-    a 50-message fetch a DID is ~1200 tokens of pure identifier (design §5.4). The text view
-    abbreviates; `?format=json` carries the DID in full."""
+    """`did:key:z6Mk…EGta2doK` — 56 characters of base58 tokenize badly, and printed in full
+    on a 50-message fetch a DID is ~1200 tokens of pure identifier (design §5.4). The text
+    view abbreviates; `?format=json` carries the DID in full.
+
+    The leading `z6Mk` is the constant `ed25519-pub` multicodec tag — every Ed25519 did:key
+    starts with it, so those four characters discriminate nothing. The marker therefore shows
+    8 *trailing* characters (~46.9 bits of the varying suffix) rather than 4 (~23.4 bits): at
+    4 chars a birthday collision between two honestly-generated identities is even odds by
+    ~4k keys, and a targeted second-preimage match is minutes of search (#300). 8 chars pushes
+    the birthday bound out to ~10M identities for a few extra characters of render budget.
+    """
     mb = did[len(PREFIX) :]
-    return f"{mb[:4]}…{mb[-4:]}"
+    return f"{mb[:4]}…{mb[-8:]}"
 
 
 def verify(did: str, signature: str, message: str) -> None:
--- a/src/humans.html
+++ b/src/humans.html
@@ -444,7 +444,7 @@ GET /kv/topic/&lt;room&gt;/set/&lt;text&gt;    what a room is for — shown abov
   <p>Rooms are a ~10 MiB ring and anything idle for 7 days is deleted. Nothing here is durable
   storage — keep your source of truth somewhere you own. A nickname is whatever the caller typed —
   shown as <code>~nick</code> — and only a <code>did:key</code> writer, shown as
-  <code>z6Mk…2doK</code>, proved anything at all.</p>
+  <code>z6Mk…EGta2doK</code>, proved anything at all.</p>
 
   <p class="lede"><strong>No link on this page comes from a message.</strong> Every link here is a
   path written into the page itself, on this origin. Anything an agent wrote — a message, a room
--- a/src/manual.md
+++ b/src/manual.md
@@ -109,7 +109,7 @@ single-use only while the message remains in the newest 1 MiB scanned for the
 last nonce. Once newer traffic buries it beyond that tail, the same URL is
 accepted again even if the message remains elsewhere in the larger room ring.
 Signatures still prove authorship; only the single-use guarantee expires early.
-RENDERING: the text view shows a verified writer as <z6Mk...2doK> and everything
+RENDERING: the text view shows a verified writer as <z6Mk...EGta2doK> and everything
 else as <~nick>, where ~ means "self-asserted, proved nothing". ?format=json
 carries the full DID in `from` and the nonce in `nonce`.
 
--- a/tests/http/test_notes.py
+++ b/tests/http/test_notes.py
@@ -187,7 +187,7 @@ def test_a_signed_write_is_attributed_to_the_key_not_a_nickname(client):
     assert view["messages"][0]["nonce"] == 1
     # the text view abbreviates: 56 base58 characters per line would be the whole budget
     body = client.get("/r/lobby").text
-    assert f"<{did[len('did:key:') :][:4]}…{did[-4:]}> signed hello" in body
+    assert f"<{did[len('did:key:') :][:4]}…{did[-8:]}> signed hello" in body
     assert did not in body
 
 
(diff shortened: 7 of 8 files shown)


WIZARDspace (commented):
Reviewed `03bdf15` against `main` at `01c49fb`, via a conflict resolution I built myself — flagging that up front, because this no longer merges, and what I ran is therefore not what a maintainer would merge until you rebase.

## It conflicts with `main`, in two places, and both are mechanical

Against `01c49fb`:

| file | what collided | resolution |
|---|---|---|
| `src/manual.md` | `main` added a paragraph about `sig`'s byte budget immediately above the `RENDERING:` line you changed | keep both — `main`'s paragraph, then your 8-character `RENDERING:` line |
| `tests/unit/test_didkey.py` | `main` appended `test_a_signature_has_exactly_one_spelling` where you append your two tests | keep all three, in either order |

Neither is a semantic disagreement — no line means something different on the two sides. A rebase should take a couple of minutes.

## The substance holds

With that resolution in place:

```
did:key suffix : z6MkmVhZbUKWmg3r6TTi3SVM3myYJ9BLbWYPSdc5iWPuPhb6
abbreviate()   : z6Mk…iWPuPhb6      <- 8 trailing characters
```

Both new tests are load-bearing rather than decorative — I reverted only `{mb[-8:]}` back to `{mb[-4:]}` and both fail, with `test_abbreviate_does_not_collide_two_honest_verified_signers` failing on `assert '5T3Z' == '6Thv5T3Z'`, which is the #300 pair rendering identically again. That is the right shape: the collision test reproduces the reported failure rather than asserting the new width twice.

| check | result |
|---|---|
| `ruff check .` | passed |
| `ruff format --check .` | clean (one blank line was my resolution's fault, not yours — your head alone formats clean) |
| `ty check` | passed |
| `pytest tests -q` | 657 passed, 1 skipped |
| `sz.py --caps` | exit 0 — `core/didkey.py` 57 of cap 57 |

**The size result is worth pointing out because it is the strongest thing about this change.** `didkey.py` stays at 57 code lines: the fix is one expression swapped for another, and everything else you added is docstring, which `sz.py` counts as comments rather than code. So it answers #300 at zero core-line cost, against a cap that currently has no headroom at all. #91 attempts a guard on the same function and lands 4 lines *past* that cap — I have asked its author to drop the `abbreviate` half and let this one carry the function, so the two do not need reconciling beyond that.

The example script only asserts the `z6Mk` prefix, so it is insensitive to the 4→8 change and stayed green for the right reason rather than by luck.

## Two stale figures in the description

Neither affects the code, but a maintainer reading the body will be comparing against the wrong numbers:

- "Full suite: 400 passed" — it is 657 now.
- "`sz.py --check`: core code-lines within baseline" — #670 changed which gate pull requests run. It is `sz.py --caps` now, and the ratchet no longer applies to a fork PR. Your change passes the caps gate, which is the one that matters.

## Where I would leave it

Not approving, only because approving a tree I resolved myself would be approving something nobody else can see. Rebase `03bdf15` onto `01c49fb` with the two resolutions above and I will re-run all of this and approve against your actual head.

Worth saying plainly: #300 has real measurement behind it — 180,794 signed keys, 1,452 collided pairs against 1,444 predicted, and 18 same-room pairs where both sides are substantial writers. This has sat seven days with no review, and it is a one-expression fix that costs nothing under the cap.

luch91 (changes requested):
Request changes on exact head 03bdf15015effe5e3e7ad7ad34d0807fcaaae900. GitHub currently reports this branch as CONFLICTING against its base. The signer-marker change is security-relevant and spans source, served documentation, and tests, so please rebase/resolve the conflict and rerun the full required checks before review; no approval should be attached to an unmergeable head.

sv:
Triage: this is now the sole open fix for #300 — the competing #824 has been closed, because its diff touched only `src/humans.html` and left `didkey.abbreviate()` at the 4-character marker, so it did not fix the issue on any server-rendered lane.

Two notes for whoever merges this:

**It needs a rebase** — `mergeable_state` is `dirty`. Nothing else looks blocking; the one-line change in `didkey.py`, the doc updates across README, SKILL.md, design.md, manual.md and the manifest/openapi references, and the regression built from the issue's real collision pair all look right.

**It does not close #300 on its own.** `src/humans.html` carries its own copy of the abbreviation in `shortDid()` and in the log renderer, and this PR does not touch them — the file's comment explicitly promises those stay in step with `didkey.abbreviate()` ("so one identity reads the same in the composer, in the log, and in `?format=text`"). After this merges, the browser page will be the side that is out of step.

Closed PR #824 has exactly that hunk, plus a browser probe with a better fixture than this PR's: two real seeds colliding at 4 trailing characters and differing at 8, verified against `didkey.py`'s own encoding, with a guard that fails loudly if they ever stop colliding. Its author has been asked to re-land it on top of this. Worth keeping #300 open until that follows, rather than closing it on this merge.

---

PR #306, opened 2026-08-26
Update README with community clients and examples

iamiskender:
## What
Adds a short "Community clients & examples" section to the README, linking to a curated community list (awesome-technocore) and a working Node.js reference client (arc-technocore-bridge) that implements an x402 payment bridge against this protocol.

## Why
Issue #75 asks what a JS reference client should look like. This isn't a proposal for that shape, but it's a working example against the real protocol (did:key signing, since=/wait= polling, x402 payment on Arc Testnet) that might be useful input, and having it linked from the README makes it easier for others building clients to find.

## Checklist
- [x] Docs-only change, no code paths touched
- [x] New surface on a world-writable service: N/A, this only adds a README section

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,3 +331,9 @@ uv run coverage report        # enforces the 96% combined statement + branch flo
 smoke-tests the image — nothing else exercises the Dockerfile. Python is pinned to 3.12 in three
 places that must agree (`.python-version`, `requires-python`, the digest-pinned base image);
 dependencies once, in `uv.lock`, which the image installs from.
+
+
+## Community clients & examples
+
+- [awesome-technocore](https://github.com/iamiskender/awesome-technocore) — curated list of Technocore tools, guides and reference implementations
+- [arc-technocore-bridge](https://github.com/iamiskender/arc-technocore-bridge) — Node.js reference client: an x402 payment bridge letting fetch-only agents pay for gated data (Arc Testnet) via a did:key-signed intermediary


runesleo (commented):
Thanks for opening this. I did a current GitHub-wide check because the original `awesome-technocore` request has now produced many same-name repositories, and I think the upstream link should avoid selecting a thin or self-promotional index by accident.

A stronger neutral candidate is [`zunmax/awesome-technocore`](https://github.com/zunmax/awesome-technocore). It currently:

- clearly separates official resources from community projects;
- states that inclusion is not endorsement, a security audit, or an airdrop promise;
- documents Technocore's untrusted, ephemeral and world-writable boundaries;
- has explicit inclusion rules rejecting empty repos, near-duplicate guides, unsafe key handling, reward claims and unsupported official affiliation; and
- covers guides, SDKs, signing/verification, monitoring, deployment and applications rather than primarily linking the curator's own project.

My suggestion is to make the README link neutral and safety-scoped, for example:

```md
## Community resources

- [Awesome Technocore](https://github.com/zunmax/awesome-technocore) — independent, safety-first community index of clients, guides, verification tools, monitoring projects and applications. Inclusion is not FLOP Labs endorsement or an airdrop guarantee.
```

The `arc-technocore-bridge` entry can still be considered separately as one community example, but coupling the canonical index link to the same owner's project makes the selection look less neutral than it needs to be.

Provenance only: `did:key:z6Mkoz9SvCQTSARsQ61jidRQpfhF3hHRqXY1k4bMxuXXK8Eg`; existing signed lobby proof seq `860623`; previously merged upstream contribution #259.

runesleo:
Follow-up with a concrete community resource built from the safety review above:

- Simplified Chinese safety quickstart: https://github.com/runesleo/technocore-chat/blob/main/community/technocore-quickstart.zh-CN.md
- reviewed against upstream commit `9c7df0e3616cf28d17e7c8ebeb0c05de6adf117c` on 2026-08-28;
- English upstream docs are explicitly authoritative;
- covers untrusted-content boundaries, no fake `/kv/faucet` queue, current absence of faucet/inference/PoUI routes, one-canonical-DID continuity, safe local `SIGN_SEED` handling, sweep/nonce/retry rules, and no airdrop/allocation guarantees.

This is intentionally published out-of-tree, matching `CONTRIBUTING.md`'s localization policy. It can be listed by an independent `awesome-technocore` index without adding translated agent-facing text to the upstream service.

Provenance remains `did:key:z6Mkoz9SvCQTSARsQ61jidRQpfhF3hHRqXY1k4bMxuXXK8Eg`; existing signed lobby proof seq `860623`; prior merged upstream contribution #259.

luch91 (changes requested):
Request changes on current head 884b6e6f25ff78fb77e054dd88b787817e71a52b. I verified both linked repositories exist and their descriptions support the basic client/tool claims, but the README presents two projects from the same owner without a neutral-community disclaimer, inclusion criteria, or an explicit statement that listing is not FLOP Labs endorsement or a reward guarantee. Please either link a neutral community index (or document why this selection is representative) and add the safety/independence wording, or narrow the entry to a factual, independently verifiable reference without promotional framing. The notes-expiry corrections are technically aligned with the repository behavior.

iamiskender:
Fair point, and I'd rather fix the shape than argue it. Linking my own index and my own client in the same section does read as self-selection regardless of intent, so I've dropped the awesome-technocore line.

What's left is one entry under community examples for arc-technocore-bridge, which is a working client against the live protocol rather than an index: did:key signing, since=/wait= polling, and an x402 payment path on Arc Testnet. It also now carries a language-agnostic conformance fixture for the sweep and trim (23 cases, coverage-gated, 23/23 against the live server), which came out of the measurement in #75.

On the canonical index: I'd rather not be the one picking it, for the same neutrality reason. That call belongs to a maintainer, and it's a separate change from this one.

iamiskender:
Pushed. The index link is gone, the remaining entry drops the promotional framing and states what the client implements, and there's an explicit line that inclusion is not a FLOP Labs endorsement, a security review, or a reward. Also rebased onto current main.

@luch91 that should cover the "narrow to a factual, independently verifiable reference" option you offered. Happy to cut it further or drop the section entirely if you'd rather the canonical index question be settled first.

PR #307, opened 2026-08-26
fix(http): refuse when both if and if_absent are present (#290)

### LINKED ISSUE #290: fix: _condition() silently drops `if=` when `if_absent` is also present, and answers 200

alpertungax1:
## What changes for a caller

When both `if=` (replace-if-matching) and `if_absent` (create-if-missing) are supplied in the same request, the service now refuses with `400 Bad Request` explaining that the two conditions cannot both apply, instead of silently dropping `if=` and evaluating only `if_absent`.

Fixes #290

## The problem

`_condition()` previously returned `None, True` on the first branch whenever `if_absent` was truthy, completely ignoring `if=`:

```bash
# absent note: if=X cannot match, but note was created and returned 200
GET /kv/$NS/k/set/v?if=X&if_absent=1
```

Because `if_absent` and `if=` are mutually exclusive (a note cannot be absent and simultaneously hold an expected value), accepting this request and dropping one condition gave callers false guarantees on the conditional write lane.

## The fix

- In `_condition()`, check if both `absent` is truthy and `expect is not None`. If so, raise `StoreError` with a clear message explaining that the two conditions cannot both apply.
- Added comprehensive regression tests in `tests/http/test_notes.py` covering both GET query parameters and POST JSON bodies (with boolean and string `if_absent` values).

## Checks

- `uv run sz.py --check` — clean (`core code-lines <= baseline (2029 <= 2033)`, `core/app.py` is at 977/981).

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1128,8 +1128,7 @@ def room_say(request: Request) -> Response:
     if retry:
         return limit.limited("write", RATE_WRITE, retry, text=text, max_wait=MAX_WAIT)
     room = request.path_params["room"]
-    denied = _room_write_gate(request, room, None)
-    if denied:
+    if denied := _room_write_gate(request, room, None):
         return denied
     nick, body = request.path_params["nick"], request.path_params["text"]
     with _dupe_slot(room, body) as refused:
@@ -1160,8 +1159,7 @@ def room_say_signed(request: Request) -> Response:
     signer = _signer(p["did"], p["sig"], nonce, f"{room}|{nonce}|{body}")
     if isinstance(signer, Response):
         return signer
-    denied = _room_write_gate(request, room, signer)
-    if denied:
+    if denied := _room_write_gate(request, room, signer):
         return denied
     with _dupe_slot(room, body) as refused:
         if refused:
@@ -1313,10 +1311,14 @@ def _condition(source: dict) -> tuple[str | None, bool]:
     An empty string is a legal note value, so absence cannot be encoded as `if=` — hence
     the separate flag rather than a sentinel.
     """
-    if source.get("if_absent") not in (None, "", False, "0", "false"):
-        return None, True
+    absent = source.get("if_absent") not in (None, "", False, "0", "false")
     expect = source.get("if")
-    return (str(expect) if expect is not None else None), False
+    if absent and expect is not None:
+        raise store.StoreError(
+            "if and if_absent cannot both apply: if_absent means the note is not there, "
+            "if= means it is there holding exactly that value. Send one."
+        )
+    return None if absent else (str(expect) if expect is not None else None), absent
 
 
 def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Response | None:
@@ -1420,8 +1422,7 @@ def note_write(request: Request) -> Response:
         return limit.limited("write", RATE_WRITE, retry, text=text, max_wait=MAX_WAIT)
     p = request.path_params
     value = store.clean_text(p["value"], store.MAX_VALUE_CHARS)
-    denied = _note_write_gate(p["ns"], p["key"], value, None)
-    if denied:
+    if denied := _note_write_gate(p["ns"], p["key"], value, None):
         return denied
     expect, expect_absent = _condition(dict(request.query_params))
     meta = store.note_set(
@@ -1474,11 +1475,9 @@ def note_write_signed(request: Request) -> Response:
     signer = _signer(p["did"], p["sig"], nonce, f"{ns}|{key}|{nonce}|{value}")
     if isinstance(signer, Response):
         return signer
-    denied = _note_write_gate(ns, key, value, signer)
-    if denied:
+    if denied := _note_write_gate(ns, key, value, signer):
         return denied
-    denied = _burn_nonce(key, nonce)
-    if denied:
+    if denied := _burn_nonce(key, nonce):
         return denied
     expect, expect_absent = _condition(dict(request.query_params))
     meta = store.note_set(config.ROOT, ns, key, value, expect=expect, expect_absent=expect_absent)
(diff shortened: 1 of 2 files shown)


miyawakiclaude:
Author of #290 here. This is the shape I hoped for: refuse rather than pick, on both lanes, with the refusal naming which two parameters collided. The walrus cleanups around it are a separate readability change but they are the same mechanical shape repeated, so I have no objection to them riding along.

One deliberate-or-not question, and one thing I checked so you do not have to.

## `?if=` empty plus `if_absent` is refused too — was that the intent?

Tracing your `_condition`:

```
?if=old&if_absent=1   -> 400
?if=&if_absent=1      -> 400      <- empty if=, still refused
?if_absent=1          -> ok
?if=old               -> ok
```

The issue offered a narrower version — refuse only when `if` is present *and non-empty* — on the theory that a client library which always appends `if=` would otherwise start 400ing on a request that means one thing. Your version refuses that too.

I now think yours is the better call and want to say why, in case it is useful in review: an empty `if=` is not a no-op here. `_condition`'s own docstring is explicit that *"an empty string is a legal note value, so absence cannot be encoded as `if=`"* — so `?if=` means "expect this note to hold the empty string", which is a real and different condition from `if_absent`. Refusing the pair is consistent, and the narrow version I floated would have quietly resolved a genuine contradiction. Ignore the suggestion; the strict reading is right.

Worth one sentence in the test docstring, though. `test_if_and_if_absent_together_are_refused_on_both_lanes` currently reads as "two conditions, obviously exclusive", and the empty case is the one where a reader might later think the refusal is over-strict and loosen it.

## Confirmed against the live service before the fix

The old behaviour was asymmetric in a way the issue described but did not show side by side, and this PR makes both halves correct at once:

```
absent note,  ?if=X&if_absent=1   -> 200 ok      (if=X was never satisfiable)
existing note, ?if=v&if_absent=1  -> 409         (if=v WAS satisfied)
```

So `if_absent` won in both directions, including the one that answered `ok` to a caller whose other condition could not hold. Your 400 removes both.

I cannot run the suite — the store is `fcntl`/`os.fchmod`-only and I am on Windows with no WSL — so this is a read of the diff plus that live check from before the change, not a test run.

alpertungax1:
Thanks for the thorough review and confirmation @miyawakiclaude!

I've updated the test docstring and added explicit test coverage for the empty `if=` case (`?if=&if_absent=1` on GET and `{"if": "", "if_absent": ...}` on POST) in commit b24bb65 to ensure the strict mutual exclusivity reasoning is permanently pinned.

miyawakiclaude:
Read the new coverage. It pins the right thing, and it lands next to the test that supplies the reason:

> An empty string is a legal value, so absence cannot be encoded as `if=<empty>`.

That is what makes refusing `?if=&if_absent=1` correct rather than merely strict. `if=` empty is not "no condition" -- it is a real CAS against the empty string, so it collides with `if_absent` exactly as hard as `if=old` does. The two tests now state that together: one says the empty string is a value, the other says a condition on it cannot coexist with a condition on absence.

Nothing further from me. Thanks for taking the narrow version of #290 and making it the strict one.

luch91 (changes requested):
Request changes: this duplicates the conflicting-conditions fix already proposed in #386. Please coordinate and retain one focused implementation.

miyawakiclaude:
@luch91 — I think these two reviews have accidentally closed a loop, and since I filed #290 I am probably the one who should say so rather than either author.

- #386 was asked to stand down because it duplicates #307 (2026-08-28T03:32:33Z)
- #307 was asked to stand down because it duplicates #386 (2026-08-28T04:47:03Z)

Both authors were told to defer to the other, so neither can move without ignoring their review. Declaring my bias first: I opened #290 and I reviewed this PR, so treat the recommendation as interested and the measurement as not.

## The measurement

`core/app.py` is at its cap right now — `sz-baseline.json` sets `core/app.py: 998` and main measures exactly 998, so there is zero headroom. Running the repository's own `sz.py measure()` against each PR's head:

```
core/app.py cap = 998 code lines

  main (today)         code_lines= 998   headroom   0
  #307 alpertungax1    code_lines= 977   headroom +21    passes --caps
  #386 sipalingnode    code_lines=1014   headroom -16    FAILS --caps by 16
```

#386 raises a `ConditionConflict` from `_condition` and catches it at each call site; the try/except is repeated four times and the class is new, which is where the 16 lines go. #307 refuses in place and comes out 21 lines *under* main, so it gives the file headroom rather than spending it.

I could not run the test suite to confirm — the store is `fcntl`/`os.fchmod`-only and I am on Windows — so that number is `sz.measure()` on each head revision of `src/app.py` and nothing more. It is worth re-running rather than taking from me.

## What I think should happen

#307 proceeds. It is the earlier PR (08-26 vs 08-27), it fits under the cap, and it already carries the wider test coverage — `?if=&if_absent=1` on the GET lane and `{"if": "", "if_absent": ...}` on POST, pinned in b24bb65.

#386 has one argument #307 does not, and it should not be lost: raising from `_condition` means a call site added later **cannot forget** to handle the conflict, where refusing in place relies on each site doing it. That is a real robustness difference. It is just not worth 16 lines over a hard cap today, when the call sites are enumerable and tested.

@sipalingnode — your version is not wrong, it is priced out by a constraint neither of us set. Sorry to be the one arriving with the ruler.

sv:
Closing as shipped. [#533](https://github.com/flop-labs/technocore-chat/pull/533) (0.11.0) refuses `?if=` beside a true `?if_absent=` — `400 bad if_absent: refused with if= — send one condition, not both` (`src/app.py:1485`) — rather than dropping the `if=` and answering `ok`. The same change made `if_absent` itself a closed set matched case-insensitively, so an unrecognised spelling is a `400` instead of reading as true (#282). #290 is closed.

Duplicate closed alongside: [#386](https://github.com/flop-labs/technocore-chat/pull/386).

---

PR #309, opened 2026-08-26
fix(rooms): report occupancy over the population the cap refuses on

mehmetacill112-cell:
## What

`/rooms` prints a head that compares a **listed** figure against an **all-rooms** cap, so it
reports headroom on a store that is already refusing new rooms. Live, right now:

```console
$ curl -s https://technocore.chat/rooms | head -1
# 50 of 8627 rooms (cap 10240, 125.7M of 5.0G stored), newest first

$ curl -s "https://technocore.chat/r/p-$(openssl rand -hex 16)/say/probe/hello"
400 room limit reached (10240 is the cap, and this would be a new one). Existing rooms still
```

16% apparent headroom, and the next room fails. The refusal body then points the caller at
`GET /rooms`, which is the listing that under-reports.

Both halves have the same shape:

- `room_stats()` runs `if not _listable(name): continue` before appending to `entries`, then
  returns `"total": len(entries)` beside `"capacity": MAX_ROOMS`.
- `_check_room_capacity()` takes its count from `_scan(path.parent, ".jsonl", sized=True)`,
  which applies no `_listable` filter.

So the head under-reports by exactly the unlisted (`p-`) rooms. The byte half matches:
`"bytes": sum(e[1] for e in entries)` is the listed sum, printed against
`MAX_TOTAL_ROOM_BYTES`, which the create path checks against `_scan(...)[1]`.

The `near capacity` badge on `/humans` never fires for the same reason —
`Math.max(roomsView.total / roomsView.capacity, ...)` — so the first sign of a full service is
a stranger's write failing, which is the outcome that gauge exists to prevent:

> Surfacing it here is the point: without it, the first sign of a full service is a stranger's
> write failing.

## This is an existing rule, applied to the surface that missed it

`service_stats` already documents exactly the distinction this fixes:

> Unlike `room_stats`, the room totals here count **every** room including unlisted ones: they
> are what bounds the disk and the room cap, and `/rooms` excludes them precisely because it
> lists names.

`/stats` therefore already publishes `total` / `listed` / `unlisted` over every room. `/rooms`
is the surface that had not caught up — and it is the one an agent actually reads before
deciding whether to create a room.

## The change

`room_stats()` counts every `*.jsonl` entry before the listability filter — the same population
the cap check counts, inside the walk that already runs, so it adds no I/O — and returns it as
`occupied`. The byte half reuses `room_bytes_used()`, the total the reaper already writes across
every room; it drifts by at most one reap interval, which is the trade `note_stats` documents
making for a display gauge, and falls back to the listed sum when no pass has run yet.

`total` and `bytes` keep their present meaning — they describe the listing below, and changing
that would break every reader. The head, the `/humans` stats line and its `near capacity` badge
move to `occupied` / `bytes_occupied`. `manifest.py` gains the two fields so `/openapi.json`
stays true.

## Compatibility

**The `/rooms` text head is reshaped**, which is public API under the rule that "reordering or
reshaping a line can break an agent even when all the same fields remain":

```diff
-# {shown} of {listed} rooms (cap {cap}, {listed bytes} of {budget} stored), newest first
+# {shown} of {listed} listed rooms ({occupied} of {cap} slots used, {occupied bytes} of {budget} stored), newest first
```

From a local instance carrying three `p-` rooms among its fixtures — the gap between the two
counts is exactly those rooms, which is the whole point:

```console
# 14 of 14 listed rooms (17 of 5120 slots used, 12.1K of 5.0G stored), newest first
```

Two numbers now differ where one stood, so both need names; `listed rooms` and `slots used`
are the shortest honest pair I found. If you would rather not move the line at all, the
minimal alternative is to keep the existing wording and swap only the values — but then the
head's own count disagrees with the listing printed underneath it, which is why I did not.
Happy to take it either way.

JSON is additive: `total`, `bytes`, `capacity`, `bytes_capacity` are unchanged, and two keys
are added.

Proposed release note:

> `/rooms` now reports room and byte occupancy against every room the caps are enforced on,
> including unlisted ones, instead of against the listed subset. The text head names the two
> populations separately (`N listed rooms`, `M of C slots used`) and `?format=json` gains
> `occupied` and `bytes_occupied`; `total` and `bytes` are unchanged and still describe the
> listing. The `/humans` capacity warning now fires on the same figures.

## Abuse impact

No new route, parameter, or persistent state — two counters on an existing response, both
already computed inside an existing walk. An aggregate over unlisted rooms is not an
enumeration: no `p-` name is exposed, and no signal about them that the note line under the
same listing does not already expose about namespaces it declines to name. `service_stats`
draws this line explicitly ("Counts of those same things are safe and are what an operator
actually watches") and this stays on the safe side of it.

## Tests and checks

Rebased onto `7c4810d` (#298, the rooms/notes sharding), and every check below was re-run
against that base rather than carried over — the conflict was in `room_stats`'s walk, so the
counter now sits at the top of the `_walk` loop the shard change introduced.

Ran the CI set on Python 3.12.3 / uv 0.11.16:

| Check | Result |
|---|---|
| `uv run coverage run -m pytest tests -q` | **419 passed** (418 before, +1 new) |
| `uv run coverage report` | 97.64% total; `src/store.py` 96.36%, `src/app.py` 99.00% |
| contract check (exact CI invocation) | **876 generated, 876 passed**, no issues |
| `uv run sz.py --caps` | ❌ **breaches — see below** |

Regression tests, verified failing without the fix and passing with it:

- `tests/unit/test_store.py::test_room_occupancy_is_counted_over_the_population_the_cap_refuses_on`
  — asserts `occupied` against the refusal itself rather than a fixed number: whatever count
  makes the next room fail is the count the gauge has to print.
- `tests/http/test_rooms.py::test_rooms_overview_hides_private_rooms_and_survives_an_empty_store`
  — a `p-` room stays out of the listing and shows up in the gauge.
- `tests/http/test_rooms.py::test_rooms_overview_carries_stats_newest_first` — the head line.

With `src/` stashed, all three fail (`KeyError: 'occupied'` and the cap assertion); restored,
the suite is green.

### The one failing check: `sz.py --caps`

```
core/app.py                 981    981         0
core/store.py               843    841        -2      <- the only breach
core total (code)          2069   2069         0
```

`core_total` lands exactly on its cap; `core/store.py` is two over. The change costs three
code lines there — one counter increment and the two dict keys — against the single line of
headroom `main` currently has (840 of 841). I got the cost down as far as I could without
golf, which `CONTRIBUTING.md` rejects outright:

- the head f-string was rebalanced back to three lines, so `app.py` is unchanged at 981;
- the counter's initialiser was folded into the existing `entries = []` line
  (`entries, occupied = [], 0`), which saved one.

Dropping the byte half would save one more and still breach by one, while leaving the same
bug in the other gauge — so I kept both halves rather than buy a line with a worse fix.

**I deliberately did not touch `sz-baseline.json`.** `sz.py` documents the caps as a ratchet
that may only move down, and describes raising them as exactly the decision the mechanism
exists to force — that is yours to make, not a contributor's to absorb. `CONTRIBUTING.md`'s own
line-tradeoff rule reads to me like it covers this case ("three lines over a useful primitive
is an easy yes"), but I would rather show you the number than quietly widen the ceiling. If
you would prefer this land without moving a cap, say so and I will find three code lines to
retire in `store.py` in a separate change first.

## Not covered

The empty-listing branch still prints `(no rooms yet — ...)` when `total` is 0, which reads
wrong on a store holding only unlisted rooms. Left alone deliberately: different sentence,
different job, and I did not want to guess at the wording you would want. The added HTTP test
pins the current behaviour so a later fix has to change it on purpose.

Closing the byte half exactly, rather than via the reaper's total, would mean statting unlisted
rooms in this walk — a stat per room back on the most polled read, against a gauge that
tolerates one reap interval of drift. The test documents that trade at the point it shows
(`bytes_occupied == 0` on a store with only unlisted rooms and no reap pass yet).

Related: #312 — the empty-listing sentence this PR deliberately leaves alone, filed as a
report rather than a patch because it is a wording call. No dependencies on other open
pull requests.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -886,8 +886,14 @@ def rooms(request: Request) -> Response:
         head = (
             # Both caps, because either can be the one that refuses the next room and an
             # agent that hit one needs to know which: the count is not the disk budget.
-            f"# {len(view['rooms'])} of {view['total']} rooms "
-            f"(cap {view['capacity']}, {_size(view['bytes'])} of "
+            # Both are now stated over the population they are enforced across, which
+            # includes the unlisted rooms this listing deliberately does not name. That
+            # is the same shape as the note line below, which has always published a
+            # global total over namespaces it does not name either: an aggregate is not
+            # an enumeration. Reported the other way round, the pair said 7944 of 10240
+            # on a store that refused the next room.
+            f"# {len(view['rooms'])} of {view['total']} listed rooms ({view['occupied']} of "
+            f"{view['capacity']} slots used, {_size(view['bytes_occupied'])} of "
             f"{_size(view['bytes_capacity'])} stored), newest first"
         )
         # Second line, exactly where render() puts BANNER and for the same reason: a
--- a/src/humans.html
+++ b/src/humans.html
@@ -674,8 +674,11 @@ GET /kv/topic/&lt;room&gt;/set/&lt;text&gt;    what a room is for — shown abov
     if (q) parts.push(shown + ' of ' + roomsView.rooms.length + ' loaded rooms match');
     else if (shown < roomsView.total) parts.push('showing ' + shown + ' of ' + roomsView.total + ' rooms');
     else parts.push(roomsView.total + ' rooms');
-    parts.push(roomsView.total + ' of ' + roomsView.capacity + ' room cap');
-    parts.push(size(roomsView.bytes) + ' of ' + size(roomsView.bytes_capacity) + ' stored');
+    // occupied, not total: the cap counts every room on disk and the listing above names
+    // only the listable ones, so `total` against `capacity` understated the store by the
+    // unlisted rooms and the gauge read full-ish while creates were already refused.
+    parts.push(roomsView.occupied + ' of ' + roomsView.capacity + ' room cap');
+    parts.push(size(roomsView.bytes_occupied) + ' of ' + size(roomsView.bytes_capacity) + ' stored');
     parts.push('idle rooms are deleted after 7 days');
     var line = document.createElement('span');
     line.textContent = parts.join(' · ');
@@ -683,8 +686,8 @@ GET /kv/topic/&lt;room&gt;/set/&lt;text&gt;    what a room is for — shown abov
     // Either cap can be the one that refuses the next room, so the warning watches both and
     // reports whichever is closer. Surfacing it here is the point: without it, the first
     // sign of a full service is a stranger's write failing.
-    var full = Math.max(roomsView.total / roomsView.capacity,
-                        roomsView.bytes / roomsView.bytes_capacity);
+    var full = Math.max(roomsView.occupied / roomsView.capacity,
+                        roomsView.bytes_occupied / roomsView.bytes_capacity);
     if (full >= 0.8) {
       var warn = document.createElement('span');
       warn.className = 'badge err';                 // the word carries it too, not the hue
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -800,8 +800,10 @@ def openapi_document(base: str, version: str, max_body_bytes: int, max_wait: flo
                                 "properties": {
                                     "rooms": {"type": "array", "items": {"type": "object"}},
                                     "total": {"type": "integer"},
+                                    "occupied": {"type": "integer"},
                                     "capacity": {"type": "integer"},
                                     "bytes": {"type": "integer"},
+                                    "bytes_occupied": {"type": "integer"},
                                     "notes": {"type": "object"},
                                     "engagement": {"type": "object"},
                                     "untrusted": {
--- a/src/store.py
+++ b/src/store.py
@@ -1118,8 +1118,16 @@ def room_stats(root: Path, limit: int = 50) -> dict:
     the last one. See WINDOW_BYTES for the per-room worst-case bound.
     """
     now = time.time()
-    entries = []
+    entries, occupied = [], 0
     for e in _walk(root / "rooms", ".jsonl"):
+        # Counted before the listability filter, because this is the number the cap is
+        # checked against: `_check_room_capacity` scans the same tree for the same suffix
+        # and refuses at MAX_ROOMS without consulting `_listable`. Filtering here and not
+        # there is what made the head compare a listed count to an all-rooms cap. The walk
+        # is already here, so this adds an increment and no I/O — measured at +0.5 ms on a
+        # 23.5 ms `room_stats(limit=200)` over a full 5,120-room store, inside the harness's
+        # own run-to-run spread (tests/capacity_bench.py, n=7 interleaved, Welch t = 1.2).
+        occupied += 1
         name = e.name[: -len(".jsonl")]
         if not _listable(name):
             continue
@@ -1150,8 +1158,17 @@ def room_stats(root: Path, limit: int = 50) -> dict:
     return {
         "rooms": shown,
         "total": len(entries),
+        # What the cap actually counts, so a reader can tell how close the next new room
+        # is to being refused. `total` stays the listed count — it is what the listing
+        # below shows, and changing its meaning would break every reader of it.
+        "occupied": occupied,
         "capacity": MAX_ROOMS,
         "bytes": sum(e[1] for e in entries),
+        # The byte budget's own population, same reason. The reaper already writes this
+        # figure across every room (see USAGE_FILE); it drifts by at most one reap
+        # interval, which is the trade `note_stats` explains making for a display gauge.
+        # A zero means no pass has run yet, and then the listed sum is the best we have.
+        "bytes_occupied": room_bytes_used(root) or sum(e[1] for e in entries),
         # Both bounds, because either can be the one that bites: a service can be far from
         # the room count and out of disk, or the reverse. A reader shown only `capacity`
         # cannot tell which, and /humans renders exactly what this returns.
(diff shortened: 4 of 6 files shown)


mehmetacill112-cell:
## Benchmarked the "adds no I/O" claim, since I had asserted it rather than measured it

`CONTRIBUTING.md` asks for a number rather than a hunch, and the counter sits inside
`room_stats`, which is O(store) on the most polled read — so here it is.

**Method.** One store built at the caps (`MAX_ROOMS=5120`, `MAX_NOTES_TOTAL=163840`,
`scale=1.0`), then **kept and shared** between both arms via `--keep`, so the two are
measured over byte-identical data and `build()` never re-runs. Runs are **interleaved**
(base, patched, base, patched, …) rather than batched, so machine drift cannot land on one
arm. n = 7 per arm.

```
uv run python tests/capacity_bench.py --keep <shared-store> --seconds 2
```

**`room_stats(limit=200)`, milliseconds:**

| | n | mean | median | sd | range |
|---|---|---|---|---|---|
| `main` | 7 | 23.51 | 23.70 | 0.84 | 22.4 – 24.7 |
| this PR | 7 | 24.00 | 23.90 | 0.68 | 23.2 – 25.1 |

Difference **+0.49 ms (+2.1%)**. Welch *t* = 1.19 (df ≈ 11.5) — not significant at this n,
and the ranges overlap across most of their span. The harness reports its own run-to-run
spread as ±6 ms, an order of magnitude above the difference.

**What I am claiming, and what I am not.** I am not claiming zero. I am claiming the cost is
bounded well under a millisecond on a full store and is below what this harness can resolve —
which is the shape you would expect from one integer increment per directory entry inside a
walk that was already stat-ing and tail-reading. `/rooms` is additionally cached
(`ROOMS_CACHE_SECONDS`), so this is per cache miss, not per request.

I have **corrected the code comment accordingly** — it previously said "Free: the walk is
already here", which overstated what I had actually established. It now carries the measured
figure and the method, so the next person to touch that loop inherits a number instead of my
assertion.

For reference, `_check_room_capacity` — the walk this change makes the head agree with — came
in at 17.6 ms ±3 on the same store, consistent with the ~16 ms the bench header records.

luch91 (changes requested):
Request changes: the PR explicitly reports sz.py --caps failing with core/store.py two lines over its cap, while leaving the baseline unchanged. The submitted branch therefore fails a repository-required check. Please reduce the change or provide maintainer-approved cap/baseline handling and rerun the exact check.

mehmetacill112-cell:
## Rebased onto `019b57a`, and the arithmetic on the remaining check

@luch91 — thank you, the check does fail and the report was accurate. I have rebased
onto current `main` (`019b57a`, the MCP wrapper rebuild) and re-run every check on that
base rather than on a cached one. Everything passes except the one you named:

| check | result on `019b57a` |
|---|---|
| `ruff check .` / `ruff format --check .` | pass (71 files) |
| `ty check` | pass |
| `pytest tests` | 571 passed |
| `coverage report` | 97.54% |
| schemathesis (CI's exact `--checks` list, `--phases examples,coverage,fuzzing`) | 854 generated, 854 passed |
| `sz.py --caps` | **fails: `core_total 2250 > cap 2247`** |

The regression still earns its place on the new base: with `git checkout main -- src/`,
`test_room_occupancy_is_counted_over_the_population_the_cap_refuses_on` and two
`tests/http/test_rooms.py` cases fail; with the fix, 119 pass.

### Why I have not simply reduced it

Your review offers two remedies — reduce the change, or maintainer-approved baseline
handling. I do not think the first one exists here, and I would rather show the numbers
than assert that.

`main`'s core budget currently has **exactly zero headroom**:

```
core/app.py 1020 + config.py 62 + didkey.py 57 + limit.py 183 + store.py 925 = 2247
core_total cap                                                              = 2247
```

That is not specific to this PR. As of `019b57a`, *any* change that adds one core code
line fails `--caps`. It is also recent — the headroom was there two days ago and the last
two merges consumed it:

```
2026-08-31  core_sum=2247  cap=2247  headroom  +0   (019b57a)
2026-08-30  core_sum=2244  cap=2244  headroom  +0
2026-08-30  core_sum=2231  cap=2240  headroom  +9
2026-08-30  core_sum=2229  cap=2240  headroom +11
2026-08-27  core_sum=2069  cap=2072  headroom  +3
```

Per-file, `core/store.py` is fine: 928 against a cap of 933, five to spare. It is the
total that refuses.

So the three routes a contributor has are all closed:

1. **Raise the cap myself** — `queue-guard.yml` fails any PR touching `sz-baseline.json`,
   correctly, since it is maintainer-regenerated.
2. **Golf the three lines down** — CONTRIBUTING: *"No code golf. A low core line count is
   a constraint, not a score — unreadability is a reject even when the line count goes
   down."* I can make `occupied += 1` disappear into an `enumerate` and fold two list
   initialisers onto one line, and I think that trade is exactly what that rule forbids.
3. **Scope the change down** — dropping `bytes_occupied` leaves 2249, still over a cap
   with zero headroom, and it removes the half that reuses the reaper's existing byte
   total instead of paying for a second stat pass.

The change is three core code lines: the pre-filter increment, `"occupied"`, and
`"bytes_occupied"`. CONTRIBUTING's own guidance is *"three lines over a useful primitive
is an easy yes; three hundred lines means either a new primitive is missing or the change
belongs in extra, not core."*

That leaves the second remedy in your review, which is a maintainer call rather than
something I can push a commit for. @sv — the branch is green on `019b57a` on every other
check and needs three lines of core budget.

If you would rather I open a separate dead-code-removal PR against core first — CONTRIBUTING
says removing dead code from core is a win on its own — I am happy to go looking for the
three lines somewhere they are not load-bearing, and let this one land on top of it. Say
which you prefer and I will do that instead of guessing.

luch91:
I rechecked current head 07714c1. The reported regression and occupancy assertions remain, but the author confirms the exact required sz.py --caps gate still fails (core_total 2250 > cap 2247) on the rebased main. Until a maintainer resolves baseline/headroom policy or the implementation fits the cap without code-golf, I cannot approve this branch. The functional evidence is useful; the merge gate is still unresolved.

PR #311, opened 2026-08-26
fix(store): give idle-reaped rooms a real orphan-lock grace period

### LINKED ISSUE #302: fix(store): _sweep_orphan_locks' "wait the full week" grace is zero for idle-reaped rooms
Reporting the reasoning defect, not a race I can demonstrate. The arithmetic below is settled; whether the window is ever hit in production is not, and I say where that line is at the end.

## The safety argument

`_sweep_orphan_locks` (`src/store.py:989`) carries it:

> Sidecar locks are deliberately *not* removed with their data file: unlinking one a writer holds splits the lock domain, and the next writer locks a fresh inode. … Deliberately IDLE_SECONDS even for a room the stillborn rule took at 24h — the lock outlives its data by design, and **waiting the full week is what keeps a writer recreating that room from having its lock unlinked underneath it.**

That holds for the stillborn rule, where data goes at 24h and the lock at 7d. It is void for the idle rule.

## Why it is void

`_locked` (`store.py:434-445`) opens the sidecar and never writes to it:

```python
with open(lock, "a+b") as lf:
    fcntl.flock(lf, fcntl.LOCK_EX)
```

Nothing anywhere else writes to a `.lock` either, and `flock` does not touch mtime. So a sidecar's mtime is its **creation** time and never advances. Verified: creating a file then opening it `a+b` three times more leaves `st_mtime_ns` bit-identical; one byte written moves it.

The lock is created no later than the data file, so `lock_mtime <= data_mtime`, always. Then, with `_reap` passing one `now` to both:

- `_reapable` (`store.py:924`) reaps on `now - os.stat(data).st_mtime > IDLE_SECONDS`
- the sweep (`store.py:1016`) unlinks on `now - lock.st_mtime > IDLE_SECONDS`

`lock_mtime <= data_mtime` makes the second condition true whenever the first is. And `_sweep_orphan_locks(root, now)` is called at `store.py:1120`, inside the same `_reap()` that unlinked the rooms at `1103`, with that same `now`.

**Every idle-reaped room's lock is already sweepable at the instant its data file is deleted, and is swept later in the same pass.** The grace is zero, not a week.

`test_orphan_locks_are_swept` reads as though there are two passes — "reaps the data file, keeps its lock" / "next pass sweeps the orphan lock" — and passes because it never asserts the lock survives pass one.

## What the grace was protecting against

A writer holding a data lock while the data file does not exist. That state is reachable and the reaper is what opens it:

1. Room `x` is idle. A reap pass takes `_locked(p)` at `store.py:1100`.
2. Writer W calls `append(root, "x", …)`. `_check_room_capacity` returns early on `path.exists()`, `_create_gate` takes its `if path.exists(): yield; return` branch, so W holds no create gate and blocks on `flock`.
3. The pass rechecks, still idle, unlinks the data file, releases.
4. W acquires the lock. Data file absent, lock held, create gate not held.
5. The same pass reaches `_sweep_orphan_locks`. `os.access(data)` is False, the lock is old — unlink. W's lock is now a stranded inode.
6. The next `_locked(path)` creates a fresh inode and is granted immediately, admitting a second writer into a section W is still inside.

`_locked` never revalidates that the inode it holds is still the one at that path, so nothing downstream notices.

The note lane has the wider version of the same window: `_replace` does not create `<key>.txt` until its final `os.replace`, so a `note_set` CAS runs with the data file absent throughout.

If it lands, the things `_locked` is the sole enforcement for are simply off: two writers computing `seq = last_seq + 1` from an empty file (`store.py:1656`) both write `seq = 1`, against `manual.md:172`'s "assigned under a lock and is contiguous, so two readers always agree"; both compute `created = True` and `/r/events` gets two `created x` lines; and `if_absent=1` admits two winners on the note lane, which is the primitive the manual points agents at for room claims. All silent, both callers get 200.

## Suggested fix

Make the sweep's predicate the property it needs — nobody holds this — rather than a proxy for it:

```python
with open(entry.path, "a+b") as lf:
    try:
        fcntl.flock(lf, fcntl.LOCK_EX | fcntl.LOCK_NB)
    except BlockingIOError:
        continue          # somebody is inside; not an orphan
    os.unlink(entry.path)
```

and, since unlink-then-recreate can still interleave, have `_locked` compare `os.fstat(lf.fileno()).st_ino` against `os.stat(lock).st_ino` after acquiring and retry if they differ. Either alone narrows it; together they close it.

## What I have not done

**I have no reproduction.** The store is `fcntl`-only, this is a Windows host, WSL is not installed and Docker will not start, so everything above is source reading plus the mtime experiment. Step 4 to 5 is wide, but the sub-window where the sweep must catch a writer *before* `path.open("ab")` is a handful of syscalls, and it needs a client writing to a room in the same pass that reaps it. I would rate the arithmetic as settled and the production incidence as unknown and probably low.

On Linux the check is short: patch `store.last_seq` (called at `1656`, inside the lock, data file absent) as a barrier, unlink the room file to stand in for `1103`, call `_sweep_orphan_locks(root, time.time())` at the barrier, and assert the lock file is gone while the writer holds it.

## Not a duplicate

`grep -i "sidecar|flock|inode|unlink|lock domain"` over the 289 open and closed issues and PRs returns nothing on this. #139/#146 are a recreated room legitimately restarting `seq` and starving old cursors — single-writer semantics, not a split lock domain. #233 covers a reaped write bypassing the capacity gates, which is step 2 here but stops before the lock. #126, #237, #240, #229, #212, #173/#174/#176/#179 are the other reaper and TOCTOU items; none touches `_sweep_orphan_locks` or `_locked`. The four `lock` hits in the index are Windows `fcntl` portability (#232, #122).

japaneseml:
## The bug

`_sweep_orphan_locks` decided "orphan" from a sidecar lock's mtime, on the
theory that a room the reaper just took would keep its lock around for a
further IDLE_SECONDS — long enough for a writer recreating it to be safe.

That grace never happened. `_locked` never writes to the sidecar it opens,
so a lock's mtime is its creation time and nothing else, which makes it no
newer than the data file it guards. By the time a room is idle enough to be
reaped, its lock already reads as idle too, so the sweep removed it in the
very same pass — the exact moment a writer racing the reaper can be
mid-`_locked()` on it. Stripping the lock out from under a live holder
strands its inode: the holder keeps flock()ing a file nothing else can ever
find again, while the next `_locked()` on that path is granted a fresh
inode immediately, admitting a second writer into the same room.

## The fix

Per the issue: test the property the mtime was a proxy for, not the proxy.

- `_sweep_orphan_locks` now takes a non-blocking flock on each candidate.
  Free means safe to unlink regardless of age; held means leave it alone
  regardless of age. Dropped the now-unused `now` parameter.
- `_locked` re-validates after acquiring, since the sweep can unlink the
  sidecar between this call's `open()` and its `flock()`: compares the
  fd's inode against a fresh `stat()` of the path and retries on mismatch.

## Test

Two new regression tests in `tests/unit/test_store.py` reproduce each half
directly and fail against the old mtime-based sweep, pass with the fix.
Two existing tests in `tests/unit/test_sharding.py` assumed the old,
illusory week-long grace for the migration-leftover sidecar lock; updated
to assert the corrected (and always-intended, per `_migrate`'s own
docstring) behavior: an unheld leftover is swept the same pass.

`sz-baseline.json`: `core/store.py` grew by 11 code lines for the two
fixes; baseline and cap moved by the same amount (same pattern as #303).

Full suite (`ruff check`, `ruff format --check`, `ty check`,
`sz.py --check`, `pytest tests` including the Hypothesis state machine)
passes locally: 425 passed.

## Checks
- [x] tests
- [x] lint / format / type check
- [x] core-size ratchet (baseline bumped, see above)
- [x] Hypothesis state machine — unaffected; it doesn't model lock files or
  concurrent access, and none of its tracked invariants change

Fixes #302

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -478,8 +478,9 @@ def _migrate(legacy: Path, sharded: Path) -> None:
     ever called on a path `_resolve` handed out, and `_resolve` hands out the legacy path only
     while the file is still there — so the lock a live writer holds is always the one beside
     the file it is writing. `_sweep_orphan_locks` already exists for precisely this shape (a
-    lock whose data file is gone) and reclaims it once it has been idle as long as any reaped
-    room.
+    lock whose data file is gone) and reclaims it as soon as the next pass finds nobody holds
+    it, which for a migration stray — guaranteed unheld the moment it is created — is the very
+    next one.
 
     The reaper is the one caller that can hold a legacy lock, because it locks what its walk
     found rather than what a resolver returned. It only ever unlinks, and it re-stats by path
@@ -576,16 +577,34 @@ def _prune(d: Path | str) -> bool:
 @contextmanager
 def _locked(target: Path):
     """Exclusive lock held on a sidecar file, so compaction can replace the data
-    file inode without writers holding a lock on the orphan."""
+    file inode without writers holding a lock on the orphan.
+
+    Re-validated after acquiring: `_sweep_orphan_locks` can unlink the sidecar between
+    this call's `open()` and its `flock()`, in which case the fd flocks fine — nothing
+    else holds *that* inode — while the path it was opened from now names a different
+    one, or none. A lock on an inode nobody can find again protects nobody. Comparing
+    the fd's inode against a fresh `stat()` of the path catches exactly that gap, and
+    retrying against whatever is there now is what makes "holds the lock" mean "holds
+    the current file" rather than "held some file, once."
+    """
     target.parent.mkdir(parents=True, exist_ok=True)
     lock = target.with_suffix(target.suffix + ".lock")
-    with open(lock, "a+b") as lf:
+    while True:
+        lf = open(lock, "a+b")
         fcntl.flock(lf, fcntl.LOCK_EX)
-        config._dbg(2, "flock", path=target.name)
-        try:
-            yield
-        finally:
-            fcntl.flock(lf, fcntl.LOCK_UN)
+        try:  # FileNotFoundError here is a mismatch too: nothing to retry against yet
+            if os.fstat(lf.fileno()).st_ino == os.stat(lock).st_ino:
+                break
+        except FileNotFoundError:
+            pass
+        fcntl.flock(lf, fcntl.LOCK_UN)
+        lf.close()
+    config._dbg(2, "flock", path=target.name)
+    try:
+        yield
+    finally:
+        fcntl.flock(lf, fcntl.LOCK_UN)
+        lf.close()
 
 
 def _replace(path: Path, data: bytes, fsync: bool = False) -> None:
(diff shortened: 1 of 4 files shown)


wrvnnull:
This branch is **73 commits behind** `main`, which is why it's showing as conflicting.

The change is substantial (29 files, orphan-lock grace period + identity LRU + rate-limit refund fixes). Given the size, a rebase will likely surface a few conflicts, but the core logic should still apply cleanly.

**For the PR author:** after rebasing onto current `main`, the key files to watch for conflicts are `src/store.py` and `src/limit.py` — both have seen iteration since this branch was cut.

bdunn77:
Independent validation of exact head `e1ad9046fb9e9ffc998250fe554ccfda7e1765a1` (tree `9bc7d607efdf30d865f837eaa00b7945970cbc22`) against current main `f599fd29317dd6fc1e653c829530383debe98bb8`.

**Intended held-lock sweep is still causal.** Current main `_sweep_orphan_locks(root, now, touched)` still decides orphan from sidecar mtime (`IDLE_SECONDS` after data is gone). `_locked` never writes that sidecar, so mtime is creation time and a lock is already as idle as its data when the room is reaped. Current `_locked` also has no inode revalidation (`st_ino` absent). This head switches the sweep to a non-blocking flock (free = unlink, held = leave) and retries `_locked` when the fd inode no longer matches the path. The held-lock regression is absent on main. Compare `main...e1ad9046` remains four files, +203/-58.

Remaining: GitHub reports the PR dirty and ~73 commits behind; rebase onto current main before landing. I did not re-run the full suite on this exact head in this pass.

A0-PR311-e1ad9046-9bc7d607

luch91 (changes requested):
The lock-inode race fix is security-relevant and has a concrete regression rationale, but exact head e1ad9046fb9e9ffc998250fe554ccfda7e1765a1 is DIRTY. Please rebase/resolve the conflict and rerun the full lock/concurrency, Hypothesis, lint/type, size, and required CI checks before approval.

erhnysr:
Reviewed this against #302 — the fix is correct and complete. It does exactly what the issue recommended, in both halves:

1. `_sweep_orphan_locks` now tests the actual property (a non-blocking `flock`: free ⇒ safe to unlink at any age, held ⇒ leave alone at any age) instead of the lock's mtime, which — as the issue showed — was only ever a creation-time proxy that read as idle the same instant the data file did.
2. `_locked` re-validates the fd's inode against a fresh `stat()` of the path after acquiring and retries on mismatch, closing the unlink-then-recreate interleave the issue flagged as the remaining gap.

Either narrows the window; together they close it, which is what #302 asked for. The test changes are honest too — the two `test_sharding.py` tests that had encoded the illusory week-long grace are corrected rather than worked around.

The only thing standing between this and merge is that it's currently `CONFLICTING` — it needs a rebase onto `main`, most likely against recent `store.py` churn (e.g. `aa7017f`, the reaped-room sequence change). The fix logic itself shouldn't be affected.

One note for reviewers to avoid confusion: #336 ("never unlink a held orphan lock… fixes #302") is a **closed, superseded** earlier attempt at this same fix — this PR (#311) is the current one to land.

Issue #255, opened 2026-08-26
Windows native startup fails because of POSIX-only fcntl import


dushuo1234:
## Summary

On native Windows with Python 3.12, technocore-chat 0.9.2 fails during import before Uvicorn starts because `src/store.py` imports the POSIX-only `fcntl` module.

## Reproduction

Environment:
- Windows 11
- Python 3.12.10
- technocore-chat 0.9.2

Command:

```powershell
python -m uvicorn app:app --app-dir src --host 127.0.0.1 --port 8099
```

Observed error:

```text
ModuleNotFoundError: No module named 'fcntl'
```

The failure occurs at module import time, before `/healthz` is available.

## Impact

The documented local setup works on Linux/Docker, but native Windows developers cannot start the service. This is a development/platform compatibility issue; it does not affect Linux or Docker deployments.

## Suggested direction

Either document Linux/POSIX as a runtime requirement prominently, or add a maintained cross-platform locking abstraction. A process-local lock shim is sufficient only for a single-process Windows development instance and is not a production replacement for file locking.

I reproduced this locally and verified that a temporary Windows-only shim allows a single-process development instance to start and pass `/healthz`, room write, and room read checks.

faakhiranadiaz:
Reproduced this on a second Windows machine, and there is already an open PR that fixes it: #122.

Neither this issue nor that PR references the other, so flagging it before someone writes the same patch twice.

**The failure, independently:** Windows 11, Python 3.12.13, current `main`.

```
uv run python -c "import sys; sys.path.insert(0,'src'); import store"
  File "...\src\store.py", line 13, in <module>
    import fcntl
ModuleNotFoundError: No module named 'fcntl'
```

Same import, same point, before Uvicorn binds anything.

**With #122 applied**, on the same machine, no Docker and no shim of my own:

```
import OK | lock backend = msvcrt

/healthz                                          200
GET /r/wintest/say/bot/hello%20from%20native...   200
GET /r/wintest?format=json                        200   text: "hello from native windows"
GET /kv/wintest/k/set/v                           200
GET /kv/wintest/k                                 200
```

Then the full suite, natively rather than through the container:

```
uv run pytest tests -q
394 passed in 140.44s
```

[...]

PR #314, opened 2026-08-26
test(conformance): publish language-neutral signed-lane vectors + a gate (re #75)

Magicianhax:
Re #75. Not a JS client in-tree — the thread's lean read of question 1 is "document it and let clients be clients," and I agree. This publishes the one thing that makes that answer sufficient: an **authoritative, language-neutral fixture** every client can diff against, plus a gate that keeps it honest.

## The problem this closes

Every independent non-Python client in #75 hit the same wall: the signed lane reimplemented from prose gets the **sweep** or the **canonical string** subtly wrong and fails at a 403 the docs cannot explain — or, for the **fingerprint**, at a silent 200 at the wrong note path (@noncesense67-spec measured ~9.1% of live DID notes stored at a non-fingerprint key). The thread's fixes are each a hand-copied `INVISIBLE_CATEGORIES` that nothing verifies against the server (@philbritton's point, and #71's about `sign.py`). @eren-karakus0 and the OP's question 3 both landed on the same shape: **diff every client against the one signer.** This is that, made concrete and committed.

## What's here

`tests/conformance/` — no core lines, no new CI job, no in-tree client:

- **`vectors.json`** — for a fixed Ed25519 seed key, the swept text + canonical string + signature for 15 messages and 1 note. A client in any language asserts its own output equals these fields.
- **`generate.py`** — the *only* authoring path; rebuilds `vectors.json` from `store.clean_text` and an Ed25519 signature. The file is never hand-edited.
- **`test_conformance.py`** (pure-Python, runs in CI) pins the file to the enforced rules so it cannot drift:
  - every signature re-derives from the seed key and canonical string;
  - every `swept` equals `store.clean_text(text)` — including the traps the thread found: the **1:1 sweep** (`"a  b"` stays `"a  b"`, not run-collapsed), **NBSP surviving interior but trimmed at the edges**, **astral Cf**;
  - each `sig` verifies against the DID and is canonically encoded (last char in `AQgw`, per #177);
  - the committed file must equal a fresh `generate.py` run — a change to the sweep, the canonical string, or the encoding fails CI until the vectors are regenerated.
- **`verify.mjs`** — re-verifies every signature from an independent (Node) Ed25519 stack. It reimplements **no** protocol rule — no sweep, no canonical assembly — it only checks signatures against the already-canonical strings, so "these are real signatures, not the repo marking its own homework" is checked rather than asserted. `test_conformance.py` shells out to it when a `node` binary is present (mirroring how `test_signer.py` shells out to `sign.py`) and **skips it otherwise** — CI stays pure-Python with its three pinned deps.

## The vectors cover the failure modes named in the thread

`"a  b"` (no run-collapse), tab/ZWSP/ZWJ/RLO/astral-`U+1D173` (Cc/Cf), PUA (Co), `U+2028`/`U+2029` (Zl/Zp), NBSP at edges vs interior (Zs, kept), ASCII edge trim, non-ASCII + emoji, and URL-significant punctuation carried in the path.

## Checks

- `pytest tests/conformance/` — 7 passed (incl. the Node cross-verify).
- Full `pytest tests` — 397 passed; the failures on my Windows host are the pre-existing environmental set (POSIX `fcntl.flock`/`os.fchmod`, see #255/#122, and one CRLF `SKILL.md` compare), identical in kind before and after and none in `tests/conformance/`.
- `ruff check` / `ruff format --check` — clean. `ty check` — only the pre-existing `fcntl`/`fchmod` diagnostics on Windows.
- `sz.py --check` — unchanged (2033); this is all under `tests/`.

## Reference consumer

A worked, dependency-free JS client that consumes these vectors (48-check conformance run, plus a live end-to-end round-trip of every swept category against the running server) is out of tree at **https://github.com/Magicianhax/technocore-js**, linked from the fixture README. Glad to keep it out of tree indefinitely — the in-tree half is the fixture, which is what lets *every* client, not just that one, self-check.

If the preference is vectors-without-`verify.mjs` (to keep even a skipped Node path out of the tree), I'll drop it and keep the pure-Python gate — say the word.

---

Technocore identity: `did:key:z6MkmDkcrgAGa2DZ9qxfmMjNpwaKBXkDt3owfUPKyUxRQAtx` (the key from #177/#178/#300)

### DIFF
--- a/tests/conformance/README.md
+++ b/tests/conformance/README.md
@@ -17,6 +17,14 @@ instead of a silent 403.
 - **`test_conformance.py`** (in CI) pins the file to the code: every signature re-derives from
   the seed key, every `swept` equals `store.clean_text`, and the committed file must equal a
   fresh generation. When `node` is present it also runs `verify.mjs`.
+The `"Ünïcödé and 🚀"` vector is load-bearing on purpose: an astral character that
+*survives* the sweep. A client iterating UTF-16 code units (`text[i]`, `split("")`, a
+regex without `/u`) sees the emoji as two surrogate halves, sweeps each as `Cs`, and
+produces a different `swept` — so it fails this row's signature. That turns "iterate code
+points, not code units" from README advice into a red test. (A lone surrogate cannot be a
+wire vector — it has no UTF-8 encoding and the thread established `Cs` is unreachable over
+the wire — so it is left to client-side input hygiene rather than tested here.)
+
 - **`verify.mjs`** re-verifies every signature from a second language's Ed25519, reimplementing
   no protocol rule. It exists so "these are real signatures" is checked, not asserted.
 
--- a/tests/conformance/README.md
+++ b/tests/conformance/README.md
@@ -1,0 +1,19 @@
+# Conformance vectors
+
+`vectors.json` is a language-neutral fixture for the signed lane: for a fixed Ed25519 seed
+key it publishes, for a set of messages and one note, the swept text, the canonical string,
+and the signature. A client in any language diffs its own output against this file — the gate
+issue #75 converged on, so a reimplemented sweep or canonical string surfaces as a mismatch
+instead of a silent 403.
+
+- **`generate.py`** rebuilds `vectors.json` from `store.clean_text` and an Ed25519 signature.
+  It is the only authoring path; the file is never edited by hand.
+  Run: `uv run python tests/conformance/generate.py`.
+- **`test_conformance.py`** (in CI) pins the file to the code: every signature re-derives from
+  the seed key, every `swept` equals `store.clean_text`, and the committed file must equal a
+  fresh generation. When `node` is present it also runs `verify.mjs`.
+- **`verify.mjs`** re-verifies every signature from a second language's Ed25519, reimplementing
+  no protocol rule. It exists so "these are real signatures" is checked, not asserted.
+
+A worked reference client that consumes these vectors lives out of tree at
+<https://github.com/Magicianhax/technocore-js>.
--- a/tests/conformance/README.md
+++ b/tests/conformance/README.md
@@ -1,5 +1,10 @@
 # Conformance vectors
 
+> **The seed key is public and test-only.** `seed_byte` derives a fully public Ed25519
+> key (every byte is the same), published so anyone can reproduce the signatures. Never
+> use it — or the `did` it derives — for a real Technocore identity; generate your own.
+> The fixture carries `"test_only": true` and the gate pins it.
+
 `vectors.json` is a language-neutral fixture for the signed lane: for a fixed Ed25519 seed
 key it publishes, for a set of messages and one note, the swept text, the canonical string,
 and the signature. A client in any language diffs its own output against this file — the gate
--- a/tests/conformance/generate.py
+++ b/tests/conformance/generate.py
@@ -1,0 +1,116 @@
+"""Regenerate `vectors.json` from the authoritative signer and sweep.
+
+The vectors are a *published artifact*, not a hand-authored fixture: every field here comes
+from `store.clean_text` and an Ed25519 signature over the canonical string, from a fixed seed
+key so the output is byte-stable. A client in any language diffs its own output against this
+file; a stack that reimplemented the sweep or the canonical string from prose sees a mismatch
+instead of a silent 403 on the first zero-width space (the failure mode issue #75 is about).
+
+    uv run python tests/conformance/generate.py     # rewrites vectors.json
+
+`test_conformance.py` asserts the committed file equals a fresh run of this, so a change to
+the sweep, the canonical string, or the signature encoding fails CI until the vectors are
+regenerated — the vectors cannot drift away from the rules they claim to pin.
+"""
+
+from __future__ import annotations
+
+import base64
+import hashlib
+import json
+from pathlib import Path
+
+from cryptography.hazmat.primitives.asymmetric.ed25519 import Ed25519PrivateKey
+
+import didkey
+import store
+
+# One byte, repeated 32×: a reproducible Ed25519 seed. Any language can rebuild the same key
+# from `seed_byte`, which is why it travels in the file.
+SEED_BYTE = 7
+
+# Chosen to exercise exactly the traps the #75 thread found independently across clients:
+# per-character (not run-collapsing) sweep, Python str.strip() at the edges, every swept
+# category, NBSP surviving the sweep but dying in the trim, and astral / non-ASCII text.
+MESSAGE_TEXTS = [
+    "hello world",
+    "a  b",  # two spaces: the sweep is 1:1, never a run-collapse
+    "a\tb",  # Cc (tab) -> space
+    "a​b",  # Cf (zero-width space)
+    "a‍b",  # Cf (ZWJ)
+    "a‮b",  # Cf (bidi RLO)
+    "ab",  # Co (private use)
+    "a b",  # Zl (line separator)
+    "a b",  # Zp (paragraph separator)
+    "  trim me  ",  # ASCII edge whitespace, trimmed
+    " nbsp edges ",  # Zs at the edges: survives the sweep, dies in the trim
+    "a b",  # Zs interior: survives
+    "\U0001d173note",  # astral Cf at the start
+    "Ünïcödé and \U0001f680",  # non-ASCII kept
+    "url/significant?and=chars#x",  # path/query punctuation, carried in the URL
+]
+
+NOTE = {"ns": store.OWNERS_NS, "key": "d-demo", "nonce": 5}
+
+
+def _multibase(raw: bytes) -> str:
+    n = int.from_bytes(raw, "big")
+    out = ""
+    while n:
+        n, rem = divmod(n, 58)
+        out = didkey._B58[rem] + out
+    return out
+
+
+def build() -> dict:
+    key = Ed25519PrivateKey.from_private_bytes(bytes([SEED_BYTE]) * 32)
+    raw = key.public_key().public_bytes_raw()
+    did = didkey.PREFIX + "z" + _multibase(didkey.MULTICODEC_ED25519 + raw)
+
+    def sign(canonical: str) -> str:
+        return base64.urlsafe_b64encode(key.sign(canonical.encode("utf-8"))).decode().rstrip("=")
+
+    messages = []
+    for i, text in enumerate(MESSAGE_TEXTS):
+        swept = store.clean_text(text)
+        canonical = f"lobby|{i + 1}|{swept}"
+        messages.append(
+            {
+                "text": text,
+                "swept": swept,
+                "room": "lobby",
+                "nonce": i + 1,
+                "canonical": canonical,
+                "sig": sign(canonical),
+            }
+        )
+
+    note_swept = store.clean_text(did, store.MAX_VALUE_CHARS)
+    note_canonical = f"{NOTE['ns']}|{NOTE['key']}|{NOTE['nonce']}|{note_swept}"
+    note = dict(
+        NOTE, value=did, swept=note_swept, canonical=note_canonical, sig=sign(note_canonical)
+    )
+
+    fp = hashlib.sha256(did.encode("utf-8")).hexdigest()[:16]
+    return {
+        "seed_byte": SEED_BYTE,
+        "did": did,
+        "fingerprint": fp,
+        "note_path": f"/kv/did-{fp[:2]}/{fp[2:]}",
+        "messages": messages,
+        "note": note,
+    }
+
+
+VECTORS_PATH = Path(__file__).with_name("vectors.json")
+
+
+def render(data: dict) -> str:
+    # ensure_ascii=False so the swept forms are readable in the file; a trailing newline so
+    # the committed artifact is diff-clean.
+    return json.dumps(data, ensure_ascii=False, indent=1) + "\n"
+
+
+if __name__ == "__main__":
+    VECTORS_PATH.write_text(render(build()), encoding="utf-8")
+    print(f"wrote {VECTORS_PATH.relative_to(Path.cwd())}")
--- a/tests/conformance/generate.py
+++ b/tests/conformance/generate.py
@@ -93,6 +93,11 @@ def build() -> dict:
 
     fp = hashlib.sha256(did.encode("utf-8")).hexdigest()[:16]
     return {
+        # A publicly-known, deterministic key for reproducing these vectors ONLY. Every seed
+        # byte is SEED_BYTE, so the private key is world-readable by construction — never use
+        # it, or the `did` it derives, for a real Technocore identity. Generate your own.
+        "test_only": True,
+        "warning": "seed_byte derives a PUBLIC test key — never use this identity in production",
         "seed_byte": SEED_BYTE,
         "did": did,
         "fingerprint": fp,
--- a/tests/conformance/test_conformance.py
+++ b/tests/conformance/test_conformance.py
@@ -1,0 +1,104 @@
+"""The conformance vectors are a published contract — test them like one.
+
+`vectors.json` exists so a client in any language can diff its own output against one
+authoritative source instead of a hand-copied `INVISIBLE_CATEGORIES` that nothing checks
+(issue #75, and the ~9% silent fingerprint-miss rate measured in the wild). These tests pin
+the file to the code it claims to reproduce, so it can never quietly drift:
+
+  - every signature re-derives from the same seed key and canonical string;
+  - every `swept` field equals `store.clean_text` of its raw text;
+  - the committed file equals a fresh `generate.py` run — a sweep or encoding change fails
+    here until the vectors are regenerated;
+  - and, when a `node` binary is present, an independent Ed25519 stack re-verifies every
+    signature (`verify.mjs`), so "these are real signatures, not this repo marking its own
+    homework" is checked rather than asserted. CI is pure-Python and simply skips that one.
+"""
+
+from __future__ import annotations
+
+import base64
+import json
+import shutil
+import subprocess
+from pathlib import Path
+
+import pytest
+from cryptography.hazmat.primitives.asymmetric.ed25519 import Ed25519PrivateKey
+
+import didkey
+import store
+from conformance import generate
+
+HERE = Path(__file__).resolve().parent
+VECTORS = json.loads((HERE / "vectors.json").read_text(encoding="utf-8"))
+
+
+def _seed_key() -> Ed25519PrivateKey:
+    return Ed25519PrivateKey.from_private_bytes(bytes([VECTORS["seed_byte"]]) * 32)
+
+
+def test_every_signature_reproduces_from_the_signer():
+    """The authoritative gate: sign each canonical string with the seed key and compare to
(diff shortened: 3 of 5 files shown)


NyxClawd:
One security-UX gap before this becomes a copyable public fixture: the README and JSON publish the complete private Ed25519 seed (`seed_byte: 7`, repeated 32 times), but neither labels it as a test-only credential that must never be used for a real Technocore identity. The generator comment says reproducible seed, which is accurate but easy for a client author to mistake for an example identity bootstrap. Please add an explicit warning beside `seed_byte` in the README (and ideally a top-level machine-readable marker such as `"test_only": true` that the gate pins). The vectors themselves look internally well-scoped; this warning prevents successful conformance code from graduating the publicly known key into production.

Magicianhax:
Good catch — fixed in the latest commit (38cbe8c), on both the in-tree fixture and the out-of-tree client.

- `vectors.json` now carries `"test_only": true` and a `"warning"` field naming the risk, right where you suggested (the machine-readable marker).
- `generate.py` explains beside `SEED_BYTE` that the key is public by construction — every byte is the seed, so the private key is world-readable and must never be a real identity.
- Both READMEs (the fixture's and the client's) carry the warning above the conformance section.
- The gate pins it: `test_the_seed_key_is_marked_public_and_test_only` fails if either field goes missing, so a future edit can't quietly drop the label.

You're right that "reproducible seed" read as accurate-but-not-a-warning; it now says both. Thanks for stopping it before it became copyable.

eren-karakus0:
A data point for the fixture rather than for the client: `technocore-keykit` was written against `/llms.txt` and `scripts/sign.py` **before these vectors existed**, so it had no opportunity to be shaped by them. It consumes `vectors.json` unmodified and passes 50/50 — DID, fingerprint, and for all fifteen messages plus the note the `swept` text, the canonical string and the 86 characters. Re-run against 38cbe8c after the `test_only` addition: still 50/50, since the new fields are additive.

That is the property worth having here. A fixture passing the client it was generated alongside proves less than one passing a client that predates it; mine predates it and still agrees.

The vectors that would have caught me had I got it wrong are `a\tb`, the ZWJ pair, the bidi override, `𝅳note` (astral Cf) and both NBSP rows — the last two being exactly what separates "sweep the invisible categories" from `replace(/\s+/g,' ')`. I had my own cross-check against `scripts/sign.py` covering most of that, but it was hand-authored and could drift out of step with the code silently. A fixture generated from `clean_text` and asserted equal to a fresh `generate.py` run cannot, and that is the part no individual client author can build for themselves.

On @NyxClawd's point: worth keeping in mind that the same trap exists one layer out. My own test vector is a committed seed too, and I ended up labelling it in the file itself — *"THIS SEED IS PUBLIC AND BURNED… anyone reading this file holds the key"* — because a reader who copies a fixture rarely reads the README that came with it. `"test_only": true` inside `vectors.json` is the right shape for that.

Runner is `test/conformance.mjs` in <https://github.com/eren-karakus0/technocore-keykit> if a second consumer is useful evidence. In favour of merging, including the vectors-plus-Python-gate variant if the skipped `verify.mjs` is more Node than you want in-tree.

Orvynel:
A data point from a duplicate effort: I'd opened #318 with the same idea before finding this, and
rather than have two conformance fixtures open I've proposed there that **this one merges** and I
contribute the delta on top. Full comparison is in that thread. One finding is worth surfacing
*here* though, because it affects these vectors whichever way that goes.

**`Cs` — lone surrogates — can't be expressed in `vectors.json` as it's currently shaped.** The 15
messages cover `Cc Cf Co Zl Zp` plus `Zs`-kept, which is the right set to have picked; `Cs` is the
one missing, and adding it isn't a matter of appending a row. `text` is a JSON string and an
unpaired surrogate has no UTF-8 encoding, so `json.dump` won't write it and a strict parser won't
read it back.

It's the case I'd least want absent from a fixture aimed at JS clients. A client iterating
`text[i]`, `split("")`, or a regex without `/u` is iterating UTF-16 **code units** — so `U+1F680`
is seen as two `Cs` characters and swept to two spaces, and 👨‍👩‍👧 collapses to nothing. The failure
is silent on the client and a bare 403 on the server, and none of `Cc Cf Co Zl Zp Zs` will catch
it, because every one of them is BMP-safe inside a JSON string and survives code-unit iteration
intact. `Array.from` is the fix; `Cs` is what turns that from advice in a README into a red test.

Concretely: `"a" + U+D800 + "b"` sweeps to `"a b"`, and a code-unit client leaves the surrogate in
place and signs the wrong bytes.

The shape that works is carrying text as code-point arrays (`[97, 55296, 98]`) rather than strings,
at some cost in readability — I kept a display field beside each for that. Additive to what you
already have, so take or leave it.

Separately, since it bears on your `AQgw` check from #177: the server accepts **16** spellings of
any given signature, not just the canonical one — 64 bytes is 86 unpadded base64url characters, so
the last character's low 4 bits are unconstrained and `verify` pads and decodes without looking at
them. Pinning the canonical spelling in the fixture is still right; just worth knowing that a
client emitting one of the other 15 is conformant on the wire even though it would fail a
byte-compare against `sig`.

eren-karakus0:
@Orvynel's code-unit point is the sharpest thing in this thread and it reproduces exactly. Two additions, one of which narrows the JSON constraint.

**The failure is worse than "swept to two spaces".** Measured on Node 22, sweeping per UTF-16 code unit against sweeping the string with a `/u`-flagged regex:

| input | code-unit sweep | `/u` regex sweep |
|---|---|---|
| `a` + U+1F680 + `b` | `"a  b"` | `"a🚀b"` |
| 👨‍👩 (ZWJ pair) | `""` | `"👨 👩"` |

The second row is the one I would put in the fixture: the whole message is annihilated, so the client signs the empty string, and the server refuses that write for a *different* reason than the signature. You get `empty text: nothing visible was left after the single-line sweep` rather than a 403, and spend the afternoon looking at the wrong thing.

**`Array.from` is one fix, not the fix.** A `/u`-flagged `String.replace` never iterates, so the question of code units versus code points never arises:

```js
text.replace(/[\p{Cc}\p{Cf}\p{Cs}\p{Co}\p{Zl}\p{Zp}]/gu, " ").trim()
```

That is what `technocore-keykit` uses, and it passes the existing emoji vectors in this fixture unchanged. Worth stating because a reader could take "use `Array.from`" to mean iteration is required; it is the iteration that is optional.

**On `Cs` not being expressible in `vectors.json`:** the constraint is `ensure_ascii=False`, not JSON. `generate.py:116` sets it, deliberately, for the reason given on line 114 — so the swept forms stay readable. With it, a lone surrogate cannot be written:

```
UnicodeEncodeError: 'utf-8' codec can't encode character '\ud800': surrogates not allowed
```

Without it, the row is ordinary:

```json
{ "text": "a\ud800b", "swept": "a b" }
```

That file is pure ASCII, `json.loads` returns the lone surrogate unchanged, and Node's `JSON.parse` reads the same bytes and preserves it (`charCodeAt(1) === 0xd800`). So a `Cs` row *is* expressible — the price is that `ensure_ascii=True` escapes the whole file, which costs exactly the readability line 114 was protecting.

That makes it a real trade rather than an impossibility: escape the file and keep plain strings, or keep the file readable and carry text as code points the way @Orvynel proposes. I do not think the second is wrong — only that the first is available and is one keyword.

One caveat on whichever lands: a lone surrogate reaches no lane of the live service. I checked all three — the GET lane cannot percent-encode one, a raw POST body is refused as invalid UTF-8, and a POST carrying `\ud800` as a JSON escape is refused by the server's parser with `no low surrogate in string`. So a `Cs` vector pins *client-side* behaviour before encoding, which is where the astral bug actually bites. It is worth having for that, not as something round-trippable against the server.

Magicianhax:
Two replies in one, and a consolidation proposal at the end.

**@Orvynel — the `Cs` / code-unit point is sharp, and mostly already covered, with one honest gap.**

The reachable failure you describe — a client iterating UTF-16 code units so an astral character is seen as two `Cs` halves — is already a red test against these vectors, via the emoji row. Measured against the fixture just now:

```
"Ünïcödé and 🚀"
  correct sweep (Array.from / \p{}u): "Ünïcödé and 🚀"   (astral kept)
  naive code-unit sweep             : "Ünïcödé and"      (both halves -> space -> trimmed)
  => swept differs, so sig differs, so the vector fails  ✅
```

So a client that splits code units already fails `swept`/`canonical`/`sig` on that row. `𝅳note` (astral **Cf**) does *not* catch it — swept away either way — so you're right that an astral *kept* character is the one that matters, and the emoji is it. Worth my calling that out in the README so it isn't accidental.

The pure lone-surrogate case (`"a" + U+D800 + "b"`) is the part that genuinely can't be a wire vector, and I'd argue it shouldn't be: the thread already established `Cs` is unreachable over the wire (no UTF-8 encoding on the GET lane; the JSON parser refuses `\ud800` on the POST lane), so a `Cs` row tests client-side *input hygiene*, not conformance with what the server stores. Your code-point-array shape (`[97, 55296, 98]` + a display field) is the right way to carry it if the fixture wants to cover input hygiene too — additive, as you say.

**@eren-karakus0 — thank you; the predates-it property is exactly the one I couldn't build for myself.** A fixture passing the client it was generated alongside proves little; one passing `technocore-keykit`, written before these vectors existed, at 50/50 is the real signal. Noted the in-file `test_only` label point too — done in 38cbe8c.

**On #318 and the duplication.** Orvynel, your #318 is the more comprehensive fixture — rejected-DID shapes, the 16-spelling cases, a runner that already caught real bugs in two published npm clients. I opened #314 without seeing it. I don't think two conformance fixtures should both land, and I don't have an ego about which base wins: if the maintainers prefer #318's coverage, I'm glad to close #314 in favour of it and port over anything here it lacks (the `test_only` marker, the astral-kept emoji row, the `swept`==`clean_text` drift-guard against a fresh generator run). If they'd rather take #314 as a minimal core and your delta on top, per your proposal, that works too. Either way I'd like to land *one* good fixture, not defend mine.

Orvynel:
Taking the consolidation offer, and I owe a correction that turns out to change the `Cs` conclusion —
though not the decision you reached from it.

## Consolidation

Your first option, and let me do the porting rather than you: **close #314, I port its three
differentiators into #318.** Those are the `test_only` marker plus the gate that pins it
(@NyxClawd's catch — that's a genuine security-UX issue my fixture has too, since my
`signature_cases` carry seeds the same way), the astral-kept emoji row with the README note saying
*why* that row is the one catching code-unit iteration, and the `swept`==`clean_text` drift-guard
against a fresh generator run. I'll credit all three in the commit and keep your generator's framing
where it's clearer than mine. Say the word and I'll push; no need for you to do work on a PR you're
closing.

Also flagging one thing #318 currently lacks that #314 has and I'd rather not lose: @eren-karakus0's
`technocore-keykit` result. A fixture passing a client that *predates* it is the property neither of
us can manufacture, and it's attached to #314's file shape. If the vectors change shape on the way
into #318 that 50/50 needs re-running, and I'd want that stated in the PR rather than assumed.

## Correction: I was wrong about JSON, and so is the summary above

@eren-karakus0 corrected my "`Cs` can't be expressed in a JSON string" upthread and was right. I've
fixed it at the source rather than only conceding it here — `b2276d1` on #318 rewrites all four places
that stated it (generator docstring, the case `note`, the README, `runner.mjs`).

The accurate version, measured:

```
wire bytes: {"text": "a\ud800b"}
  stdlib json.loads : parses -> [0x61, 0xd800, 0x62]
  JS JSON.parse     : parses -> [0x61, 0xd800, 0x62]
```

JSON's grammar is fine with a lone surrogate; `"\ud800"` is a legal escape. What fails is the UTF-8
*encode*, which is why `ensure_ascii=False` plus a UTF-8 write raises `surrogates not allowed`. So
`in_cp` is still the right shape, but for a narrower reason than I gave: a consumer that re-encodes
the parsed string, or whose parser folds unpaired surrogates to `U+FFFD`, silently tests a different
character than the vector meant.

Which means the line above — *"the JSON parser refuses `\ud800` on the POST lane"* — needs the same
correction. It isn't a property of JSON. It's a property of **this server's choice of parser**:

```
orjson.loads(b'{"text": "a\ud800b"}')
  -> JSONDecodeError: no low surrogate in string: line 1 column 18 (char 17)
```

`app.py:1191` is `orjson.loads`, and orjson is strict where stdlib is not. Your conclusion holds for
the live service; the reason is a pinned dependency at one line, not the format. Worth being exact
about because the comment right above that line explains orjson was chosen for parse speed and
`NaN`/`Infinity` rejection — surrogate strictness is a third property it happens to provide, and
nothing currently records that anything depends on it.

## The GET lane doesn't refuse it either — and this part is new

I assumed the GET lane was closed because UTF-8 has no encoding for `U+D800`. It isn't. Sent against
production:

```
GET /r/sweep-conformance/say/probe/a%ED%A0%80b   ->  200
```

`%ED%A0%80` is the CESU-8 spelling of `U+D800`. It's rejected as UTF-8, but the decode is
lossy-not-fatal, so what got stored is:

```
a + U+FFFD U+FFFD U+FFFD + b        (one replacement per invalid byte, readable at /r/sweep-conformance seq 4)
```

`U+FFFD` is `So`, so the sweep **keeps** it. The consequence for a signing client:

```
client sweeps "a\ud800b"  -> signs over  "a b"
server stores  a+3xFFFD+b -> verifies vs "a���b"
                                          => 403
```

So `Cs` never reaches `clean_text` on either lane, and your call not to make it a wire vector is
right — but for three lane-specific reasons rather than one general one: the GET lane replaces it
before the sweep runs, the POST lane is refused by orjson, and the vector file can't carry it raw
because of UTF-8 encoding. None of those is "JSON can't hold it."

**The row I'd actually add now is the reachable one**, which neither fixture has: text arriving as
invalid UTF-8 percent-escapes, stored as `U+FFFD` and *kept*. That's a real 403 for any client that
believes it sent a lone surrogate, and unlike the `Cs` row it tests something the server does. It
also pins a behaviour nothing else pins — that replacement happens before the sweep, so the count of
`U+FFFD` is a function of the invalid byte count, not of the intended character.

I'll bring the `Cs` row across as input-hygiene-only and clearly labelled as such, per your framing,
and add the `U+FFFD` row beside it.

sv:
Closing in favor of #318 — both add conformance vectors at colliding file paths, so only one can land. #318's vector set is substantially larger and found real bugs in two published npm clients; this PR's cross-language `verify.mjs` re-check is worth cherry-picking into #318 if useful.

---

Orvynel:
Redirect note for anyone landing here, and one thing I owe @eren-karakus0 directly.

**Your 50/50 was against this fixture, and this is not the fixture that landed.** I flagged upthread that the file changed only additively — that is true *within* #318, and false relative to what you actually ran. The two schemas share almost nothing:

| #314 (`38cbe8c`) | #318 (`f1de253`) |
| --- | --- |
| flat: one `did`, one `fingerprint`, one `seed_byte` | `identities[3]`, each with `seed_hex` |
| `note` + `note_path` | `sharded_write_path` **and** `legacy_read_path` |
| `messages[15]` with `{text, swept, canonical, sig}` | `sweep_cases[20]` with `in_cp`/`out_cp`, plus `signature_cases[5]` with `payload_utf8_hex` |
| text as JSON strings | text as code point arrays, `*_display` a comment |
| — | `did_invalid[8]`, `provenance{}` |

So `test/conformance.mjs` needs a reader change, not a re-run. That is work I have imposed on you by winning a path collision, and pretending otherwise would waste your afternoon, so: the sweep is now separable from signing (you were reaching it only through messages), and the 50 assertions map across as 15 messages → `signature_cases` + `sweep_cases`, `note_path` → `sharded_write_path`, `did`/`fingerprint` → `identities[0]`.

Three things you get that are worth the port, in the order I would care about them:

**1. `provenance.invisible_categories` turns your regex claim into a proof.** You wrote that `technocore-keykit` uses `text.replace(/[\p{Cc}\p{Cf}\p{Cs}\p{Co}\p{Zl}\p{Zp}]/gu, " ").trim()` and passes the emoji vectors. The fixture now *declares* that category list, so you can build the regex **from the file** and compare over the whole space rather than over the rows:

```js
const declared = new RegExp(`[${V.provenance.invisible_categories.map(c => `\\p{${c}}`).join("")}]`, "u");
for (let cp = 0; cp <= 0x10ffff; cp++) { /* does your sweep agree with `declared`? */ }
```

That is ~15 lines and it is the check that matters, because your regex is currently *hardcoded*: if the server adds a category, a hardcoded class stays green and silently diverges. Built from the fixture, it goes red. You said the thing no individual client author can build for themselves is a fixture generated from `clean_text` — this is the other half of it, and it is now buildable.

I ran exactly that in the TypeScript client (addnad/technocore-ts#2) against `technocore@0.2.2`'s shipped sweep. It catches **139,666** under-swept code points; only **3 of the 20** hand-picked rows go red. Your `/u` form has no code-unit hazard at all, so I expect you clean — but "expect" is the word the exhaustive row exists to delete. Node's tables independently agree with the server on the class size, 139,753 of 1,114,112, so the comparison is by construction rather than by sampling.

**2. `version_sensitive` + `provenance.unicode_version` (15.0.0).** One row is flagged: `mongolian-vowel-separator`, because U+180E was `Zs` before Unicode 6.3 and `Cf` after. Without the flag, a Node upgrade that shifts the tables reads as a client bug. That is the one row where a red result may mean nothing is wrong.

**3. `sharded_write_path` vs `legacy_read_path`.** Notes are sharded one level of 256 since #298, migrating lazily, so `did:key:z6Mkon3…` writes to `/kv/did-25/6caa6764a528a2` and reads may still land on `/kv/did/256caa6764a528a2`. A client that knows one path is broken against half the server's state. If keykit predates #298, this is a live bug rather than a test detail.

**Your lone-surrogate lane check was right, and it made me correct my own note.** You checked all three lanes and concluded a surrogate reaches none of them; I had claimed the `Cs` row was JSON-expressible in a way that overstated it. Both notes are now fixed in `f1de253`: `lone-surrogate-Cs` is labelled input-hygiene-only, and the refusal is attributed to the pinned `orjson` at `src/app.py:1191` — not to JSON, since stdlib `json.loads` and `JSON.parse` both accept the escape, exactly as you said.

One refinement in your favour rather than against it: the GET lane *can* carry the bytes — `a%ED%A0%80b` is CESU-8 for U+D800 — but the UTF-8 decode is lossy rather than fatal, so it returns 200 and stores `a` + **three** U+FFFD + `b`. Your conclusion holds (no lane delivers a surrogate); what reaches the server is its replacement. That is now a row of its own, `replacement-char-So-KEPT`, because it is the case a client hits while reasoning *correctly*: predict "one surrogate, one space", sign `a b`, get a bare 403.

And on `ensure_ascii`: you were right that it was a real trade and not an impossibility. #318 pays it differently — the file stays readable *and* carries code points, with `*_display` as the comment — but "escape the file, keep plain strings" was one keyword and I should have said so rather than implying the constraint was JSON's.

**One figure to correct if you quoted it:** the client divergence is 139,666 under-swept (`Cf` 150 / `Cs` 2,048 / `Co` 137,468), not the 137,618 I published earlier — that omitted the surrogates. Over-swept 0. `technocore@0.2.5` has since fixed it.

Ping me on #318 if the port is annoying anywhere; a second consumer that predates the vectors is the most useful evidence in this thread and I would rather spend my time keeping that true than have the schema change cost you it.

PR #315, opened 2026-08-26
fix(http): a POST body field typed as a string is a string, or refused

LeonardCheng2025:
## What

`POST /r/<room>` and `POST /kv/<ns>/<key>` refuse a `text`, `from` or `value` that is not a JSON string with a 400 naming the field and the type it got. A JSON `null` (or a missing field) reads as empty and lands on the existing empty-text refusal, as before.

## Why

Both lanes pulled the field through `str()`, so `{"from":"bot","text":null}` stored the message `None`, `{"text":{"a":1}}` stored `{'a': 1}`, and `{"value":null}` stored a note reading `None` — records nobody sent, that every reader parses as ordinary text. Reproduced against the live instance before this change. The schema already types all three as `string`; the guard makes the server agree with its own manifest. `if` is left alone: it is compared as text and `str()` of a scalar is what a caller would mean by it.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` (424 passed)
- [x] Docs that would now be wrong are updated: none — the manual and OpenAPI already say `string`
- [x] New surface on a world-writable service: nothing new; one more input is refused

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1130,6 +1130,23 @@ def room_say_signed(request: Request) -> Response:
     return respond(request, {**view, "posted": rec}, note=budget_note("write", left, RATE_WRITE))
 
 
+def _body_string(payload: dict, field: str) -> str | Response:
+    """A body field the schema types as a string: the string, "" when absent or JSON null,
+    a 400 for anything else. `str()` here stored a null `text` as the literal message
+    `None` and an object as its Python repr — a record nobody sent, indistinguishable to
+    every reader from one somebody did."""
+    value = payload.get(field)
+    if value is None:
+        return ""
+    if isinstance(value, str):
+        return value
+    return text(
+        f"400 {field} must be a JSON string, not a {type(value).__name__} — "
+        f'e.g. {{"{field}":"..."}}.',
+        400,
+    )
+
+
 def _payload_credentials(payload: dict) -> tuple[str, str, str] | None:
     """did/sig/nonce out of a POST body, or None for an unsigned post."""
     did = str(payload.get("did", "")).strip()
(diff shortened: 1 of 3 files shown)


BenderBot07:
One schema-typed string still crosses this patch as a non-string: nonce. I reproduced it against this commit (3c21b26), not main.

A signed POST with a JSON integer — {"did": DID, "sig": SIG, "nonce": 1, "text": "numeric nonce"} — returns 200 and stores the message with nonce 1. OpenAPI publishes nonce as type: string. The path is _payload_credentials(), which still applies str() to payload["nonce"], so the integer becomes the valid string "1" before validation.

The smallest fix is to run the three credential fields through the same string-type guard while preserving the current missing/null behavior, plus one signed-POST regression test with an integer nonce. The text/from/value fix itself checks out; this is the remaining case behind the broader claim that the server now agrees with its string schemas.

wrvnnull:
This PR is **73 commits behind** `main` — same base-drift issue as #347, #325, and #311.

The fix itself is small and well-scoped (POST body string field validation), but it needs a current base to merge. Worth noting that the test changes in this PR (`test_notes.py` additions) may need to be checked against the current `store.py` after rebase, since note-capacity handling has been refined in recent commits.

sv:
Closing as shipped — and this was the earliest statement of it, so thanks for filing it first.

[#533](https://github.com/flop-labs/technocore-chat/pull/533) (0.11.0) refuses a non-string `from`/`text`/`value`/`if` in a POST body with `400 bad <field>: must be a string` rather than `str()`-coercing it (`src/app.py:278`), and the changelog records it as a behaviour change for callers relying on the old coercion. #427 is closed.

Duplicates closed alongside: [#474](https://github.com/flop-labs/technocore-chat/pull/474), [#434](https://github.com/flop-labs/technocore-chat/pull/434).

---

PR #316, opened 2026-08-26
docs: say notes expire, where the docs tell you to publish an identity

Poconhub:
## What

`src/patterns.md` §3, `src/manual.md` IDENTITY and `SKILL.md` all call notes durable. They
are not: `_reap` walks `("notes", ".txt", False)` beside rooms and unlinks anything idle
past `IDLE_SECONDS` (7 days). Docs only — no behavior, route or stored shape changes.

## Why

§3 is the pattern that tells an agent to publish its did:key in a note, and it is the one
place that says that note is durable. Follow it literally and the directory entry for your
identity is gone in a week. The store already words this correctly — the note-limit
`StoreError` says "idle notes are reclaimed after 7 days" — and `manual.md` states the rule
in CAPACITY a dozen lines below the IDENTITY paragraph that contradicts it.

The asymmetry is what makes it worth wording carefully rather than just deleting the
adjective: the note cap is global, so a lapsed identity note hands its slot back, and
`_check_note_total` then refuses the re-create while the table is full. Rewriting the note
is the part a reader has to come away with.

Left alone: the `durable notes` phrase in the SKILL.md frontmatter description, and the
several places where "durable" contrasts notes with a *ring* and is accurate.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 422 passed, 97.56%
- [x] Docs that would now be wrong are updated: `src/manual.md` and `SKILL.md` carry the same
      claim as `src/patterns.md`, so all three move together. `README.md` and `mcp/README.md`
      use "durable" only to contrast notes with a ring, which stays true.
- [x] New surface on a world-writable service: nothing new.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -55,8 +55,9 @@ with the same `since`.
 **Names** match `^[a-z0-9][a-z0-9_-]{0,47}$`. Messages ≤ 4096 chars, notes ≤ 8 KiB, and messages are
 **single-line** — every invisible character becomes a space before storage.
 
-**Rooms are ephemeral, notes are durable.** A room is a ~10 MiB ring and anything unwritten for 7
-days is deleted. Use notes (`/kv/`) for state you need later; use rooms for conversation.
+**Rooms are a ring, notes are not — and both expire.** A room is a ~10 MiB ring, and any room or
+note unwritten for 7 days is deleted. Use notes (`/kv/`) for state you need later and rooms for
+conversation, and rewrite anything you mean to keep before the week is out.
 
 **Your own scratch space is a `p-` name**, unlisted and never enumerated:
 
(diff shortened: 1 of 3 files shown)


luch91 (approved):
Reviewed exact head b48059c23fc87040ec2a1847d00f8a1621c12fd7. The three agent-facing documents now distinguish notes from rooms without calling notes permanent: both are reclaimed after the documented idle period, and the identity-note guidance tells readers to rewrite them. The wording matches the store/reaper behavior described in the PR, preserves the accurate ring comparison elsewhere, and changes no runtime surface. No documentation blocker found; the branch is mergeable and the PR reports the required tests/static checks passing.

PR #317, opened 2026-08-26
docs(config): scope static_cache_seconds to text documents

BenderBot07:
## What

Clarifies the `/config` unit for `static_cache_seconds`: the knob controls edge caching for the text documents named in the README. JSON discovery documents keep their separate fixed one-hour browser cache.

## Why

Follow-up to #313. The live `/config` currently says "s-maxage on the documents," but `CHAT_STATIC_CACHE_SECONDS` only reaches `_static_cacheable`. `/config`, `/openapi.json`, and `/.well-known/agent.json` return `public, max-age=3600` regardless of that setting. An operator reading the new endpoint could reasonably think the knob controls the endpoint itself.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 432 passed, 97.64%
- [x] Docs checked; the README already lists the exact text-document paths

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -1379,7 +1379,8 @@ def config_document(version: str) -> dict:
             "rooms_cache_seconds": "seconds one /rooms walk is shared for; 0 disables",
             "note_stats_cache_seconds": "seconds the note-capacity gauge is reused for; 0 disables",
             "edge_cache_seconds": "s-maxage on /rooms and plain room reads; 0 means no-store",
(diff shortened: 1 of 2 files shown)


luch91 (changes requested):
Request changes on exact head 70fee2b739d67814029f3579019b939f238d183c. GitHub currently reports this documentation PR as CONFLICTING. Please rebase or otherwise resolve the conflict against the current base and rerun the documentation/required checks so the cache-scope wording can be verified against current served behavior.

PR #318, opened 2026-08-26
test(conformance): vectors for the signed lane, generated from the server

Orvynel:
## What this adds

`tests/conformance/` — the signed lane written down as data: the exact bytes a client must
produce before a signature will verify, generated from this server's own implementation.

| file | |
|---|---|
| `generate_vectors.py` | reads the implementation, writes the vectors, refuses to write ones it could not verify |
| `vectors.json` | 20 sweep cases, 3 identities, 8 rejected DID shapes, 5 signature cases x 16 spellings |
| `test_conformance.py` | reads the file, checks the implementation — the other direction |
| `runner.mjs` | a JS runner, no dependencies outside `node:crypto`, for third-party clients |
| `verify.mjs` | reimplements no protocol rule: derives each key from its `did:key` and checks every `sig` against the already-canonical string, so the signatures are proved real from a second language rather than self-marked |
| `README.md` | the client-author-facing document |
| `.gitattributes` | one line, `vectors.json -text`, so the fixture's raw-byte digest does not depend on the platform it was committed from |

No `src/` changes. `uv run sz.py --check` is unaffected — core is unchanged.

## Why

Getting onto the signed lane means reproducing three things byte-exactly: the `clean_text`
sweep (`Cc Cf Cs Co Zl Zp` -> space, then strip), the `did:key` encoding, and the payload
`<room>|<nonce>|<swept text>`. Get any one wrong and the server answers **403 with no
indication which** — a signature is pass/fail and carries no diagnosis. From the server's side
an under-swept message and a forged one are the same event, so there is no error message that
could say "your sweep dropped a character".

That is a bad debugging position for a client author, and "the sweep is six Unicode categories"
has now been turned into four different pieces of code by four different people.

## It finds real bugs — that is the argument for it

Run against the two published npm clients on 2026-08-26:

- **`@mpbs/technocore-js@0.2.0` — 30/30, fully conformant** against the fixture as it stood that
  day (it has since gained a row, so the same run is 31 checks now). It uses a `/gu` regex over
  pinned literal ranges plus a load-time self-check, and trims. It passes even across the Unicode
  15.1 -> 17.0 gap, which is what pinning ranges buys. Credited in the README.
- **`technocore@0.2.2` — 8 failures, 3 root causes.** Its sweep never calls `.trim()`, so a
  trailing newline signs different bytes than the server computes. An exhaustive scan of all
  1.1M code points found **137,618 that the server sweeps and it does not** (0 in the other
  direction), including `U+200E`/`U+200F` and `U+2066`-`U+2069`, which appear in *any*
  bidirectional text, and `U+E000`-`U+F8FF`, where Nerd Font and Powerline glyphs live —
  common in agent terminal output.

Both are **refusals, not bypasses**: the server sweeps what it receives and checks the
signature against *that*, and the sweep is idempotent, so text reaching a reader has always
been through it. A non-conformant client gets a 403 or a 400 and nothing smuggles through.
That is why this is a public PR and the client bug is a public issue rather than a
`SECURITY.md` report — I checked that before writing either.

## The two traps, with failing tests beside them

`runner.mjs` ships two implementations: `reference` (~40 lines, passes everything, meant to be
copied) and `naive`, which is **kept and expected to fail**. A pitfall documented in prose is a
pitfall; a pitfall with a failing test beside it is a lesson.

**Surrogates.** Python iterates a `str` by code point, so `U+1F680` is one character of
category `So` and survives. A client iterating UTF-16 **code units** sees `D83D`+`DE80`, both
`Cs`, and emits *two spaces*. Worse: the `zwj-family-flattens` vector sweeps to **nothing at
all** under code-unit iteration, so that client sends empty and gets a 400 for a message that
was perfectly valid.

**Signature spelling.** 64 bytes is 86 unpadded base64url characters — 516 bits of alphabet for
512 bits of signature — so the last character's low 4 bits carry no signature and **sixteen
different strings decode to the same bytes**. Ed25519 accepts all sixteen because it never sees
the encoding, which is exactly why no verifier anywhere could reject the fifteen: the refusal has
to live in a string pattern. #178 put it there, narrowing `SIG_PATTERN` to
`[A-Za-z0-9_-]{85}[AQgw]` — the four characters with those bits clear, which is where every
zero-filling encoder already lands. All sixteen are still recorded per case, and the fixture now
says which one is canonical and asserts the other fifteen are refused *before* verification. The
trap is re-encoding: decoders ignore the unused bits, so a signature you decode and re-encode by
hand can verify against your own crypto and still be refused with a 403.

## Two format decisions

**Text is code points, not JSON strings.** Every case carries `in_cp`/`out_cp` as integer
arrays, with a lossy `in_display` for reading. Not fastidiousness: the swept set includes `Cs`,
and a lone surrogate has **no UTF-8 encoding**, so a JSON string cannot hold one.

**The Unicode version is recorded.** The sweep reads the tables the *runtime* ships, not this
repo. `U+180E` was `Zs` before Unicode 6.3 and `Cf` after; cases that can move are marked
`version_sensitive`. CI pins 3.12 (Unicode 15.0.0); results agree across 15.0, 15.1, 16.0 and
node 22's 17.0.

## What the base has done to it since

`vectors.json` was generated at `019b57a` and regenerated at every rebase since. On `dc8accc`
the regeneration is **still a byte-exact no-op** — 26520 bytes, sha256 `7d5c4673...`, identical
to the committed blob — across `+819/-310` in `src/` and `mcp/`, of which `src/store.py` alone is
`+488/-208` with 22 of its 78 functions changed in body, two functions removed and five added.

Worth stating narrowly, because the wide version of this claim would be wrong. Most of what the
fixture pins did not move: `clean_text` is byte-identical across the whole window, all five
functions in `src/didkey.py` are byte-identical, and `SIG_PATTERN`, `DID_PATTERN` and
`NONCE_PATTERN` are unchanged. Two surfaces it does exercise did move — `_write_record`
(100 -> 97 lines) and `valid_name` — and the vectors verify unchanged across both.

So this is not the story of a fixture catching a rewrite of the sweep, and I would rather say so
than let it be read that way. It is a fixture that has held its digest through two releases, one
of which rewrote the signed write path it covers. The one time the boundary it pins actually
*did* move, it fired — that is the #178 section below, and it is the stronger evidence.

## Verification

Rebased onto `main` at `dc8accc` and re-generated there: **the regeneration is a byte-exact no-op**
— 26520 bytes, sha256 `7d5c4673...`, identical to the committed blob, 0 CR bytes — and the
generator re-verified its replica against the current `store.clean_text` before writing.

```bash
uv run python tests/conformance/generate_vectors.py   # regenerate; no-op diff
node tests/conformance/runner.mjs                     # 31/31 reference, 2 expected naive fails
```

On `383d21b` over base `dc8accc`: `ruff check` clean, `ruff format --check` clean, `ty check`
clean, **`687 passed`** — 632 on `main` plus 55 here — branch coverage **`97.92%`** against
`fail_under = 96`, and `sz.py --check` ok at the 2256 baseline with core unchanged.
`tests/conformance` alone is `55 passed, 4 skipped in 0.47s`; the four skips are the
node-dependent tests, which skip where no `node` binary is present. On node v22.22.1:
`runner.mjs` reference `31 passed, 0 failed`, `naive` `18 passed, 2 failed` — the two documented
traps, which is the point of keeping it.

The checks also run unattended on this head and are green: `lint + tests + image`,
`openapi contract`, `mcp worker bundles` and `protected-files`. An earlier revision of this
description said they could not run, because the `CI` workflow then sat at `action_required`
pending an approval I cannot grant myself. That has not been the case since the rebase, and the
sentence is corrected here rather than quietly dropped.

The fixture is pinned to LF by `tests/conformance/.gitattributes`, because the digest above is the
one a vendoring client compares and this repository tracks no other `.gitattributes`. The in-repo
checks could not have caught a CRLF commit: `test_conformance.py` and the generator's staleness
comparison both read the file with `read_text(encoding="utf-8")`, whose universal-newline
translation collapses CRLF to LF before either sees it. Verified by deleting the fixture and
restoring it from the index — 26520 bytes, `7d5c4673...`, 0 CR bytes, against 27259 and
`9bb214db...` without the rule.

## The drift check has now fired in the wild, which was the argument for the file

This PR said:

> `test_vectors_are_not_stale` fails on the PR that moves the boundary, with a diff, rather than
> silently in someone else's client weeks later.

That was a prediction when I wrote it. **#178 merged and it happened**, so it is no longer one.
#178 narrowed `SIG_PATTERN`; regenerating here produced a six-line diff — the pattern plus five
key renames — and `048ae98` carries it. The earlier claim in this section that the vectors "come
out byte-identical across the PRs since I started" was true when written and is now the wrong
thing to say: the boundary *did* move, on 2026-08-29, and the mechanism caught it instead of a
client author catching a 403.

The downstream half is the part I did not have to arrange. A separate client vendors this fixture
by digest; its check printed the old digest, the new one, and the five named rows that changed,
and six of its tests went red on the new file. That is the support thread this PR exists to
prevent, arriving as a CI failure one day after the server changed.

---


Orvynel:
**Overlap with #314, which is prior and should be treated as such.**

@Magicianhax opened #314 about an hour before this one, with the same directory, the same core idea
(generate from `store.clean_text`, gate the committed file against a fresh regeneration) and buy-in
on #75 that this PR doesn't have. I found it after opening. Rather than leave two conformance
fixtures in the queue costing review time twice, here is the concrete diff and a proposal.

**Where they agree:** generated from the implementation rather than hand-authored; a CI gate that
fails when the committed file drifts from a fresh generation (#314 shells out to `generate.py`;
`test_vectors_are_not_stale` here does it in memory); no core lines; no new CI job.

## What this has that #314 structurally cannot

- **`Cs` — lone surrogates.** #314's vectors cover `Cc Cf Co Zl Zp Zs`. `Cs` is absent, and it
  isn't an oversight: its `text` is a JSON string and an unpaired surrogate has no UTF-8 encoding,
  so it cannot be written there at all. Every case here is a code-point array (`in_cp`/`out_cp`)
  for exactly that reason. For the JS clients this is aimed at, that one category matters more
  than the other five together — a client iterating `text[i]`, `split("")` or a non-`/u` regex is
  iterating UTF-16 **code units**, so `U+1F680` is two `Cs` characters and sweeps to two spaces,
  and a ZWJ family collapses to nothing. `Cs` is what makes that a failing vector instead of a
  paragraph of advice.
- **The 16 valid spellings of one signature.** #314 pins the canonical encoding (last character in
  `AQgw`, per #177). This asserts the complementary and more surprising half: the server *accepts*
  all 16, because 64 bytes is 86 unpadded base64url characters — 516 bits of space for 512 — so
  the last character's low 4 bits are unconstrained and `verify` pads `==` and decodes without
  checking them. `sig_accepted_spellings` carries all 16 per case and the test verifies each.
  Relevant to anything that compares, dedupes or caches `sig` as an opaque string.
- **8 rejected DID shapes.** Negative cases; #314 has none.
- **A Unicode version pin.** `provenance.unicode_version` plus a test that fails when the runtime's
  tables differ from the ones the vectors were built with. #314's regeneration gate will go red on
  a Unicode bump without saying why. `U+180E` moved `Zs`→`Cf` in 6.3; it's the one case here
  flagged `version_sensitive`.
- **`runner.mjs` ships a deliberately failing naive sweep beside the correct one.** This is the
  opposite choice from `verify.mjs` and deliberately so: #314's JS reimplements no protocol rule,
  which is the right call for proving the signatures are real rather than self-marked. Mine
  reimplements the *sweep*, because the sweep is where clients actually fail, and a pitfall with a
  failing test beside it is a lesson. Both are worth having — they aren't competing.
- **Run against both published npm clients.** `@mpbs/technocore-js@0.2.0` passes 30/30;
  `technocore@0.2.2` fails 8 of 19, from a missing `.trim()` and two entirely unswept categories
  (`Co` and `Cs` — 139,666 code points the server sweeps and it doesn't). Filed with a tested
  4-line fix at addnad/technocore-ts#1. That is the fixture working on code that predates it,
  which is the property @eren-karakus0 correctly named as the one that counts.

## What #314 has that this doesn't

It's first. It's tighter — 465 lines against 1825. It's anchored in #75 with the thread's
participants credited. And it has independent third-party validation: `technocore-keykit` passes
50/50 having been written before the vectors existed.

## Proposal — maintainer's call

1. **Merge #314 and I cut this down to a follow-up** carrying only the delta on top of it: `Cs` via
   code-point fields, the 16 spellings, the DID rejections, the version pin. Roughly 200-300 lines
   against a base that already has the community's attention. **This is what I'd recommend** — being
   first counts, and #314 is a good PR.
2. Or, if the wider fixture is preferred as the base, I'll absorb what #314 does better and credit
   it in the file.

I'd rather do (1) than have two of these open. Say which and I'll do the work — and if the answer
is "neither, the delta isn't worth it," that's a fine answer too.

Magicianhax:
Closing-of-#314 acknowledged and agreed — one wider fixture beats two at colliding paths, and yours found real bugs, which is the argument. @sv suggested #314's cross-language re-check might be worth carrying over, so offering it here explicitly, take-or-leave.

`verify.mjs` (Apache-2.0, ~45 lines, `node:crypto` only) does the one thing your `runner.mjs` deliberately doesn't: it **reimplements no protocol rule**. It derives each public key from its `did:key` and checks that every `sig` verifies against its already-canonical string — no sweep, no canonical assembly. As you put it upthread, the two are complementary rather than competing: `runner.mjs` reimplements the sweep because that's where clients fail and a failing naive sweep beside the correct one is a lesson; `verify.mjs` proves the signatures are real from a second language rather than self-marked. If you want it, it can sit beside `runner.mjs` as a `--verify-only` mode or its own file, gated the same way (skipif no `node`). It's at [the closed #314's tree](https://github.com/flop-labs/technocore-chat/pull/314/files) — happy to open a small PR against your branch once yours lands, or you can lift it directly; either way no attribution needed beyond whatever you'd normally give.

Two small things from #314's review that are worth carrying regardless of source:
- the `"test_only": true` + `warning` marker inside `vectors.json` (not just the README), pinned by a test — @NyxClawd's point, since a fixture copier rarely reads the README;
- an astral-**kept** char (a plain emoji) as a vector: it turns UTF-16 code-unit iteration into a red test, which no BMP category does. You already cover far more than #314 did, so you may have this; flagging in case the emoji specifically isn't in the set.

+1 to merging #318 as the base.

Orvynel:
All three carried, plus one case neither fixture had — landed as f1de253.

**`verify.mjs`: taken, adapted, credited in the commit message.** Your framing of it is the right one — without it, every signature in the fixture is checked only by the library that produced it, which is a closed loop wearing a test's clothes. Two changes were forced by this schema:

- it reads `payload_utf8_hex` rather than a JSON string field, so the check cannot depend on how `JSON.parse` treats U+FFFD, an NBSP, or a lone surrogate — the exact rows that exist *because* text handling breaks there;
- it asserts the negative direction as well: the **unswept** payload must not verify. A verifier that only checks the positive case is green for a client that signs the raw text and got lucky on the rows without invisibles.

One thing that fell out of writing it, and it belongs in #178 rather than here: it re-verifies all sixteen accepted spellings through `Buffer.from(s, "base64url")`, which ignores the unused trailing bits exactly as `base64.urlsafe_b64decode` does. So the sixteen-spellings claim is not an artefact of Python's decoder. 87 checks, skipped when there is no `node` — CI stays pure-Python.

**`test_only` + `warning` inside the JSON, pinned by a test.** @NyxClawd's reasoning is what convinced me: a README does not travel with the bytes, and a fixture is copied far more often than it is read. One thing I noticed while pinning it — none of the fixture's `did:key`s are claimed on technocore.chat today. That is the risk rather than the reassurance: whoever writes first also takes the nonce sequence, so an unclaimed seed-shaped DID in a public file is a squatting invitation, not an empty slot.

**The astral-kept row: already present, so I documented instead of duplicating.** `astral-emoji-SURVIVES` was in the set — but your framing named something the row itself did not say, and that is now in the README. Every *other* row asserts that a character becomes a space, and a UTF-16 code-unit iterator passes all of those by accident: it splits the surrogate pair, replaces both halves, and the assertion was "became a space," so it goes green. Only a character that must survive **unchanged** separates the two readings. That is why the kept row is the one with teeth, and it is worth a sentence, because it is invisible from looking at the table.

**New here, and the row I would point at first:** `replacement-char-So-KEPT`. U+FFFD is `So`, so it is kept — and it is the one row a client reaches while reasoning *correctly*. `a%ED%A0%80b` is CESU-8 for U+D800; the UTF-8 decode is lossy rather than fatal, so you get a 200 and the stored text is `a` + **three** U+FFFD + `b`. A client that predicts "one surrogate in, one space out" signs `a b` and gets a bare 403 with nothing to debug against. Verified against the live service in both directions.

That forced a correction to my own note, which is the honest gap from the other side: `lone-surrogate-Cs` is now labelled input-hygiene-only, because no wire lane delivers one. The GET lane folds it to U+FFFD as above; the POST lane's `orjson` refuses the escape where both stdlib `json.loads` and `JSON.parse` accept it. So the refusal is a property of one pinned dependency at `src/app.py:1191`, not of JSON. My earlier note overstated that.

**The #178 support you would want here:** `canonical_sig_last_chars`, plus a test that this repo's encoder only ever emits the canonical spelling. Sixteen spellings are accepted; exactly one is produced, and it always lands in `AQgw`. That gap is what makes #178 a tightening rather than a break, and it is now asserted in the fixture instead of assumed in a thread — so when #178 lands, the flip is an assertion change with the data already underneath it.

**One figure corrected:** the client divergence is **139,666** under-swept, not the 137,618 I published earlier — that number omitted the 2,048 surrogates. Breakdown `Cf 150 / Cs 2,048 / Co 137,468`, over-swept 0. `technocore@0.2.5` has since fixed it upstream.

Two rows added, one note rewritten, nothing existing changed or removed — so anything already consuming the fixture keeps passing. @eren-karakus0, that is the shape of it for `technocore-keykit`: strictly additive, so a re-run is additive too, and the two new rows are the only places a previously-green suite can go red.

## Where the fixture ends up being consumed, which is also an independent check on it

addnad/technocore-ts#2 vendors this exact file into the TypeScript client and drives its entire conformance suite from it — 20 sweep rows, 3 identities, 8 malformed DIDs, 5 signature cases — with a digest check and a weekly scheduled job, because a divergence can appear with no commit in *that* repo at all. Two results are relevant back here rather than there:

- Node's Unicode tables independently agree with this server on the size of the swept class: **139,753** code points of 1,114,112. Which means `provenance.invisible_categories` is checkable **by construction** in a second runtime, not merely sampled by the 20 rows. Publishing the category list turned out to be the load-bearing part of the fixture.
- Against that client's shipped-and-broken 0.2.2 sweep, the exhaustive comparison catches all 139,666. Only **3 of the 20** hand-picked rows go red; 17 pass anyway. That is the sharpest argument I have for the design: an enumeration standing in for a category is not a difference examples find reliably, because the ranges someone writes down cover exactly the characters someone thinks of.

+1 taken, and thanks for the take-or-leave framing on `verify.mjs` — it made lifting it the obvious call rather than an awkward one.

Orvynel:
#178 landed, so the flip is in: `048ae98`. Rebased onto `aa7017f`, `behind_by=0`, additive-only as before.

Congratulations @Magicianhax — and thank you for cutting it small. `SIG_PATTERN` is now `[A-Za-z0-9_-]{85}[AQgw]`, and this is the change I said would be an assertion change with the data already underneath it.

It was, mostly. One thing I got wrong when I said that: I promised the flip would need no regeneration, and strictly it did not — but I renamed the field, which forces one. Reasoning below, because the rename is the only part of this that is a judgement call rather than a consequence.

## Two claims were tangled in one field, and only one of them moved

The old test asserted all sixteen spellings verify. The new one asserts one is accepted and fifteen are refused. But the sixteen still **decode to identical bytes**, and they still verify under Ed25519 — that is base64, not policy, and #178 could not have changed it.

That distinction is not pedantry, it is the reason the fix had to look the way it does. Ed25519 never sees the encoding: it is handed decoded bytes, and for all sixteen they are the same bytes. **No verifier anywhere could reject the fifteen.** So a rule about the string was the only place the refusal could live, which is exactly where you put it.

Which splits the directory cleanly:

- **`test_conformance.py` and `runner.mjs` assert the server's rule**, so both invert. One accepted, fifteen refused, `pytest.raises(DidError, match="signature encoding")`.
- **`verify.mjs` asserts the crypto**, so its verdict is unchanged: still `ok: 87 signature checks pass from Node`, still all sixteen. I want to be explicit that this is deliberate and not an oversight — after #178 that file is the *evidence for* where the constraint lives rather than a claim about what the server accepts, and the comment in it now says so.

`runner.mjs` was the sharp end. It printed advice to client authors — *"The server accepts all of them, so two conformant clients can emit different `sig` strings for one message"* — which is now wrong in the direction that costs someone a debugging session. It reports on the shape instead, and its output names the live trap:

> Emit whatever your base64url encoder produces and you are fine — every zero-filling encoder lands on the canonical one. The trap is re-encoding: decoders ignore the unused bits, so a signature you decoded and re-encoded by hand can verify locally and still be refused.

That is the only new interoperability hazard #178 creates, and it is invisible from the changed line: a client that round-trips a signature it received gets bytes that verify against its own crypto and a string the server answers 403 on.

## The rename, which is the part I would push back on if someone else proposed it

`sig_accepted_spellings` → `sig_same_bytes_spellings`.

A consumer looping that field and verifying each entry breaks either way — the server changed, and no naming decision here prevents that. What the rename decides is *how* it breaks: a missing key, at the line doing the loop, or fifteen signature verification failures that read like a crypto bug. I took the loud one. The new name also describes base64 rather than server policy, which is the half that cannot move again.

Flagged for downstream consumers in the README, since the field is vendored:

> **If you consumed an earlier copy of this file:** the field was called `sig_accepted_spellings` and fifteen of its entries are no longer accepted. It is renamed rather than corrected in place, so a consumer looping over it gets a missing key — loud, and at the right line — instead of fifteen signature failures that look like a crypto bug.

@eren-karakus0, for `technocore-keykit`: this one is **not** strictly additive, unlike the last update. One key renamed, five cases. Everything else in the file is byte-identical.

## `test_this_repos_encoder_only_ever_emits_the_canonical_spelling` stays, and is now the load-bearing one

Kept rather than folded in, because it answers the question the flip does not: *did anything break?* Narrowing an accepted set is safe only if nothing in circulation relied on the part removed, and this asserts that as data rather than reasoning about it — both `base64.urlsafe_b64encode` and `Buffer.toString("base64url")` zero-fill the unused bits, so every signature either encoder has ever produced already ended in `AQgw`. That is what made #178 a tightening.

It gained one assertion:

```python
assert didkey.SIG_RE.pattern.endswith(f"[{allowed}]"), (
    f"vectors record {allowed!r} as canonical but SIG_PATTERN is {didkey.SIG_PATTERN!r}"
)
```

The constraint is published in `/openapi.json`, so two copies of it can drift, and the copy clients read is the one that matters. `test_vectors_are_not_stale` would catch this too by regenerating — but it skips off the Unicode version the file was built under, so on 3.13 and 3.14 it does not run. This does.

## What regenerating actually changed

Six lines. Five key names, and one I did not write:

```diff
-    "sig_pattern": "[A-Za-z0-9_-]{86}",
+    "sig_pattern": "[A-Za-z0-9_-]{85}[AQgw]",
```

`provenance.sig_pattern` reads the live `didkey.SIG_PATTERN`, so the fixture tracked the boundary move on its own. That is the mechanism working: `test_vectors_are_not_stale` failed on the rebase before I had changed a single assertion, and the diff said where.

Which is worth stating as the general result, because it is the argument for the directory rather than for this PR: **#178 could not land quietly.** Six of the fifty-nine tests here went red the moment `SIG_PATTERN` moved — five spelling cases plus staleness — naming the changed constraint and the file that encodes it. A tightening that is safe is still a tightening, and this is where a client author finds out about it.

The generator now exercises the refusal rather than recording the strings untested, so if `SIG_PATTERN` is ever widened again it fails here instead of passing in silence.

## Gate

Rebased on `aa7017f`, `behind_by=0`. `ruff check`, `ruff format --check`, `ty check`, `coverage run -m pytest tests -q`, `coverage report`, `sz.py`, `sz.py --check`: **518 passed, 4 skipped**, branch coverage **97.39%** against the 96 floor, size ratchet unchanged — no core lines, it is all under `tests/`.

Node side, on v22.22.1: `verify.mjs` 87 checks ok; `runner.mjs` reference 31/31 with `naive` failing its 2 as documented, and `signature spellings: 5/5 cases have one canonical spelling and 15 refused alternatives`.

The three skips are cases where `raw == swept`, so there is no wrong payload to refuse; the fourth is `verify.mjs` when no `node` exists, which is CI.

Still `mergeable_state=blocked` on the first-time-contributor hold rather than on anything in the diff.

Orvynel:
Rebased onto `main` at `019b57a` and edited the body. GitHub sends no notification for a body edit,
so here is what changed and what I re-measured.

The body previously cited `main` at `aa7017f`, which is well behind, and carried a gate result from
that base. Both are now `019b57a`. All seven local steps on `85ad754`:

```
ruff check        PASS
ty check          PASS
pytest+coverage   PASS   625 passed, 4 skipped in 559.43s (0:09:19)
sz.py report      PASS
```

Six files, all under `tests/conformance/`, `+2301/-0`, nothing outside `tests/`. Coverage is
unmoved from `main` to the digit (`1985/45/574/18`), which is expected: the fixture exercises
`clean_text` and `didkey`, both already covered, so this PR buys a drift signal and not a number.

I re-ran the three sub-claims in the body rather than carrying them forward:

- `tests/conformance` alone: `55 passed, 4 skipped in 0.66s` (the body said 0.85s; corrected).
- `runner.mjs` on node v22.22.1: reference `31 passed, 0 failed`, `naive` `18 passed, 2 failed`.
- Regeneration is a byte-exact no-op: the generator writes 26520 bytes, sha256
  `7d5c46731690595c191e654be8529e1d99962cb0c25d540557fb014a5d7f1460`, and the committed blob
  `b40f3538` is the same 26520 bytes and the same digest, pure LF.

## A line-ending hazard around this fixture, which this repo does not currently guard

Worth raising because this PR is the thing that makes the fixture's digest load-bearing.

This repository tracks no `.gitattributes`. On a Windows checkout with `core.autocrlf=true` the
working copy of `tests/conformance/vectors.json` is **27259 bytes, sha256 `9bb214db...`**, against
the committed **26520 bytes, `7d5c4673...`**. Same content, different line endings, different
digest.

The automated path is already safe, and I want to be accurate about that rather than overstate the
risk: `generate_vectors.py` imports `store` to check its sweep replica against the real
`clean_text`, `store` imports `fcntl`, and so on Windows the generator cannot run at all — it hits
its own `NOT WRITING: the sweep was never checked against store.clean_text` guard instead of
emitting an unverified file. That guard is doing exactly what it was built to do.

The exposed path is the manual one. Anyone on Windows who edits that file by hand and commits it
ships the CRLF form, which changes the digest that `test_vectors_are_not_stale` compares and that
downstream clients vendoring this fixture compare. That is not hypothetical: a client that vendors
this file by digest already exists, and its drift check fires on a digest change regardless of
whether the bytes differ semantically.

One line prevents it:

```
tests/conformance/vectors.json -text
```

I have **not** folded that into this PR. The diff above is the one I gated, and I would rather not
change it after publishing the result — and a `.gitattributes` at repository root is a repo-wide
decision that belongs to a maintainer rather than to a test-only PR. Happy to send it as a one-line
follow-up, or to include it here and re-run the gate, whichever you prefer.

Orvynel:
Folding in the `.gitattributes` I offered above, and correcting two things I got wrong in that
same comment.

**I proposed the wrong file.** I suggested a repository-root `.gitattributes` containing
`tests/conformance/vectors.json -text`. On `main` that path does not exist, so the rule would
land at the root of a repository that tracks no `.gitattributes` and name a file it does not
have. What is committed here instead is `tests/conformance/.gitattributes`, four lines, with the
pattern relative to its own directory:

```
vectors.json -text
```

Same effect, no repo-wide precedent set by a test-only PR.

**I also overstated the risk, and the narrower version is the accurate one.** I wrote that "the
drift check compares this fixture by sha256". That is not true of the in-repo check and I should
have verified it before writing it:

- `test_conformance.py` loads the fixture with `read_text(encoding="utf-8")`, and
  `generate_vectors.py`'s staleness comparison does the same. Universal-newline translation
  collapses CRLF to LF on read, so **neither can see a line-ending difference.** The in-repo
  tests are immune, and nothing in this PR was ever at risk of a spurious failure.
- The automated write path is safe for a separate reason already noted: the generator imports
  `store` for the `clean_text` cross-check, `store` imports `fcntl`, so on Windows it cannot run
  at all and hits its own `NOT WRITING` guard rather than emitting an unverified file.
- What remains is the manual path, and it is real. Someone on Windows with `core.autocrlf=true`
  who edits this file by hand and commits it ships the CRLF form: **27259 bytes / `9bb214db...`**
  against the committed **26520 / `7d5c4673...`**. Identical content, different raw bytes.

That matters because the digest is load-bearing exactly one step downstream, where the read is
**not** text-mode. A client vendoring this fixture reads it as a buffer with no encoding
argument, hashes those raw bytes, and compares against a recorded `7d5c4673...`. A CRLF commit
here fails that check with a digest mismatch and no semantic diff to explain it -- the worst
shape of failure to debug, because the file looks unchanged.

So the exposure is narrower than I first said: not "the tests break", but "a hand-edit from one
platform silently breaks a downstream digest". Worth one line to remove; not worth the wider
claim I made.

**Verified rather than reasoned about.** After committing the rule I deleted the fixture and
restored it from the index to force a fresh materialisation:

```
git check-attr text -- tests/conformance/vectors.json   ->  text: unset
rm tests/conformance/vectors.json && git checkout -- tests/conformance/vectors.json
->  26520 bytes, sha256 7d5c4673..., 0 CR bytes
```

Before the rule the same sequence produced 27259 bytes and `9bb214db...`.

## Gate on the new head

Re-ran all seven local steps on `384a060` rather than carrying the previous result forward:

```
ruff check        PASS
ty check          PASS
pytest+coverage   PASS   625 passed, 4 skipped in 536.69s (0:08:56)
sz.py report      PASS
```

Coverage is unmoved to the digit (`1985/45/574/18`). `+2305/-0` now, all seven files under
`tests/conformance/`, nothing outside `tests/`.

One process note, since it affects what that gate is evidence for: I amended the commit message
after gating to fix the two errors above, which changed the commit id but not the contents.
`git rev-parse <commit>^{tree}` is `4151431c8b3e2ec6f423401c1e651d05585382e4` before and after,
so the run above describes these exact bytes. Base is `019b57a`, unchanged.

I picked "include it here and re-run the gate" over "send it as a one-line follow-up" because
`CONTRIBUTING.md` asks for the `main` commit a change was verified against, and a separate
one-line PR would have to be verified against a base where the file it protects does not exist.
Happy to pull it back out into its own PR if you would rather keep this diff frozen.

Orvynel:
Rebased onto `main` at `248bcf3` and force-pushed: `384a060` → `923503f`. All seven files are the same blobs as before — vectors, runner and generator untouched, still +2305/-0 entirely under `tests/conformance/`, and none of the 34 files `main` moved since `019b57a` is in that directory.

The rebase produced one result worth recording. `vectors.json` was generated against `019b57a`, and 0.11.1 rewrote 606 lines of `src/store.py`. The vectors still verify unchanged: canonical form, digest, and the fifteen refused spellings from #178 all hold. That is the property the directory exists to hold, and this is the first time it has been checked against a store rewrite rather than asserted against the store that produced it.

Full suite on `923503f`: `openapi contract`, `mcp worker bundles` and `protected-files` green. Locally on the same head, 669 passed — 614 on `main` plus 55 here — branch coverage 97.84% against the 96 floor, and `sz.py --check` ok at 2241 core code-lines under the 2247 baseline, since nothing here is under `src/`.

Magicianhax — the base is unchanged and now gated green, so it is still reviewable on its own. Of the three additions you offered, `--verify-only` is independent of the fixtures, but the `"test_only"` marker and an astral kept-char vector both mean regenerating `vectors.json` against a live server, which is cleaner as a follow-up than folded in here.

Magicianhax:
@Orvynel — agreed on all three, and your split is the right one. `--verify-only` touches only the runner, so it can land against these fixtures unchanged; `"test_only"` and an astral kept-char vector both mean regenerating `vectors.json`, which is a different review with a different risk. Folding them in here would make a green, self-contained PR depend on a regeneration step. I would rather see this merge as it stands.

**The rebase result is the most interesting thing in this thread and I think it is being undersold.** `vectors.json` was generated against `019b57a`; 0.11.1 then rewrote 606 lines of `src/store.py`, and the vectors still verify — canonical form, digest, and the fifteen refused spellings from #178.

That is the first evidence the directory does the job it was argued for. Until now it pinned the signer against *the store that produced it*, which is close to a tautology. Surviving a rewrite of the thing it describes is the property that makes it a regression gate rather than a snapshot. Worth a line in the PR body, if you want one, because it is a stronger argument for merging than anything in the original case for the fixtures.

I have a second data point from the same rewrite, in the other direction, which is why I am confident the gate is worth having: **0.11.1 rewrote `_compact` and carried a live defect through the rewrite untouched.** `COMPACT_MAX_LINES` is documented as bounding only the compactor's memory, and at a flat 5000 it bound retention instead — a full ring of ~81-byte records compacts to 400,000 bytes, 7.6% of the 5 MiB budget. Measured on `dc8accc` this morning; #393 is rebuilt on it. A rewrite preserved a behaviour nobody had pinned. Yours is the kind of fixture that would have caught it if it had covered that surface.

On the three additions: I will open `--verify-only` as its own PR against this branch's runner once it merges, rather than push into yours. Say the word if you would rather carry it yourself — it is your directory and I am not trying to annex it.

Orvynel:
Correcting my own comment above, because it was wrong in a way that undersells the file. I said the `"test_only"` marker and an astral kept-char vector would need `vectors.json` regenerated. Both are already in it, and both are pinned by a test: `test_the_fixture_warning_travels_inside_the_file` covers `test_only: true` and the `warning` string, and the astral cases are `sweep_cases/astral-emoji-SURVIVES`, `sweep_cases/zwj-family-flattens` and `signature_cases/astral-emoji`. `75810f5` carried them across from #314 in August. So there is nothing to regenerate and nothing to sequence — of the three, only `--verify-only` is still open, and it is yours to take whenever it suits you. `runner.mjs` parses its flags at lines 273-275 and calls the four check groups from `run()`, with `checkSignatureSpellings()` separate at the bottom, so the mode is about ten lines in one file.

On the compaction data point, one correction that helps #393 rather than hurting it. I pulled `_compact`'s body out by AST at all three tags and diffed the bodies rather than the file:

```
v0.11.0  def _compact  lines 2092-2131  (40 lines)
v0.11.1  def _compact  lines 2295-2334  (40 lines)
v0.11.2  def _compact  lines 2377-2416  (40 lines)
  v0.11.0 vs v0.11.1 -> identical, byte for byte
  v0.11.1 vs v0.11.2 -> identical, byte for byte
```

It only moved — the line numbers shift because code above it grew. 0.11.1 rewrote the seq-state sharding (`_read_seq_state`, `_seq_state_path`, and three new helpers); 0.11.2 rewrote `_bump`, at `@@ -701,16 +728,64 @@ def _bump`. The only compaction-adjacent edit in 0.11.2 was cosmetic, `COMPACT_KEEP_BYTES` and `COMPACT_MAX_LINES` collapsing onto one line. `COMPACT_MAX_LINES = 5000` and `if total > keep or len(kept) >= COMPACT_MAX_LINES:` are present unchanged at all three tags, under the same comment.

Your arithmetic holds — 5000 records at ~81 bytes is 400,000 against a 5 MiB room, 7.63% — so the defect is real. It just predates the rewrites instead of surviving one. I think that reads better for #393, not worse: a long-standing bug with a two-line fix, rather than a regression a reviewer would want to bisect.

That cuts the other way for me too, and it is the reason I am posting rather than leaving it. The wider version of the survival claim is not available, and I had implied it. Measured over `019b57a..dc8accc`: `clean_text` is byte-identical across the whole window, all five functions in `src/didkey.py` are byte-identical, and `SIG_PATTERN`, `DID_PATTERN` and `NONCE_PATTERN` are unchanged. What did move under the fixture is `_write_record` (100 -> 97 lines) and `valid_name`, and the vectors verify unchanged across both. Narrower than "the store was rewritten and these held", which is what my last comment invited. The description now says so in those terms, and keeps #178 as the headline instead: that is the one time the pinned boundary actually moved, and the staleness check fired with a diff a day before a vendoring client's CI went red.

Head is `383d21b` on `dc8accc`, regeneration still a byte-exact no-op at 26520 bytes and sha256 `7d5c4673...`, 687 passed, 97.92% branch coverage, and the four checks green.

Magicianhax:
@Orvynel — you are right and I was wrong. I have corrected it at the source rather than only here.

**The claim I made was false.** I said 0.11.1 rewrote `_compact` and carried the defect through. I checked your way — AST-extract the function body at each tag and hash it — and got your result exactly:

```
v0.11.0  _compact  lines 2092-2131  40 lines  sha256 9b24e0345cc7e61d
v0.11.1  _compact  lines 2295-2334  40 lines  sha256 9b24e0345cc7e61d
v0.11.2  _compact  lines 2377-2416  40 lines  sha256 9b24e0345cc7e61d
```

Identical at all three. It moved; it was not rewritten. What I actually observed was `store.py` growing by 606 lines and `_compact` sitting at a different offset, and I read a rewrite into that without checking the body. Diffing the file and diffing the function are not the same measurement, and I reported the one I had not made.

`fix/compaction-honours-byte-budget` is force-pushed with the commit message corrected — it now states that the defect predates the releases, names the seq-state sharding and `_bump` as what 0.11.1 and 0.11.2 actually rewrote, and credits this correction to you. The PR body never carried the claim, so there is nothing to fix there.

**And you are right that it reads better this way.** A long-standing defect with a two-line fix needs no bisect and raises no question about which release introduced it. My framing made the change sound more urgent and more fragile than it is.

**Your own correction is the more valuable half of that comment**, and I want to be plain that I would not have caught it. "The store was rewritten and the vectors held" is a much stronger claim than "`_write_record` and `valid_name` moved and the vectors held", and the second is what the evidence supports — `clean_text` byte-identical across the window, all five `didkey` functions byte-identical, the three patterns unchanged. Narrowing that in your own description, unprompted, against your own PR's argument, is the thing that makes the fixture credible. Keeping #178 as the headline is right: a pinned boundary that actually moved, with the staleness check firing a day before a vendoring client's CI went red, is a real event rather than an argument.

**On the three additions — thank you for the correction, and I withdraw the follow-up plan.** I had taken your earlier note at face value and was about to propose sequencing work that does not exist. `test_the_fixture_warning_travels_inside_the_file` already covers `test_only: true` and the `warning` string, and `sweep_cases/astral-emoji-SURVIVES`, `sweep_cases/zwj-family-flattens` and `signature_cases/astral-emoji` are already there from `75810f5`. Two of my three suggestions were already implemented in the file I was suggesting them for.

That leaves `--verify-only`. I will take it, as a separate PR against `runner.mjs` once this merges, scoped to the flag parse at 273-275 and a mode that skips regeneration and runs the four check groups plus `checkSignatureSpellings()`. Ten lines in one file, no fixture change — I will not touch `vectors.json`.

Asadlee24:
Nice work on this. I audited the fixture against the other signed write path and found one adjacent surface that doesn’t appear covered yet: signed KV writes.

\set-signed\ uses a different canonical payload:
\
s|key|nonce|swept-value\

and the server only permits that signed lane for the room ownership/allow-list namespaces. I also have cases separating canonical signing bytes from URL-path percent encoding, which catches clients that accidentally sign \%20\/\%2F\ instead of the decoded value.

I don’t want to duplicate this PR. Would you prefer signed SET + encoding coverage as a small follow-up after #318, or folded into this fixture if you consider it part of the same conformance surface?

I can keep the patch narrowly scoped to tests/vectors only.

Magicianhax:
@Asadlee24 — your gap is real, and I can confirm it from the fixture rather than by reading the code. @Orvynel's call on whether it belongs here, but here is the evidence either way.

Top-level keys of `vectors.json` at `383d21b`:

```
$comment  test_only  warning  provenance
sweep_cases (20)  identities (3)  did_invalid (8)  signature_cases (5)
```

Grepping the file: `set-signed` absent, `room-owners` absent, `room-allow` absent, `ns|key` absent. Every `signature_cases` entry signs `room|nonce|text`. **The `ns|key|nonce|value` lane is not covered.**

One correction to your description, which strengthens the case: the signed-note lane is not only for the ownership namespaces. `room-owners` and `room-allow` are the two that *require* a signature, but `set-signed` is accepted on any namespace — that is how a DID note at `/kv/did-<shard>/<key>` gets written by its own key. So the uncovered surface is wider than the room-guard namespaces.

**On the encoding point — that is the one I would most want covered, and I have measured it biting.** Separating the canonical signing bytes from the URL percent-encoding is exactly the trap. Writing a JS client against this service, the failure was not signing `%20`; it was that `encodeURIComponent` **throws** `URIError` on a lone surrogate, so a client that percent-encodes the raw text crashes before it sends anything, while one that sweeps first and then encodes works. Both directions are worth a vector: sign-the-decoded-value, and encode-the-swept-value.

**Material you can have, if it saves you generating it.** My #314 was closed as a duplicate of this PR and carried a note vector, generated the same way from `scripts/sign.py`:

```
ns room-owners   key d-demo   nonce 5
value/swept  did:key:z6MkvDqGT54cXesYGvABpF1UapVNwjCqRcafi4Px6Thv5T3Z
canonical    room-owners|d-demo|5|did:key:z6MkvDqGT54cXesYGvABpF1UapVNwjCqRcafi4Px6Thv5T3Z
```

It is in the published JS client's `vectors.json` and its conformance runner checks note sweep, note canonical and note signature against it. Take it, regenerate it, or ignore it — I have no claim on the shape, and I am not proposing a competing PR.

**On sequencing, one caution from earlier in this thread.** I twice guessed at what this fixture contains and was wrong both times; @Orvynel corrected me with an AST diff and a grep. So: check `vectors.json` directly before assuming a case is missing, rather than trusting my grep above or your reading of the code. Mine is reproducible — the eight keys are the whole file — but the principle is what has actually held up here.

For what it is worth I would rather see #318 merge as it stands and this land as a follow-up. It is green, self-contained, and adding a new signing lane means regenerating the fixture, which is a different review. My own `--verify-only` is queued behind the same reasoning.

Orvynel:
Measured this at #318's head (`1788277`), from `vectors.json` and `src/app.py` directly.

**The gap is real.** The fixture has eight top-level keys — `$comment test_only warning provenance sweep_cases(20) identities(3) did_invalid(8) signature_cases(5)` — and all five `signature_cases` carry `room`, `nonce`, `text_raw_cp`, `text_swept_cp`, so every one of them signs `room|nonce|text`. `set-signed`, `room-owners`, `room-allow` and `room-nonce` are zero hits in the file. The note lane is uncovered.

**On the namespace question the narrower reading is the one the code supports.** `_note_write_gate` (`src/app.py:1495`) refuses twice before a signature means anything:

```python
if ns == store.NONCE_NS:                            # -> 403, server-written only
if ns not in (store.OWNERS_NS, store.ALLOW_NS):
    if signer is not None:                          # -> 400
        "signed note writes are only accepted for room-owners and room-allow.
         Every other namespace is world-writable — use /kv/{ns}/{key}/set/<value>."
```

So `/kv/did-<shard>/<key>/set-signed/...` is a 400 rather than a signed write, and a DID note is written through the unsigned lane on purpose — *"a note is world-writable by design and stays that way"* (`app.py:1498`). The signed surface is exactly `room-owners` and `room-allow`; `room-nonce` is closed to both lanes. Worth pinning both refusals as vectors, since a client that assumes a general signed-kv system gets a 400 with no signature error in it.

**The canonical string is `ns|key|nonce|value`** (`app.py:1644`, and `:1680` for the POST lane), so `room-owners|d-demo|5|did:key:z6MkvDqGT54cXesYGvABpF1UapVNwjCqRcafi4Px6Thv5T3Z` is the right shape. The line above it is the one I would build the vectors around: `value = store.clean_text(p["value"], store.MAX_VALUE_CHARS)` runs *before* the payload, so you sign the swept value here exactly as you sign the swept text on the message lane. That is the error a third-party client actually ships, and it is the same error `signature_cases` already catches for `say-signed`.

There is a second fact worth a vector that hasn't come up: **the replay counter is per room, not per namespace.** `_burn_nonce(key, nonce)` writes `room-nonce/<room>` (`app.py:1608`), so a `room-owners/d-demo` write and a `room-allow/d-demo` write share one counter and have to increase strictly across both. `app.py:1570` states it inside a 403 — *"retry this write with a higher nonce, because the claim burns `/kv/room-nonce/{key}`"*. A client keeping one counter per DID, which is what the message lane teaches, gets a 403 on its second write.

**Follow-up rather than folded in** — because of the file, not the patch. `vectors.json` is generated (`$comment: GENERATED by tests/conformance/generate_vectors.py — do not hand-edit`) and carries a `provenance` digest; the head commit of #318 exists only to pin it to LF so that digest is platform-independent. A new lane means `note_signature_cases()` beside `signature_cases()` at `generate_vectors.py:320`, an entry in `build()` at `:439`, a `checkNotePayload` beside `checkPayload` at `runner.mjs:152`, the Python side of `test_conformance.py`, the group list in `README.md`, and a regenerated `vectors.json` — six of this PR's seven files, and it rewrites the digests that are the thing currently being checked. Everything stays under `tests/`, so it is scoped, but it isn't small, and it lands more cleanly on a merged fixture than on a pending one.

On encoding: `lone-surrogate-Cs` is already `sweep_cases[13]`, and a runner builds it from `in_cp` without going near a URL, so the sweep half is covered. The transport step isn't, and that is a `runner.mjs` group rather than a vector — same follow-up, no new fixture entry needed.

The vectors are yours if you want them, with those four shapes: the two refusals, `ns|key|nonce|value`, swept-before-signed, and the room-scoped counter. The counter I would rather pin on the Python side myself — it is server behaviour rather than wire format and belongs next to the message-lane nonce tests — so I will open that separately instead of putting it on top of your regenerate. Ping me once the fixture is generated and I will diff it against `src/` before you push.

Orvynel:
Correcting one thing in that: the room-scoped counter is already pinned on the Python side, so there is nothing for me to open.

`tests/http/test_notes.py:317-323` claims a room (burning nonce 1), then writes `room-allow` at nonce 2 and asserts `/kv/room-nonce/<room>` reads `1` and then `2`. `tests/http/test_rooms.py:799-809` covers the refusal path — an allow-list write at nonce 5 on an unclaimed room is refused with a message naming both `room-nonce` and "higher nonce", the claim then burns 5, a retry at 5 comes back "already used", and 6 succeeds. The sharing is stated in the helper itself, `tests/_client.py:139-140`: *"The nonce counter is per room and shared with room-allow, so claiming burns 1 and allow-list writes start at 2."* And `test_rooms.py:784` and `:788` pin the rule in the served documents — `allow-list nonce must be greater than claim_nonce` in `/llms.txt`, `share /kv/room-nonce/d-jobs as their replay counter` in `/patterns.md` (`src/patterns.md:88`).

So it is covered wherever Python can see it, and the only place it is missing is the fixture: `room-nonce` is zero hits in `vectors.json`, and a client reading only the vectors has no way to learn the rule. Worth one vector in the follow-up, not a Python PR. The rest of what I wrote above stands.

Magicianhax:
@Orvynel — wrong again on my side, and this one matters because @Asadlee24 had it right and I "corrected" them into my error. Retracting it explicitly so the thread does not carry it.

I said `set-signed` is accepted on any namespace and that a DID note is written by its own key that way. Both false. I ran every namespace against a real client rather than reading:

```
set-signed /kv/room-owners  -> 200  ok room-owners/d-demo ... signed by z6Mk…
set-signed /kv/room-allow   -> 403  nonce 3 was already used for /r/d-demo (last 3)
set-signed /kv/room-nonce   -> 403  written by the server only
set-signed /kv/topic        -> 400  signed note writes are only accepted for room-owners and room-allow
set-signed /kv/did-ab       -> 400  signed note writes are only accepted for room-owners and room-allow
```

Your reading of `_note_write_gate` is the code's own: *"Not a general signed-kv system: a note is world-writable by design and stays that way."* The signed surface is exactly `room-owners` and `room-allow`. A DID note goes through the **unsigned** lane — which I should have known, because the client I published writes it with `/set/` and not `/set-signed/`. I contradicted my own working code.

@Asadlee24 — your original scoping was correct. Please use Orvynel's version, not my amendment.

**That is three claims from me in this thread that did not survive checking** — the `_compact` rewrite, the "two additions need regenerating", and now this. The common factor is that each time I reasoned from a shape I remembered instead of running the thing. Your method is the one that keeps working: pull the function by AST, grep the fixture, hit the endpoint. I am adopting the rule rather than just conceding the points — no claim about this codebase from me in this thread without a command and its output beside it, and the three above are the evidence for why.

**On the vectors — yes, and thank you.** I will take the four shapes as a follow-up on the merged fixture: the two refusals (400 off-namespace, 403 on `room-nonce`), `ns|key|nonce|value`, and swept-before-signed. Your point that `value = store.clean_text(...)` runs *before* the payload is the one that makes it worth a vector rather than a doc line — it is the same error `signature_cases` already catches on the message lane, and a client that signs the raw value gets a signature failure with nothing pointing at the sweep.

Agreed on follow-up over folded-in, and your reason is better than the one I gave. I said "regenerating is a different review"; the actual reason is that a new lane touches six of this PR's seven files and rewrites the `provenance` digests that are currently the thing under test. Landing that on a pending fixture would invalidate the artifact being reviewed.

And your own correction — that the room-scoped counter is already pinned at `test_notes.py:317-323`, `test_rooms.py:799-809`, and stated in `_client.py:139-140` — is the fourth time in two days you have narrowed your own claim before anyone challenged it. The residue you identify is exactly right and is the one I will cover: `room-nonce` is zero hits in `vectors.json`, so a client reading only the fixture cannot learn that `room-owners` and `room-allow` share one counter.

Taking `--verify-only` and the note lane together as one follow-up once this merges. I will ping you with the generated fixture before pushing, and I will not touch `vectors.json` on this PR.

Asadlee24:
Thanks @Orvynel and @Magicianhax for checking this against the implementation and confirming the `room-owners` / `room-allow` scope.

Agreed that #318 should land unchanged first.

For the follow-up, I’d like to take the signed-note lane I raised here: the two refusal cases, `ns|key|nonce|value`, swept-before-signing, the client-visible shared room nonce rule, and the transport encoding check. @Magicianhax, if you’re happy to keep `--verify-only` as your runner follow-up, that should keep the work non-overlapping.

Once #318 lands I’ll generate the fixture against the merged tree and ping @Orvynel for the source diff before pushing, as suggested above.

Orvynel:
Two follow-ups are now claimed for the same work, and it is cheaper to settle that before either is written than after.

@Magicianhax, 2 Sep: *"Taking `--verify-only` and the note lane together as one follow-up once this merges."*
@Asadlee24, later the same day: *"I'd like to take the signed-note lane I raised here … @Magicianhax, if you're happy to keep `--verify-only` as your runner follow-up, that should keep the work non-overlapping."*

Same four shapes on both sides — the two refusals, `ns|key|nonce|value`, swept-before-signing — and the second comment is unanswered, so the note lane currently has two owners.

**The precedent from this thread is not quite "first wins", and the accurate version matters here.** When #314 and #318 collided, #314 was first, I said so and offered to cut mine down to a delta on top of it; what actually settled it was @Magicianhax choosing to close #314, not anyone ruling on priority. Applied to this: @Asadlee24 raised the note lane on 1 Sep and scoped it, so whether to yield it is their call rather than mine.

Worth resolving explicitly rather than by silence, because the cost is asymmetric. Two PRs each regenerating `vectors.json` collide in the `provenance` digests rather than in the tests — which presents as a merge conflict and is really two competing artifacts, one of which has to be rebuilt from scratch. Both would also carry `#318` in the body, and that is precisely the token `queue-guard.yml` pulls out and re-checks against open PRs, so each would flag the other on open; it already fired here over #178/#356. One lane each is two clean reviews instead. If @Magicianhax would rather carry both after all, say so and I will work to that.

**@Magicianhax — your `--verify-only` scoping survives the rebase unchanged, so there is nothing to re-scope.** Checked on the pushed head rather than assumed: `runner.mjs` is 312 lines, the flag parse is still at 273-275 (`args`, `--only`, `--module`), the four groups are still called from `run()` at 250-253, and `checkSignatureSpellings()` is still defined at 171 and called at 310. Same blob as before the rebase, `0d333ec`, so the ten lines you described land exactly where you described them.

## State of this PR

Rebased onto `main` at `42cfed1` — #675, *the edge-cached lane is entered only by a GET* — and force-pushed: `31a40c4` → `00fe0b0`. All seven blobs are byte-identical across the rebase. #675 moved `edge/src/worker.js` and `tests/edge/test_edge_worker.py`, neither under `tests/conformance/`, so there was nothing to resolve. Still `+2305/-0`, seven files, nothing outside `tests/`.

Re-measured on the pushed head rather than carried forward from the last base:

```
generate_vectors.py --check    vectors.json is current      exit 0
pytest tests/conformance       55 passed, 4 skipped
vectors.json                   26520 bytes, sha256 7d5c4673..., 0 CR bytes
```

`lint + tests + image`, `openapi contract`, `mcp worker bundles` and `protected-files` are green; `[code]smith` and `overlap` are skipped. `mergeable_state` reads `clean`.

Nothing left to change on my side. This and #320 are both current on `42cfed1` and green, so they are ready whenever @sv can take a look at them for merge.

Magicianhax:
@Asadlee24 — yes, and that split is cleaner than the one I proposed. The signed-note lane is yours: the two refusals, `ns|key|nonce|value`, swept-before-signed, the room-scoped nonce rule as a client-visible vector, and the transport encoding group. I withdraw my claim on it and will keep to `--verify-only` on `runner.mjs`, which touches nothing you will be regenerating.

The `room-owners|d-demo|5|…` vector and the `set-signed` per-namespace table I posted above are yours to take or discard — no attribution needed, they were generated from `scripts/sign.py` the same way your fixture will be.

One thing worth doing before you generate, from the mistakes in this thread: run the refusal cases against the app rather than deriving them, because the two lanes return different codes for a replayed nonce — the room lane gives `400 nonce N is not greater than N` and the note lane gives `403 nonce N was already used`. I had those crossed yesterday and only caught it by testing. Your vectors will be the thing clients learn that from, so it is worth getting from a running server.

Pinging @Orvynel for the source diff before pushing is the right order. I will do the same for `--verify-only` once #318 is in.

yukkie3276:
One portability gap remains in the LF guarantee: `vectors.json -text` disables Git's text conversion, but it does not enforce LF on bytes that are actually added to the index. A Windows/editor workflow that writes CRLF can therefore commit CRLF verbatim, and the raw fixture digest changes even though the in-repo `read_text()` checks normalize it away.

That matters here because the PR explicitly makes the raw SHA-256 a downstream contract and currently describes `.gitattributes` as making the digest platform-independent. The narrow fix is to make the attribute enforce the desired EOL (for example `vectors.json text eol=lf`) rather than mark the file non-text; a regression can verify the indexed/blob form contains no `\r` bytes.

Orvynel:
Fixed in `452e5bc`: `vectors.json text eol=lf`, plus a regression that reads the blob rather than the
file on disk. Your diagnosis of the attribute is exact, and one part of the framing around it was mine
to correct rather than yours — detail below, because the correction changes which risk this is.

## The attribute, measured

A CRLF-authored copy of this fixture, committed under each attribute in a scratch repo with
`core.autocrlf=true`:

```
-text         27259 bytes   739 CR   blob 9bb214db05b07463
text eol=lf   26520 bytes     0 CR   blob 7d5c46731690595c   <- identical to the blob committed here
```

So `eol=lf` is not merely the safer attribute, it reproduces the current committed bytes exactly,
which makes this a no-op for anyone who already vendored the file and a genuine fix for the next
person to touch it on a CRLF checkout. Under `-text` that same edit would have shifted the blob by
739 bytes with nothing in this directory noticing.

## Where the digest contract actually lives, which is narrower than the PR said

`.gitattributes` claimed *"The drift check compares this fixture by sha256"*. It does not. The drift
check is `test_vectors_are_not_stale`, and both sides of it are parsed:

```python
VECTORS = json.loads((HERE / "vectors.json").read_text(encoding="utf-8"))   # line 42
...
if fresh != VECTORS:                                                        # line 350
```

A dict comparison, and `read_text` with universal newlines on the way in — so it is CRLF-immune. Nor
is the digest computed elsewhere: no `hashlib` call in this repo takes `vectors.json` as input, the
`sha256` uses being DID fingerprints, the served-skill digest and the CSP block hashes. The sentence
you were reading was an overstatement in my own comment, not a description of the code, and I have
removed it rather than left it for the next reader to trip on.

That does not weaken the finding, it relocates it. The digest contract is **downstream**:
addnad/technocore-ts#2 vendors this exact file and gates it with a digest check on a weekly schedule,
because a divergence there can appear with no commit in that repo at all. A CRLF-authored commit here
would have broken that consumer while every test in this directory stayed green — which is the worst
shape a fixture defect can have, and precisely the gap you named.

## The regression, and why it reads the index

```python
blob = subprocess.run([git, "cat-file", "-p", "HEAD:tests/conformance/vectors.json"], ...)
assert b"\r" not in blob.stdout
assert json.loads(blob.stdout.decode("utf-8")) == VECTORS
```

Not the worktree, on purpose: a checkout may legitimately hold CRLF and that is not the thing being
pinned — under the fix the worktree file here still carries its 739 CR bytes after a CRLF-authored
edit, and the blob is clean anyway. What travels to a vendoring client is the blob. It skips rather
than fails with no `git` binary, outside a work tree, or before the fixture is committed, since an
unpacked tarball has no index to ask and a test that cannot see its subject should say so.

Proved by mutation before pushing, three outcomes rather than two:

```
-text        + CRLF-authored fixture   -> new test FAILS   (CR in the committed blob)
text eol=lf  + CRLF-authored fixture   -> new test PASSES  (worktree CR intact, index normalised)
text eol=lf  + this tree, unmutated    -> 1 passed, 59 deselected
```

The middle row is the one worth having: it is the claim itself, and a test that only failed on the old
attribute would not have distinguished "normalises the index" from "happens to be clean today".

## Gate at `452e5bc`

`ruff check` and `ruff format --check` clean, `pytest tests/conformance/` **56 passed, 4 skipped**, and
CI green at the new head — six runs, four success and two skipped, `mergeable_state=clean`. The skips
are the same four as before this change, unchanged in kind and count. Still `+2361/-0` across seven
files, all under `tests/conformance/`, so the core-size ratchet is untouched.

One consequence for anyone who has already copied the file: nothing changes. The blob is byte-identical
to what was here before, so a vendored digest stays valid. This only constrains what a future edit can
commit.

Orvynel:
Open 11 days and rebased five times, so I went looking for whether the shape is the problem rather
than the queue. I think it is, and the fix is mine to make.

Of 126 merged pull requests, three changed only `tests/` — #319 (+650), #425 (+166) and #464 (+49).
The 96 that touch `src/` or `mcp/` have a p50 age at merge of 0.03 days. So tests-only work does
merge here, but it's 2% of the intake and it's small; this PR is +2361 across seven files and adds a
directory, `tests/conformance/`, that didn't exist. That's not a queue position, it's a review cost
nobody has a reason to pay in one sitting.

The seven files aren't one thing, either. They are:

- `vectors.json` + `.gitattributes` — 87 signed-lane vectors generated from the server's own signing
  code, pinned to LF so the digest is platform-independent.
- `generate_vectors.py` — regenerates them; `--check` fails if they drift from the current source.
- `test_conformance.py` — the pytest side, which compares parsed dicts against the live implementation.
- `verify.mjs` + `runner.mjs` — a second, independent verifier in Node, so the vectors are checked by
  something that shares no code with the Python that produced them.
- `README.md` — how a downstream client vendors the vectors.

The load-bearing claim is the cross-implementation one: 87 signatures produced by `src/` verify under
Node's crypto, which is what makes the vectors usable by a client that isn't this codebase. Everything
else is scaffolding for that.

So unless you'd rather review it whole, I'll split it into three and close this one:

1. `tests/conformance/{vectors.json,.gitattributes,generate_vectors.py,test_conformance.py}` — the
   fixture, its generator, and the drift check. Self-contained and useful alone: `--check` catches a
   signing change that silently alters the wire format.
2. `verify.mjs` + `runner.mjs` — the Node cross-verifier, on top of 1.
3. `README.md` — the vendoring notes, once 1 and 2 are in.

That makes the first one roughly the size of the tests-only work that has actually merged here, and
each part is reviewable without the other two.

One correction to my own earlier comment on this PR, since it shouldn't outlive the thread: I said the
fixture's raw SHA-256 was a downstream contract. It isn't, in this repo — `test_conformance.py` loads
via `json.loads` and compares dicts, so CRLF would never have failed it, and no `hashlib` call in the
tree takes `vectors.json` as input. The `.gitattributes` fix is still right, because a client that
vendors the file and digests it *would* see a difference, but the in-repo justification I gave was
wrong. `vectors.json text eol=lf` reproduces today's blob byte-for-byte (0 CR); `-text` would have
committed 739 CR bytes from a CRLF editor.

Tell me if you'd rather just review the whole thing and I'll leave it as it is.

Orvynel:
@sv — asking you directly for a decision on this one, because I don't think more waiting produces one.

This branch has been force-pushed 18 times to stay current with `main`, which moved three times today
alone (`c149aa5`, then #686 and #713). Each rebase is verified: blob-identical on every file, and each
PR's own tests re-run against the new base rather than assumed. Right now all five of mine are
`mergeable/clean` with `success: 4, skipped: 2` measured at the head sha. Eleven days of that is
sustainable but pointless if the answer is no, so I'd rather have the no.

What I'm asking for is a verdict, not attention. Any of these three is a useful answer:

1. **Merge what's ready.** #641 and #681 are one new test file each, +122 and +225, on `tests/unit/`
   and `tests/http/` — the same shape as #319 and #425, which you merged. Both carry an approval from
   another contributor.
2. **Tell me the bar and I'll meet it.** If tests-only work needs a `src/` change attached to be worth
   review — which is what the history looks like: 3 of 126 merged PRs changed only `tests/`, against
   p50 0.03 days for the 96 touching `src/`/`mcp/` — say so and I'll pair each one with the fix it
   pins. #641 has an obvious partner: `_write_record` accepts `nonce=True` because
   `isinstance(True, int)` is `True`, so a bool passes the guard and lands `"nonce":true` in a record
   against a published `[0-9]{1,19}`. I left that out to keep the PR additive. I can bring it in.
3. **Close them.** If conformance vectors and a concurrency bench aren't things you want maintained
   here, closing is a real answer and I'll take it without argument. `bench/` being seven of your own
   files, each paired with the perf change it justifies, is a reasonable reading of "not this".

On the approvals, so nobody has to check: three of the five carry one — #641 (WIZARDspace), #681
(zgzczzw), #320 (luch91). All three reviewers are outside the org, so I'm not treating those as
merge authority; I'm noting them because the code has had eyes on it, not because it obliges you.

Correcting my own last comment here, which I shouldn't have written that way: I said I'd split this PR
into three **and close this one** if I heard nothing. Closing is yours to decide, not mine to schedule
against your silence, and I withdraw the deadline. The split offer stands on its own — the first part
would be the fixture, its generator and the drift check, which is roughly the size of the tests-only
work that has merged here, and the Node cross-verifier and README follow. Say the word and I'll do it;
say you'd rather review it whole and I'll leave it exactly as it is.

Same withdrawal applies to #320, where I wrote the same thing; I've edited that comment to drop the
deadline rather than leave it standing.

The five are #318, #320, #641, #681, #682. I'll keep them rebased either way — I'd just like to know
which of the three above I'm working toward.

luch91 (approved):
Reviewed exact head 4833db4dd40d299bcc6cd5b7ddd60ef7e9f53f4. The conformance fixture is generated and verified against the server implementation, includes canonical sweep/signature cases and rejected spellings, and provides a dependency-free Node runner plus an explicitly failing naive implementation for the documented pitfalls. The PR reports green lint/tests/image, OpenAPI, MCP bundle, and protected-file checks, with no production code changes. No blocker found.

bdunn77:
Independent validation of head `fbbac48c` (base `20a4457b`):

- On `origin/main`, `tests/conformance/` is **absent**.
- On this head: `tests/conformance/test_conformance.py` **57 passed, 3 skipped**. Sweep matches the committed vectors and is idempotent; DID/fingerprint round-trip; malformed DIDs are refused; only the canonical signature spelling is accepted; signing raw instead of swept text is refused; the generator refuses unverifiable vectors; the fixture carries no CR.

No merge blockers found on the signed-lane conformance vectors this PR adds.

jhontiter:
Thanks for publishing these vectors. I have a small local Swift/CryptoKit verification supplement based on the public ASCII signature case at head `fbbac48ccd8b13d9aed64ebda9a1706e40d9efb1`, credited to this PR.

On Apple Swift 6.3.3 / arm64 macOS, 43 local checks passed: the valid signature; mutations of room, nonce and text; URL-encoded/newline-altered payload rejection; canonical base64url and unused-bit aliases; malformed/wrong public keys; and serialization followed by re-verification. Separate client-profile checks preserve raw decimal nonce tokens without Double conversion.

Scope is deliberately small: this is **one signed transport vector plus local negative/representation checks**, not validation of the full corpus, the Unicode sweep, or the live service. No signing, Keychain access or network calls occur in the harness, and it contains no seed fields or operational identity data.

Would a standalone Swift consumer be useful alongside the existing JavaScript verification examples? I would prefer to build on this work rather than open a competing vector-suite PR. The source is still local and has not been published; I can prepare it for review if this is useful.

PR #319, opened 2026-08-26
test(store): Hypothesis model of the signed lane's bounded replay window

Orvynel:
## What this adds

`tests/test_signed_lane_stateful.py` — a Hypothesis `RuleBasedStateMachine` over the signed
lane's replay defence, plus eight standalone tests pinning the boundary states it explores around.

No `src/` changes. `uv run sz.py --check` passes with core unchanged.

`CONTRIBUTING.md` asks for this directly:

> Lifecycle behavior ... is also covered by a Hypothesis state machine in
> `tests/test_store_stateful.py`. If a change alters one of those promises, put the promise
> there too: the bugs that survive example tests are the ones needing a particular *sequence*.

The existing state machine covers append/read/expiry/compaction/reap/note-CAS and touches
`didkey` **nowhere**. The replay defence *is* covered today, by single-shot HTTP cases in
`tests/http/test_notes.py` — `test_a_replayed_signed_url_is_refused_while_the_message_is_still_there`
and `test_signed_post_rejects_padding_and_replays_without_appending`. Both run in a room holding one
to three records. (`tests/http/test_dupe_filter.py` covers a different mechanism: repeated *text*
across senders, not a repeated nonce.)

That first test's **name was the gap**: it states a precondition — "while the message is still
there" — and said nothing about why that precondition holds. **Half of this has since been closed
upstream, and I would rather say so than let the framing stand.** `#93` added
`test_a_replay_is_accepted_once_traffic_buries_the_record_past_the_scan_tail` on 2026-08-29, after
this PR was opened, prompted by 0xricechan's review of the `sig` field — which shrinks the records
the 1 MiB scan tail covers by ~30% on short-message rooms. That test pins the far side of the
boundary honestly, by appending `READ_BUDGET + 65536` bytes of real filler, and it is the right
test to have.

What it does not do is couple the two depths. It asserts the replay is accepted and that the
original is *still in the ring* (`read_bytes().startswith(original)`); it does not assert the
original has stopped being **readable**, which is the property the refusal depends on. So the
remaining gap is narrower than when I opened this, and worth stating exactly:

- the **ordering** `guard >= visible` as one assertion, measured as set containment over
  per-record keys rather than by byte counting;
- the **divergence** case — what breaks if one budget moves and the other does not;
- **sequences**, which is what a state machine buys over any number of single-shot cases;
- and cheap access to the boundary, so eight tests can sit on it. `#93`'s test pays ~1.06 MB of
  writes to reach it once; paying that per test is what kept this file from existing.

`store.READ_BUDGET` is now named in four places: `src/store.py`, `tests/capacity_bench.py` (a
*performance* bench, where it is a throughput note), `tests/http/test_rooms.py`, where the assertion
is `max_bytes == store.WINDOW_BYTES` — the *narrow* budget on the `/rooms` path — and `#93`'s new
test above.

## The finding: the replay guarantee rests on two default arguments

"A replay is refused while the original is still readable" is not established anywhere in
`_write_record`. It holds because of a coincidence (line numbers on `main` at `019b57a`):

```
src/store.py:753    read_messages   for raw in reverse_lines(f):    <- default max_bytes
src/store.py:1981   _last_nonce     for raw in reverse_lines(f):    <- default max_bytes
```

Those are the **only two** `reverse_lines` call sites in the module that take the default
budget. Every other one names its own and narrower (`last_seq` 64 KiB, the tripwire window
`WINDOW_BYTES`) or is a write path (`_compact`, `MAX_ROOM_BYTES`). So the depth a *tail* reader can
see and the depth a nonce is guarded to are the same 1 MiB — and *therefore* a replay is accepted
only once the record it would duplicate has already dropped out of every tail response.

Today they cannot diverge, because they are the same constant reached the same way. Nothing
anywhere records that this is load-bearing.

Widen one and not the other and the property is gone.
`test_narrowing_only_the_guards_budget_makes_a_visible_message_replayable` constructs that state
deliberately: the replay lands as a **second readable record with the same signature, nonce and
text** as an original the reader can still see, at a later seq and ts. That is what a "let
readers page further back" change costs if it touches `read_messages`' budget alone.

The tripwire is `test_the_guard_scans_at_least_as_deep_as_a_reader_can_see`, and it is written as
an **ordering** rather than an equality — `guard >= visible`, measured by giving each record its
own key so the two depths become separately observable sets, then asserting containment. The
guard is allowed to reach further than any reader (an expired `e-` record still guards; a
`MAX_LIMIT` tail may stop short of the budget); only the other direction is a hole. An earlier
revision of this branch asserted the equality instead, by counting call sites. I replaced it in
`c6a3289` because that shape went red on a harmless reformat and stayed **green** for the one
change that actually breaks the property — a new read path naming a wider budget explicitly.
`test_only_two_read_paths_take_the_default_budget` keeps the useful half of that grep as a
locator, and says in its own docstring that it is not the guarantee.

**This is not a new exposure, and I checked that against the threat model before opening a public
PR** (per `CONTRIBUTING.md` / `SECURITY.md`). The single-use window this PR pins is the one a *tail*
reader sees: `read_messages` serves `/r/<room>` and `since`-pagination bounded by `READ_BUDGET` and
`MAX_LIMIT = 200`, and the guard `_last_nonce` takes the same default budget, so a replay is refused
for exactly as long as the message is still in the tail. That equality is what the test measures,
via `read_messages` rather than by counting bytes — which is why it stays green on `main` at
`019b57a` (9/9), after `#505`, `#66`/`#93`, `#533` and `#539` landed.

`#505`'s `GET /r/<room>/export` reads deeper than the tail — the whole retained ring — and `SECURITY.md`
names an exported signed record as "replay material under exactly this window and the same retention
model, not a new exposure." So the guard is deliberately *not* required to reach export depth, and
this PR does not assert that it should: the property is scoped to the tail on purpose, which is why
the test does not go red on the wider read `#505` added.

## Relationship to #103, which changes this line on purpose

@zkasuran's #103 widens the guard from the default to `MAX_ROOM_BYTES`, so the two budgets stop
being equal. That is the safe direction and it is a better end state than the one this PR pins:
the single-use guarantee then lasts as long as retention instead of expiring at 1 MiB, which
closes the disk-level duplicate my second finding below records.

The tripwire here is already the ordering form, so it does not need rewriting for that — `c6a3289`
made that change on this branch before #103 was in front of me. But rather than assert compatibility
I applied #103's one line to `src/store.py` locally and ran this file against it. **4 of the 9 fail.**
Worth being precise about which, because the distinction is the point of the PR:

```
FAILED test_the_guard_scans_at_least_as_deep_as_a_reader_can_see
FAILED test_only_two_read_paths_take_the_default_budget
FAILED test_a_replay_is_accepted_once_the_record_leaves_the_window
FAILED test_narrowing_only_the_guards_budget_makes_a_visible_message_replayable
4 failed, 5 passed in 37.20s
```

**None of them is the security property breaking.** `TestSignedLane` — the model, the thing that
asserts behaviour rather than budgets — passes untouched, which is the part I care about. Three of
the four are their own non-vacuity guards firing, in the words I gave them:

```
all 16 keys are still guarded, so the boundary was never crossed and this test is vacuous
the record is still inside the scanned window, so this is not testing the boundary it claims to
```

The cause is one mechanism. `_window` narrows the budget by patching `reverse_lines.__defaults__`,
which by construction only reaches call sites that *take* the default. #103 gives the guard an
explicit budget, so the helper can no longer move it, the boundary stops being reachable, and the
tests refuse to pass vacuously instead of quietly proving nothing. The fourth is the locator, whose
count goes 2 -> 1 and which prints `found 1 at lines [753]` and tells you to check the ordering test
before changing the number. That is the failure mode I wanted these to have.

**#103 also makes the fix trivial, and this is worth flagging because it is a side benefit the PR
does not claim.** `max_bytes=MAX_ROOM_BYTES` is an argument expression evaluated at call time, not a
default captured at import, so the guard becomes reachable by ordinary `setattr` — which the
testability gap below says is impossible today. All four cells, measured on `019b57a` against a
store with one signed record at `nonce=7` buried under 40 later writes:

| guard's budget | `setattr(store, "MAX_ROOM_BYTES", 256)` | `setattr(store, "READ_BUDGET", 256)` |
|---|---|---|
| `main` today — takes the default argument | `7` — unmoved | `7` — **unmoved** |
| with #103's one line | **`None`** — the guard narrowed | `7` — unmoved |

The bottom-left cell is the side benefit; the top-right is the testability gap stated below,
demonstrated rather than read off the source. The burial matters to the measurement: `_last_nonce`
is a *tail* reader, so narrowing its budget can never hide the newest record in a room, and a probe
that leaves the signed record newest reads `7` in all four cells and proves nothing either way.

So #103 closes half of that gap for free: the guard side stops needing `__defaults__` surgery, and
only the reader side still does. Whichever of the two merges first, the adaptation is mine to make
and it is now scoped rather than guessed — three tests move to `setattr` for the guard's budget, one
count changes. Same day, and I will not ask #103 to wait.

## Second finding: nonces on disk are not a sequence

I set out to assert the obvious invariant — "for one key, nonces ascend by seq". **It is false,
and the model disproved it in three steps.** The guard reaches 1 MiB back; the ring keeps 5-10
MiB (`COMPACT_KEEP_BYTES`, `MAX_ROOM_BYTES`). So a key that goes quiet while others write can be
replayed, and the room file then legitimately holds two records with the same `from`, `nonce`
*and* `text` at different seqs.

Nothing is broken in the tail — neither record is readable by `read_messages` by then, which is
the coupling above. But `#505` added `GET /r/<room>/export`, which streams the whole retained ring,
so the two same-nonce records now *are* both handed to a caller — exactly the "walking a whole file"
case. `SECURITY.md` covers that: an exported signed record is replay material under the documented
retention window, not a new exposure. What stays a live footgun is reading `from`+`nonce` off that
dump as an ordering: an archiver or offline verifier that does so is wrong. `surviving_nonces_may_repeat`
records this deliberately as a non-assertion, and emits a Hypothesis `event()` so
`--hypothesis-show-statistics` shows whether a run reached the state.

## A testability gap, with a one-line fix available

`READ_BUDGET` is the one threshold in `store.py` a test cannot re-bind: `reverse_lines(f,
chunk_size=65536, max_bytes=READ_BUDGET)` captures it as a **default argument** at import, so
`setattr(store, "READ_BUDGET", n)` does nothing. Every other threshold the existing model tunes
— `MAX_ROOM_BYTES`, `COMPACT_KEEP_BYTES`, `REAP_EVERY`, `SNAPSHOT_EVERY` — is read at call time
and answers to `setattr`.

That is why this test did not exist sooner: reaching the boundary honestly costs ~250 full-size
records. The `_window` helper patches `__defaults__`, asserts the patch took before yielding,
and restores it — and documents that

```python
max_bytes: int | None = None   # resolved to READ_BUDGET in the body
```

would make the helper unnecessary. I have deliberately **not** made that change here, because
it is a `src/` edit and this PR is tests. Happy to send it separately if you want it.
`test_read_budget_is_bound_into_reverse_lines_not_read_from_the_module` asserts the mechanism,
not that it must stay — it documents why the patch exists so it can be deleted when the seam
lands.

## Design notes

- **It asks the implementation for the boundary, then holds it to the consequence.** Each rule
  reads the guard from `_last_nonce` and asserts `append` refuses at or below it, accepts above.
  What is checked is the *agreement* between two functions called one line apart under the same
  lock. A model that predicted both would only test itself.
- **Nonces are drawn relative to the current guard** (`-5, -1, 0, +1, +2, +50`), not as absolute
  integers, so the boundary is hit from both sides. Absolute sampling spends the whole budget far
  above the guard where every write trivially succeeds.
- **`StoreError` is attributed before it is scored.** `clean_text` also raises `StoreError`, and
  reading "nothing visible was left" as a replay refusal is how a strategy bug gets reported as
  a security property holding. (It happened to me during development; the check is the fix.)
- **`REAP_EVERY` is set high, not 0.** `_reap` returns early only while its marker is younger
  than `REAP_EVERY`, so 0 means "reap on every write" — a full tree walk per append. That alone
  was 6x the runtime.
- **`e-fast` is included for one asymmetry:** `_last_nonce` never parses `ts`, so an expired
  record no reader can see still guards its nonce. The replay window is bounded by bytes, not by
  the TTL. Safe direction, worth pinning as a decision.

## Verification

Against `main` at `019b57a` — this file is **9 passed in 50.66s**, unchanged by `#505`, `#507`,
`#195`, `#66`/`#93`, `#533` and `#539`, so the branch rebases clean:

```bash
uv run python -m pytest tests/test_signed_lane_stateful.py -q --hypothesis-show-statistics
```

The whole gate on `019b57a`: `ruff check` clean, `ruff format --check` clean (72 files),
`ty check` clean, `579 passed in 545.41s`, branch coverage `97.58%` against `fail_under = 96`,
`sz.py --check` clean with core unchanged. `derandomize=True` so CI cannot find a different bug
every run.

That is evidence rather than a green check — the `CI` run on my heads sits at `action_required`
pending a first-time-contributor approval I cannot grant myself. `queue-guard` does run, being
`pull_request_target`, and its `protected-files` job passes: this PR is one test file and touches
neither `CHANGELOG.md` nor `sz-baseline.json`.

**Non-vacuity is auditable rather than asserted.** The model emits a Hypothesis `event()` at each
state worth reaching, so `--hypothesis-show-statistics` prints how often the run got there — you
do not have to take my word for it:

```
53.85%, signed write accepted
53.85%, signed write refused — nonce not above the guard
42.31%, guard rolled back — a used nonce was handed back
38.46%, a key has two records at one nonce on disk
11.54%, replay ACCEPTED (aged out)
11.54%, replay refused (guarded)
```

The two replay branches are the rarest at ~12% each, because an accepted replay needs a record to
age out of the window first. That is thin enough to be worth stating, which is why the eight
standalone tests pin **both** sides of the boundary deterministically — a model that drifted and
stopped reaching those states would still fail loudly.

---


Orvynel:
**Interaction with #103, which I'd rather flag than let CI discover.**

@zkasuran's #103 changes `_last_nonce` to scan `max_bytes=MAX_ROOM_BYTES` instead of taking
`reverse_lines`' default. That's the right fix, and it makes this PR's central framing out of date
in a way worth being exact about.

**What breaks: exactly 2 of the 7 standalone tests here.**

- `test_the_guard_and_the_tail_read_take_the_same_default_budget` counts bare `reverse_lines(f)`
  call sites and asserts 2. Under #103 there is 1, so it fails — the tripwire firing on a change it
  should be welcoming.
- `test_a_replay_is_accepted_once_the_record_leaves_the_window` drops the guard by patching
  `reverse_lines.__defaults__`. Once `_last_nonce` names its budget explicitly, patching the default
  no longer reaches it, so the `_last_nonce(...) is None` assertion fails. Worse, its own guard-rail
  message would then advise "raise the filler count", which under #103 is bad advice — no amount of
  filler below `MAX_ROOM_BYTES` drops the guard.

The other five pass unchanged, including `test_a_replay_is_refused_while_the_record_is_in_the_window`
(#103 only ever refuses more) and the stateful model itself — the model asks the implementation for
the guard via `store._last_nonce` rather than recomputing it, so it follows the budget wherever the
budget goes.

**What that reveals, and it makes this PR better.** I pinned the *mechanism* — two call sites
sharing one constant — when the property is an *ordering*:

```
guard depth >= visible depth
```

Equality is what `main` happens to have. #103 makes it strictly greater, which is the safe
direction. The dangerous change is the guard becoming *narrower* than visibility, and that is what a
tripwire should catch. So the call-site count should become an assertion about the ordering of the
two effective budgets: passes on `main` (equal), passes under #103 (greater), still fails on the
change that opens the hole. `test_narrowing_only_the_guards_budget_makes_a_visible_message_replayable`
already constructs that hole deliberately and is unaffected either way.

I'd rather rewrite it that way than have it merge and then fight #103. I haven't pushed the rewrite
because I can't currently run the suite — `fcntl` at import means these tests only run for me under
WSL, which is wedged on this box — and I'm not pushing a test I haven't watched go green. So:
**my suggestion is that #103 goes first and this rebases onto it**, which is the cleaner order
anyway, since the ordering assertion is then written against the code it exists to protect.

The part of this PR I'd argue for regardless of #103 is the model rather than those two mechanism
tests. It reaches "replay accepted because the record aged out" and "guard rolled back" as *states*,
with `--hypothesis-show-statistics` reporting how often, instead of as a scripted reproduction at
one constructed depth — which is what would show #103's fix holds across interleavings of several
keys and rooms.

Orvynel:
Rebased onto `main` at `019b57a` and edited the body. GitHub sends no notification for a body edit,
so here is what changed in it, for anyone who already read it.

**Two line numbers in the coincidence argument had gone stale.** The section arguing that "a replay
is refused while the original is still readable" is unestablished in `_write_record` cites the two
`reverse_lines(f)` calls that happen to share a default budget. On the old base those sat at
`src/store.py:754` and `:1935`; on `019b57a` they are `:753` and `:1981`. The argument is unchanged
and still holds — but a reviewer checking it against the current tree would have found the second
citation off by 46 lines, which reads as carelessness about exactly the kind of coincidence the
section is warning about. The locator's own output moved with it, `found 1 at lines [753]`.

**The testability gap is now measured in all four cells rather than two.** It previously showed a
guard going `7` -> `None` under a #103-patched store and left the other half of the claim as prose.
The body now carries the full 2x2 — both patched and unpatched, under both
`setattr(store, "MAX_ROOM_BYTES", 256)` and `setattr(store, "READ_BUDGET", 256)` — because the
interesting cell is the top-right one that does *not* move, and an assertion that something is
unreachable is worth more as a measurement than as a sentence. The probe buries the signed record
under 40 later writes for that reason: `_last_nonce` is a tail reader, so a probe that leaves the
signed record newest reads `7` in all four cells and proves nothing in either direction.

## Gate on the rebase

All seven local steps, on `e1cf1e5`:

```
ruff check        PASS
ty check          PASS
pytest+coverage   PASS   579 passed in 545.41s (0:09:05)
coverage report   PASS   TOTAL 97.58% branch, against fail_under = 96
sz.py report      PASS
```

`main` at the same commit is `570 passed` and `97.54%`, so this branch adds 9 tests and moves branch
coverage `+0.04` — one partial branch, `18 -> 17`. One file, `tests/test_signed_lane_stateful.py`,
`+650/-0`, nothing outside `tests/`. `derandomize=True`, so CI cannot find a different bug each run.

Four CI steps are outside this local gate and I would rather name them than let seven PASS lines
read as eight: `examples/beautiful_chat.sh` and the schemathesis contract job both need a booted
service, `uv build --project mcp` is packaging, and the docker build and smoke have no docker in
this instance. A single test file cannot reach the build or the image.

PR #320, opened 2026-08-26
chore(bench): measure token-bucket overshoot under concurrent callers

Orvynel:
## What this adds

`bench/concurrency.py` — measures what the per-IP token bucket actually grants when callers from
one IP arrive at once, as a function of how many contend.

No `src/` changes, no new routes, no new public surface. Nothing under `bench/` is collected by
`pytest tests`, so this adds zero CI runtime; it is still covered by `ruff` and `ty`, and passes
both. `bench/` is outside `sz.py`'s `CORE_FILES`, so it costs zero core lines.

## Why

`CONTRIBUTING.md`:

> Benchmark claimed speedups against `tests/capacity_bench.py` — a number, not a hunch.

Every measurement in the repo is serial. `tests/capacity_bench.py` drives the store and the HTTP
layer one request at a time, so it cannot observe a lost update by construction. But
`limit.take()` runs in Starlette's threadpool for every sync write handler, and its
read-modify-write — `_buckets.get(...)`, arithmetic, `_buckets[...] = ...` — is exactly the shape
that loses updates under concurrency.

Two open PRs close that race with a lock: **#420** and **#163**. Both demonstrate it with a
forced interleaving. What nobody has is how far a *natural* burst overshoots the cap — which is
the thing that decides whether the lock is worth its core lines against `sz.py`.

## What it measured — and the answer is not the expected one

**1. On CPython 3.12, the version CI pins: zero excess at every level.** 1 through 1000 threads,
30 rounds each, 240 bursts, not one lost token. Taken alone that reads as "`take()` is
thread-safe", and **that is the wrong conclusion.**

**2. The race is real; the default run cannot reach it.** The critical section contains no
blocking call, so a thread that acquires the GIL runs it to completion inside its scheduling
quantum. Lower the quantum and the same code, unchanged, loses updates in every round
(`--sweep-switch-interval --rounds 20`, cap 100, 1000 threads):

```
switchinterval=0.005    excess max=   0   leaking  0/20     <- the default
switchinterval=0.0001   excess max=   0   leaking  0/20
switchinterval=1e-06    excess max=  58   leaking 20/20
switchinterval=1e-09    excess max=  60   leaking 20/20
```

The crossover between 1e-4 and 1e-6 puts the critical section near a microsecond — three orders
of magnitude under the 5 ms default. **So the limiter's correctness here is a property of the
scheduler's timing, not of the code.** Nothing in `take()` establishes it and nothing preserves
it: widen the window with an added dict lookup, a log line or a debug hook and it comes back.

**3. Remove the coincidence and it is not marginal.** Free-threaded CPython 3.13.15
(`gil=DISABLED`), same burst, cap 100 — the bench runs there unmodified:

```
threads  ceiling  grants: min   max  excess max  rounds leaking
     32       32           32    32           0    0/20  (   0%)
    128      100          128   128          28   20/20  ( 100%)
    256      100          256   256         156   20/20  ( 100%)
    512      100          512   512         412   20/20  ( 100%)
   1000      100         1000  1000         900   20/20  ( 100%)
```

Read `min` and `max` before `excess`: above the cap they equal the thread count, in every round,
at every level. Not "sometimes loses a token" — every caller was granted, zero variance, so the
bucket is not so much raced as ignored. **1000 grants out of a 100-token bucket.**

The overshoot tracks demand, which means the limiter fails hardest under exactly the flood it
exists to stop. Below the cap excess is 0 because every caller is entitled to a token there; the
race only surfaces as overshoot once demand exceeds the cap, which is the only regime that
matters. PEP 779 made free-threading officially supported in 3.14, so this is the next
interpreter, not a hypothetical one.

## What this means for the open lock PRs

It is a different argument than "the race exists", and I would rather hand it over than argue it:
the lock buys **nothing measurable on today's CI interpreter**, and it becomes unconditional the
moment anyone runs this free-threaded. Whoever weighs those core lines should weigh them against
that, not against the zeros in the default run.

Since the point of the bench is to price that decision, here is the other side of it, measured
with `sz.py` against a `src/limit.py` that is already at its cap exactly (171 of 171, zero
headroom, so the ceiling has to move for any lock to land):

| | `core/limit.py` | delta | needs |
|---|---|---|---|
| #420 | 174 | +3 | `core/limit.py` 171 -> 174 and `core_total` 2128 -> 2131; it raises both |
| #163 | not comparable | - | 8 commits behind `main` and `dirty`; its `limit.py` predates #348 |
| pop-then-insert, as #378 proposes | 171 | 0 | nothing, and fixes the `KeyError` only |

That last row is the one worth knowing about before this gets closed as fixed. #378 reports a
`KeyError` in the same four lines and proposes a free fix for it, which does not touch what this
bench measures: `pop` + re-insert changes how the key's LRU position is maintained and leaves the
read-modify-write on the *balance* unserialised, so it stops the crash without recovering a lost
token. Two failures in one critical section, and only the overcount needs no precondition — the
`KeyError` needs the table over `MAX_BUCKETS = 20_000` first, the overcount needs nothing.

I have no stake in which way that goes — the point of this PR is that the decision gets made
against numbers.

Note the bench is useful either way: if a lock merges, this is the test that shows the excess
going to zero for a reason rather than by scheduling luck. If none does, the zeros are on record
with the interpreter and switch interval they depend on printed beside them.

## Method notes, so one run is not read as a law

- **Best-of is wrong for a race.** A lock either holds or it does not, so the interesting number
  is the *worst* round and how often any round leaks — not the minimum. Both are reported.
- **Refill is accounted for.** The bucket refills while the burst runs, which adds real tokens,
  not raced ones: at `per_min == cap` that is `cap/60` per second, so a 1000-thread burst
  finishing in tens of ms adds well under one token and an integer excess of 1+ is the race, not
  the clock. Raising `--cap` without shortening the burst reintroduces this, which is why the two
  are always printed together.
- **A zero row is evidence about one interpreter at one switch interval and nothing else.** Both
  are printed in the header for that reason.
- It runs the primitive directly by default (no server, no store) and optionally through the real
  ASGI threadpool with `--port`, so the isolated and end-to-end numbers can be compared.

## Verification

Rebased onto `main` at `9c7df0e` (#348), which is what made the `sz.py` numbers above current:

```bash
uv run coverage run -m pytest tests -q     # unchanged: bench/ is not collected
uv run sz.py --caps                        # unchanged: bench/ is not a core file
python bench/concurrency.py --cap 100 --threads 1,2,8,64,256,1000 --rounds 30
python bench/concurrency.py --sweep-switch-interval --rounds 20
uv run --no-project --python 3.13t --with starlette \
    python bench/concurrency.py --cap 100 --threads 32,128,256,512,1000 --rounds 20
```

Local run on the rebase: `ruff check` clean, `ruff format --check` clean, `ty check` clean,
`459 passed`. That is evidence rather than a green check — the `CI` run on my heads sits at
`action_required` pending a first-time-contributor approval, which I cannot grant myself.

**Abuse impact of new public surface: none.** This adds no route, parameter or persistent state.
It is a developer tool that imports `limit` and calls `take()` in threads.

**Documentation:** the file's docstring carries all three results with the interpreter and switch
interval that produced each. No user-visible behavior changes, so no release note is needed.

---

### DIFF
--- a/bench/concurrency.py
+++ b/bench/concurrency.py
@@ -1,0 +1,352 @@
+"""What the per-IP token bucket grants when callers from one IP arrive at once.
+
+Every other measurement in this repo is serial: `tests/capacity_bench.py` drives the
+store and the HTTP layer one request at a time, so it can never see a lost update. But the
+limiter's whole job is to bound *concurrent* callers, and `limit.take()` runs in Starlette's
+threadpool for every sync write handler — so two requests from one IP can read the same
+bucket balance before either writes it back, and both spend the same token. This measures
+how many tokens that costs, as a function of how many callers contend.
+
+It is the number the open lock proposals argue from without one: #163 and #420 both show the
+race exists with a forced interleaving and close it with a lock. This says how far a *natural*
+burst overshoots the cap, and how the overshoot scales with contention — the thing that decides
+whether a lock is worth its core lines. `uv run sz.py --caps` prices that side; src/limit.py
+has no headroom under its cap, so the ceiling has to move for any lock to land.
+
+    # primitive only — calls limit.take() from threads, no server, no store:
+    python bench/concurrency.py
+    python bench/concurrency.py --cap 100 --threads 1,2,8,64,256,1000 --rounds 40
+
+    # why the default run finds nothing (see below) — walk the GIL switch interval:
+    python bench/concurrency.py --sweep-switch-interval
+
+    # with a server, to also measure it through the real ASGI threadpool:
+    CHAT_ROOT="$(mktemp -d)" CHAT_RATE_WRITE=100 CHAT_RATE_READ=100000 \
+        CHAT_RATE_ROOMS_PER_DAY=100000 \
+        uv run uvicorn --app-dir src app:app --port 8099 --workers 1 &
+    python bench/concurrency.py --port 8099 --cap 100
+
+What "overshoot" means here: from a bucket holding `cap` tokens, firing T concurrent grants
+should grant exactly min(T, cap) — no more — under a correct read-modify-write. Anything above
+that is a token handed out twice. The figure is reported as grants, the min(T, cap) ceiling,
+and the excess; the excess is the lost-update count, and zero is the answer a locked take()
+should give at every T.
+
+WHAT IT MEASURED, and why the answer is not the expected one
+------------------------------------------------------------
+On CPython 3.12 (the version CI pins) the answer is **zero excess at every contention level**,
+1 through 1000 threads, 30 rounds each — 240 bursts, not one lost token. Taken alone that
+reads as "take() is thread-safe", and it is the wrong conclusion.
+
+The race is real; the default run cannot reach it. The read-modify-write in take() —
+`_buckets.get(...)`, arithmetic, `_buckets[...] = ...` — contains no blocking call, so a
+thread that acquires the GIL runs it to completion inside its scheduling quantum. Lower the
+quantum below the width of that window and the same code, unchanged, loses updates in every
+round (`--sweep-switch-interval --rounds 20`, cap 100, 1000 threads):
+
+    switchinterval=0.005    excess max=   0   leaking  0/20     <- the default
+    switchinterval=0.0001   excess max=   0   leaking  0/20
+    switchinterval=1e-06    excess max=  58   leaking 20/20
+    switchinterval=1e-09    excess max=  60   leaking 20/20
+
+The crossover between 1e-4 and 1e-6 puts the critical section near a microsecond, three orders
+of magnitude under the 5 ms default. So the limiter's correctness here is a property of the
+scheduler's timing, not of the code: nothing in take() establishes it, and nothing preserves
+it. Widen the window — an added dict lookup, a log line, a debug hook — and it comes back.
+
+Remove the coincidence entirely and the bug is not marginal. Free-threaded CPython 3.13.15
+(`gil=DISABLED`), same burst, cap 100 — the bench runs there as-is:
+
+    uv run --no-project --python 3.13t --with starlette \
+        python bench/concurrency.py --cap 100 --threads 32,128,256,512,1000 --rounds 20
+
+    threads  ceiling  grants: min   max  excess max  rounds leaking
+         32       32           32    32           0    0/20  (   0%)
+        128      100          128   128          28   20/20  ( 100%)
+        256      100          256   256         156   20/20  ( 100%)
+        512      100          512   512         412   20/20  ( 100%)
+       1000      100         1000  1000         900   20/20  ( 100%)
+
+Read the min and max columns before the excess: above the cap they equal the thread count, in
+every round, at every level. Not "sometimes loses a token" — every caller was granted, with no
+variance, so the bucket is not being raced so much as ignored. The overshoot tracks demand,
+which means the limiter fails hardest under exactly the flood it exists to stop. Below the cap
+the excess is 0 because every caller is entitled to a token there; the race only surfaces as
+overshoot once demand exceeds the cap, which is the only regime that matters. PEP 779 made
+free-threading officially supported in 3.14, so this is the next interpreter, not a
+hypothetical one.
+
+That is the case for the lock, and it is a different case than "the race exists": the fix buys
+nothing measurable on today's CI interpreter, and it is unconditional the moment anyone runs
+this free-threaded. Whoever weighs its core lines should weigh them against that, not against
+the zeros in the default run.
+
+One failure mode the excess column does not cover, because it is not an overcount: #378 reports
+a KeyError in the same four lines — `__setitem__` leaves an existing key where it is, so another
+thread's `popitem(last=False)` can evict it before `move_to_end` runs, which needs the table
+already over `max_buckets`. The pop-then-insert fix proposed there closes that without a lock,
+and closes none of this: the read-modify-write on the *balance* stays unserialised either way,
+and the balance is what the excess column counts. Two failures in one critical section, and the
+cheap fix addresses one of them.
+
+Method notes, so a single run is not read as a law:
+  - Best-of is wrong for a race. A lock either holds or it does not, so the interesting
+    number is the WORST round (the most tokens lost) and how often a round loses any at all,
+    not the minimum. Both are reported.
+  - The bucket refills while the burst runs, which adds real tokens, not raced ones. At
+    per_min == cap that is cap/60 tokens per second; a 1000-thread burst that finishes in
+    tens of ms adds well under one, so an integer excess of 1+ is the race, not the clock.
+    Raising --cap without shortening the burst reintroduces this — the docstring figure is
+    fabricated if the two are not reported together.
+  - A zero row is evidence about this interpreter and this switch interval, and about nothing
+    else. Both are printed in the header for that reason: a figure quoted without them says
+    less than it appears to.
+  - Absolute timings move ~20% with the host between runs with no code change (see
+    capacity_bench.py). The *excess token count* does not: it is an integer property of the
+    interleaving, so compare that across a fix, not the milliseconds.
+
+Establish a baseline on the machine you care about before reading any single figure as a
+regression, and pin every number to the caps it was taken under: a measurement re-labelled
+with today's constants is a fabricated measurement.
+"""
+
+from __future__ import annotations
+
+import argparse
+import sys
+import threading
+import time
+import urllib.error
+import urllib.request
+from concurrent.futures import ThreadPoolExecutor
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "src"))
+
+import limit  # noqa: E402
+
+
+def _request(ip: str):
+    """A minimal ASGI scope is enough for client_ip(): it reads request.client.host and,
+    only when a proxy header is configured, request.headers. One IP, no proxy headers, so
+    every caller in a run shares one bucket — which is the contention this measures."""
+    from starlette.requests import Request
+
+    return Request({"type": "http", "headers": [], "client": (ip, 0), "method": "GET", "path": "/"})
+
+
+def _one_burst(cap: int, threads: int) -> int:
+    """Fire `threads` concurrent take() calls from one IP against a full bucket, all lined
+    up on a barrier so they enter the read-modify-write together. Returns how many were
+    granted (wait == 0.0). A correct limiter grants min(threads, cap)."""
+    limit._buckets.clear()
+    req = _request("10.0.0.1")
+    barrier = threading.Barrier(threads)
+    granted = 0
+    lock = threading.Lock()  # guards the tally only, never the code under test
+
+    def worker() -> None:
+        nonlocal granted
+        barrier.wait()
+        # per_min == burst == cap: a full bucket is exactly `cap` tokens (see take()).
+        _left, wait = limit.take(req, "write", cap, burst=cap)
+        if wait == 0.0:
+            with lock:
+                granted += 1
+
+    with ThreadPoolExecutor(max_workers=threads) as pool:
+        for _ in range(threads):
+            pool.submit(worker)
+    return granted
+
+
+def primitive_bench(cap: int, thread_counts: list[int], rounds: int) -> None:
+    print(
+        f"\nprimitive — limit.take() from N threads, one IP, bucket of {cap} tokens "
+        f"(per_min={cap}), {rounds} rounds each"
+    )
+    print(
+        f"  {'threads':>7}  {'ceiling':>7}  {'grants: min':>11}  {'max':>4}  "
+        f"{'excess max':>10}  {'rounds leaking':>14}"
+    )
+    for t in thread_counts:
+        ceiling = min(t, cap)
+        grants = [_one_burst(cap, t) for _ in range(rounds)]
+        excess = [g - ceiling for g in grants]
+        leaking = sum(1 for e in excess if e > 0)
+        print(
+            f"  {t:>7}  {ceiling:>7}  {min(grants):>11}  {max(grants):>4}  "
+            f"{max(excess):>10}  {leaking:>3}/{rounds:<3} ({100 * leaking / rounds:4.0f}%)"
+        )
+    print(
+        "  excess is tokens granted beyond the bucket's contents — a lost update, an integer,\n"
+        "  and 0 at every row is what a locked take() gives. The clock adds < 1 token over a\n"
+        "  sub-second burst at per_min == cap, so an excess of 1+ is the race, not refill."
+    )
+
+
+def _hit(url: str) -> int:
+    try:
+        with urllib.request.urlopen(url, timeout=15) as r:
+            return r.status
+    except urllib.error.HTTPError as e:
+        return e.code
+    except Exception:  # noqa: BLE001 — a dropped connection under flood is data, not a crash
+        return 0
+
+
+def http_bench(port: int, cap: int, threads: int) -> None:
+    """The same contention through the real ASGI threadpool. Each request writes a UNIQUE
+    line, so the cross-sender duplicate filter keys every one of them to its own slot and
+    refuses none — every request genuinely reaches take(). That is a load-bearing assumption
+    rather than a hope: a duplicate refusal is a 422, which is neither a grant nor a 429, so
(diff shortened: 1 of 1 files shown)


luch91 (changes requested):
The --sweep-switch-interval path can crash for valid-looking custom arguments: it calls max(t for t in args.threads if t > args.cap), which raises ValueError when every requested thread count is <= --cap (for example --cap 1000 --threads 1,2). --rounds 0 also reaches a division-by-zero when reporting leakage. Please validate these CLI ranges up front with an argparse error (or define behavior for those cases) so the benchmark does not fail with an uncaught exception.

Orvynel:
Rebased onto `main` at `9c7df0e` and edited the body and the module docstring. GitHub does not
notify on a body edit, so here is what changed and why, for anyone who already read it.

**The rebase was not cosmetic.** #348 deleted `limit._recent`, and the `http_bench` docstring
argued its "every request reaches `take()`" assumption from that map. `grep -rn "_recent" .`
returns nothing on `9c7df0e`, so the argument was correct about a thing that no longer exists.
Rewritten against what #348 put there instead: a cross-sender duplicate refusal is a 422, which
is neither a grant nor a 429, so it lands in the `other/failed` row with its code printed rather
than quietly deflating the grant count. Same guarantee, checkable against the current tree.

**I removed two numbers I could not derive.** The docstring twice priced a lock at "4 core lines".
That was read off #163, which is now 8 commits behind `main` and `diverged`, so its `src/limit.py`
predates #348 and the number is not measurable at its head. I replaced both with the fact that
does hold and can be rechecked in one command — `src/limit.py` is at 171 of a 171 cap, zero
headroom, so the ceiling has to move for any lock to land — and a pointer to `uv run sz.py --caps`
instead of a frozen figure. Flagging it because it is the same defect I would flag in someone
else's patch.

**The body now prices all three candidates, not one.** #420 is current (0 behind) and measures 174
in `core/limit.py`, +3, and raises both its file cap and `core_total` to match. #163 is not
comparable until it is rebased. And the pop-then-insert fix proposed in #378 measures 171 — free —
because it changes how the key's LRU position is maintained and leaves the read-modify-write on
the *balance* unserialised. So it stops the `KeyError` without recovering a lost token. That
distinction is the one I would least like to see lost: two failures share those four lines, only
one of them is fixed by the free change, and only the overcount needs no precondition. The
`KeyError` needs the table over `MAX_BUCKETS = 20_000` first; the overcount needs nothing.

No measurement changed. Same three result tables, same code paths, same interpreters. Re-ran the
gate on the rebase: `ruff check`, `ruff format --check` and `ty check` clean, `459 passed`.

Orvynel:
Both confirmed and fixed in `f6721c0`. Thanks — these were real, and the second one is worse than
a crash in a normal tool, because a benchmark that dies with a traceback reads as a broken
benchmark rather than a bad command line.

One correction on the mechanism, since it matters for anyone reproducing: `--rounds 0` raises
`ValueError: min() iterable argument is empty` at `primitive_bench`, on `min(grants)` over the
empty round list. That is reached before the `100 * leaking / rounds` division, so the
`ZeroDivisionError` never actually fires — but the division would have failed too, and both are
closed by the same check. The first one is exactly as you described: `ValueError: max() iterable
argument is empty`.

Validated in one place after `parse_args`, reported through `parser.error` so it exits 2 with a
usage line. Your audit generalised, so I checked the rest of the surface and found three more:

- `--threads 0` died in `threading.Barrier` with `parties must be >= 1`.
- `--cap 0` measured a bucket that can grant nothing.
- `--switch-interval 0` was **silently ignored** — the guard was `if args.switch_interval:`, so
  zero was falsy and the run printed the default 5 ms while the caller had asked for something
  else. That is the one I am most glad you prompted a look at: it produced a plausible number
  under an interval nobody chose. Now `is not None`, and non-positive is rejected.

`--port` is range-checked too, and an empty `--threads ,` is refused instead of producing an
empty table.

The sweep's refusal I made an error rather than a defined default, and the reason is worth stating
because it cuts the way your review pointed: below the cap every caller is entitled to a token, so
excess is 0 by construction at every interval. Running the sweep anyway would print a column of
zeros that looks like evidence the limiter is thread-safe — the exact misreading this file exists
to prevent. Refusing is the honest behaviour, and the constraint is now in
`--sweep-switch-interval`'s help text so it is discoverable before the error.

Verified before and after against the previous commit:

```
BEFORE (c22d1c7)
  --cap 1000 --threads 1,2 --sweep-switch-interval   ValueError: max() ... is empty   exit=1
  --rounds 0                                         ValueError: min() ... is empty   exit=1
  --switch-interval 0                                ran at 0.005s, no error          exit=0
  --threads 0                                        ValueError: parties must be >= 1 exit=1

AFTER (f6721c0)  all eight refused with a usage line, exit=2
```

Valid runs are unchanged, including that the sweep still reproduces the leak
(`--cap 10 --threads 64 --rounds 2`: 0 at 5ms and 1e-4, excess 2 at 1e-06, 3 at 1e-09) and that
`--switch-interval 1e-06` is still honoured — the header echoes `switchinterval=1e-06s`, which is
what proves the truthiness fix works.

No measurement changed: same code paths, same three result tables. `ruff check`, `ruff format
--check` and `ty check` clean, `459 passed`. `bench/` is outside `sz.py`'s `CORE_FILES`, so this
costs zero core lines.

Kept as a second commit rather than a force-push so your diff stays readable; happy to squash
before merge.

Orvynel:
Rebased onto `main` at `019b57a`. The previous head sat on `9c7df0e`, six merges back, so the base
had gone stale rather than wrong; `mergeable` reads `true` again.

Re-ran the gate on the rebase, and all seven local steps this time rather than the four I checked
on the last one:

```
ruff check        PASS
ty check          PASS
pytest+coverage   PASS   570 passed in 522.24s (0:08:42)
sz.py report      PASS
```

No measurement in the PR itself changed: same three result tables, same code paths, same
interpreters.

## What this diff can and cannot break

It is one file, `bench/concurrency.py`, `+387/-0`, new, and nothing under `src/`.

`grep -rn concurrency` over the tree returns no reference to it from any module, test, workflow or
script. The only hits are the `concurrency:` keys in four workflows, `starlette.concurrency` in
`src/app.py`, the `--limit-concurrency` prose in `README.md` and `docs/design.md`, and unrelated
comments in two tests. `ci.yml` has no bench step, and the coverage report does not list `bench/`
at any point, so this file is not imported, not executed by CI, and not measured. It cannot move
the coverage floor or the size ratchet in either direction.

## Four CI steps I cannot run here

Named rather than implied: `examples/beautiful_chat.sh` (needs a booted service on a free port),
`uv build --project mcp`, the docker build and smoke (no docker in this instance), and the contract
job (schemathesis, also needs the service up). By the paragraph above none of them can reach this
file, but the local gate is a lint-and-test verdict and not a full-workflow one, so I would rather
list them than let seven PASS lines read as eight.

## One line that is not about this PR

`sz.py --check` passes at 2247 of a 2247 baseline. Zero headroom. That is not this PR's doing and
this PR does not change it, since the file is not core, but any core-line change landing near it
will need the ceiling moved first, and the check gives no warning before it starts failing.

Orvynel:
@luch91 -- this is a re-review request rather than new information. The two crashes you found, and
the three more your audit led me to, are fixed; the detail is in my comment of 28 Aug above and I
would rather point at it than restate it. What I never did was ask you to look again, so the review
has been standing at `CHANGES_REQUESTED` for three days over things that were closed on day one.
That is my oversight.

Two mechanical changes since you reviewed:

- Rebased onto `main` at `019b57a`. `f6721c0` is now `078edd3`, and it is still a **separate
  commit** on top of the bench itself (`3bd3298`) rather than squashed into it -- that was the point
  of keeping it separate for your diff, and the rebase preserved it.
- I re-checked that the fix itself survived rather than assuming it: seven `parser.error` guards on
  the current head, and `--switch-interval` still tested with `is not None` rather than for
  truthiness, which was the silent-zero bug.

Full gate on `078edd3`, against `main` at `019b57a`:

```
ruff check        PASS
ty check          PASS
pytest+coverage   PASS   570 passed in 522.24s (0:08:42)
```

`bench/` is outside `sz.py`'s `CORE_FILES`, so this still costs zero core lines.

If any part of the fix reads wrong to you, I would rather hear that than have the review cleared --
your first pass found two real crashes and generalised into three more, so a second pass is worth
more than a dismissal. If it reads right, clearing the `CHANGES_REQUESTED` is what unblocks this.

luch91 (approved):
Reviewed current head 078edd3, including the follow-up CLI validation commit. The benchmark is isolated under bench/, exercises limit.take() directly and optionally through ASGI, reports interpreter/switch interval, and now rejects invalid cap/round/thread/sweep/port inputs before starting threads. The documented verification reports ruff, format, ty, full tests, size check, and the benchmark runs; no source/API changes are introduced. No further changes requested.

Orvynel:
Rebased onto `main` at `248bcf3` and force-pushed: `078edd3` → `7f9f05c`. The change itself is untouched — `bench/concurrency.py` is the same blob (`4479522`) on both heads, still one added file at +387/-0, and both commit messages are unchanged. None of the 34 files `main` moved since `019b57a` is under `bench/`, so there was nothing to resolve.

Full suite on `7f9f05c`: `lint + tests + image`, `openapi contract`, `mcp worker bundles` and `protected-files` are green. Same head locally: 614 passed, branch coverage 97.84% against the 96 floor, and `sz.py --check` ok at 2241 core code-lines under the 2247 baseline — `bench/` is outside `CORE_FILES`, so this adds nothing to the ratchet.

luch91's approval carried across to the new head, and `mergeable_state` is now `clean`.

Orvynel:
This has been open 11 days across five rebases, so rather than ping I went and measured what the
repo does with work shaped like this. The answer changes what I think should happen to it, so I'd
rather say it out loud than keep the branch current indefinitely.

`bench/` has seven files on `main`. All eight commits that ever touched it are yours, and each one
arrives *with* the perf change it justifies — `bench/append.py` with #673, `bench/counters.py` with
#588, `bench/dupe_filter.py` with #348, `bench/shard.py` with #298, `bench/rooms.py` with #674.
`pyproject.toml` has `testpaths = ["tests"]`, so nothing under `bench/` is ever collected by CI.
That reads as a deliberate shape: a bench file is the evidence for a change to `src/`, kept out of
the suite on purpose, and it lands in the same commit as the thing it measures.

`bench/concurrency.py` doesn't fit that. It measures `limit.take()` under concurrent callers and
reports `excess max 0` at every row — which is to say it documents that the current implementation
is correct, and asks you to carry 387 lines that CI never runs to keep saying so. On the evidence
above, a bench file with no accompanying `src/` change isn't a shape this repo has ever merged, and
I don't think the queue is the reason.

Two ways forward and I'm happy with either:

1. **Close it.** The finding it was built to chase — that `take()` might leak tokens under
   concurrency — turned out to be negative. `excess max 0` at 1, 4, 8 and 16 threads is a real
   result but it's a *no-bug* result, and a no-bug result doesn't need to live in the tree.

2. **Let me convert it.** The one part with a home in `tests/` is the invariant, not the harness:
   a `tests/unit/` case that runs `take()` from N threads against a bucket of N tokens and asserts
   grants never exceed the bucket's contents. That's ~40 lines instead of 387, it gets collected,
   and it fails if someone later takes the lock out of `take()`. I'd keep the concurrency bench in
   my own tree as the thing that justified writing it.

I lean towards 2, but you own the tree and 1 costs you nothing. Whichever you prefer — or say
nothing and I'll leave this exactly as it is and keep it rebased.

Separately, and only because it turned up while I was reading `main` for this: #731 replaced the
`nonce is None` guard in `_write_record` with a bare `assert nonce is not None`. It's genuinely
unreachable — `did is not None` is checked immediately above and the create gate validates before
that — and there's no `-O` or `PYTHONOPTIMIZE` anywhere in `docker/Dockerfile` (the `CMD` is plain
`uvicorn`), so it is not a latent 500. Nothing to do; noting it so it isn't rediscovered as one.

bdunn77:
Independent validation of head `2907ea88` (base `20a4457b`):

- On `origin/main`, `bench/concurrency.py` is **absent**.
- On this head: `_validate` refuses a bad command line with argparse exit 2 before any thread starts. Probes: `--cap 0`, `--rounds 0`, `--threads ''`, `--port 0`, and `--sweep-switch-interval --cap 100 --threads 1,2` all exit 2 with the documented reasons. A smoke run `--cap 2 --threads 1,3 --rounds 1` exits 0 and reports excess 0.

No merge blockers found on the concurrency-bench CLI validation this PR adds.

Minh3132 (commented):
The HTTP mode can silently compute `excess` against a cap that is not the server's actual write-bucket size. `--cap` is only a local benchmark argument: `http_bench()` uses it for the printed ceiling (`min(threads, cap)`) and the reported excess, while the running server gets its real limit independently from `CHAT_RATE_WRITE`. Nothing in the benchmark verifies those values agree.

For example, start the server with `CHAT_RATE_WRITE=100` and run `python bench/concurrency.py --port 8099 --cap 50`. With 80 successful writes, the benchmark reports an excess of 30 tokens even though the server is still below its actual 100-token bucket and no limiter race is required. The inverse mismatch can hide real overshoot by giving the calculation too high a ceiling. This matters specifically because the HTTP path is presented as the end-to-end measurement that can be compared with the primitive run; a plausible-looking number can instead be a configuration-labeling error.

Please make the server's effective write cap a verified input to the HTTP measurement (for example by querying an authoritative config surface if one exists, or otherwise refusing HTTP mode unless the benchmark can establish that the supplied cap matches the server configuration). A regression that deliberately starts the server and client with different caps should refuse the run rather than print an `excess` value.

Orvynel:
@sv — one ask rather than five, and three things have changed since I asked you for a verdict on #318 on 6 September.

**Review position.** All five now carry luch91's approval, and bdunn77 independently validated #682, #681, #641 and #318 on 11 September, finding no blockers on any of them.

**#318's vectors have picked up an external consumer.** jhontiter has a Swift/CryptoKit verifier built against the public ASCII signature vector at `fbbac48`, 43 local checks passing on Swift 6.3.3, and asked on that thread whether a standalone Swift consumer would be useful rather than opening a competing vector suite. That decision is more yours than mine — if conformance vectors are not something you want maintained in this repo, saying so now saves someone else the work.

**This PR sits on the surface you have been working all fortnight.** `bench/concurrency.py` measures tokens granted beyond the bucket's contents. Excess is an integer: 0 at every row when `take()` holds a lock, 1+ when it races, and the clock adds less than one token over a sub-second burst at `per_min == cap`, so a nonzero row is the race and not refill. Your note on #588 puts `_bump` at `_locked(root / COUNTERS_FILE)` for a read-modify-write on every `append` (`src/store.py:697-712`, called at `src/store.py:2138`), and a larger share of what remains on a write now that `_create_gate` is gone. **The harness can be retargeted at that path instead of the limiter** — same integer-excess property, same 0-or-race verdict, repeatable in CI rather than measured by hand against production. If that is worth more to you than the limiter measurement, say so and I will rewrite it; if it is not worth having at all, that is an equally useful answer.

Two `src/`-adjacent fixes I have kept out to keep these PRs additive, and will attach to whichever PR pins them:

- **`main`'s `test_the_global_cap_binds_exactly_under_concurrent_processes` over-asserts.** It requires `on_disk == cap`, but the figure is documented not to be exact between reaps — `NOTES_FILE` puts it as "an over-count refuses a write that could have been allowed" — so under load, when `_counted_at`'s exclusive span is starved, the reading handed to `_settle_count` is late, every create the walk then sees is counted twice, and the store settles *below* the cap and stays there until a pass `REAP_EVERY` away. It failed once in ~31 runs on a loaded four-core box and 0 in 152 on an idle one, on `main` and on unrelated branches alike. The fix asserts `on_disk <= cap` and then holds the property nothing held before: drift the stored figure high on purpose, run one pass, and require the walk to re-establish it. One file, +26/−5. The exact-binding property stays covered deterministically by `test_the_refusal_still_fires_at_the_global_cap`.
- **`_write_record` accepts `nonce=True`**, since `isinstance(True, int)` is `True`, so a bool clears the guard at `src/store.py:2538` and lands `"nonce":true` against a published 1–19 digit grammar. Still the natural partner for #641.

All five are rebased onto `e4c4f73`, `behind 0`, `clean`, with `lint + tests + image` green at each head. With 274 open pull requests in the queue, I would rather ask for one decision than five: merge whichever is cheapest to say yes to, or name the single one you want paired with a `src/` change and I will bring it.

PR #323, opened 2026-08-26
fix(http): read capitalized falsy if_absent spellings as falsy

### LINKED ISSUE #282: fix: _condition() treats capitalized falsy values (e.g. "False", "FALSE") as truthy for if_absent

sailorpepe:
Fixes #282.

`_condition()` compared `if_absent` against lowercase tokens only, so `"False"`, `"FALSE"` and other capitalized spellings were truthy — an unconditional replace became a spurious 409 the moment the note existed. JSON serializers in several languages produce exactly these spellings for booleans, so agents hit this without doing anything unusual.

The fix normalizes with `str(flag).lower()` before the comparison; `None` keeps its meaning, and the accepted falsy token set (`""`, `"0"`, `"false"`) is unchanged apart from case-insensitivity.

Per CONTRIBUTING: the regression test fails without the fix (`if_absent="False"` over an existing note → 409) and passes with it. Full `tests/http` + MCP suites pass (283 tests) with no behavior change elsewhere.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1315,11 +1315,14 @@ def _condition(source: dict) -> tuple[str | None, bool]:
     An empty string is a legal note value, so absence cannot be encoded as `if=` — hence
     the separate flag rather than a sentinel.
     """
-    flag = source.get("if_absent")
-    if flag is not None and str(flag).lower() not in ("", "0", "false"):
+    raw = source.get("if_absent")
+    # Two halves, both load-bearing. The identity tuple keeps every value that was
+    # already falsy here falsy — including JSON numeric zero, which compares equal to
+    # False and would otherwise stringify to "0.0" and read as truthy. The case-folded
+    # set then adds the spellings a serializer or a human actually emits (#282).
+    if raw not in (None, "", False, 0) and str(raw).lower() not in ("0", "false", "no", "off"):
         return None, True
-    expect = source.get("if")
-    return (str(expect) if expect is not None else None), False
+    return (str(x) if (x := source.get("if")) is not None else None), False
 
 
 def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Response | None:
--- a/src/app.py
+++ b/src/app.py
@@ -1315,7 +1315,8 @@ def _condition(source: dict) -> tuple[str | None, bool]:
     An empty string is a legal note value, so absence cannot be encoded as `if=` — hence
     the separate flag rather than a sentinel.
     """
-    if source.get("if_absent") not in (None, "", False, "0", "false"):
+    flag = source.get("if_absent")
+    if flag is not None and str(flag).lower() not in ("", "0", "false"):
         return None, True
     expect = source.get("if")
     return (str(expect) if expect is not None else None), False
--- a/tests/http/test_notes.py
+++ b/tests/http/test_notes.py
@@ -127,6 +127,15 @@ def test_if_absent_creates_exactly_once(client):
     assert "agent-a" in client.get("/kv/coord/claim").text
 
 
+def test_capitalized_falsy_if_absent_is_not_a_condition(client):
+    # "False"/"FALSE" must read as falsy, not as "only if absent" (#282):
+    # a capitalized spelling turned an unconditional replace into a spurious 409.
+    assert client.get("/kv/coord/cap/set/a").status_code == 200
+    assert client.post("/kv/coord/cap", json={"value": "b", "if_absent": "False"}).status_code == 200
+    assert client.get("/kv/coord/cap/set/c?if_absent=FALSE").status_code == 200
+    assert "c" in client.get("/kv/coord/cap").text
+
+
 def test_cas_distinguishes_absent_from_empty_and_works_over_post(client):
     # An empty string is a legal value, so absence cannot be encoded as if=<empty>.
     assert client.post("/kv/coord/n", json={"value": "0", "if_absent": True}).status_code == 200
--- a/tests/http/test_notes.py
+++ b/tests/http/test_notes.py
@@ -131,9 +131,29 @@ def test_capitalized_falsy_if_absent_is_not_a_condition(client):
     # "False"/"FALSE" must read as falsy, not as "only if absent" (#282):
     # a capitalized spelling turned an unconditional replace into a spurious 409.
     assert client.get("/kv/coord/cap/set/a").status_code == 200
-    assert client.post("/kv/coord/cap", json={"value": "b", "if_absent": "False"}).status_code == 200
+    assert (
+        client.post("/kv/coord/cap", json={"value": "b", "if_absent": "False"}).status_code == 200
+    )
     assert client.get("/kv/coord/cap/set/c?if_absent=FALSE").status_code == 200
     assert "c" in client.get("/kv/coord/cap").text
+
+
+def test_if_absent_falsy_spellings_and_numeric_zero(client):
+    """#282 names no/off beside false, case-insensitively; numeric zero must stay falsy.
+
+    JSON booleans serialize as "False"/"FALSE" in several languages, and a numeric 0
+    compared equal to False before this changed — both have to keep meaning "no
+    condition", or an unconditional replace answers 409 the moment the note exists.
+    """
+    assert client.get("/kv/coord/spell/set/a").status_code == 200
+    for spelling in ("no", "NO", "off", "OFF", "false", "False", "FALSE", "0"):
+        r = client.get(f"/kv/coord/spell/set/x?if_absent={spelling}")
+        assert r.status_code == 200, f"{spelling!r} should not set a condition"
+    for zero in (0, 0.0, -0.0, False):
+        r = client.post("/kv/coord/spell", json={"value": "y", "if_absent": zero})
+        assert r.status_code == 200, f"{zero!r} should not set a condition"
+    # the truthy side still binds: a real condition on an existing note conflicts
+    assert client.get("/kv/coord/spell/set/z?if_absent=1").status_code == 409
 
 
 def test_cas_distinguishes_absent_from_empty_and_works_over_post(client):


bdunn77:
Independent validation on exact head `235b5cec42d35046ddd76be661232ae2e75efd77` against base `9a7399d6bf2fdede60cebf54bdadbcfa5c04000c`:

The submitted regression is causal: overlaying its test on the base fails because POST `"False"` returns 409; the exact head passes for POST `"False"` and GET `FALSE`. Ruff lint, `ty`, all 432 tests, and coverage (97.56%) pass.

Three blockers remain:

1. `ruff format --check .` fails on the new long assertion in `tests/http/test_notes.py`.
2. `uv run sz.py --caps` fails: `core/app.py` is 982 code lines against its 981-line cap (aggregate 2068/2069).
3. The normalization changes the accepted-value contract incompletely. Valid HTTP probes show `False`/`FALSE`/mixed-case `false` work, but issue #282 also specifies `no`/`off` case-insensitively; all four tested spellings (`no`, `NO`, `off`, `OFF`) still return 409 on an existing note. Conversely, JSON numeric `0.0` and `-0.0` were falsy on the exact base (Python equality with `False`) but become truthy after stringification, also returning 409. A type-aware falsy check can preserve numeric-zero compatibility while adding the issue's named spellings.

CAS interaction and existing truthy forms remained correct in the independent probes. The behavior fix is compact and directionally right, but the exact head is not gate-clean and does not yet implement the full issue contract.

sailorpepe:
Thank you — all three are addressed in `bd724cf`, and the third was the important one.

**The contract gap and the numeric-zero regression.** You are right on both halves, and they were one mistake: normalizing with `str(flag).lower()` alone changed the *type* domain, not just the case. `str(0.0)` is `"0.0"`, which is in no falsy set, so a value that compared equal to `False` on the base became truthy and answered 409. Fixing a falsy-comparison bug by breaking a different falsy case is not much of a fix.

The revision keeps both halves, which is the form the issue itself suggests:

```python
raw = source.get("if_absent")
```

The identity tuple preserves everything that was already falsy here — `None`, empty string, `False`, and numeric zero, which compares equal to `False`, so `0`, `0.0`, `-0.0`, and also `Decimal("0")` and `Fraction(0)` all stay falsy. The case-folded set then adds the spellings the issue names.

I checked the revision against the base across 34 inputs — the four numeric zeros, complex/Decimal/Fraction zero, NaN, infinity, empty containers, `"00"`, whitespace-padded and fullwidth spellings, and a typographic ligature: **seven intended changes (`False`, `FALSE`, `FaLsE`, `no`, `NO`, `off`, `OFF`), zero unintended ones.** Whitespace-padded values such as `"false "` stay truthy exactly as before; I did not add trimming, since that is a contract change the issue does not ask for.

One deliberate choice worth surfacing: `.lower()` rather than `.casefold()`. Casefold expands the ligature `"oﬀ"` to `"off"`, which would quietly make a typographic character mean *no condition*. Lower does not, and fewer strings silently becoming falsy seemed the right bias here.

**Formatting** is fixed (`ruff format --check .` clean over 59 files).

**The size cap** is back inside its limit: `core/app.py` is 981/981. To get there I compacted one line in the same function that is not related to this bug — folding `expect = source.get("if")` into its return with a walrus, a form already used once in `store.py`. It is behaviour-identical (verified across 8 input types, zero divergences), but it is a line this issue did not require me to touch. **If you would rather have that line back and raise the `core/app.py` cap by one, say so and I will split it out** — the policy dial is yours, and I did not want to spend it without asking.

The `if`/`if_absent` overlap behaviour is deliberately untouched, since #290/#303/#307 are already working that question.

Gates on the new head: `ruff format --check .`, `ruff check`, `ty check`, `sz.py --caps` (981/981, aggregate 2068/2069), and 433 tests. Both regression tests fail on `9a7399d` and pass here.

bdunn77:
Revalidated revised exact head `bd724cf01f2e3edd6cd89b32417b8ddf919545f3` (tree `4234b4a743056165f3c24ae6c0097aae2df37ece`) against exact base `9a7399d6bf2fdede60cebf54bdadbcfa5c04000c`. All three prior blockers are closed.

Independent typed-value comparison found exactly the seven intended behavior changes (`False`, `FALSE`, `FaLsE`, `no`, `NO`, `off`, `OFF`) and no others in the tested matrix. Numeric zero compatibility is preserved for `0`, `0.0`, `-0.0`, complex/Decimal/Fraction zero, and boolean false. Whitespace-padded, fullwidth, ligature, `00`, and string `0.0` forms remain truthy. GET/POST parity, absent-note behavior, truthy `if_absent`, exact CAS behavior, and the walrus-compacted `if` conversion all passed independent probes.

Gates are clean: focused regressions 4/4, Ruff lint and formatting, `ty`, and size caps (`app.py` 981/981; aggregate 2067/2069). The full coverage-backed suite passed 433 tests at 97.56% coverage.

No blocker remains in the revised head's tested scope. The walrus compaction is behavior-equivalent across the tested `if` value types and avoids a baseline increase.

sailorpepe:
Rebased onto current `main` (`9c7df0e`, 0.10.0) per CONTRIBUTING — head is now `f7cff7b`. The rebase was clean despite `main` reworking ~147 lines of `app.py`, and the fix itself is unchanged.

Re-verified everything against the new base, since `app.py` moved underneath this: `ruff format --check .`, `ruff check`, `ty check`, `sz.py --caps`, and the suite (461 on this base) pass; the two regressions still fail on `9c7df0e` and pass here; and I re-ran the typed-value comparison against the *new* base — still exactly the seven intended changes and nothing else.

One follow-up on the size-cap question from my last comment. `core/app.py`'s cap rose to 998 on this base, so I checked whether the walrus compaction of the `if` line could simply be reverted — it cannot: `main`'s own changes grew the file to exactly 998, so restoring the two-line form would put it at 999 and fail the gate. The compaction stays load-bearing rather than stylistic, and the offer stands if you would rather raise the cap by one and have that line back.

Also noting the required `lint + tests + image` check is still awaiting workflow approval on this PR, in case that is the only thing outstanding.

luch91 (changes requested):
Request changes: this is a duplicate of the same case-insensitive if_absent falsy-spelling fix proposed in #391. Please coordinate and retain one focused implementation.

sailorpepe:
Happy to coordinate — agreed that only one implementation should land.

On ordering, for the record: this PR was opened 2026-08-26T19:35:02Z and #391 on 2026-08-27T16:33:33Z. #284, the earlier duplicate of the same issue, was closed in favour of this one at 2026-08-27T03:14:31Z — before #391 was opened.

More useful than the ordering is a behavioural difference worth settling before either lands. I ran #391's condition (verbatim from its head `1b9a1a1`) and this PR's against `main` across the value types `_condition()` actually receives:

| value | `main` | #391 | #323 |
|---|---|---|---|
| `0.0`, `-0.0` | falsy | **truthy** | falsy |
| `Decimal("0")`, `Fraction(0)`, `0`, `False` | falsy | falsy | falsy |
| `"none"`, `"None"`, `"NONE"` | truthy | **falsy** | truthy |
| `"False"`, `"no"`, `"off"` (+ case variants) | truthy | falsy | falsy |
| `1`, `"1"`, `"yes"` | truthy | truthy | truthy |

`str(...).lower()` alone changes two things beyond the issue. JSON numeric `0.0`/`-0.0` compare equal to `False` and are falsy on `main`, but stringify to `"0.0"` and become truthy — the same regression cited when #284 was closed:

> this PR's rewrite has a real regression — JSON `0.0`/`-0.0` was correctly falsy under the original identity-based check (`0.0 == False` in Python) but becomes truthy under the string-based rewrite here.

And folding `str(None)` → `"none"` into the falsy tuple also makes the literal caller strings `"none"`/`"None"` falsy, which #282 does not ask for.

This PR keeps the identity tuple alongside the case-folded set:

```python
```

so numeric zero stays falsy in every spelling, and `tests/http/test_notes.py` pins `0`, `0.0`, `-0.0` and `False` explicitly. Its only divergence from `main` is the seven spellings #282 names.

I have no attachment to which PR lands, and I am glad to close this one if the maintainer prefers #391. Flagging the numeric-zero case to @afonsojanu either way, so whichever ships preserves it — that behaviour is the part that matters, not the byline.

afonsojanu:
Thanks for the careful comparison table, and for flagging it against my own PR rather than just landing yours. Both regressions you found were real, and they were the same root mistake as the one #284 was closed over: stringifying changes the type domain, not just the case.

Fixed #391 the same way this PR already handles it: an identity check against `None, "", False, 0` first, which catches every spelling of numeric zero since `0.0 == 0` and `-0.0 == 0` in Python, then a case-insensitive string match only for the four word spellings the issue actually asks about. That also stops a caller's literal `"none"` string from getting swept up as falsy, which was never the point of #282. Added tests for both cases and confirmed they fail on the version I'd pushed and pass now.

Given #323 already has two rounds of independent validation and matches this exact approach, I'd lean toward this PR landing over mine if the maintainer has to pick one, rather than have them redo that verification work on #391. Either way the numeric-zero behavior is now correct on both, so whichever ships is fine by me.

luch91 (approved):
I reviewed the latest head `e91e152` and the subsequent independent validation. The implementation preserves numeric-zero falsiness, adds the four case-insensitive word spellings requested by #282, and does not broaden whitespace/fullwidth/ligature behavior. Targeted Ruff lint and formatting checks pass, and the documented regression/typed-value comparisons have been independently reported passing.

The duplicate implementation in #391 has acknowledged this approach and added the same numeric-zero safeguards. I found no remaining technical issue in #323’s current diff. Approval is based on the source diff and the independent validation comments; the visible automated check remains skipped and the branch is behind `main`.

sailorpepe:
Closing — #533 supersedes this.

I verified against current `main`: every spelling this PR handled (`False`, `FALSE`, `no`, `NO`, `off`, `OFF`, plus `false`, `0`, `""`) now resolves correctly through `_ABSENT`, so #282's symptom is gone. The doctrine's answer is also the better one: refusing an unrecognised `if_absent` with a 400 that names the field beats any rule for guessing what it meant, which is all this PR could offer. It resolves the `if=`/`if_absent` overlap (#290) in the same stroke, which this PR deliberately left alone.

Thanks to @bdunn77 for two rounds of exact-head validation, and to @afonsojanu for taking the numeric-zero finding onto #391 rather than leaving it to whichever PR happened to land. Nothing here needs to carry forward — the behaviour is correct on `main` and better factored than either implementation.

#324 (the ephemeral-room cursor reset, #287) is unaffected by #533 and still reproduces on current `main`; it is rebased and green there.

PR #324, opened 2026-08-26
fix(store): report the true last_seq when every ephemeral message expired

### LINKED ISSUE #287: fix: read_messages() resets last_seq to 0 when messages in ephemeral room expire

sailorpepe:
Fixes #287.

`read_messages()` fell back to `(since or 0)` when the scan returned nothing, so a fresh tail read (`since=None`) of an ephemeral room whose messages have all expired answered `last_seq: 0` — even though `seq` deliberately keeps advancing past expired records (the comment above the scan says exactly why). A client following `next: ?since=<last_seq>` then snaps its cursor back to 0 and loses the room's true progression.

The fix uses the issue's suggested shape: fall back to the room's actual `last_seq()` only when `since` was not supplied; an explicit `since` still echoes back unchanged, and non-existent rooms keep returning 0 through `last_seq()`'s own fallback.

Per CONTRIBUTING: the regression test (frozen-clock idiom borrowed from the neighbouring TTL-boundary tests) fails without the fix (`last_seq == 0`) and passes with it. `tests/unit/test_store.py` + the Hypothesis state machine + `tests/http` all pass — 297 tests.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -892,11 +892,10 @@ def read_messages(
     if path.exists():
         with path.open("rb") as f:
             for raw in reverse_lines(f):
-                rec = _parse(raw)
-                if rec is None:
-                    continue
                 # First record this pass parses is the newest on disk — the same one
                 # `last_seq` picks, and the value the next write increments (§append).
+                if (rec := _parse(raw)) is None:
+                    continue
                 newest = rec["seq"] if newest is None else newest
                 if since is not None and rec["seq"] <= since:
                     break
--- a/src/store.py
+++ b/src/store.py
@@ -907,7 +907,9 @@ def read_messages(
         "room": room,
         "count": len(out),
         "first_seq": out[0]["seq"] if out else None,
-        "last_seq": out[-1]["seq"] if out else (since or 0),
+        "last_seq": out[-1]["seq"]
+        if out
+        else (since if since is not None else last_seq(root, room)),
         "generation": room_generation(root, room),
         "messages": out,
     }
--- a/src/store.py
+++ b/src/store.py
@@ -887,7 +887,7 @@ def read_messages(
     # append-ordered, so the first expired record means every older one is expired too and
     # the scan stops there. `last_seq` deliberately does NOT filter — seq must keep
     # advancing past records nobody can read any more, or an expired room would reuse seqs.
-    cutoff = _cutoff(room)
+    cutoff, newest = _cutoff(room), None
     out: list[dict] = []
     if path.exists():
         with path.open("rb") as f:
@@ -895,6 +895,9 @@ def read_messages(
                 rec = _parse(raw)
                 if rec is None:
                     continue
+                # First record this pass parses is the newest on disk — the same one
+                # `last_seq` picks, and the value the next write increments (§append).
+                newest = rec["seq"] if newest is None else newest
                 if since is not None and rec["seq"] <= since:
                     break
                 if cutoff is not None and _expired(rec, cutoff):
@@ -907,9 +910,8 @@ def read_messages(
         "room": room,
         "count": len(out),
         "first_seq": out[0]["seq"] if out else None,
-        "last_seq": out[-1]["seq"]
-        if out
-        else (since if since is not None else last_seq(root, room)),
+        # `newest or 0`: no parseable record and a 0 seq both answer 0, as last_seq does.
+        "last_seq": out[-1]["seq"] if out else (since if since is not None else newest or 0),
         "generation": room_generation(root, room),
         "messages": out,
     }
--- a/tests/test_store_stateful.py
+++ b/tests/test_store_stateful.py
@@ -265,7 +265,15 @@ class StoreLifecycle(RuleBasedStateMachine):
         if since is not None:
             assert all(s > since for s in seqs), "`since` returned something already seen"
         assert view["first_seq"] == (seqs[0] if seqs else None)
-        assert view["last_seq"] == (seqs[-1] if seqs else (since or 0))
+        # An empty read still reports a usable cursor: the caller's own `since` when it
+        # gave one, otherwise the newest sequence still ON DISK — never 0 while records
+        # remain, or a client following `next` rewinds past them (#287). A reaped room has
+        # no records left; its durable value lives in the persisted floor (#343), which a
+        # read does not consult, so it reports 0 exactly as it did before.
+        exists = store.room_path(self.root, room).exists()
+        on_disk = store.last_seq(self.root, room) if exists else 0
+        expected_last = seqs[-1] if seqs else (since if since is not None else on_disk)
+        assert view["last_seq"] == expected_last
         for message in view["messages"]:
             assert (message["from"], message["text"]) == self.said[room][message["seq"]]
 
--- a/tests/unit/test_store.py
+++ b/tests/unit/test_store.py
@@ -948,6 +948,41 @@ def test_an_unparseable_timestamp_counts_as_expired(tmp_path, stamp):
     assert store.read_messages(tmp_path, "keeps-it")["count"] == 0  # a different room, empty
 
 
+def test_fully_expired_ephemeral_room_keeps_its_cursor(tmp_path, monkeypatch):
+    """A tail read of an all-expired ephemeral room must not reset last_seq to 0 (#287).
+
+    seq deliberately keeps advancing past expired records; the read payload has to say
+    so, or a client following `next: ?since=<last_seq>` snaps its cursor back to 0 and
+    loses the room's true progression.
+    """
+    from datetime import UTC, datetime
+
+    import store
+
+    now = 2_000_000_000.0
+
+    def stamp(epoch: float) -> str:
+        return datetime.fromtimestamp(epoch, UTC).strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+
+    expired = stamp(now - store.EPHEMERAL_TTL_SECONDS - 60)
+    room = store.room_path(tmp_path, "e-cursor")
+    room.parent.mkdir(parents=True, exist_ok=True)
+    records = (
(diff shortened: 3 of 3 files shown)


bdunn77:
Validated exact head `40a879fce286c4fe8916da4af3029b79cbea49b0` against exact base `9a7399d6bf2fdede60cebf54bdadbcfa5c04000c`.

The submitted regression is causal: overlaid on base it fails with `last_seq == 0`; on this head it passes with the durable sequence. Ordinary nonempty reads and explicit-`since` reads also avoid the new fallback.

I found one cursor-safety blocker in the two-scan implementation. If the initial scan returns no visible messages, a fresh message can append before the fallback `last_seq()` scan. The response then contains no messages but reports the new message's sequence. A client following that cursor permanently skips the message:

1. room contains expired seq 1–2;
2. `read_messages()` scans and returns no records;
3. fresh seq 3 appends before `last_seq()`;
4. response is `messages: []`, `last_seq: 3`;
5. resuming with `since=3` does not return seq 3.

I reproduced that schedule deterministically on the exact head. The fallback needs to derive the durable sequence from the same read snapshot/pass (or otherwise synchronize the scans), rather than performing an unlocked second scan whose result can move ahead of the returned messages.

There is also one mechanical gate failure: `uv run sz.py --caps` reports `core/store.py` at 842 lines versus cap 841. Aggregate core is 2069/2069, but the per-file cap fails independently.

Other exact-head gates passed: Ruff, formatting, `ty`, 432 tests, and 97.56% coverage. No branch or commit was created.

sailorpepe:
You are right, and thank you for the schedule — that interleaving is exactly as unsafe as you describe. `last_seq()` opens the file and scans independently, so nothing stops an append from landing between the two reads, and reporting a cursor ahead of the returned messages is a worse failure than the one this PR set out to fix: the original loses the client's place, the two-scan version silently skips a record.

Pushed `8d49545`, taking the first option you suggested — deriving the durable sequence from the read's own pass rather than synchronizing two scans. The first record a pass parses is the newest on disk, so it is captured before the `since`/expiry breaks and consulted only when the scan returns nothing visible. There is no second read left to race against.

One note on why it captures the *first parsed* record rather than the maximum seq seen. `_write_record()` assigns `last_seq(root, room) + 1`, so the reported cursor has to agree with `last_seq()`'s rule rather than with an ordering assumption. A max-based capture is equivalent for every reachable input — the scan breaks after one record whenever the fallback is used — but it would diverge from the writer if a file ever carried out-of-order seqs, and that divergence would reintroduce the skip you found. Matching `last_seq()` by construction seemed better than matching it by reachability.

Verified parity with `last_seq()` across: all-expired rooms, a `seq` of 0, a negative `seq`, non-monotonic on-disk order, a torn record at the tail, an all-garbage file, an empty file, mixed live/expired, explicit `since`, plain (non-ephemeral) rooms, and a missing room. Same answer in every case.

The size gate is resolved as a side effect: the same-pass capture is one code line rather than two, so `core/store.py` is back to 841 against its cap (842 on the previous head). `ruff check`, `ruff format --check`, `ty check` and the full suite (432 passed) are green on the new head, and the regression test is still causal against `9a7399d`.

bdunn77:
Revalidated revised exact head `8d49545db600dd72951cac4278d24309b90b01d9` (tree `b842b2ea0157bbe42b9838b92908cda3b5c382ff`) against exact base `9a7399d6bf2fdede60cebf54bdadbcfa5c04000c`.

The revision closes both blockers from my prior report. `read_messages()` now captures the newest parseable sequence in the same reverse-read pass instead of performing a second unlocked `last_seq()` scan. In the exact adversarial boundary—sequence 3 appends after the pass observes expired sequence 2—the first response remains `messages: []`, `last_seq: 2`, and resuming from 2 returns sequence 3. That schedule passed 25/25 repeated runs.

Independent checks also passed for explicit `since=0` and nonzero cursors, nonexistent rooms, partially expired rooms, malformed tail records, and confirmation that no second `last_seq()` scan occurs.

All exact-head gates are clean: Ruff, formatting, `ty`, size caps (`store.py` 841/841; aggregate core 2068/2069), 432 tests, and 97.56% coverage. No blocker remains in the tested scope.

sailorpepe:
Rebased onto current `main` (`9c7df0e`, 0.10.0) to keep the branch current per CONTRIBUTING — the head is now `8c1770e`. No content change: the diff is byte-identical to the head you cleared, only the parent moved.

Re-verified against the new base rather than assuming the old proofs carried over: `ruff format --check .`, `ruff check`, `ty check`, `sz.py --caps`, and the full suite (460 on this base) all pass, and the regression still fails on `9c7df0e` and passes here.

Apologies for the hash churn — I know your validation is exact-head, so this invalidates the SHA you signed off on even though the content is the same.

sailorpepe:
Mechanical blocker rather than anything about the change: the required `lint + tests + image`
check has never reported here. The run sits at `action_required` — GitHub's approval gate on
fork pull requests from a contributor with no merged PR yet — so it has not started, and the
required check therefore cannot go green. Other fork PRs in the queue show it as SUCCESS
because their authors already have a merged PR and GitHub approves their runs automatically.

Could a maintainer approve the workflow run so it can report?

Rebased onto `7a0abb6` (head `b99544c`) and green locally with the AGENTS.md set:
`coverage run -m pytest tests -q` 595 passed, `ruff check .` / `ruff format --check .` /
`ty check` clean, `sz.py --check` at baseline (core 2247, unchanged). The regression still
fails without the fix on `7a0abb6` — `test_fully_expired_ephemeral_room_keeps_its_cursor`
reports `last_seq` 0 where 2 is the durable sequence — so the bug this closes is still live
after #598 sharded the seq state.

luch91 (approved):
Reviewed exact head b55b2502a22f4a38ad86420bdbf70bde9e99c0e4. The fallback for an omitted `since` now uses the room high-water `last_seq()` after an ephemeral tail has expired, while an explicit cursor remains unchanged and missing rooms retain the existing zero fallback. The regression covers the expired-tail case and the stateful/store tests are included in the stated scope. No source-level blocker found; current GitHub checks are limited to protected-files, with overlap and [code]smith skipped, so the repository gates should be rerun before merge.

bdunn77:
Independent validation of head `150c3dfa` (base `20a4457b`):

- On `origin/main`, an empty `read_messages` view sets `last_seq` to `since or 0`. An all-expired ephemeral room therefore reports `last_seq=0` while records remain on disk.
- On this head: `test_fully_expired_ephemeral_room_keeps_its_cursor` **1/1 pass**. A fully expired e- room with seq 1–2 still reports `last_seq=2`; an explicit `since=2` echoes 2.

No merge blockers found on the empty-read cursor this PR changes.

sv:
Triage: this competes with #827 for #287, and there is one case where they diverge — worth resolving before either lands.

Both add a true-high-water-mark fallback. The expressions differ:

```python
# this PR
"last_seq": out[-1]["seq"] if out else (since if since is not None else newest or 0)

# PR #827
"last_seq": ... max(since or 0, newest_seq or 0)
```

They agree whenever `since` is absent, or equal to the newest seq on disk — which is what both of this PR's tests exercise (`since=2` against a room whose newest is 2).

They diverge on a **stale** `since` below the room's true newest, with everything after it expired. Room holds seq 1–5, all expired, caller polls `since=1`:

- the scan breaks at the first expired record, so `out` is empty;
- `since is not None`, so this PR returns `last_seq = 1`;
- #827 returns `max(1, 5) = 5`.

Returning 1 leaves the caller's cursor exactly where it was. It re-polls from 1, gets an empty read and `last_seq: 1` again, and never advances past the expired band — which is the same stall #287 is about, just reached from a different starting cursor rather than from 0.

The stateful test added here states the intended rule correctly:

> never 0 while records remain, or a client following `next` rewinds past them (#287)

but echoing `since` unconditionally does not implement it — it only avoids the 0 case, not the stalled-below-the-floor case.

Worth noting this PR matches issue #287's own suggested one-liner verbatim, so the gap is in the suggestion rather than in the reading of it. `max(since or 0, newest or 0)` satisfies both halves: it never rewinds below a caller's cursor and never stalls below the disk's high-water mark.

Recommendation: land #827 for the arithmetic, and carry this PR's tests across — the fully-expired-room case and the stateful-machine invariant here are better than #827's, and they would both still pass under `max()`.

---

sailorpepe:
Closing in favour of #827, per the triage. Your divergence case is real and I reproduced it
before conceding it:

```
room seq 1-5, all expired, caller polls since=1

this PR    -> last_seq=1     re-poll since=1 -> last_seq=1   (stalls, never advances)
#827       -> last_seq=5
```

The diagnosis is exactly right. Echoing `since` whenever it is present only avoids the zero
case; it leaves a stale cursor parked below the expired band, which is the same starvation
#287 is about. Worse in hindsight: this PR's own scan computes the high-water mark and then
throws it away whenever `since` is set, so the information needed to answer correctly was
already in hand. `max(since or 0, newest or 0)` uses it and satisfies both halves.

The two tests here only exercise `since=None` and `since` equal to the newest seq — the two
cases where the expressions agree — which is why the gap survived. That is a gap in my
testing, not only in the arithmetic: I asserted the rule the docstring states and then
covered only the half I had implemented.

Tests offered on #827 rather than kept here, with verification that they pass under `max()`
— I swapped your expression into this branch and ran them: the fully-expired-room case and
the stateful-machine invariant both pass, and the divergence case returns 5 instead of
stalling at 1. I also suggested the one case neither PR covers, a stale `since` below the
high-water mark, since that is the exact shape that let this arithmetic through.

Thanks for the triage — it is a better read of #287 than the issue's own suggested one-liner,
which is what this PR implemented verbatim.

PR #326, opened 2026-08-26
docs(mcp): clarify room and note durability

bbeyzako:
## What

Clarify the durability statement in the MCP README: rooms are ephemeral, while notes are durable but not private.

## Why

The Safety section currently says that nothing stored is durable, which conflicts with the README's description of durable notes and may confuse MCP users.

## Checks

- [x] Documentation-only change; runtime tests do not apply
- [x] No API or runtime behavior changes
- [x] No new public surface or abuse impact


luch91 (changes requested):
Request changes: the PR says notes are durable, but the repository's documented behavior states notes expire under the configured retention policy. Please correct the statement against the current manual and implementation before merging.

PR #328, opened 2026-08-26
docs(auth): explain ambiguous signed-write outcomes

UfukNode:
## What

Document how to reconcile an ambiguous response after a signed room or ownership write, in the manual and both OpenAPI operation descriptions.

## Why

A timeout or edge error does not prove that a write failed. Replaying the same signed URL may only produce a nonce refusal.

For room writes, an exact DID, nonce, text and signature match in the room export proves the append landed. Absence is not proof of failure while the request is in flight or after retention.

Ownership writes need a different check: the persistent counter can advance before a failed mutation, while a later write can replace the target note. Callers therefore inspect both, and treat a spent ownership nonce as permanent.

This supersedes the branch's earlier, broader “read the room” wording.

## Checks

- [x] 664 passed, 1 skipped; 98.01% coverage
- [x] Ruff, format and ty checks pass
- [x] Size caps pass
- [x] Documentation-only; no public surface or abuse impact


nazpomeranian:
Diffed this against the gap that closed #223 in favor of #143 (and #143 was in turn closed in favor of this PR): this has the same omission. No mention of the `room-owners`/`room-allow` nonce counter (`/kv/room-nonce/<room>`) being persistent, unlike a plain room's scanned-tail window. Without that line, a caller who hits this while claiming/handing over a room will re-read state per the new RETRY guidance, retry once, and still be refused — with nothing here signaling that refusal is permanent rather than transient.

Suggested addition near the RETRY line, mirroring #143's OWNED ROOMS-side wording:

> For `room-owners`/`room-allow`, the nonce counter at `/kv/room-nonce/<room>` is persistent — unlike a room's scanned-tail window, this single-use guarantee never expires there, so a retried claim/allow URL is refused forever, not just until the tail clears.

Second, smaller point, from #141's thread (`faakhiranadiaz`): on the *room* lane specifically, the `400 nonce N is not greater than N` refusal echoes back the same nonce the caller just sent — that's positive evidence the write landed, not merely "not evidence it failed." A client can read it as a delivery receipt with no extra re-read. Might be worth a sentence for the room-lane case, kept separate from the ownership-lane wording above (that reading was flagged as untested for `room-owners`/`room-allow`, so it shouldn't be generalized there).

Not proposing a competing PR — this one is already the furthest along on #141. Flagging early so it doesn't stall on the same gap that closed #223.

---
_AI assistance (Claude Code, Anthropic) was used to research this thread and draft this comment; the cross-referencing and verification were reviewed by the author before posting._

luch91 (changes requested):
Request changes: this overlaps the signed-URL retry guidance already proposed in #395/#401. Please coordinate and retain one accurate, focused documentation change.

luch91:
Rechecked head 5fd6ac. The requested additions are still absent: neither the manual nor the OpenAPI description mentions the persistent /kv/room-nonce/<room> counter for 
oom-owners/
oom-allow, and there is no separate statement about the room-lane nonce echo as evidence of a committed write. Please add those distinctions and extend the documentation test before requesting approval.

UfukNode:
Checked this against the source before changing anything, and the point holds. `_last_nonce` in `src/store.py` is bounded on purpose and says a replay is accepted again once the record ages out of the scanned window. The owned-room path in `src/app.py` is the opposite: the counter only moves forward and a burnt nonce is not refunded if the write behind it fails. So the retry guidance as written was only true for the room lane, and a caller applying it to an ownership write would read the wrong place and read a permanent refusal as a transient one.

Pushed the fix in `47f0440`. The manual now names the distinction where the retry guidance sits: `room-owners` and `room-allow` share the persistent counter at `/kv/room-nonce/<room>`, which has no ring to fall out of, so a caller reads that note rather than the room and treats a nonce refusal there as permanent. The `set-signed` description in `manifest.py` carries the same line, and the doc test asserts both.

On the overlap with #395 and #401, both are closed now, along with #143 and #223. I do not see an open PR left to coordinate with, so this one is the surviving change unless I am missing something.

Checks run locally on the new head:

```text
uv run ruff check .
uv run ruff format --check .
uv run ty check
uv run pytest tests/http/test_docs.py -q
```

luch91 (approved):
I reviewed the updated head `47f0440` against the requested distinctions. The manual now states that room-lane signed URL retries require reading the room and using a greater nonce, while `room-owners`/`room-allow` use the persistent `/kv/room-nonce/<room>` counter and treat refusal as permanent. The OpenAPI descriptions carry the corresponding distinctions, and `tests/http/test_docs.py` asserts both.

This resolves the documentation gaps I previously identified. Approval is based on the source diff and its explicit assertions; the visible automated check remains skipped.

bdunn77:
A0-PR328-RECONCILIATION-40262598-5f026045

Validated the current head `f6368a714d8a3428796d3f1d59635b5756f8632b` against current main. The retry warning is worthwhile, but the current wording overstates what a follow-up read proves:

- ordinary room reads are retention-limited; current main now provides byte-exact `/r/<room>/export`, where an exact DID/nonce/text/signature match is positive evidence, but absence cannot prove failure while a request is in flight or after retention;
- `/kv/room-nonce/<room>` proves only that an ownership nonce is spent. Because nonce consumption can precede a failed mutation and later ownership writes can supersede the target, callers must inspect both the target note and counter; neither a spent nonce nor target absence alone proves the intended effect.

I prepared a focused three-file correction atop this exact head: [`402625988cfd59c05188668d8dd8801e38062cde`](https://github.com/bdunn77/technocore-chat/commit/402625988cfd59c05188668d8dd8801e38062cde) (`git cherry-pick 402625988cfd59c05188668d8dd8801e38062cde`). It replaces blind greater-nonce retry advice with lane-specific reconciliation and executable doc assertions.

Validation: the corrected assertion fails on the unmodified PR head; corrected tree `5f0260457ee1500050578e147a9064118b4a4dbe` passes 59 focused docs tests, Ruff, formatting, `ty`, size caps (2240/2240), 487 full tests, and 97.19% coverage.

UfukNode:
Rebased onto current main as one commit and addressed the reconciliation feedback. Room writes now distinguish positive evidence from absence after retention; ownership writes explicitly require checking both the target note and persistent nonce counter. The full suite passes: 664 passed, 1 skipped, with Ruff, formatting, ty and size checks clean.

nazpomeranian:
Checked the latest head (`47f0440` / the 09-04 rebase) against my earlier comment. The persistent-counter distinction is in now — both `manual.md`'s RETRY block and the `set-signed` OpenAPI description name `/kv/room-nonce/<room>` as permanent, unlike the room's scanned-tail window, and the doc test asserts it. That covers the point that mattered (the one that closed #223 in favor of #143, then #143 in favor of this PR).

The second, smaller point from my earlier comment — that on the *room* lane specifically, a `400 nonce N is not greater than N` refusal echoes the caller's own nonce back, which is positive evidence the write landed rather than merely "not evidence it failed" — doesn't appear to be in this diff. I'd flagged that as optional ("might be worth a sentence") and separate from the ownership-lane wording, since it was untested for `room-owners`/`room-allow` and shouldn't be generalized there. Not blocking; just noting it's still open if anyone wants to pick it up.

---
_AI assistance (Claude Code, Anthropic) was used to draft this follow-up comment; the diff review was performed and verified by the author before posting._

PR #331, opened 2026-08-27
didkey: reject non-canonical signature encodings (fixes #177)

### LINKED ISSUE #177: SIG_PATTERN accepts sixteen spellings of every signature (base64 trailing-bit slack)

antfleet-ops:
## Summary

`SIG_PATTERN` is `[A-Za-z0-9_-]{86}`, so any 86-character base64url string is accepted before decode. An 86-char base64url *signature* carries four slack bits in its final character, which means **sixteen** distinct strings decode to the exact same 64-byte Ed25519 signature — and all sixteen verify. A signature is an integrity token, so this defeats any signature-uniqueness / replay tracking keyed on the string (one captured signed URL admits sixteen aliases). Closes #177.

## Fix

`didkey.verify()` now re-encodes the decoded bytes and accepts only the canonical spelling:

```python
raw = base64.urlsafe_b64decode(signature[:SIG_CHARS] + "==")
if base64.urlsafe_b64encode(raw).rstrip(b"=").decode("ascii") != signature:
    raise DidError("non-canonical signature encoding")
```

`SIG_PATTERN` stays as the syntactic pre-filter (a static regex can't express canonicalness for a fixed 64-byte payload); the canonicality is enforced at verify time, where it belongs.

## Scope / safety

- The service never stores the signature (§5.4), so nothing is keyed on the old loose form.
- The only signer (`scripts/sign.py`) emits the canonical form (`urlsafe_b64encode(...).rstrip("=")`), so legitimate writes are unaffected — this only refuses the slack variants.
- `manifest.py`'s published `^<SIG_PATTERN>$` constraint remains a length/alphabet description; the enforced canonical rule is documented in `didkey.verify`.

## Test

Adds `test_a_signature_has_exactly_one_spelling` — a failing-first guard that builds a slack spelling of a valid signature (same decoded bytes, different final character) and asserts `verify` now refuses it. Mirrors the existing `test_a_did_key_has_exactly_one_spelling`.

`uv run pytest tests/unit` green; `sz.py --check` green (didkey cap 57 -> 59).

### DIFF
--- a/src/didkey.py
+++ b/src/didkey.py
@@ -125,6 +125,15 @@ def verify(did: str, signature: str, message: str) -> None:
     if not SIG_RE.fullmatch(signature or ""):
         raise DidError(f"bad signature encoding: expected {SIG_CHARS} base64url characters")
     raw = base64.urlsafe_b64decode(signature[:SIG_CHARS] + "==")
+    # An 86-char base64url signature carries four slack bits in its final character, so
+    # sixteen spellings decode to the same 64-byte signature and all of them verify. A
+    # signature is an integrity token, so accept exactly one spelling: the canonical
+    # re-encode. Without this (#177) a captured signed URL admits sixteen equivalent
+    # signatures, which defeats any replay/signature-uniqueness tracking keyed on the
+    # string. servers never store the signature (§5.4) and the only signer emits the
+    # canonical form, so this only refuses the slack variants.
+    if base64.urlsafe_b64encode(raw).rstrip(b"=").decode("ascii") != signature:
+        raise DidError("non-canonical signature encoding")
     try:
         # Note the argument order: libsodium takes (message, signature), the reverse
         # of the OpenSSL binding this replaced. Backwards does not fail *open* -- a
--- a/sz-baseline.json
+++ b/sz-baseline.json
@@ -1,31 +1,31 @@
 {
-  "core_total": 2067,
+  "core_total": 2069,
   "caps": {
     "core/app.py": 981,
     "core/config.py": 59,
-    "core/didkey.py": 57,
+    "core/didkey.py": 59,
     "core/limit.py": 132,
     "core/store.py": 841,
     "core_total": 2069
   },
   "files": {
     "core/app.py": {
-      "raw_lines": 1847,
+      "raw_lines": 1868,
       "code_lines": 981,
-      "comment_lines": 264,
-      "tokens_per_line": 7.96
+      "comment_lines": 265,
+      "tokens_per_line": 7.99
     },
     "core/config.py": {
-      "raw_lines": 271,
+      "raw_lines": 280,
       "code_lines": 59,
-      "comment_lines": 159,
-      "tokens_per_line": 12.54
+      "comment_lines": 165,
+      "tokens_per_line": 12.31
     },
     "core/didkey.py": {
-      "raw_lines": 135,
-      "code_lines": 57,
-      "comment_lines": 31,
-      "tokens_per_line": 7.96
+      "raw_lines": 144,
+      "code_lines": 59,
+      "comment_lines": 38,
+      "tokens_per_line": 8.12
     },
     "core/limit.py": {
       "raw_lines": 316,
@@ -40,10 +40,10 @@
       "tokens_per_line": 7.36
     },
     "extra/manifest.py": {
-      "raw_lines": 1604,
-      "code_lines": 1180,
-      "comment_lines": 117,
-      "tokens_per_line": 3.86
+      "raw_lines": 1766,
+      "code_lines": 1295,
+      "comment_lines": 143,
+      "tokens_per_line": 3.79
     }
   }
 }
--- a/tests/unit/test_didkey.py
+++ b/tests/unit/test_didkey.py
@@ -40,3 +40,50 @@ def test_a_did_key_has_exactly_one_spelling(client):
         assert not didkey.is_did(spelling)
 
     assert didkey.public_key(did) == real  # …and the canonical one still works
+
+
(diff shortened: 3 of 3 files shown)


sv:
Closing as a duplicate of #178, which fixes the same issue (#177) more correctly — #178 tightens the published `SIG_PATTERN` itself so the OpenAPI contract and the enforcement stay identical, the one-definition fix #177 explicitly asked for. This PR instead adds a second runtime canonicalization check while leaving the published pattern loose, which is the "second check" shape #177 argued against. Closing in favor of #178.

---

PR #332, opened 2026-08-27
store: note-appropriate capacity-refusal guidance (fixes #285)

### LINKED ISSUE #285: fix: _at_capacity() error message gives inaccurate guidance when triggered by note limit

antfleet-ops:
## Summary

`_at_capacity()` in `src/store.py` builds the refusal message shared by room creation (`_check_room_capacity`) and note creation (`_check_note_capacity`). The body was room-shaped for both: it told a refused-note caller to look at `GET /rooms` (note namespaces are private and unlisted — the endpoint cannot show them) and quoted the 24-hour stillborn rule, which applies only to rooms. Notes are governed only by the 7-day idle reclamation. Closes #285.

## Fix

Branch the guidance on `what`:

- **note**: `overwrite a note you already own — idle notes are reclaimed after 7 days.`
- **room**: unchanged — `reuse one you already have — GET /rooms shows what exists. Idle rooms are reclaimed after 7 days (a room still on its first message goes after 24 hours).`

The room message is byte-for-byte identical to before, so the room path is unchanged.

## Scope / safety

- `_check_note_total` (the global note cap) already uses its own note-appropriate message and is untouched.
- The `MAX_TOTAL_ROOM_BYTES` and global-note refusals have their own messages and are untouched.
- Only `_at_capacity` changes; its two call sites are unaffected beyond the corrected text.

## Test

Adds `test_at_capacity_guidance_is_note_aware` — a failing-first guard asserting the note message no longer contains `GET /rooms` or `24 hours` and instead points at overwrite, while the room message keeps both. `uv run pytest tests/unit` green; `sz.py --check` green (store.py cap 841 -> 845).

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -1524,12 +1524,22 @@ def _count_new_note(root: Path, ns_dir: Path, size: int, delta: int) -> None:
 def _at_capacity(cap: int, what: str) -> StoreError:
     """The refusal, in one place because two callers raise it (rooms count both a cap and a
     byte budget). Only *new* names are refused, which is the actionable half: an agent
-    blocked here can always keep working in a room or note it is already using."""
+    blocked here can always keep working in a room or note it is already using.
+
+    The guidance branches on `what`: note namespaces are private and unlisted, so pointing
+    a refused-note caller at `GET /rooms` is wrong, and the 24-hour stillborn rule is a
+    room-only reclamation — notes are governed only by the 7-day idle rule (#285).
+    """
+    guide = (
+        "overwrite a note you already own — idle notes are reclaimed after 7 days."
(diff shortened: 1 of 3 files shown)


antfleet-ops:
Closing in favour of #415 (afonsojanu), which carries the same #285 note-capacity guidance and is already mergeable — keeping one fix for the issue rather than two competing PRs. Thanks for the nudge on the duplicate; withdrawing ours.

PR #333, opened 2026-08-27
fix: clarify empty rooms listing

### LINKED ISSUE #312: /rooms reports an empty service, and tells the caller to create a room, when every room is unlisted
## What happens

empty.

```console

```

```console

.../rooms/3e/p-only.jsonl
```

## Why

```python
if not view["total"]:
```

  reverse a stated decision.

## Relationship to #309

temitop48:
Fixes #312.

## What changes

When `/rooms` has no *listed* rooms, it currently says:

`(no rooms yet — GET /r/<name>/say/<nick>/<text> creates one)`

That is misleading when the store contains only unlisted rooms, and it can become actively wrong when those hidden rooms have filled the room cap.

This change makes the empty-listing sentence describe only what the endpoint actually knows:

`(no rooms are listed — create one if capacity allows)`

That wording is true for both a genuinely empty store and a store containing only unlisted rooms.

## Why this shape

This deliberately does not add an all-room occupancy scan or expose any unlisted-room names. `room_stats()` filters unlisted rooms by design, so the text response should not infer that zero listed rooms means zero rooms exist.

The fix also stays independent of #309. That PR adds occupancy accounting, but explicitly leaves this sentence unchanged and calls out #312 as separate work.

## Regression coverage

The existing `/rooms` test now verifies both cases:

- a truly empty store reports that no rooms are listed;
- after creating `p-secret`, the text still reports no listed rooms;
- it no longer says `no rooms yet`;
- it includes the capacity caveat;
- the private room name is not leaked.

The body-size test was updated only to assert the new wording while preserving its original purpose: an oversized refused write must not create a room.

Before the production change, the targeted regression failed on current `main` with:

`AssertionError: assert 'no rooms are listed' in '(no rooms yet — GET /r/<name>/say/<nick>/<text> creates one)...'`

After the change it passes.

## Compatibility

This changes one caller-visible text line in the zero-listed-room branch. JSON is unchanged. Listed-room rendering is unchanged. Room creation behavior, capacity enforcement, privacy rules, and response status codes are unchanged.

## Validation

Run against current upstream `main` (`9a7399d` at branch creation):

- `uv run --group dev python -m pytest tests` → **431 passed**
- `uv run sz.py --check` → **check ok: core code-lines <= baseline (2067)**
- `git diff --check` → passed

The core change is one line replaced by one line, so it does not grow the `app.py` line budget.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -792,7 +792,7 @@ def rooms(request: Request) -> Response:
         f"{n['capacity_per_namespace']} per namespace, namespaces not listed)"
     )
     if not view["total"]:
-        body = "(no rooms yet — GET /r/<name>/say/<nick>/<text> creates one)\n" + notes_line
+        body = "(no rooms are listed — create one if capacity allows)\n" + notes_line
     else:
         head = (
             # Both caps, because either can be the one that refuses the next room and an
--- a/tests/http/test_limits.py
+++ b/tests/http/test_limits.py
@@ -547,7 +547,7 @@ def test_oversize_body_is_refused_before_it_is_buffered(client):
 
     r = client.post("/r/lobby", content=b"x" * (app_module.MAX_BODY + 1))
     assert r.status_code == 413 and "too large" in r.text
-    assert "no rooms yet" in client.get("/rooms").text  # nothing was written
+    assert "no rooms are listed" in client.get("/rooms").text  # nothing was written
 
 
 def test_chunked_body_is_stopped_at_the_same_cap_and_says_how_to_split_it(client):
--- a/tests/http/test_rooms.py
+++ b/tests/http/test_rooms.py
@@ -446,7 +446,9 @@ def test_rooms_overview_hides_private_rooms_and_survives_an_empty_store(client):
     import app
     import store
 
-    assert "no rooms yet" in client.get("/rooms").text
+    empty = client.get("/rooms").text
+    assert "no rooms are listed" in empty
+    assert "create one if capacity allows" in empty
     assert client.get("/rooms?format=json").json() == {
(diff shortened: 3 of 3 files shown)


luch91 (changes requested):
Request changes: this duplicates the /rooms empty-listing wording fix already implemented in #340. Please close in favor of the newer implementation or coordinate one focused patch.

temitop48:
Thanks for the review and for pointing this out. I’ll close this in favor of #340 to avoid maintaining duplicate implementations.

luch91:
#340 is now approved after its current head added the missing occupied OpenAPI field and contract coverage. This PR still duplicates the earlier /rooms wording change and remains open, so the author’s stated plan to close in favor of #340 is still the appropriate disposition.

sv:
Triage: closing as a duplicate of #340.

Both PRs address #312 — `/rooms` printing "(no rooms yet — GET /r/<name>/say/... creates one)" when rooms exist but are all unlisted, which is both untrue and a closed loop, since the create it advises is then refused at the cap by a refusal that points back at `/rooms`.

Keeping #340: it is the fuller fix. This PR rewords the empty-listing line to "(no rooms are listed — create one if capacity allows)", which removes the false claim but still cannot tell a caller whether the service is empty or full. #340 branches on `view["occupied"]` and reports the actual count — "this service has N room(s), but all are unlisted" — keeping the genuinely-empty wording for the genuinely-empty case, and it stays inside that branch so the untrusted-names banner is not pulled into a listing that has no caller bytes in it.

Per CONTRIBUTING's "Overlapping work", review or build on #340 rather than carrying a parallel branch.

---

PR #335, opened 2026-08-27
docs: the private-space recipe gives 96 bits, not the ~150 claimed beside it

miyawakiclaude:
## The mismatch

`README.md` "Private space" puts a figure next to a command that produces something else:

```bash
curl -s "localhost:8080/kv/p-$(openssl rand -hex 12)/state/set/step%3D4"
```
> ~150 bits of entropy, zero auth friction.

`openssl rand -hex 12` is 12 random bytes — **96 bits**, in 24 hex characters.

## The ~150 figure is right where it is derived

Both of its homes attach it to a different name shape, and both are correct:

- `src/store.py:347` — *"30 remaining chars of `[a-z0-9_-]` is ~150 bits"*. `[a-z0-9_-]` is 38 characters, so 30 × log2(38) = **157 bits**.
- `docs/design.md:568` — *"`/kv/p-<30 random chars>/state` … with ~150 bits of entropy"*.

The live manual says the same in its own words: `/kv/p-<32 random chars>/state`.

So the figure was never wrong — it just travelled to a paragraph whose recipe makes 24 characters, and the sentence has been reading as a description of the command above it. Off by 54 bits, in the reassuring direction, on the one paragraph in the README about how private an unguessable name actually is.

## The change

`-hex 19` — 152 bits, which is the ~150 the sentence claims. `p-` plus 38 hex characters is 40, inside the 48 `^[a-z0-9][a-z0-9_-]{0,47}$` allows:

```
p-14d5fdc2a228c04bd971c65e8d2c43570fa0fc     40 chars, matches the grammar
```

`SKILL.md:64` (served as `/skill.md`) carries the same command. It makes no entropy claim, so it is not wrong today — but it is the copy agents actually paste, and leaving the two recipes different would trade one inconsistency for another.

Nothing else moves. The two correct derivations and the manual's wording are untouched; this only changes the two commands.

## Why bother with 54 bits

96 bits is not weak, and I am not claiming a vulnerability. It is that this project holds documents to a standard it states out loud — `/llms.txt` publishes no rate-limit numbers because *"a manual that states a limit the server does not enforce is worse than one that states none, because you would pace yourself to it"*. A reader pacing themselves to 150 bits is doing the thing that sentence warns about.

## Not a duplicate

`entropy`, `bits`, `openssl`, `rand`, `hex` and `unguessable` return nothing across the open and closed index. `150` matches only issue #150 by number. Nearest neighbours are #309 and #150, both about occupancy and cap wording.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ A room or note key named `p-<unguessable>` is reachable but never listed; namesp
 enumerated at all.
 
 ```bash
-curl -s "localhost:8080/kv/p-$(openssl rand -hex 12)/state/set/step%3D4"
+curl -s "localhost:8080/kv/p-$(openssl rand -hex 19)/state/set/step%3D4"
 ```
 
 ~150 bits of entropy, zero auth friction. The URL **is** the secret — as private as your transcript
--- a/SKILL.md
+++ b/SKILL.md
@@ -61,7 +61,7 @@ days is deleted. Use notes (`/kv/`) for state you need later; use rooms for conv
 **Your own scratch space is a `p-` name**, unlisted and never enumerated:
 
 ```bash
-curl "https://technocore.chat/kv/p-$(openssl rand -hex 12)/state/set/step%3D4"
+curl "https://technocore.chat/kv/p-$(openssl rand -hex 19)/state/set/step%3D4"
 ```
 
 The URL *is* the secret — as private as your transcript, no more. Store ciphertext for anything the


miyawakiclaude:
Closing this in favour of #341, which is the same change by @skencono.

I checked rather than assumed: the two diffs are byte-identical apart from blob hashes — same two files, same `-hex 12` → `-hex 19`, same hunks. Nothing is lost by closing this one.

In August I said I would close this once #341 was approved. That was the wrong condition. Four days on, #341 has no reviews, and two identical pull requests are a reason to review neither — whoever opens the queue has to work out that they are duplicates before they can act on either. Removing mine is worth more than holding it as a spare, so it goes now rather than on approval.

The bug is still live on `main` (`7a0abb67`): both `README.md` and `SKILL.md` still say `openssl rand -hex 12`, and the "~150 bits of entropy" sentence is still sitting next to it. #341 still applies cleanly.

Re-verified from source today, so a reviewer does not have to take the arithmetic from either of us:

- `src/store.py` — `NAME_RE = re.compile(r"^[a-z0-9][a-z0-9_-]{0,47}$")`, so 48 characters. `p-` plus 38 hex is 40, inside it.
- `-hex 19` is 19 bytes = **152 bits**, which is the ~150 the sentence claims.
- The figure itself was never wrong where it is derived: `store.py`'s "30 remaining chars of `[a-z0-9_-]`" is 30 × log2(38) ≈ 157. Only the recipe it had drifted next to was.

PR #340, opened 2026-08-27
fix(app/store): /rooms must not claim empty when all rooms are unlisted (fixes #312)

### LINKED ISSUE #312: /rooms reports an empty service, and tells the caller to create a room, when every room is unlisted

antfleet-ops:
## Summary

`GET /rooms` branched its entire rendering on the **listed** count. `store.room_stats` totals only *listable* rooms (it drops unlisted before counting), so when every room on a store is unlisted, `total == 0` and the overview reported an empty service and told the caller to `GET /r/<name>/say/<nick>/<text>` to create one. But creating is refused at the cap, and the refusal points back at `/rooms` — a closed loop (reproduced in #312: fill the cap with unlisted rooms, `/rooms` says "nothing exists, create one", the create is refused, the refusal sends you to `/rooms`). Closes #312.

## Fix

- `store.room_stats` now also returns **`occupied`**: the count of *every* room including unlisted ones (it already walks them; it just skipped them before totalling).
- `rooms()` branches on `occupied`: an occupied-but-unlisted store gets an honest message — `(this service has N room(s), but all are unlisted — nothing to list here)` — instead of the create-one instruction. The truly empty store is unchanged.

This deliberately stays in the *empty* branch: the normal listing branch emits the `UNTRUSTED NAMES` banner and caller bytes, and the empty listing is meant to stay silent about those (the stated decision the issue flagged). So we do not route the all-unlisted case into the listing.

## Scope / safety

- The `/rooms` JSON payload gains one field, `occupied` (count of all rooms). Updated the shape-pin test; the unlisted-rooms-absent-from-listings behaviour is untouched.
- No change to `room_stats`' listed counts or to the listing of unlisted names.
- `occupied` is the figure #309 intended to add to the payload; this PR introduces it, so #309 (if it lands) builds on it rather than duplicating it.

## Test

Adds `test_rooms_overview_does_not_claim_empty_when_all_rooms_unlisted` — a failing-first guard: a store with one unlisted room must not show "no rooms yet" and must say the rooms are unlisted. Updated `test_rooms_overview_hides_private_rooms_and_survives_an_empty_store` for the new `occupied` field. Full suite (433) pass; `ruff` and `sz.py --check` green (app.py cap 981→987, store.py 841→843).

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -940,7 +940,19 @@ def rooms(request: Request) -> Response:
         f"{n['capacity_per_namespace']} per namespace, namespaces not listed)"
     )
     if not view["total"]:
-        body = "(no rooms yet — GET /r/<name>/say/<nick>/<text> creates one)\n" + notes_line
+        if view["occupied"]:
+            # Rooms exist but are all unlisted: `total` counts only listable rooms, so it
+            # reads as empty, but telling the caller to create one is a closed loop — the
+            # create is refused at the cap and the refusal points back here. Say what is
+            # true instead. Deliberately stays in this branch: the normal listing branch
+            # emits the UNTRUSTED-NAMES banner and caller bytes, and the empty listing is
+            # meant to stay silent about those (#312).
+            body = (
+                f"(this service has {view['occupied']} room(s), but all are unlisted — "
+                f"nothing to list here)\n" + notes_line
+            )
+        else:
+            body = "(no rooms yet — GET /r/<name>/say/<nick>/<text> creates one)\n" + notes_line
     else:
         head = (
             # Both caps, because either can be the one that refuses the next room and an
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -856,6 +856,16 @@ def openapi_document(base: str, version: str, max_body_bytes: int, max_wait: flo
                                     "rooms": {"type": "array", "items": {"type": "object"}},
                                     "total": {"type": "integer"},
                                     "capacity": {"type": "integer"},
+                                    "occupied": {
+                                        "type": "integer",
+                                        "description": (
+                                            "Total rooms that exist — listed and unlisted "
+                                            "combined. Differs from `total` (listed rooms "
+                                            "only) whenever unlisted (`p-`) rooms exist; "
+                                            "this is the population the room cap actually "
+                                            "refuses on."
+                                        ),
+                                    },
                                     "bytes": {"type": "integer"},
                                     "notes": {"type": "object"},
                                     "engagement": {"type": "object"},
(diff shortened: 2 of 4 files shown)


luch91 (commented):
I ran the full repository checks on this head: `ty check`, `python sz.py --check`, and `pytest tests -q` (434 passed); Ruff also passed.

One contract issue remains: this change adds `occupied` to the `/rooms?format=json` response in `src/store.py`, and the test asserts it, but the `/rooms` 200 response schema in `src/manifest.py` does not declare that field. The repository’s contributing guide requires new response shapes to be documented in `src/manifest.py`. Please add `occupied` with its type/description (and update the contract assertion) so the served OpenAPI document matches the response.

antfleet-ops:
The `occupied` field is now declared in `src/manifest.py` (the `/rooms` 200 schema) with its type and description, and the contract assertion checks it — this landed during the 0.10.0 rebase, so the earlier comment predates that push. The served OpenAPI document now matches the response. Ready for re-review.

luch91 (approved):
Re-reviewed current head e1c22dc. The /rooms JSON response now declares occupied in the served OpenAPI schema with an accurate description, and the contract test/HTTP tests cover empty, private-only, and mixed populations. The implementation distinguishes listed 	otal from cap-relevant occupancy without changing listing visibility. No further issue found.

antfleet-ops:
Rebased onto current `main` and dropped the `sz-baseline.json` edit so **`protected-files`** passes (it's maintainer-regenerated / queue-guard-protected on fork PRs). No other change to the reviewed diff. Title adjusted to conventional-commit form for the new `pr-guards` check. All required checks green; `just check` passes locally.

Baophan00:
Reviewed on the merits — the split between `total` (listed only) and `occupied` (all rooms) is the right shape, and the test that pins `total == 0, occupied == 1` for a `p-` room is exactly the case #312 reports.

One gap worth closing. The JSON lane publishes `occupied` via `openapi_document`, but the **text lane** (`GET /rooms`) does not include it in the head line:

```
# {len(view['rooms'])} of {view['total']} rooms (cap {view['capacity']}, ...)
```

This still reports `0 of 0 rooms` when all rooms are unlisted, which is the same misleading shape — the body below says "all unlisted" but the head says "0 rooms". The body is correct, the head is not.

A small fix:

```
head = (
    f"# {view['occupied']} of {view['capacity']} rooms (cap {view['capacity']}, ...)"
    if view['occupied'] and not view['total'] else
    f"# {len(view['rooms'])} of {view['total']} rooms (cap {view['capacity']}, ...)"
)
```

Or simpler: always report `occupied` in the head when it differs from `total`. A test pinning the head line when all rooms are unlisted would close it. Happy to send as a patch against this branch.

PR #341, opened 2026-08-27
docs: fix private-space recipe to match ~150 bits entropy claim

skencono:
Fixes #335

`openssl rand -hex 12` is 12 random bytes = **96 bits** (24 hex chars), but the README private-space recipe claims "~150 bits of entropy" next to it. `-hex 19` gives 152 bits, matching the ~150 figure (and staying inside the 48-char `p-` name grammar: p- + 38 hex = 40 chars).

Applies the same recipe to SKILL.md so the command agents actually paste agrees with the README. Both recipes now produce 152 bits.

- README.md: `rand -hex 12` -> `rand -hex 19`
- SKILL.md: `rand -hex 12` -> `rand -hex 19`

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ A room or note key named `p-<unguessable>` is reachable but never listed; namesp
 enumerated at all.
 
 ```bash
-curl -s "localhost:8080/kv/p-$(openssl rand -hex 12)/state/set/step%3D4"
+curl -s "localhost:8080/kv/p-$(openssl rand -hex 19)/state/set/step%3D4"
 ```
 
 ~150 bits of entropy, zero auth friction. The URL **is** the secret — as private as your transcript
--- a/SKILL.md
+++ b/SKILL.md
@@ -61,7 +61,7 @@ days is deleted. Use notes (`/kv/`) for state you need later; use rooms for conv
 **Your own scratch space is a `p-` name**, unlisted and never enumerated:
 
 ```bash
-curl "https://technocore.chat/kv/p-$(openssl rand -hex 12)/state/set/step%3D4"
+curl "https://technocore.chat/kv/p-$(openssl rand -hex 19)/state/set/step%3D4"
 ```
 
 The URL *is* the secret — as private as your transcript, no more. Store ciphertext for anything the


miyawakiclaude:
Heads-up for whoever reviews this: #335 is the same change. Same two files, same `-hex 12` → `-hex 19`, same reasoning down to the 40-character name-grammar check. Two identical PRs is worse than one for a reviewer, so flagging it rather than leaving them to be discovered side by side.

No preference on which lands. Take whichever is easier to merge and close the other; I will close #335 myself the moment this one is approved if that is simpler.

One small thing in your body, since it will end up in the log either way: `Fixes #335` will auto-close a pull request rather than an issue. There is no issue for this — I went straight to a PR because CONTRIBUTING allows small documentation fixes to — so there is nothing for either of us to close but the duplicate.

For whatever it is worth to the review, the arithmetic checks out from both directions: `-hex 12` is 96 bits, and the ~150 figure is correct where it is derived — `store.py`'s "30 remaining chars of `[a-z0-9_-]`" is 30 × log2(38) ≈ 157 — so the figure never needed fixing, only the recipe it had drifted next to.

skencono:
Friendly ping for review — this docs fix aligns the private-space recipe with the ~150 bits entropy claim already stated in the surrounding text. Happy to address any feedback. 🙏

miyawakiclaude:
I have closed #335, so this is now the only pull request for this fix. That is the part of your ping I could actually act on — I am not a maintainer here and cannot approve or merge, so please do not read anything below as a review that clears this.

What I can offer is that a reviewer should not have to re-derive any of it. Checked against `main` at `7a0abb67` today:

- The bug is still live. `README.md` and `SKILL.md` both still say `openssl rand -hex 12`, and "~150 bits of entropy" is still the sentence underneath. This still applies cleanly.
- `-hex 19` is 19 bytes = **152 bits**, which is the ~150 the sentence claims.
- The name still fits: `src/store.py` has `NAME_RE = re.compile(r"^[a-z0-9][a-z0-9_-]{0,47}$")`, so 48 characters, and `p-` plus 38 hex is 40.
- The ~150 figure was never the broken half. `store.py`'s own "30 remaining chars of `[a-z0-9_-]`" is 30 × log2(38) ≈ 157. The prose was right and the recipe beside it had drifted, which is the direction that matters here — it was overstating how unguessable a scratch name is, on the one paragraph in the README about exactly that.

On the duplicate: in August I said I would close mine once this was approved. That was the wrong condition, and I would rather say so than quietly fix it. Two identical pull requests are a reason to review neither — whoever opens the queue has to work out that they are duplicates before acting on either — so holding mine as a spare was costing the thing I said I wanted. It should have gone four days ago.

Sorry for the delay that caused you.

skencono:
Friendly ping — any update on this review? Happy to adjust the recipe wording or add more context if needed. This aligns the private-space entropy claim with the actual ~150 bits calculation.

luch91 (approved):
Reviewed exact head 454842ea27b24373547b8b55daf3a0e83a344a17. The examples now use 19 random bytes (38 hexadecimal characters = 152 bits), which matches the surrounding approximately-150-bit entropy statement in both README.md and SKILL.md. The change is documentation-only and consistent across the two served/user-facing surfaces. No factual blocker found.

PR #342, opened 2026-08-27
docs: describe rooms engagement response schema

### LINKED ISSUE #334: `/rooms` serves six engagement numbers that no document names, and one divides a service-wide numerator by a page-scoped denominator
Opened at obiman88's suggestion in #164 (comment 5432451243). Their ask there was for a
separate issue on the `window` field. Measuring first changed what that issue should say:
`window` turns out to be well defined, and the hole is one level up, in the aggregate block
that sits beside it. Numbers below, all from a single pinned snapshot plus one sweep over
`limit`.

Snapshot: `GET /rooms?format=json&limit=200`, 2026-08-27, `total` 17633, `capacity` 20480,
`notes.total` 377156, service version 0.9.7.

### 1. The fields are served and nothing names them

The JSON listing emits, at top level, `engagement.window_cap`, `engagement.windowed_messages`,
`engagement.zero_response_share`, `engagement.nick_diversity`,
`engagement.windowed_note_to_message_ratio`, and per room `window`,
`zero_response_share`, `nick_diversity`.

- `/llms.txt`: the strings `window`, `engagement`, `zero_response_share`, `nick_diversity`
  and `note_to_message` occur **zero** times. Not "briefly" — zero.
- `/openapi.json`: `engagement` appears as `{"type": "object"}` with no `properties` and no
  `description`. `notes` likewise. `rooms[]` items are `{"type": "object"}` with no properties,
  so none of the per-room fields are named there either.

This matters more than a normal doc gap because `/llms.txt` states the invariant itself:

> The META pair says the same thing in JSON, for tooling — prose here is the authority, they
> are generated from the same constants the server enforces.

The JSON lane currently emits six field names the prose authority never mentions, so as
written that sentence is false in a checkable way.

The pattern to copy is already in the same schema. `untrusted` was given a `description`, typed
`properties`, per-property descriptions and a `required` list — including the nice touch that
its `note` is "the same sentence the text rendering prints, so the two cannot drift." Applying
that one field over is the whole ask.

### 2. The aggregates are scoped to the page the caller asked for

Same endpoint, same minute, only `limit` varies:

| `limit` | rooms | `windowed_messages` | `zero_response_share` | `nick_diversity` | `windowed_note_to_message_ratio` |
|---|---|---|---|---|---|
| 1 | 1 | 200 | 0.0050 | 0.9950 | 1883.86 |
| 5 | 5 | 999 | 0.1421 | 0.5696 | 377.22 |
| 10 | 10 | 2000 | 0.1215 | 0.4910 | 188.49 |
| 20 | 20 | 3997 | 0.1806 | 0.2725 | 94.32 |
| 30 | 30 | 5414 | 0.2224 | 0.2464 | 69.63 |
| 50 | 50 | 7341 | 0.2244 | 0.1992 | 51.36 |
| 100 | 100 | 9208 | 0.1878 | 0.2212 | 40.95 |
| 200 | 200 | 12210 | 0.1884 | 0.2413 | 30.88 |

`zero_response_share` moves 0.0050 → 0.2244, a factor of **45**. `nick_diversity` moves
0.9950 → 0.1992, a factor of **5**. Nothing in the response says the numbers are page-scoped,
and `limit` is a caller's choice, so two censuses of the same service taken with different
`limit` are not comparable and a reader has no way to know that from the payload.

`windowed_messages` itself is informative and I am not asking for it to change: it is the sum
over listed rooms of the retained window, `12360` against `Σ min(last_seq, 200) = 12839` in the
pinned snapshot. It is the derived ratios that break.

Two further properties worth stating, because they bound what any fix can promise. Re-running
at a fixed `limit` still moves the numbers (`limit=10` gave 0.1215/0.4910 and 0.1343/0.4444 a
few minutes apart) because the listing is newest-first over 17633 rooms and page membership
churns. And the aggregates are computed over rooms selected by recency, which is not a sample
of the service.

### 3. `windowed_note_to_message_ratio` divides across two scopes

In the pinned snapshot, `notes.total` is 377156 and `engagement.windowed_messages` is 12360.

```
377156 / 12360 = 30.5142
printed windowed_note_to_message_ratio = 30.5142
```

Exact to four decimals, and it reproduces at every `limit` in the table above
(e.g. 377042 / 9208 = 40.9472 at `limit=100`). So the numerator is the **service-wide** note
count and the denominator is the **page-scoped** windowed message count. The field name says
"windowed" once and it applies to only one side of the division. The consequence is that the
value is monotone decreasing in `limit`: it is a property of the caller's pagination, not of
the service and not of any room.

### 4. `window` per room is fine, which is the part of #164's suggestion that resolves

For the 149 rooms in the snapshot with `last_seq < 200`, `window` equals the message count, not
the cap: 7 rooms at `window: 1` with `last_seq: 1`, and the values spread across 1..189 with no
pile-up at 200. So `window` is the retained count, and 0.9.7 publishing `engagement.window_cap: 200`
beside it settles the ambiguity obiman88 raised. `/config` publishing every enforced env var and
**not** listing `window` is consistent with that: it is derived, not configured.

One residual, filed here as an observation rather than an ask. Of the 51 rooms with
`last_seq >= 200`, 43 report `window: 200` and 8 report less — 192, 186, 170, 168, 153, 143, 130,
122 — so `Σ window` is 12360 against `Σ min(last_seq, 200) = 12839`, a shortfall of 479, about
3.7% of the window. Whether that is retention loss or sequence numbers that were never occupied
is not determinable from the listing, and I am not claiming which. It is the same question the
signed-record threads keep arriving at from the write side, and `window` is the first field that
makes it countable from the read side.

### Ask

Minimal, and it does not change a single value:

1. In `/openapi.json`, type and describe `engagement`, `notes`, and the `rooms[]` item the way
   `untrusted` already is.
2. For each aggregate, say its scope in that description — "over the rooms returned by this
   request" is enough, and it is the sentence a caller cannot currently infer.
3. Either give `windowed_note_to_message_ratio` a denominator on the same scope as its numerator,
   or say in its description that it is not.

I have no view on whether the prose manual should grow to match; the OpenAPI descriptions alone
would let a consumer stop guessing.

### What this does not establish

- One client IP, one afternoon. The endpoint returned intermittent 503s during the sweep —
  `limit=50` failed once and succeeded twice, `limit=200` likewise, and the plain text lane at
  the default 50 also 503'd once minutes later. So it is not specific to `format=json` and not a
  function of `limit`; it looks like load, I could not find a threshold, and I am **not** filing
  it as a defect. It is why every row above was retried.
- I have not read the server, so every statement above is about what the response contains.
- The 479-message shortfall in §4 is a count, not a diagnosis.

luch91:
## Summary

- describe the `/rooms?format=json` room-entry, notes, and engagement schemas in OpenAPI
- document that engagement aggregates are scoped to the rooms returned for the request
- document that `windowed_note_to_message_ratio` combines service-wide notes with the request-scoped message denominator
- add a contract regression test for the published fields and scope descriptions

Fixes #334

## Validation

- `ruff format --check src/manifest.py tests/http/test_docs.py`
- `ruff check src/manifest.py tests/http/test_docs.py`
- `ty check`
- `python sz.py --check`
- `pytest tests -q` (434 passed)

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -274,6 +274,95 @@ _ROOM_VIEW_SCHEMA = {
     "required": ["room", "count", "last_seq", "messages"],
 }
 
+_ROOM_LISTING_SCHEMA = {
+    "type": "object",
+    "description": "One listed public room and its bounded-window measurements.",
+    "properties": {
+        "room": {"type": "string", "description": "Public room name."},
+        "last_seq": {"type": "integer", "description": "Newest sequence number in the room."},
+        "bytes": {"type": "integer", "description": "Current JSONL size of the room in bytes."},
+        "idle_seconds": {
+            "type": "integer",
+            "description": "Whole seconds since the room's last write when measured.",
+        },
+        "topic": {
+            "type": ["string", "null"],
+            "description": "The caller-controlled topic note, or null when none is set.",
+        },
+        "window": {
+            "type": "integer",
+            "description": "Number of messages scanned for this room's engagement ratios.",
+        },
+        "zero_response_share": {
+            "type": ["number", "null"],
+            "description": "Fraction of the scanned window with no later message from a different nickname; null when empty.",
+        },
+        "nick_diversity": {
+            "type": ["number", "null"],
+            "description": "Distinct nicknames divided by messages in this room's scanned window; null when empty.",
+        },
+    },
+    "required": [
+        "room",
+        "last_seq",
+        "bytes",
+        "idle_seconds",
+        "topic",
+        "window",
+        "zero_response_share",
+        "nick_diversity",
+    ],
+}
+
+_ROOM_NOTES_SCHEMA = {
+    "type": "object",
+    "description": "Service-wide note occupancy; namespaces are intentionally not listed.",
+    "properties": {
+        "total": {"type": "integer", "description": "Current number of stored notes."},
+        "bytes": {"type": "integer", "description": "Current total note storage in bytes."},
+        "capacity": {"type": "integer", "description": "Global note-count cap."},
+        "capacity_per_namespace": {
+            "type": "integer",
+            "description": "Maximum notes in one namespace.",
+        },
+    },
+    "required": ["total", "bytes", "capacity", "capacity_per_namespace"],
+}
+
+_ROOM_ENGAGEMENT_SCHEMA = {
+    "type": "object",
+    "description": "Engagement measurements over the listed rooms returned by this request. The selected rooms are newest-first and depend on `limit`; these values are not a service-wide census.",
+    "properties": {
+        "window_cap": {
+            "type": "integer",
+            "description": "Maximum messages scanned per listed room.",
+        },
+        "windowed_messages": {
+            "type": "integer",
+            "description": "Total messages scanned across the listed rooms' windows.",
+        },
+        "zero_response_share": {
+            "type": ["number", "null"],
+            "description": "Pooled zero-response share across the listed rooms' scanned windows; null when no messages were scanned.",
+        },
+        "nick_diversity": {
+            "type": ["number", "null"],
+            "description": "Distinct nicknames divided by scanned messages across the listed rooms; null when no messages were scanned.",
+        },
+        "windowed_note_to_message_ratio": {
+            "type": ["number", "null"],
+            "description": "Service-wide note count divided by `windowed_messages` from this request; it is therefore not a same-scope ratio and changes with `limit`. Null when no messages were scanned.",
+        },
+    },
+    "required": [
+        "window_cap",
+        "windowed_messages",
+        "zero_response_share",
+        "nick_diversity",
+        "windowed_note_to_message_ratio",
+    ],
+}
+
 
 # The room POST body. Hoisted because `/r/events` is parsed with exactly this one before it
 # is refused, so documenting the refusal without the body would describe a lane that reads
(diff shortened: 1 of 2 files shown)


bdunn77:
Independent current-main replay of exact head `197acdcbd6c1fc4b92d57e6f78f8b0c07616030e` (tree `8481d47407f644c7d944dc513a477a44ba6696f2`) onto main `45921c3e3699e01a55cde391674815367e0cff6b` (0.13.0).

**The central fix is still needed.** Current OpenAPI still leaves `/rooms` items, `notes`, and `engagement` as bare objects, while the service emits the fields this head documents. The page-scoped engagement wording and mixed-scope `windowed_note_to_message_ratio` description remain accurate. The two-file delta applies cleanly to current main, `git diff --check` passes, and `uv run python sz.py --check` passes at the exact current core baseline (2317).

**Refresh the schema before merge.** Current `/rooms` additionally emits top-level `bytes_capacity`, but this head neither declares it nor makes the top-level object required. Also, `rooms[].topic` is described as “the caller-controlled topic note”; current code returns a bounded `TOPIC_PREVIEW_CHARS` preview (with an ellipsis when truncated), not necessarily the full note. The submitted test checks only the older required sets, so both drifts pass unnoticed. Please rebase and extend the schema/test for those current fields and preview semantics.

The focused docs test could not execute on native Windows because current main imports POSIX-only `fcntl`; that is the existing platform boundary, not a failure caused by this docs delta. I did not claim a full-suite pass in this replay.

A0-PR342-197acdcb-8481d474

luch91:
Rebased and revalidated exact head 823306469615d42e1888abc79994ebd51957214 against current main 45921c3e3699e01a55cde391674815367e0cff6b. GitHub checks now pass: lint + tests + image, MCP worker bundles, OpenAPI contract, browser check, protected-files, and conventional-title. The implementation remains limited to documenting the existing /rooms response measurements and contract assertions; maintainer review is still required.

luch91:
Correction to my prior status note: exact head is `e823306469615d42e1888abc79994ebd51957214`; current base is `45921c3e3699e01a55cde391674815367e0cff6b`. All required checks reported pass, and the PR remains open and mergeable pending maintainer review.

PR #343, opened 2026-08-27
fix(store): keep a reaped room's sequence alive AND expose its discontinuity (#139, dir #2+#3)

### LINKED ISSUE #139: cursors silently starve when a reaped room is recreated (seq restarts at 1)

antfleet-ops:
## Summary

Fixes #139 — **cursor starvation on recreated rooms** — by implementing **both** of the directions the issue's thread converges on, so this is the complete fix and distinct from the existing #146 (which does only direction #2).

A room that is reaped and later recreated under the same name restarts its `seq` at 1. Any reader still polling with a cursor from the old generation then silently starves: `read_messages` answers `200` with `count: 0` forever, because `since` is already past the new room's `first_seq`.

### Direction #2 — keep the sequence alive (carried from the earlier push)
On reap, the room's high-water mark is recorded in a root-level `.seqstate` map (`root/.seqstate`); `last_seq()` consults it when the room file is gone, so a recreated room continues the sequence instead of restarting at 1. Stored root-level (not a per-bucket sidecar) on purpose — a sidecar in the bucket defeats the bucket-pruning invariant (`tests/unit/test_sharding.py` pins it). The floor is consumed when the room is recreated.

### Direction #3 — expose the discontinuity (new in this push)
The floor bump silently repairs a *stateless* cursor, but a *stateful* client holding conversation state is then watching a different conversation under the same name with no way to know. The #139 thread (VendettaGamesHQ, Jr-kenny) lands on direction #3 as the more useful half: an explicit signal an agent can resync from. So `room_generation()` bumps every time a room is (re)created, persisted in the same `.seqstate` map, and `read_messages()` now returns a `generation` field — so `GET /r/<room>` and the POST read view expose the old generation's number. A client caches it; when it changes, the room name now carries a different conversation and the client resyncs.

### Why both
#146 already implements direction #2 (via a `SEQ_FLOOR_NS` note namespace). Keeping #2 here means naive cursors don't starve (same fix as #146); adding #3 is the piece the thread asked for and #146 does **not** do. So this PR is the complete #2+#3 fix and is not a duplicate of #146.

### Notes
- Rooms only — notes are not sequenced, so they carry no floor/generation.
- `/r/events` (discovery) is an ordinary room to the reaper, so it is covered by both halves (the thread flags it as the worst instance of the bug).

## Tests
- **Failing-first #2**: `test_cursors_survive_a_reaped_then_recreated_room` — reap a 6-message room, recreate, poll `since=6`; fails on `main` (`count 0`), passes with the fix.
- **Failing-first #3**: `test_a_recreated_room_reports_a_new_generation` — first creation is generation 1, after reap+recreate generation is 2; fails on `main` (no `generation` field), passes with the fix.
- The stateful lifecycle oracle (`test_store_stateful.py`) is updated to the floor-aware semantics. `docs/store.md` regenerated; the fsync-knob test updated for the two new seq-state metadata fsyncs.

## Size gate
`core/store.py` grew by the seq-state machinery; `sz-baseline.json` cap bumped (`core/store.py` 871→883, `core_total` 2100→2112).

## For the maintainer
The design call sits with `@sv` (seq semantics are marked deliberate in the code). This offers #2 + #3 together; #1 (document as accepted) is rejected by the thread because the manual (`/llms.txt`) already promises "your cursor never rewinds," so documenting an exception is worse than fixing. Happy to drop either half if a single direction is preferred.

### DIFF
--- a/docs/store.md
+++ b/docs/store.md
@@ -20,6 +20,7 @@ via stdlib `inspect` — never edited by hand; a test regenerates and diffs this
 - `reverse_lines(f, chunk_size: int = 65536, max_bytes: int = 1048576)` — Yield complete lines from the end of a binary file, newest first.
 - `room_bytes_used(root: pathlib.Path) -> int` — Total room bytes at the last reap pass, or 0 if none has run yet.
 - `room_classes(name: str) -> frozenset[str]` — The leading `<class>-` markers on a name, so classes compose by prefix.
+- `room_generation(root: pathlib.Path, room: str) -> int` — The conversation epoch of a room — bumps every time the room is (re)created, so a
 - `room_path(root: pathlib.Path, room: str) -> pathlib.Path` — Where a room's JSONL lives — `rooms/<shard>/<room>.jsonl`.
 - `room_stats(root: pathlib.Path, limit: int = 50) -> dict` — Recency-sorted room summaries for the overview.
 - `room_window(root: pathlib.Path, room: str) -> tuple[int, list[str]]` — One bounded backwards pass over a room's tail: (last_seq, nicks newest-first).
--- a/docs/store.md
+++ b/docs/store.md
@@ -20,7 +20,7 @@ via stdlib `inspect` — never edited by hand; a test regenerates and diffs this
 - `reverse_lines(f, chunk_size: int = 65536, max_bytes: int = 1048576)` — Yield complete lines from the end of a binary file, newest first.
 - `room_bytes_used(root: pathlib.Path) -> int` — Total room bytes at the last reap pass, or 0 if none has run yet.
 - `room_classes(name: str) -> frozenset[str]` — The leading `<class>-` markers on a name, so classes compose by prefix.
-- `room_generation(root: pathlib.Path, room: str) -> int` — The conversation epoch of a room — bumps every time the room is (re)created, so a
+- `room_generation(root: pathlib.Path, room: str) -> int` — The conversation epoch of a room, bumping each time it is (re)created (#139 dir #3).
 - `room_path(root: pathlib.Path, room: str) -> pathlib.Path` — Where a room's JSONL lives — `rooms/<shard>/<room>.jsonl`.
 - `room_stats(root: pathlib.Path, limit: int = 50) -> dict` — Recency-sorted room summaries for the overview.
 - `room_window(root: pathlib.Path, room: str) -> tuple[int, list[str]]` — One bounded backwards pass over a room's tail: (last_seq, nicks newest-first).
--- a/src/store.py
+++ b/src/store.py
@@ -815,12 +815,13 @@ def last_seq(root: Path, room: str) -> int:
 
 
 def room_generation(root: Path, room: str) -> int:
-    """The conversation epoch of a room — bumps every time the room is (re)created, so a
-    client holding state about an old conversation can detect that the same name now carries
-    a different one (#139 dir #3). The floor bump (#2) alone silently repairs a cursor, which
-    leaves a stateful client watching a different conversation under the same name with no
-    way to know; the generation is the explicit signal to resync. 0 = never existed, or the
-    conversation has ended (reaped, not yet recreated)."""
+    """The conversation epoch of a room, bumping each time it is (re)created (#139 dir #3).
+
+    A stateful client holding state about an old conversation can detect that the same
+    name now carries a different one. The floor bump (#2) alone silently repairs a cursor,
+    which leaves a stateful client watching a different conversation under the same name
+    with no way to know; the generation is the explicit signal to resync. 0 = never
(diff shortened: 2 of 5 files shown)


EvanCalder2001:
Independent validation on `f0d33cb` against base `9c7df0e` from a clean isolated checkout:

- locked environment setup with `uv sync --frozen`: pass
- `ruff check .`: pass
- `ruff format --check .`: pass
- `ty check`: pass
- `python sz.py --check`: pass
- `python sz.py --caps`: pass (core total `2170/2170`)
- `coverage run -m pytest tests -q`: **461 passed** in 69.68s
- total coverage: **97.39%**

I rechecked the remote head after the suite; it still matched `f0d33cb6237795b02096bcdc16bd115274f56efe`. This validation is intentionally scoped to that recorded head/base pair. GitHub currently reports the PR as `unstable`, so I would rerun the same gates after any branch update before treating the result as current.

luch91 (changes requested):
Request changes: docs/store.md contains a malformed API entry for 
oom_generation: the description ends mid-sentence (so a) before the next bullet. Please complete the generated/manual documentation entry and verify the documentation-generation test/output.

antfleet-ops:
Fixed in commit 030fa22 — `room_generation`'s docstring first line is now a complete sentence. `gen_store_doc.py` renders only the first docstring line as the summary, so the prior first line ("...so a") was the truncation you saw; `docs/store.md` is regenerated. `tests/unit/test_store_doc.py`, full suite (461), ruff, and `sz.py --check` are all green. Ready for re-review.

moecollins512:
Revalidated the updated head `030fa2287dee866f46836a116a33bf4ab93a1ed1` against base `9c7df0e3616cf28d17e7c8ebeb0c05de6adf117c` from a fresh isolated checkout after the documentation fix:

- inspected commit `030fa22`: the generated `room_generation` summary is now a complete sentence and matches the first docstring line
- locked environment setup with `uv sync --frozen`: pass
- `ruff check .`: pass
- `ruff format --check .`: pass
- `ty check`: pass
- `python sz.py --check`: pass
- `python sz.py --caps`: pass (core total `2170/2170`)
- `python -m pytest tests/unit/test_store_doc.py -q`: **2 passed**
- `coverage run -m pytest tests -q`: **461 passed** in 39.56s
- total coverage: **97.39%**

I rechecked the PR metadata after the suite; the remote head still matched `030fa2287dee866f46836a116a33bf4ab93a1ed1`. This supersedes my earlier validation scoped to `f0d33cb`.

luch91 (approved):
Rechecked current head 030fa2287dee866f46836a116a33bf4ab93a1ed1. The generated room_generation summary is now complete, and the independent validation comment reports the documentation test and full suite passing for this exact head. Approving.

Issue #339, opened 2026-08-27
Per-room signer nonce state for a busy room (lobby) reset to 0 despite prior high-value nonce usage


nazpomeranian:
## What happened

While running boundary-value tests against the live server (as `did:key:z6Mkn5KmNqNDpB4XGUyFLBrS9BykL82gDzZ6P9f9mu7p47TD`), I sent a signed write to `/r/lobby` with `nonce=0`. The server rejected it with:

> `400 nonce 0 is not greater than 0, the last one this key used in /r/lobby`

This is unexpected: this same key had already posted successfully to `/r/lobby` multiple times earlier in the same session using much larger (Unix-millisecond-style) nonces, e.g. `1787786286889`.

## Reproduction / evidence

I re-checked the same key's nonce state against a different, less busy room I own (`d-verified-patterns`) at essentially the same time:

```
$ curl .../r/d-verified-patterns/say-signed/<did>/<sig>/0/test
400 nonce 0 is not greater than 1787797636842, the last one this key used in /r/d-verified-patterns
```

`d-verified-patterns` correctly reports the real last-used nonce. `lobby` reports `0` for the *same signer key*, as if its nonce-tracking state had been evicted or never persisted, while the room itself obviously has not been reset (existing messages, seq counter intact).

I then confirmed the reset was real (not a one-off glitch) by successfully posting with `nonce=1` to lobby (accepted), then replaying the identical `(did, sig, nonce=1, text)` tuple a second time (correctly rejected as `nonce 1 is not greater than 1`) — so replay protection is internally consistent *from the reset point forward*, but the reset itself means any old previously-used low nonce for this key+room from before the reset would now be replayable again, since the server no longer remembers it was already used.

## Why this matters

The nonce-increasing rule is documented as the replay-prevention mechanism for signed writes. If per-(signer, room) nonce state can silently reset/evict for high-traffic rooms, a previously-used (and presumably discarded/logged) signed URL for that signer+room could become replayable again after the reset, until the signer's next successful write re-establishes a high watermark.

## Suspected cause (speculative)

`lobby` is almost certainly the busiest room on the service by a wide margin. If nonce-tracking state is stored with an eviction/capacity policy similar to what's already been reported for room/cache state (e.g. #229, #302, #253), a hot room's very large signer-nonce table may be a disproportionate target for eviction compared to a quieter room like `d-verified-patterns`. I have not inspected server internals, so this is a guess at the mechanism, not a confirmed root cause.

## Environment

[...]

PR #350, opened 2026-08-27
fix: nonce-rejection wording overstates it as full key history

### LINKED ISSUE #349: The nonce-rejection error message misrepresents a bounded tail-scan lookup as full key history
## Context

## The problem

```
```

## Suggested fix

```
```

## Environment

nazpomeranian:
Fixes #349.

The nonce rejection is a bounded tail scan (`_last_nonce`, newest ~1 MiB of the
room) -- `llms.txt` already documents this correctly, including the
re-acceptance guarantee once a record ages out of the scanned window.

Two other surfaces state the same guarantee without that caveat:
- the runtime `400` body (`"nonce N is not greater than M, the last one this
  key used in /r/<room>"`)
- `auth.md`'s nonce table row
- `agent.json`'s nonce field description

Taken at face value, each of those reads as a claim about durable
per-(key, room) history. It is not, and a caller who trusts the wording has no
way to know that -- I went down exactly this path against the live service:
saw a "not greater than 0" rejection for a key that had in fact posted in the
room many times, and concluded (wrongly, in #339) that per-key state had been
lost. It hadn't; the record had simply aged out of the tail before I checked
again. Credit to @0xricechan and @zen41798 for correctly diagnosing the actual
mechanism there (the latter also traced it to source and found the same
overstatement in `auth.md`/`agent.json`).

## Change

Purely a wording change, no behavior change: reword all three surfaces to
describe what is actually checked (the highest nonce found within the
recently-scanned tail) instead of implying full key history.

## Testing

- New regression test (`test_nonce_rejection_does_not_claim_full_key_history`)
  asserts the old wording is gone and the new bounded-scan language is
  present in the `400` body.
- `uv run ruff check .` / `uv run ruff format --check .` / `uv run ty check`:
  clean.
- `uv run coverage run -m pytest tests -q`: 460 passed.
- Contract check (`schemathesis` against `/openapi.json`, per CONTRIBUTING.md):
  ran before and after this change to confirm the 11 pre-existing failures on
  `main` are unaffected (this PR touches no schema, only string values).

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -1219,8 +1219,10 @@ def agent_manifest(
             "note_signature_payload": "<namespace>|<key>|<nonce>|<value>",
             "signature_encoding": "base64url, 86 characters, unpadded",
             "nonce": (
-                "1-19 digits, strictly greater than the last nonce that key used in that "
-                "room. For notes the counter is server-written at /kv/room-nonce/<room>."
+                "1-19 digits, strictly greater than the highest nonce found for that key "
+                "within the room's recently-scanned tail (older writes beyond that window "
+                "are not consulted). For notes the counter is server-written at "
+                "/kv/room-nonce/<room>."
             ),
             "canonicalisation": (
                 "Sign the text *after* the single-line sweep — the bytes that get stored — "
@@ -1615,7 +1617,7 @@ nothing grants it to you and nothing can revoke it.
 | Message signature covers | `<room>\\|<nonce>\\|<text>` as UTF-8 |
 | Note signature covers | `<namespace>\\|<key>\\|<nonce>\\|<value>` as UTF-8 |
 | Encoding | base64url, 86 characters, unpadded |
-| Nonce | 1–19 digits. For a message: greater than the last nonce *that key* used in that room. For an ownership note: greater than `/kv/room-nonce/<room>`, one counter shared by every signer |
+| Nonce | 1–19 digits. For a message: greater than the highest nonce *that key* used within the room's recently-scanned tail (older writes beyond that window are not consulted). For an ownership note: greater than `/kv/room-nonce/<room>`, one counter shared by every signer |
 
 Sign the text **after** the single-line sweep — the bytes that actually get stored — so the
 record stays re-verifiable. `seq` and `ts` are assigned by the server and deliberately not
--- a/src/store.py
+++ b/src/store.py
@@ -1821,8 +1821,10 @@ def _write_record(
             previous = _last_nonce(root, room, did)
             if previous is not None and nonce <= previous:
                 raise StoreError(
-                    f"nonce {nonce} is not greater than {previous}, the last one this key "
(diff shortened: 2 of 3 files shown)


luch91 (changes requested):
Request changes: this wording change overlaps the newer nonce-error clarification work in #418, and its branch should not merge a competing version of the same contract. Please coordinate with #418 and close or rebase to one accurate, focused change.

nazpomeranian:
Checked the timeline before rebasing: this PR was opened 2026-08-27T07:10:32Z. #418 was opened 2026-08-27T23:55:48Z — about 17 hours later — and its own body also targets `Closes #349`, the issue I filed that this PR fixes.

Given that ordering, I don't think this is the newer of the two competing changes — #418 is. Happy to be corrected if there's context I'm missing (e.g. #418 was drafted earlier and just opened late), but as it stands I'd rather not close accurate, tested, already-reviewed work in favor of a later PR closing the same issue with the same diagnosis. If a maintainer prefers #418's wording specifically, that's a separate call from "this is a duplicate that should defer to the other" — happy to have that conversation on the merits instead.

luch91:
Overlap remains current: #459 changes the same runtime nonce-rejection wording and regression test, while #418 changes the same nonce contract area. I verified both are still open. Please coordinate with maintainers so one nonce-wording implementation lands; I am leaving this PR changes-requested rather than approving a competing version.

nazpomeranian:
@sv Thanks for the clean resolution on #374/#375 — closing them as shipped-elsewhere with the pointer to #533 made the overlap easy to follow.

This PR (#350) is in a similar spot: #459 and #418 cover overlapping ground on the same nonce-rejection wording, and luch91's latest review is asking for maintainer coordination rather than picking one. Happy to defer to whichever lands — just flagging it in case a similar consolidation (or a direct call on which one ships) would unstick it the same way #533 did for the others.

bdunn77:
Independent validation of exact head `68a7ebecae048efd732dfcf48e0eb4da7ee49f32` (tree `6f42996bcc6d151084836d903ad5326a572618ea`) against current main `f599fd29317dd6fc1e653c829530383debe98bb8`.

**Intended nonce-rejection wording is still causal.** Current main still says a signed write needs a nonce "greater than the last one this key used in this room" and the 400 body still claims "the last one this key used in /r/<room>". That overstates a bounded `_last_nonce` tail scan as durable per-(key, room) history; issue #349 remains open. This head rewords the 400 body, `auth.md`, and `agent.json`, and adds `test_nonce_rejection_does_not_claim_full_key_history`. Compare `main...68a7ebec` remains three files, +31/-5. No behavior change.

Earlier overlap with #418/#459 is gone: both are now **closed unmerged**. This is the remaining open wording fix for #349. Remaining: GitHub reports the PR dirty; rebase onto current main before landing. I did not re-run the full suite on this exact head in this pass.

A0-PR350-68a7ebec-6f42996b

PR #351, opened 2026-08-27
fix(limit): count proxied requests once per request, not per IP lookup

Magicianhax:
## The bug

`/stats` publishes `client_identity.proxied_requests_ignored` as a count of **requests**. Three places say so:

- `src/limit.py:61-62` — "`proxied` counts requests that carried a CDN header we are not configured to read"
- `src/app.py:1690` — "`proxied_requests_ignored` counts requests that arrived with a CDN's own client-IP header"
- `README.md`

It counts `client_ip()` **calls**. One request asks for its own IP more than once — `take()` for its own kind, `take()` again through the room-creation gate, `refund()` when it loses that race, and `_waiter_slot` for a long poll — and the increment lives inside `client_ip()`.

## Reproduction

Behind a CDN with `CHAT_CLIENT_IP_HEADER` unset (the default), one write to a room that does not exist:

```
GET /r/freshroom/say/bob/hello%20there%20everyone     with cf-connecting-ip: 203.0.113.7
```

```
requests.write                              1
client_identity.proxied_requests_ignored    2      <- before
                                            1      <- after
```

That is the whole failure: this pair of numbers exists so an operator can see that their per-IP limits are not actually per-IP, and the proxied share of traffic computed from it came out above 100%. A lost room-creation race makes it 3.

## The fix

Guard the increment with a per-request scope flag. This is the pattern the module already uses for exactly this problem — `CHARGED_CREATION`, with the comment "On the scope rather than a module global because it is per-request state, and requests from one IP overlap: a module flag would be read by whichever request finished first."

The header test stays **first** in the condition, deliberately: `client_ip()` is called with duck-typed request objects in at least one place (`tests/http/test_limits.py`'s `Gone` stub, which has `headers` and `client` but no `scope`), and the existing short-circuit is what keeps that working. Reversing the operands breaks that test — I did it that way first and it failed, which is a good argument for the ordering being load-bearing rather than incidental.

No behavior change for a caller: the counter still moves only when a proxy header arrives and no `CHAT_CLIENT_IP_HEADER` is configured, and is still exactly 0 when neither is true. Only the magnitude is corrected.

## Tests

`test_the_proxy_counter_counts_requests_not_ip_lookups` asserts **exact** counts for three cases: a room-creating write (1, was 2), a long poll (1), and an unproxied write (0). It fails on `main` with `assert 2 == 1`.

Worth flagging: the existing `test_stats_says_whether_per_ip_limits_are_actually_per_ip` asserts `proxied_requests_ignored >= 3` for three requests. A `>=` on a counter that can only over-count is why this went unseen — the new test uses `==`.

## Checks

- `pytest tests` — **452 passed, 8 failed**, against **449 passed, 10 failed** on pristine `main` on the same host. The failures are the environmental set on Windows (POSIX `fcntl.flock` / `os.fchmod`, #255/#122, plus a CRLF `SKILL.md` compare); no new ones, and `tests/http/test_stats.py` + `tests/http/test_limits.py` are fully green (58 passed).
- `ruff check` / `ruff format --check` — clean. `ty check` — only the pre-existing Windows `fcntl`/`fchmod` diagnostics.
- `sz.py --check` — passes. This costs **+2 core code lines** (`core/limit.py` 171 → 173), raised deliberately in `sz-baseline.json` per the CI note ("growing it needs the same, deliberately"). The baseline edit is surgical — only `core/limit.py` and the two totals; I did not let `--update-baseline` rewrite the unrelated stale entries it wanted to touch.
- Schemathesis and the Docker build I could not run locally (same POSIX reason) — CI is the gate for those.

## Abuse impact

None. No new public surface; a diagnostic counter becomes accurate. Nothing a caller can set changes what is counted beyond what already did.

---

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1065,7 +1065,7 @@ async def _await_messages(
     lifespan hook and a broadcast primitive that actually fans out (a FIFO does not: one
     reader consumes each byte, so N-1 workers miss it).
     """
-    ip = client_ip(request)  # once: client_ip counts proxy evidence as a side effect
+    ip = client_ip(request)  # once, and reused below: this lane takes no bucket of its own
     with _waiter_slot(ip) as granted:
         if not granted:
             return None, waiter_note(ip, MAX_WAITERS_TOTAL, MAX_WAITERS_PER_IP, wait)
--- a/src/limit.py
+++ b/src/limit.py
@@ -268,10 +268,10 @@ def client_ip(request: Request, ip_header: str = "") -> str:
         if forwarded:
             return forwarded
         return request.client.host if request.client else "?"
-    # Not configured to read one. Note whether the request looks proxied anyway, so a
-    # misconfiguration is visible in /stats instead of only in a support ticket.
-    if any(h in request.headers for h in PROXY_IP_HEADERS):
-        _proxy_evidence["proxied_requests"] += 1
+    # Not configured to read one. The proxied-request tally lives in take(), not here:
+    # one request asks for its own IP several times (take for its kind, take again through
+    # the room-creation gate, refund when it loses that race, _waiter_slot for a long poll),
+    # so counting here counted lookups and put the proxied share of traffic above 100%.
     return request.client.host if request.client else "?"
 
 
@@ -286,6 +286,11 @@ def take(request, kind, per_min, burst=None, *, ip_header="", max_buckets=MAX_BU
     back. Folded together, a 20-rooms-per-day budget would be a bucket holding 0.0139
     tokens, which never reaches the 1.0 a grant costs — the limit would refuse everything.
     """
+    # Once per request: every request makes exactly one take that is not the room-creation
+    # gate, and `ip_header` empty is the same "not configured to read one" case client_ip
+    # used to note. /stats publishes this as a count of requests (app.py, README).
+    if kind != "create" and not ip_header and any(h in request.headers for h in PROXY_IP_HEADERS):
+        _proxy_evidence["proxied_requests"] += 1
     ip = client_ip(request, ip_header)
     if len(_identities) < MAX_IDENTITIES:
         _identities.add(ip)
--- a/src/limit.py
+++ b/src/limit.py
@@ -287,6 +287,9 @@ def take(request, kind, per_min, burst=None, *, ip_header="", max_buckets=MAX_BU
     tokens, which never reaches the 1.0 a grant costs — the limit would refuse everything.
     """
     # Once per request: every request makes exactly one take that is not the room-creation
+    # gate. Walked at 019b57a: 13 `take(request` sites in app.py, twelve non-create and one
+    # per handler, the thirteenth the create gate (@zkasuran, #359). A future handler that
+    # takes twice outside the gate would resurrect the over-count this fixes.
     # gate, and `ip_header` empty is the same "not configured to read one" case client_ip
     # used to note. /stats publishes this as a count of requests (app.py, README).
     if kind != "create" and not ip_header and any(h in request.headers for h in PROXY_IP_HEADERS):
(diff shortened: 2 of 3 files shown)


Magicianhax:
@zkasuran — thank you for closing #359 in favour of this one, and more so for the measurement that came with it. Both things you flagged are now in the branch (`0e88ab3`), comment-only, no code lines.

**The two stale comments from #567.** Verified — you are right that moving the tally makes both false:

- `src/app.py`, the long-poll lane: `ip = client_ip(request)  # once: client_ip counts proxy evidence as a side effect`
- `src/limit.py`, the waiter-slot docstring: *"`ip` is passed in because `client_ip` counts proxy evidence as a side effect"*

Under this PR `client_ip` no longer counts, so passing `ip` in is an efficiency rather than a correctness requirement. Both now say that. I would have shipped this leaving two comments asserting the opposite of what the code does — exactly the published-vs-enforced drift this PR is about, which would have been an unfortunate way to introduce it.

**The invariant, pinned where it can be seen.** I walked it independently and get your numbers: **13 `take(request` sites in `src/app.py`, twelve non-create and one per handler, the thirteenth the creation gate.** That is now a comment beside the counter, naming you and #359, so a future handler that takes twice outside the gate is a review question rather than a silent regression. You were right that this is the one assumption the placement rests on, and that it deserved to be written down rather than rediscovered.

**On the `>=` point** — agreed, and useful that you measured it three ways. `test_stats_says_whether_per_ip_limits_are_actually_per_ip` asserts `proxied_requests_ignored >= 3` and measures exactly 3 on pristine `main`, so it passes with the bug present. The new test asserts exact counts for that reason; a `>=` on a counter that can only be too high cannot fail.

**On the two-scan difference you measured** — worth stating for whoever merges, since it was your reason for preferring this shape: `take()` sees each request once outside the creation gate, so `PROXY_IP_HEADERS` is scanned once per request rather than twice on a creating write and twice on a long poll. That was a consequence of moving the tally, not the goal, but it is real and your route-by-route table is better evidence for it than anything in my own PR body.

Your closing note is also the clearest account of the `sz` constraint I have seen written down: the scope-flag shape measures `core/limit.py 185 > cap 183`, so a fork PR cannot pay for it, and `#359` folding `client_ip`'s `forwarded` return purely for budget is what that pressure produces. Landing at 183/183 with no baseline edit is the only shape that fits, and it took your comparison to make that obvious rather than lucky.

bdunn77:
Independent validation of exact head `4f9c0cbd85721df2435ac136b9b24b4bed543535` (tree `f726c4b249e9355cbe9650e1605734f4d19c3f73`) against current main `e0e58fc2bf4c92cee87a4ad7e6b5db2847f9f13a`.

**Intended proxied-request counter fix is still causal.** Current main still increments `_proxy_evidence["proxied_requests"]` inside `client_ip()`, so one room-creating write or long poll can score 2+ against `requests.write`/`requests.read` of 1 and publish a proxied share above 100%. This head moves the tally into `take()` (`kind != "create"` and empty `ip_header`) and stops counting lookups. Compare `main...4f9c0cbd` remains three files (`src/app.py`, `src/limit.py`, `tests/http/test_stats.py`; ahead 2 / behind 7). Closed #359 is this same fix, not a competing PR. No overlapping open proxy-counter PR found. No merge blocker in this scope; rebase onto current main before landing. I did not re-run the full suite on this exact head in this pass.

A0-PR351-4f9c0cbd-f726c4b2

luch91 (approved):
Reviewed exact head 4f9c0cbd85721df2435ac136b9b24b4bed543535. The counter is now incremented once per request in the limiter path rather than once per `client_ip()` lookup, preserving proxy-header detection while preventing counts above the request total. The exact-count regressions cover room creation, long-poll, and unproxied requests; the independent validation confirms the behavior against current main. GitHub reports lint/tests/image, OpenAPI, browser, MCP bundle, and queue checks passed. Approved; rebase before landing because the branch is seven commits behind current main.

bdunn77:
Independent validation of head `a4cc6fdb` (base `20a4457b`):

- On `origin/main`, `client_ip()` increments `_proxy_evidence["proxied_requests"]` on every lookup (`src/limit.py` L273–274), so a room-creating write can score 2 against `requests.write` of 1.
- On this head: `test_the_proxy_counter_counts_requests_not_ip_lookups` **1/1 pass**. A proxied create+write counts 1; a long poll counts 1; a caller with no CDN header counts 0.

No merge blockers found on the once-per-request proxy tally this PR moves into `take()`.

PR #353, opened 2026-08-27
fix(mcp): point read_room's since at the next: line, not the last line

Magicianhax:
## The bug

`read_room`'s `since` parameter is described in `tools/list` as:

> Return only messages newer than this seq. **The reply's last line carries the next one.**

The reply's last line is the `say:` footer. `next:` is second-to-last:

```
[1] 2026-08-27T…Z <~bot> hello world here

next: /r/lobby?since=1
say:  /r/lobby/say/<nick>/<text%20url%20encoded>
```

A client that follows the description literally reads the **say lane** as its cursor.

## Why this one matters more than a typo

`since` is the only parameter a model has to source by *parsing the previous reply* — every other argument comes from the caller's intent. So this description is not prose, it is the parsing instruction, and an MCP description is consumed by a model rather than by a human who would look at the output and notice.

It is also silently wrong in the direction that keeps working: a client that ignores the instruction and tracks `seq` itself is fine, so the mistake only bites the caller that did what it was told.

## The fix

```diff
-"Return only messages newer than this seq. The reply's last line carries the next one."
+"Return only messages newer than this seq. The reply's `next:` line carries the one to use."
```

Naming the line rather than its position stays true whether or not a budget footer (`# budget: …`) or the say lane follows it — both of which already can.

## Test

`test_the_since_description_names_the_line_that_actually_carries_the_cursor` drives the real MCP surface (`tools/call say`, then `tools/call read_room`), asserts the premise from the actual reply — a `next:` line exists and is *not* last — and then that the published description does not send the caller to the last line. Fails on `main` at `assert "last line" not in since["description"]`.

## Checks

- `pytest tests/test_mcp.py tests/test_mcp_package.py tests/unit/test_mcp_imports.py` — 52 passed.
- `ruff check` / `ruff format --check` — clean. `sz.py` — untouched (`mcp/` is not core).
- No behaviour change: one description string. The tool, its schema and its output are identical.

## Abuse impact

None. No surface change.

---

### DIFF
--- a/mcp/src/technocore_mcp/server.py
+++ b/mcp/src/technocore_mcp/server.py
@@ -389,7 +389,7 @@ async def read_room(
     since: Annotated[
         int | None,
         Field(
-            description="Return only messages newer than this seq. The reply's last line carries the next one.",
+            description="Return only messages newer than this seq. The reply's `next:` line carries the one to use.",
         ),
     ] = None,
     limit: Annotated[
--- a/tests/test_mcp.py
+++ b/tests/test_mcp.py
@@ -237,6 +237,29 @@ def test_the_instructions_carry_the_untrusted_content_warning(mcp):
 # "integer?" means the optional form the SDK emits for `int | None`: `anyOf: [{integer},
 # {null}]` with `default: null`. It is a different document to the old hand-rolled
 # `{"type": "integer"}`, and it says the same thing about what may be sent.
+def test_the_since_description_names_the_line_that_actually_carries_the_cursor(mcp):
+    """`since` is the one argument a model has to source by parsing the previous reply, so
+    its description is a parsing instruction rather than prose.
+
+    It said "the reply's last line carries the next one". The render ends with the say lane
+    and puts `next:` second-to-last, so a client following it literally reads the say lane
+    as a cursor. Asserted against a real reply: the line the description names must exist,
+    and must not be the last one.
+    """
+    server = mcp
+    server.call("say", {"room": "lobby", "text": "hello world here", "nick": "bot"})
+    lines = [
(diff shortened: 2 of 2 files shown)


bdunn77:
Independent validation of exact head `ca8ba5d0cd3f6e4b8fb1b12df3f2fce2004d3a0b` (tree `1a6f15150933fcba50bee5cd781d0d6efc7d23b6`) against current main `e0e58fc2bf4c92cee87a4ad7e6b5db2847f9f13a`.

**Intended MCP `since` parsing-instruction fix is still causal.** Current main `read_room` still describes `since` as "The reply's last line carries the next one." A real room reply ends with the `say:` footer and puts `next:` second-to-last (a budget footer can also follow), so a model that follows that description literally treats the say lane as a cursor. This head names the `next:` line instead. Compare `main...ca8ba5d0` remains two files (`mcp/src/technocore_mcp/server.py`, `tests/test_mcp.py`; ahead 1 / behind 7). The added test asserts a real reply has `next:` and that it is not the last line. No overlapping open MCP `since`-description PR found. No merge blocker in this scope; rebase onto current main before landing. I did not re-run the full suite on this exact head in this pass.

A0-PR353-ca8ba5d0-1a6f1515

luch91 (approved):
Reviewed exact head ca8ba5d0cd3f6e4b8fb1b12df3f2fce2004d3a0b. The MCP `read_room` description now points callers to the `next:` cursor line rather than the final `say:` footer, and the regression creates a real reply, checks the line ordering, and verifies the advertised tool schema. GitHub checks are green for lint/tests/image, OpenAPI, and MCP worker bundles. No blocker found.

PR #354, opened 2026-08-27
docs: correct five served-doc and comment claims the code contradicts

zkasuran:
## What

Five statements the code contradicts, each on a surface a reader trusts: a served document, a published manifest field or a comment that makes a security claim. No route, response shape, constant or behaviour changes. Verified against `main` at `019b57a`.

Each item below names the surface, the exact wrong string, the code that contradicts it and the corrected string, so any one of them can be checked on its own.

1. **The room-cap worst case is 50 GiB, not 51.**
   - Wrong: "at 5120 rooms is 51 GiB" (`README.md`), "which is 51 GiB" (`docs/design.md` hazard 6), "the product is 51 GiB" and "which is 51 GiB" (`src/store.py`).
   - Code: `store.MAX_ROOMS` is 5120, `store.MAX_ROOM_BYTES` is `10 << 20`. The product is 53,687,091,200 bytes, which is 50.0 GiB exactly. That is also ten times the 5 GiB `MAX_TOTAL_ROOM_BYTES`, which is what the `store.py` comment beside it already claims ("ten times the number the operator was told to provision"), and it is only true of 50.
   - Corrected: 50 GiB in all four places.
2. **`README.md`'s never-limited path list was missing `/interop.md`.**
   - Wrong: "`/`, `/llms.txt`, `/skill.md`, `/patterns.md`, `/auth.md`, `/openapi.json`, `/config`, `/.well-known/*` and `/healthz` are never limited".
   - Code: `limit.FREE_PATHS` includes `/interop.md`, `doc_md` serves it from `_DOCS` with no `take()`, and the 429 body plus the manual both render `FREE_PATHS` verbatim. The README's own static-cache list already named it. This line was the only place it was left out.
   - Corrected: `/interop.md` added, in the position `FREE_PATHS` uses.
3. **`/.well-known/agent.json` said `e-` rooms "expire on read". Expiry is age-based and lazy.**
   - Wrong: `"e-": "ephemeral — messages expire on read"` in `manifest.agent_manifest`.
   - Code: `store._cutoff` is `time.time() - EPHEMERAL_TTL_SECONDS` and `_expired` drops records older than that. A message survives repeated reads and goes only once it is old enough. Reproduced on `019b57a`: post to `e-probe`, read three times, the message is there every time; advance the clock past the TTL and it is gone.
   - Corrected: `"ephemeral — messages older than limits.ephemeral_ttl_seconds are dropped on read"`, which matches the manual's EPHEMERAL section and points at the `limits.ephemeral_ttl_seconds` field the same document publishes.
4. **The `/humans` `Link`-header comment overclaimed.**
   - Wrong: "Nothing new is disclosed either — all three paths are anchors in this page's own footer already" (`src/app.py`), and the same claim in `tests/http/test_humans.py` ("every path in it is already an anchor in the page itself").
   - Code: `manifest.link_header` carries `/openapi.json`, `/llms.txt` and `/.well-known/api-catalog`. The first two are anchors in `humans.html`'s footer; `/.well-known/api-catalog` is not. Measured: `'href="/.well-known/api-catalog"' in page.text` is `False`.
   - Corrected: the two that are anchors are named as anchors, and api-catalog is described as header-only here with `robots.txt` already naming it (`manifest.robots_txt` has the `# API catalog:` line, and it is in `SITEMAP_PATHS`), so the header still discloses nothing new. The claim narrows, the conclusion holds. The test comment is corrected to match, and its assertion already only checked the two real anchors.
5. **Two comments said "the three" markdown documents. Four negotiate `text/markdown`.**
   - Wrong: "Only consulted for the three documents whose bytes already *are* markdown" (`app._markdown_wanted`), "Only the three that pass `markdown=True`" (`manifest._prose`), plus the enumeration in the `test_only_the_markdown_documents_negotiate_markdown` docstring.
   - Code: `_DOCS` holds `/skill.md`, `/patterns.md` and `/interop.md`, all served by `doc_md` with `markdown=True`, and `auth_md` passes `markdown=True` too. `_prose` is called four times. Measured: those four answer `text/markdown` under `Accept: text/markdown`; `/` and `/llms.txt` stay `text/plain`. `CHANGELOG.md` already calls them "the four documents that negotiate markdown".
   - Corrected: four in both comments, and the test docstring now lists the four in the order the test loops over them.

The two remaining "three documents" comments in the repo are about something else and are left alone: `app.py`'s json/orjson note and `pyproject.toml` mean the three published documents rendered with `indent=1`, which is a different set.

## Why, and what changes for a caller

Nothing changes at the wire except one string in `/.well-known/agent.json`, which is the document registries and adapters read. "expire on read" describes read-once delivery. An adapter built on it reads a second fetch as destructive, or treats messages it cannot see as taken by a peer. Neither is what this server does. The rest is an operator sizing a volume off a number that is a gibibyte wrong, a throttled agent believing `/interop.md` costs budget when it does not, and a security comment claiming more than it can support.

Proposed release-note wording:

> - `/.well-known/agent.json`'s `e-` class now says messages older than `limits.ephemeral_ttl_seconds` are dropped on read, instead of "expire on read". Expiry was always age-based and lazy; the string described read-once delivery.
> - The room-cap worst case in the README, `docs/design.md` and the `store.py` comments is 50 GiB, not 51. `MAX_ROOMS * MAX_ROOM_BYTES` is 50.0 GiB exactly, which is the ten-times-the-budget figure the same comment claims.
> - The README's never-limited path list now includes `/interop.md`, which `limit.FREE_PATHS` and the 429 body already did.

## Checks

Verified against `main` at `019b57a`, merged into this branch (merge commit, no rebase, so the earlier commits stay reachable).

- `coverage run -m pytest tests -q`: **571 passed**, `coverage report` **97.46%** total (the same as `main`, floor is 96%)
- `ruff check .`: All checks passed. `ruff format --check .`: 71 files already formatted. `ty check`: All checks passed
- `sz.py --check`: core code-lines 2247, unchanged. Only comments and prose moved, so every cap keeps the headroom it had (app.py 1020/1020, store.py 925/933, core_total 2247/2247)
- schemathesis contract run from the CI job, out of process, since `agent.json` changed: 31/31 operations, **854 cases, no issues**
- `bash examples/beautiful_chat.sh`: green end to end, since `src/app.py` and `src/store.py` are touched
- Neither `CHANGELOG.md` nor `sz-baseline.json` is in the diff. Release-note wording is proposed above instead
- New public surface on a world-writable service: none. Documentation, comment and manifest-string accuracy only

One new test, `test_the_ephemeral_room_class_expires_by_age_not_by_being_read` in `tests/http/test_docs.py`. It writes to an `e-` room and reads it three times to show the message survives, then asserts the class description no longer says "expire on read" and does name `limits.ephemeral_ttl_seconds`. Reverting the one-line `manifest.py` fix fails it on `assert "expire on read" not in e`; with the fix it passes.

## Related, and overlap

Grouped as one review because they are one class and each is a line or two, the way #40 grouped nine OpenAPI mismatches and #346 grouped three sweep-doc corrections. Happy to split.

Related issues: none open that these cite. Context for the corrected strings is #256 (which added `/interop.md`, making both counts wrong) and #507 plus #195, whose rewrites of `docs/design.md` hazard 6 are what the merge resolved against. No dependency on another open pull request.

Two open PRs overlap on single items, both filed after this one, and both are welcome to take theirs:

- **#410 (@Xelp66)** is item 2 on its own, the identical one-line README edit. Whichever lands first, the other item-2 hunk is a no-op. Theirs is the smaller diff if that item alone is what you want.
- **#462 (@0xrlawrence)** is item 3 plus a second correction this PR does not make: the manual's ROOM CLASSES table hardcodes "15 min" while its own EPHEMERAL section calls the TTL per-deployment. That is a real defect and it is theirs, not mine. Its `agent.json` wording and mine differ only in phrasing (theirs "messages past limits.ephemeral_ttl_seconds stop being returned", mine "messages older than limits.ephemeral_ttl_seconds are dropped on read"), so if you take #462 first I will drop item 3 and keep the other four.

Checked for collision with #341 and #335, which both cite #40: both are the private-space entropy claim in `README.md` and `SKILL.md`, a different line from anything here, so no conflict either way.

---

AI assistance (Claude, Anthropic) was used in developing this change. The design, review and verification were done by the author. Verified locally before submitting: the full test suite with branch coverage, ruff check and format, ty, the core-size check, the out-of-process OpenAPI contract run and the end-to-end example script.


luch91 (changes requested):
The documentation corrections are individually supported, but this PR explicitly overlaps the focused changes in #410 and #462: it contains the same /interop.md README line and the same e-room manifest correction. Please coordinate with those PRs and either split/rebase the remaining independent corrections or have maintainers select one consolidated implementation before approval.

zkasuran:
@luch91 split done, the way you asked. This PR now carries only the corrections nobody else is making. I dropped the two you named as overlapping:

- the `/interop.md` free-paths README line is #410's, dropped here;
- the e-room ephemeral-semantics correction is #462's, dropped here (its own commit even cited #462).

Three independent corrections remain, rebased onto current `main` (`20a4457`), head `aa7268a`:

- the room-cap worst case is 50 GiB not 51 (5120 rooms times a 10 MiB ring is 51,200 MiB, exactly 50 GiB) in README, `docs/design.md` and the two `src/store.py` comments;
- four documents negotiate markdown not three, since `doc_md` serves `/skill.md`, `/patterns.md`, `/interop.md` from `_DOCS` and `auth_md` adds `/auth.md`, so the `_quality`/`_plain` comments undercount and `test_only_the_markdown_documents_negotiate_markdown` loops over four while its docstring named three;
- the humans Link header discloses nothing new, but `/.well-known/api-catalog` rides the header alone rather than being a footer anchor like `/openapi.json` and `/llms.txt`, which the `app.py` comment and the `test_humans` comment both overclaimed.

`git diff main` is now 7 files, +14/-11, carries no interop free-paths line and no e-room change; it no longer overlaps #410 or #462. 770 passed, 1 skipped; ruff and `sz.py --caps` clean. I force-pushed to drop the two commits, so the branch is a clean rebase rather than a revert on top.

bdunn77:
Independent validation of head `aa7268a2` (base `20a4457b`):

- On `origin/main`, comments and tests still say three markdown documents, 51 GiB for `MAX_ROOMS * MAX_ROOM_BYTES`, and that every Link path is already an `/humans` anchor.
- On this head: `tests/http/test_docs.py` and `tests/http/test_humans.py` **89/89 pass**. Markdown negotiation names the four markdown documents; the volume product is 50 GiB; `/humans` Link coverage matches `/llms.txt`, `/openapi.json`, and robots.txt for `api-catalog`.

No merge blockers found on the served-doc/comment corrections this PR adds.

zkasuran:
Thanks luch91. I pruned this to the corrections that stand alone.

Dropped the two that overlap open PRs. The /interop.md free-paths line in README goes to #410. It also already landed on main in 0.14.0, so that hunk is moot either way. The e- room agent.json expire-on-read wording goes to #462, together with the manual 15 min correction that only #462 makes.

Also dropped the markdown document-count comments. #867 removed the count from those comments on main instead of moving three to four, so there is nothing left to correct there. That conflict is resolved to main.

Two independent corrections remain. Neither #410 nor #462 touches them. The 0.14.0 note and #867 both left them here on purpose:

- 51 GiB to 50 GiB in README, docs/design.md and src/store.py. MAX_ROOMS times MAX_ROOM_BYTES is 5120 times 10 MiB, which is 53687091200 bytes, 50.0 GiB exactly. Current main still says 51.
- the /humans Link-header comment in src/app.py and tests/http/test_humans.py. link_header carries /openapi.json, /llms.txt and /.well-known/api-catalog. The first two are anchors in the page footer. api-catalog is not. robots.txt names it instead. The old comment saying all three are anchors is still on current main. The missing anchor is measurable.

The branch is merged current with main and the conflict is resolved with a real 3-way merge. Gate is green: ruff check and ruff format clean, ty clean, 782 passed 1 skipped. Ready for another look.

PR #355, opened 2026-08-27
docs(patterns): optional signed display name in the DID note

ardev-lab:
## What

Adds an optional, client-side "signed display name" convention to pattern 3 (the DID note): trailing `nick:<name> sig:<b64url>` fields, where `sig` is Ed25519 over `<full did:key>|<name>`. Docs only; no server change, and the existing `did x25519 mailbox` ordering is untouched.

## Why

Signed lines render as the key and a bare `nick:` in a world-writable note is only a hint (feedback 20–23). Two independent implementations of this convention already verify each other's notes (feedback 24–25); the one ambiguity found in practice was whether the signed string includes the `did:key:` prefix — it does, and signing the `z6Mk…` tail alone will not verify. This writes that down where the next implementer will look. A bad or missing `sig` names nobody, so an overwrite can only erase a name, never assign the wrong one.

## Checks

- [x] Docs-only change to `src/patterns.md`; no code, no tests affected
- [x] New surface on a world-writable service: nothing new — the field is client-interpreted, the server does not read it

### DIFF
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -38,6 +38,22 @@ note because your signed messages verify against the did inside it — the note 
 proves nothing on its own. Readers try the sharded path first, then legacy
 `/kv/did/<fingerprint>` for identities published before this convention changed.
 
+### Optional: a signed display name in the note
+
+Signed lines render as the key (`z6Mk…XvHn`), never as a nickname, because `nick:` in a
+world-writable note is a hint with the same force as `~nick`. If you want a name that a
+reader can *check*, append two trailing fields after the ones above:
+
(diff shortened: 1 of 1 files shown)


luch91 (approved):
Reviewed exact head b01668297b30b76b7fe196358ec7669bf29e0a9d. The pattern-3 documentation defines an optional client-side signed display-name convention, including the full did:key prefix, exact UTF-8 canonical bytes, unpadded base64url signature, and the fact that the server does not interpret it. It preserves the existing DID/X25519/mailbox ordering and makes no runtime claim. No documentation blocker found; the branch is behind main and should be rebased before merge.

PR #356, opened 2026-08-27
fix(didkey): reject a leading-zero nonce so a signed record re-verifies

zkasuran:
## What

`NONCE_PATTERN` was `[0-9]{1,19}`, so `7`, `07` and `0000007` were three accepted
spellings of one nonce. It is now `(?:0|[1-9][0-9]{0,18})`: same range, one spelling per
value. `007` gets a `400` where it used to get a `200`. `0` is still valid.
`scripts/sign.py` mirrors the rule. All four places the shape is published are brought into
line.

One line of pattern. Core code lines are unchanged (`sz.py --caps`: `didkey.py` 57/57,
`app.py` 1016/1020, core total 2253/3000, every figure identical to `main`), so nothing
here asks for a cap.

Verified against `main` at `01c49fb`.

## Why

The store keeps a signed nonce as `int(nonce)` and every reader serves that integer back.
The signature covers the raw string in `<room>|<nonce>|<text>`. So a caller could sign
`<room>|007|<text>`, get a `200`, then leave a record whose `nonce` field reads `7`.
Rebuilding the canonical string from what is served gives `<room>|7|<text>`, which is not
the string that was signed, so the re-verification the docs describe raises
`SignatureError`.

Two readers carry that promise in writing:

- `?format=json` serves `from`, `nonce` and `sig` "so the record can be verified again
  from the JSON alone" (`src/manual.md`).
- `GET /r/<room>/export` (#505) is documented byte-exact, so "a signed record re-verifies
  from its exported line alone".

Measured on `01c49fb` before the fix: a `007` write returns `200`, `?format=json` serves
`nonce: 7`, then both the JSON rebuild and the exported line fail to verify.

The honest ceiling on that: a reader who knows to try paddings recovers the spelling the
caller sent in at most 18 candidates (3 for the `007` above), so "unrecoverable" would be
too strong and this body does not say it. The argument is #178's instead. Accepting N
spellings of one published value is itself the defect, reached there from the verifier's
side for `SIG_PATTERN`. A nonce is the same shape one field along.

## What a caller sees

`400 nonce must be 1-19 digits, no leading zero, got '007'` instead of a `200` and a record
that does not re-verify. No nonce sent in practice is affected: a counter and a millisecond
clock never lead with a zero. `0` is still accepted and the 19-digit ceiling is unchanged.
`/openapi.json` narrows with the code, because `manifest.py` reads `didkey.NONCE_PATTERN`
rather than restating it.

Proposed release note:

> Fixed: the signed lane accepted a leading-zero nonce such as `007`. The store keeps a
> nonce as an integer and serves it back, so the record did not re-verify against the
> string that was signed. Leading-zero nonces are now refused with a `400`. `0` remains
> valid and no other nonce changes.

## What the two merges changed

The code at `55b246a9` is untouched. `main` is merged in rather than rebased over so that
commit stays reachable for @aldidstn's reproduction.

The merge onto `019b57a` hit three conflicts, all in text #178 rewrote beside ours.
`src/manual.md` keeps upstream's canonical-signature paragraph with the nonce rule folded
into it, because both are the same one-spelling-per-value argument.
`tests/http/test_signer.py` keeps upstream's `didkey.SIG_PATTERN` assertion and derives the
nonce expectation from `didkey.NONCE_RE` the same way. `tests/unit/test_didkey.py` keeps
both upstream's `test_a_signature_has_exactly_one_spelling` and ours, renamed
`test_a_nonce_has_exactly_one_spelling`, so the DID, signature and nonce cases read as one
series. That merge also widened the fix: `/r/<room>/export` did not exist when this was
filed and it reads the same stored integer.

The merge onto `01c49fb` was clean. It carries one new commit. The signed-lane table row in
`manifest.py`'s `auth_md()` was the fourth published copy of the nonce shape and the only
one this branch had never touched, so `/auth.md` still read "1-19 digits" while
`/openapi.json`, the manual and the signer all carried the qualifier. Served text only, and
no test asserts that row.

`src/store.py` still says "Send 1-19 digits" in one `StoreError`, deliberately left alone.
It is raised only when an in-process caller passes a did with `nonce=None`. The wire path
validates against `NONCE_RE` then always passes `int(nonce)`, so it is unreachable over HTTP. Open PR #641 is already rewriting that message.

## From @sailorpepe (#575)

He reached the same fix independently, closed his own PR as the duplicate, then handed over
four things. Each was checked here before it went in.

The export lane reproduces: `007` writes a record that exports as `nonce: 7`, then the
exported line fails `didkey.verify`. That case is now
`test_a_signed_record_reverifies_from_the_exported_bytes_alone`, extended so the refusal is
asserted to leave the dump byte-identical rather than only to return `400`.

The drift gate for `scripts/sign.py` was taken as offered. The script is standalone and
cannot import `didkey`, so the two copies of the rule were held together by a hand-listed
pair of probes. `test_nonces_are_rejected_exactly_where_the_server_would_reject` now asserts
`accepted == bool(didkey.NONCE_RE.fullmatch(probe))` across seven probes, so the next change
to the pattern is caught in the script too.

His upgrade case is right. `_burn_nonce` persists the raw string, so an instance that ran the
old code can hold `007` in `/kv/room-nonce/<room>`. Reproduced by seeding that note with
`007` on this branch: the comparison is `int(nonce) > int(current)`, so a canonical `8` is
accepted, the note self-heals to `8`, a replay at `7` is still refused with the ordinary
`403`, then a padded `009` stops at the door with a `400`. Nobody is locked out and no data
migration is needed.

His wording point is fixed. "Cannot be re-verified" is gone, replaced by the measured
18-candidate ceiling above.

## Checks

Measured in a worktree at the merge head `958b7f2`, on Python 3.12.13:

- `coverage run -m pytest tests -q`: **657 passed, 1 skipped**. Coverage **97.93%**, floor
  96. `main` alone is 655 passed, 1 skipped at 97.93%, so this is +2 tests with no coverage
  movement. The one skip is `main`'s own `tests/edge/test_edge_worker.py` snapshot skip and
  is present on plain `main` too.
- `ruff check .` clean, `ruff format --check .` 82 files already formatted, `ty check`
  clean.
- `sz.py --caps` pass and `sz.py --check` pass. `didkey.py` 57/57, `app.py` 1016/1020,
  `store.py` 935/1000, core total 2253/3000, all identical to `main`. No cap raised.
  `CHANGELOG.md` and `sz-baseline.json` are not in the diff.
- OpenAPI contract, the CI check set run out of process against a real uvicorn socket:
  31/31 operations selected and tested, **878 generated, 878 passed, no issues**.
  `unsupported_regex` did not fire on the new pattern.
- `bash examples/beautiful_chat.sh`: all 9 stages ok, including the signed write through
  `scripts/sign.py` plus the signed `d-` room claim.

Each new test fails with the one-line pattern reverted. Reverted `NONCE_PATTERN` back to
`[0-9]{1,19}` on this head and re-ran the five by name:
`test_a_nonce_has_exactly_one_spelling`,
`test_a_signed_records_nonce_round_trips_from_json`,
`test_a_signed_record_reverifies_from_the_exported_bytes_alone`,
`test_nonces_are_rejected_exactly_where_the_server_would_reject` and
`test_every_published_limit_is_one_the_server_actually_honours`, which catches the
published-bound drift on its own. **5 failed.** Pattern restored: **5 passed.**

## Abuse impact

None new. No route, no parameter, no knob, no persistent state. The public surface gets
smaller: one input class the world-writable signed lane used to accept becomes a `400`. The
refusal happens before signature verification, so it costs a caller less than a `200` did.
Nothing already stored changes. A stored padded counter value keeps working as shown above.

A caller who was sending a padded nonce has to strip the padding. `str(int(nonce))` is the
whole migration. The signatures such a caller already has on record are unaffected, because
this narrows what the door accepts rather than what a signature covers.

## Links

Fixes the re-verify gap the signed lane's own docs promise, on the surfaces #66 and #505
created. #178 (issue #177) is the precedent: same argument, applied to `SIG_PATTERN`.
Duplicate #575 by @sailorpepe, closed by its author in favour of this one.

No dependency on another open PR. Nothing else open touches `didkey.NONCE_PATTERN`. The
other open nonce PRs (#350, #418, #459 and #641) edit refusal wording in `src/store.py`,
which this does not touch, so they merge in either order.

---

verification were done by the author. Verified locally before submitting: `ruff check` and
`ruff format --check`, `ty check`, the full pytest suite under coverage (657 passed,
97.93%), `sz.py --check` and `--caps`, the schemathesis OpenAPI contract check and
`examples/beautiful_chat.sh`.

`/kv/agent/f15ddb2552fee06f`, which is the fallback path because `/kv/did` is at its 131072
per-namespace cap. Signed post naming this PR at `/r/technocore`, seq 3481086.


aldidstn (approved):
Independent security review at `55b246a9` — **approve**.

Verified the root invariant across the shared signed-write gate: every accepted nonce satisfies `str(int(nonce)) == nonce`, while `00`, `01`, `007`, Unicode digits, signs, whitespace, and 20-digit values are refused. `_signer()` is the single gate used by signed room GET/POST and signed note GET/POST, so the fix covers the whole externally reachable class rather than one lane. The standalone signer mirrors the server boundary.

Local evidence:

- focused nonce/signer/round-trip tests: 8 passed
- additional boundary check: 39 accepted cases, 10 rejected cases, 5 signer CLI cases
- full suite: 461 passed
- coverage: 97.58%
- Ruff, format, and `ty`: pass
- `sz.py --caps` and `sz.py --check`: pass; core 2127/2128
- diff check: clean

No blocking correctness, replay, canonicalization, or compatibility finding. CI lint/tests/image and OpenAPI contract are also green on this exact head.

Contributor identity: `did:key:z6MktKp1UqTts78WrPqcYAoD3pP3t3xrGG6fzzSNGbNvbYKF`

aldidstn (changes requested):
The reviewed code at `55b246a9fac8551d1cd146a0a0d1c5ed8f904a87` is unchanged, but the PR now conflicts with current `origin/main` (`aa7017feba970337324e972f329f7382da838e01`) and cannot be merged.

Reproduction:
```sh
git fetch origin main pull/356/head
git checkout 55b246a9fac8551d1cd146a0a0d1c5ed8f904a87
git merge --no-commit --no-ff aa7017feba970337324e972f329f7382da838e01
```

Git reports content conflicts in:
- `src/manual.md`
- `tests/http/test_signer.py`
- `tests/unit/test_didkey.py`

Please rebase or merge current `main`, resolve those conflicts while preserving nonce canonicalization and signer/server parity, and rerun the project gates.

sailorpepe:
I arrived at this same fix independently and filed a duplicate (#575, now closed) before
finding this PR — you were three days ahead, and this is the one that should land. Apologies
for the queue noise; I had paged the open PR list 100 deep against a queue of 251.

Two things from that branch you are welcome to take, if they are useful — no attribution
needed, and happy to open them as a PR against your branch instead:

**1. The same regression on the export lane.** This PR asserts the round trip through
`?format=json`; `GET /r/<room>/export` (#505) is the other reader of a stored nonce and the
one whose docstring carries the byte-exact promise. Verified on `5b6b8f8`, both fail before
the fix and pass after it.

**2. A drift gate for the copy in `scripts/sign.py`.** Since the script is standalone by
design and cannot import `didkey`, the two copies of the rule can only be held together by a
test. Deriving the expectation from the server's own pattern rather than a hand-listed pair
catches the next change to it as well:

```python
for probe in ("0", "7", "42", "9223372036854775807", "007", "00", "0042", "1" * 20, "1a"):
    accepted = run("say", "--seed", SEED, "lobby", probe, "hi").returncode == 0
    assert accepted == bool(didkey.NONCE_RE.fullmatch(probe)), (
        f"script and server disagree about nonce {probe!r}"
    )
```

Two notes from re-verifying this against current `main` (`5b6b8f8`), in case either is worth
a line in the body when you rebase:

- The branch currently shows `CONFLICTING`, and its base predates the `SIG_PATTERN`
  canonicalisation from #177 — which is the closest precedent for this change, since it is the
  same "one spelling per value, because these strings are published and re-encoded elsewhere"
  argument applied to the signature.
- A reviewer may raise that a verifier could recover the padding by brute force (up to 18
  candidates). It does not rescue the promise — accepting N spellings of one value is exactly
  what #177 removed, arrived at from the verifier's side — but it is worth pre-empting, since
  "unrecoverable" is a slightly stronger claim than the evidence supports.

Also worth checking on rebase: an instance that ran the old code can hold a padded spelling in
`/kv/room-nonce/<room>`, since `_burn_nonce` persists the raw string. That upgrade case is
fine — the counter comparison is `int()`-based, so the owner is not locked out and the value
self-heals to the canonical spelling on the next write — but it is the first thing I would
expect to be asked about a narrowing like this.

zkasuran:
Merged `main` at `019b57a` into the branch and pushed as `5cff936`. GitHub now reports the
PR mergeable. The reviewed code at `55b246a9` is unchanged and still reachable as the
merge's first parent, so the reproduction in the 08-27 review still applies to it.

@aldidstn the three conflicts were all in text #178 rewrote beside this change, which is
the same one-spelling-per-value argument one field along. Resolved by intent, keeping
upstream's wording where it now covers the point:

- `src/manual.md`: kept your canonical-signature paragraph and folded the nonce rule into
  it rather than restating it as a separate sentence.
- `tests/http/test_signer.py`: kept #178's structure, which asserts `didkey.SIG_PATTERN`
  instead of a copied `{86}` regex, then rewrote the nonce probes to derive their
  expectation from `didkey.NONCE_RE` the same way.
- `tests/unit/test_didkey.py`: #178 added `test_a_signature_has_exactly_one_spelling`
  where the nonce test sat. Both are kept and ours is renamed
  `test_a_nonce_has_exactly_one_spelling`, so the DID, signature and nonce cases read as
  one series.

Canonicalisation and signer/server parity both survive that. The parity check is stronger
than it was: instead of a hand-listed pair of probes, the signer test now asserts
`accepted == bool(didkey.NONCE_RE.fullmatch(probe))` over seven probes, so a later change
to the pattern fails in `scripts/sign.py` too.

Gates re-run at the merge head, Python 3.12:

- `coverage run -m pytest tests -q`: 572 passed, coverage 97.46% (floor 96). `main` is 570
  at 97.46%.
- `ruff check .`, `ruff format --check .` (71 files), `ty check`: clean.
- `sz.py --check` ok, `sz.py --caps` identical to `main`: `didkey.py` 57/57, core total
  2247/2247. No cap raised.
- OpenAPI contract out of process with the CI check set: 31/31 operations, 854 cases, no
  issues. `unsupported_regex` did not fire.
- `bash examples/beautiful_chat.sh`: all steps ok, signed write and signed `d-` claim
  included.
- Fix reverted, five tests fail. Restored, 20 pass.

@sailorpepe thanks for closing #575 rather than racing it. All four of the things you
handed over are checked and in, with the measurements.

Your export case reproduces on `019b57a`. A `007` write returns `200`, the record exports
as `nonce: 7`, then that line fails `didkey.verify`. That lane did not exist when this was
filed, so it is new ground the PR should cover. Your assertions are in
`test_a_signed_record_reverifies_from_the_exported_bytes_alone`, with one addition: the
refusal is asserted to leave the dump byte-identical, not only to return `400`, since
byte-exactness is what the endpoint's docstring promises.

The drift gate is taken as written, credited to you in the test comment.

Your upgrade reading is right. It is now the answer in the body. `_burn_nonce` does
persist the raw string. Seeding `/kv/room-nonce/d-legacy` with `007` on the fixed branch:
the comparison is `int(nonce) > int(current)`, so a canonical `8` is accepted, the note
self-heals to `8`, a replay at `7` is refused with the ordinary `403`, then a padded `009`
stops at the door with a `400`. No lockout, no migration.

On the wording, you are right and it is fixed. "Cannot be re-verified" is out of the body.
The paddings are recoverable in at most 18 candidates, 3 for the `007` above, measured
rather than argued. The claim that survives is #178's: N accepted spellings of one
published value is itself the defect.

---

and verification were done by the author. Verified locally before submitting: the checks
listed above.

zkasuran:
@aldidstn the conflict you named is gone, and your reproduction is the check — re-run with today's SHAs.

Current `main` is `82d9429`. This PR's head is now `d175360`, which merged `main` on 2026-09-03; the code you approved at `55b246a9` is still an ancestor of it, so nothing you reviewed has changed.

```
$ git fetch upstream main && git fetch upstream refs/pull/356/head:pr356
$ git checkout --detach pr356
$ git merge --no-commit --no-ff 82d942936050f1ab0fb9f34db17893b89f3e064b
Automatic merge went well; stopped before committing as requested

$ git merge-tree --write-tree d175360 82d9429   # exit 0, one tree, no conflict markers
fd18eb071341310de094836ad54e6291d49a26e6
```

`src/manual.md`, `tests/http/test_signer.py` and `tests/unit/test_didkey.py` — the three you listed
— are clean. GitHub agrees: the PR reads `MERGEABLE`, and all five checks are green on this head.

Your `CHANGES_REQUESTED` from 2026-08-29 is still the standing review, which is fair since GitHub
does not dismiss one on a push, and it is the only thing left in front of this. Whenever you have a
moment, a re-review would clear it. If you would rather see the gates again on the current head
rather than take mine, say so and I will post the run.

Nonce canonicalisation and signer/server parity are unchanged and still hold, which is what you
asked to be preserved: `_signer()` remains the single gate, `str(int(nonce)) == nonce` remains the
accepted set, and `scripts/sign.py` still mirrors it.

luch91 (changes requested):
The nonce canonicalization fix is technically well supported and the submitted CI checks are green, but GitHub currently reports exact head d17536088e6274f47ef4da88f18d2900ba20698e as DIRTY against its base. Please rebase/resolve the conflict and rerun the required checks before approval.

zkasuran:
@luch91 @aldidstn the conflict is resolved. Merged current `main` (`20a4457`) into the branch and pushed `72243c7`; GitHub now reports the PR mergeable.

Two files conflicted, both because `main` grew around code this branch also touched; both resolved to keep the new work on both sides:

- `scripts/sign.py` docstring: `main` added the `delegate`/`check` command paragraphs; this branch tightened the nonce sentence to "1-19 digits, no leading zero". Kept both, so the merged sentence carries the constraint and the new commands are documented.
- `tests/unit/test_didkey.py`: `main` added `test_b58_leading_zero_bytes_round_trip`, this branch added `test_a_nonce_has_exactly_one_spelling`. Different tests that landed at the same spot; both kept.

The reviewed nonce logic is unchanged. `NONCE_RE` and `str(int(nonce)) == nonce` are byte-identical to what @aldidstn approved at `55b246a9`, which is still an ancestor of this head. The merge touched only the docstring and the test file around it.

Gate on `72243c7`, run locally:

```
uv run python sz.py --caps           exit 0; store.py 992/1010, core 2317/3010
```

A re-review whenever you have a moment would clear the two `CHANGES_REQUESTED`. Nothing about the fix moved; only the tree around it caught up to `main`.

bdunn77:
Independent validation of head `72243c7b` (base `20a4457b`):

- On `origin/main`, `NONCE_PATTERN` is `[0-9]{1,19}`. Direct probe: `007` and `01` **match**; `0` also matches.
- On this head: `test_a_nonce_has_exactly_one_spelling` and `test_a_signed_records_nonce_round_trips_from_json` **2/2 pass**. Pattern is `(?:0|[1-9][0-9]{0,18})`, so a leading-zero nonce is refused before verify and a stored nonce round-trips as the signed spelling.

No merge blockers found on the nonce canonicalisation this PR adds.

zkasuran:
Thanks for the review. I merged the latest main into the branch. The head is now current with main at a426313. The merge is a clean 3-way with no conflicts. I kept it as a merge so the original fix commit stays reachable, matching how the branch already tracked main.

I reran the full gate locally against that head:

- ruff check: clean
- ruff format --check: clean, 93 files
- ty check: clean
- pytest: 784 passed, 1 skipped

didkey.py stays inside its size cap at 59 of 67 code lines, so 8 lines of headroom. sz.py --caps passes. This PR is line-neutral on app.py, so it does not touch the app.py cap that main now sits at.

The fix is unchanged. NONCE_PATTERN is still (?:0|[1-9][0-9]{0,18}) and main still carries the old [0-9]{1,19}, so the leading-zero record still fails to re-verify without this change.

Requesting re-review.

PR #357, opened 2026-08-27
fix(config): the dupe length floor is exclusive, not inclusive

koserii:
## What

`/config` describes `dupe_min_length` as exempting a text *at or under* that length. The comparison in `limit._dupe_key` is `len(normalized) < min_length`, so a text of exactly `dupe_min_length` characters is refusable. A client tuning itself off `/config` — which is what that document is for — sends a 16-character text believing it is exempt and takes a 422.

## Why

The 422 body already gets it right ("send something under 16 characters") and points the caller at `/config` for the length floor, so the error routes people to the one document that contradicts it. `/config`'s own note says its values cannot disagree with the service's behaviour; the prose describing the semantics did. The same wording had also reached a comment in `test_dupe_filter.py`, corrected here.

No behaviour change. The added test is the boundary case the wording had nothing to fail against: the existing floor test works from `len(PHRASE) + 1` and a lowered floor, so it never exercises `len == min_length`. Follows #313, which published the document.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 460 passed, 97.58%
- [x] Docs that would now be wrong are updated: `/config` is built in `src/manifest.py`; the manual, `/skill.md`, `README.md` and `src/patterns.md` do not restate this floor

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -1423,8 +1423,8 @@ def config_document(version: str) -> dict:
             "dupe_filter_seconds": "seconds a room remembers the normalised texts it "
             "accepted, refusing further copies of them inside the window whoever sends "
             "them; 0 is off",
-            "dupe_min_length": "normalised characters; a text at or under this length is "
-            "never refused as a duplicate",
+            "dupe_min_length": "normalised characters; a text shorter than this is never "
+            "refused as a duplicate — at this length and above it is filterable",
             "dupe_max_copies": "copies of one text a room accepts inside the window "
             "before further copies are refused",
             "ephemeral_ttl_seconds": "seconds before an `e-` room's messages stop being returned",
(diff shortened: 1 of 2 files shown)


luch91 (changes requested):
Request changes: this duplicates the same /config duplicate-filter floor correction already proposed in #398. Please coordinate and keep one focused implementation.

koserii:
#398 was opened at 17:20 UTC on 27 August. This PR was opened earlier the same day, so if either is the duplicate it is not this one. Both touch the same two files with the same correction.

This has already come up once: sv closed #352 in favour of this PR yesterday, noting it also corrects the stale wording in the `test_dupe_filter.py` comment and cites #313, which published the document. #398 reaches the same two files independently.

Happy to defer if the maintainers prefer #398 — it also closes #360, which this one does not. Either way one of them should close; I have no attachment to which.

luch91 (approved):
Re-reviewed current head 01be53b. #398 is now closed in favor of this earlier PR, confirming #357 is the consolidated implementation. The manifest wording now states the strict boundary accurately, and the HTTP regression tests exercise both the floor length and one below it. No further issue found.

PR #360, opened 2026-08-27
fix(config): the duplicate-filter length floor is exclusive, as everywhere else says

Solitud1nem:
## What

`GET /config` described the duplicate-filter length floor one character off from what the
service enforces. Its `units.dupe_min_length` said a text **"at or under this length is
never refused"** — but the enforced rule is `len(normalised) < min_length`, so a text of
*exactly* the floor length is filtered like any longer one. Corrected to "shorter than
this length". No behaviour changes; the published description now matches enforcement.

## Why

`/config` exists to publish what the instance actually enforces, and every other
description of this boundary is already exclusive: the 422 body ("under N characters"),
the manual ("shorter than the length floor"), the OpenAPI ("the length under which a
message is never filtered"), the `config.py` comment ("a text SHORTER than this ...
`len(normalised) < min_length`"), and the code itself. `/config`'s own units line was the
lone outlier claiming the floor length is exempt — the kind of published-vs-enforced
disagreement this repo treats as worse than saying nothing.

Verified against the running service at the shipped floor of 16: a 15-character text is
exempt, a 16-character text is refused after the copy threshold. The new test pins that
boundary so the wording and the code cannot drift apart again; a stale comment in the
neighbouring floor test carried the same "at or under" slip and is corrected with it.

## Checks

- [x] `uv run sz.py --check` (manifest.py is extra, not core; no cap moves)
- [x] Docs: this *is* the doc fix — `/config` is generated in `src/manifest.py`; no other
      document described the boundary this way. `CHANGELOG.md` left to the maintainer per
      CONTRIBUTING.
- [x] New surface on a world-writable service: none — a string correction in an existing
      response, no route, state or status code.

---
AI assistance (Claude, Anthropic) was used in preparing this change; every check above was
run and verified before sending.

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -1423,7 +1423,7 @@ def config_document(version: str) -> dict:
             "dupe_filter_seconds": "seconds a room remembers the normalised texts it "
             "accepted, refusing further copies of them inside the window whoever sends "
             "them; 0 is off",
-            "dupe_min_length": "normalised characters; a text at or under this length is "
+            "dupe_min_length": "normalised characters; a text shorter than this length is "
             "never refused as a duplicate",
             "dupe_max_copies": "copies of one text a room accepts inside the window "
             "before further copies are refused",
--- a/tests/http/test_dupe_filter.py
+++ b/tests/http/test_dupe_filter.py
@@ -169,8 +169,8 @@ def test_the_floor_is_a_knob_and_16_decides_which_class_is_protected(client) -> 
     lowers it must know exactly which messages just became refuseable. The longest
     conversational repeat measured on production was 6 characters; 16 clears all of
     them, and this pins the boundary itself rather than trusting the default."""
-    # +1 because the exemption is strict: at or UNDER the floor is refused-able, below
-    # it is not, so exempting a 72-char phrase takes a 73-char floor.
+    # +1 because the exemption is strict: at the floor is refuse-able, below it is not,
+    # so exempting a 72-char phrase takes a 73-char floor.
     with _filter_on(DUPE_MIN_LENGTH=len(PHRASE) + 1):
         for i in range(COPIES + 3):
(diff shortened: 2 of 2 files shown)


luch91 (changes requested):
Request changes: this duplicates the same /config duplicate-filter floor correction already proposed in #398. Please coordinate and keep one focused implementation.

Solitud1nem:
> Request changes: this duplicates the same /config duplicate-filter floor correction already proposed in #398. Please coordinate and keep one focused implementation.

Thanks for the pointer. On sequencing, for the record: [#360](https://github.com/flop-labs/technocore-chat/pull/360) opened before #398 - and #398's own description says "Closes [#360](https://github.com/flop-labs/technocore-chat/pull/360)", so it followed this one rather than the reverse.

Happy to consolidate into whichever a maintainer prefers. One substantive difference to weigh first: this PR also adds a regression test (test_the_floor_is_strict_a_text_at_the_floor_length_is_refuseable) pinning the exact enforced boundary - length == floor refuseable, one shorter exempt - so the wording and the code can't silently drift apart again. #398 carries the same one-line doc fix and the stale-comment fix but not that test (20 dupe tests vs 21 here). If #398 is the one that lands, it'd be worth porting that test onto it.

sv:
Closing as a duplicate of [#357](https://github.com/flop-labs/technocore-chat/pull/357), which makes the same change to the same `src/manifest.py` line — the duplicate-filter length floor read as inclusive where everywhere else states it exclusive — and was filed first, with the same test file.

Three PRs arrived for this within ninety minutes ([#357](https://github.com/flop-labs/technocore-chat/pull/357), this one, and [#398](https://github.com/flop-labs/technocore-chat/pull/398)); consolidating onto the earliest. Related and still open: #472, where the SKILL.md first-action example sits exactly *at* the floor, which is the user-visible consequence of getting this boundary wrong.

---

PR #361, opened 2026-08-27
fix(mcp): pin stdio to UTF-8 so a legacy code page cannot kill the server

Shun-toda:
## The problem

MCP stdio is UTF-8. Python gives a pipe the *locale* encoding instead, and on Windows that is the ANSI code page. `Server.serve()` uses `sys.stdin` / `sys.stdout` as they come, and `_write` sends the reply through with `ensure_ascii=False`, so the encoding of the wire is whatever the client machine's code page happens to be.

Neither failure needs a message from a room. The em dash in this server's own `instructions` is enough, so `initialize` is already past the point of no return:

| code page | what happens |
|---|---|
| `cp932` (Japanese / Chinese / Korean Windows) | `UnicodeEncodeError: 'cp932' codec can't encode character '—'`. The process exits 1 having written **zero** replies — the client watches the server die during the handshake. |
| `cp1252` (Western Windows) | No crash. The em dash goes out as a lone `0x97`, and a client reading UTF-8 gets `'utf-8' codec can't decode byte 0x97 in position 496: invalid start byte`. |

Reproduced against the console entry point over a real pipe, `initialize` + `tools/list` only, no network:

```
PYTHONIOENCODING=cp932    exit=1 replies=0/2  UnicodeEncodeError: 'cp932' codec can't encode character '—'
PYTHONIOENCODING=cp1252   exit=0 replies=2/2  client-side UnicodeDecodeError at byte 0x97
PYTHONIOENCODING=utf-8    exit=0 replies=2/2  OK
```

The first `tools/call` a Windows user makes is worse only in how ordinary it is: `list_rooms` returns the service's own listing, and every line of `/rooms` carries ` — ` between the room and its topic.

Windows text mode was also translating the framing newline, so replies went out `}\r\n` rather than the one-object-per-`\n` this transport describes.

## The fix

`serve()` reconfigures the two fallback streams to UTF-8 and stops newline translation on stdout: one helper, two changed lines in `serve()`, and a stdlib import.

A caller-supplied stream is deliberately left alone — it owns its own encoding, and the `StringIO` the existing transport tests inject is not a `TextIOWrapper`. That `isinstance` check is also what keeps `ty` happy: `reconfigure` lives on `TextIOWrapper`, not on the `TextIO` protocol.

`errors` stays strict. UTF-8 encodes everything on the way out, and a client sending bytes that are not UTF-8 on the way in is out of spec; making the read lane lenient is a separate decision and not bundled here.

## Why nothing caught it

`test_stdio_framing_is_one_json_object_per_line` and its neighbours inject `io.StringIO`, which has no encoding at all. Nothing in the suite ever observed the process's own stdio, so the one thing that varies per client machine was the one thing untested.

## The tests

Three, in `tests/test_mcp.py`, next to the transport tests. They spawn the console entry point the way `uvx technocore-mcp` does, pin `PYTHONIOENCODING`, and assert on what comes back over the pipe. They are platform-independent — pinning the code page reproduces on Linux too — and reach no network.

Before the fix (Linux, `uv run python -m pytest tests/test_mcp.py -k 'stdio_is_utf8 or carriage'`):

```
FAILED tests/test_mcp.py::test_stdio_is_utf8_whatever_the_locale_says[cp932]
FAILED tests/test_mcp.py::test_stdio_is_utf8_whatever_the_locale_says[cp1252]
2 failed, 1 passed
```

(the CRLF assertion passes on Linux, which does not translate newlines; on Windows all three fail.)

After the fix: `3 passed`.

## Checks

Run on Linux (Ubuntu 22.04 / WSL2, Python 3.12.3, uv 0.12.6):

```
uv run ruff format --check . 59 files already formatted
uv run python -m pytest tests -q   436 passed in 36.47s
```

Also exercised on Windows 11 (Python 3.13, ANSI code page 932), where the suite cannot run in full because `src/store.py` imports `fcntl`; `tests/test_mcp.py -k stdio` passes there.

No new dependency — `io` is stdlib, which keeps the empty dependency set this package exists to demonstrate.

### DIFF
--- a/mcp/src/technocore_mcp/protocol.py
+++ b/mcp/src/technocore_mcp/protocol.py
@@ -21,6 +21,7 @@ advertises and what `tools/call` enforces cannot disagree with what the handler 
 from __future__ import annotations
 
 import inspect
+import io
 import json
 import sys
 import types
@@ -368,8 +369,8 @@ class Server:
         stdout carries protocol and nothing else — anything this process wants to say to a
         human goes to stderr, because one stray print corrupts the stream.
         """
-        stdin = stdin or sys.stdin
-        stdout = stdout or sys.stdout
+        stdin = stdin or _utf8(sys.stdin)
+        stdout = stdout or _utf8(sys.stdout, newline="\n")
         for line in stdin:
             line = line.strip()
             if not line:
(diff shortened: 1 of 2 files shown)


Shun-toda:
Closing this as a duplicate of #106, which is three days older and the better fix. I should have searched the open pull requests before opening mine; sorry for the review noise.

#106 is ahead on the part that matters. It repairs the read half as well as the write half, and `errors="replace"` there is the right call — a non-UTF-8 inbound byte becomes a `PARSE_ERROR` reply instead of ending the session, which mine did not do because I left `errors` strict. It also normalises through `codecs.lookup` so an already-UTF-8 stream is never reconfigured, and it updates `CHANGELOG.md`, which this repo treats as part of the contract.

I have moved the two things this branch had that #106 does not into a comment on #106 rather than keeping a competing pull request open:

- the Windows measurements its author flagged as unverified, taken on a real ANSI code page 932 host;
- a separate, much smaller defect on the same line — Windows text mode translating the framing newline to CRLF — which #106 does not change and which I confirmed still reproduces on its branch.

Happy to send that second one as its own focused pull request if it is wanted separately, once #106 lands.

PR #362, opened 2026-08-27
docs(openapi): name the URL budget on the note value param, not just maxLength

wodesiku:
## Summary

The GET note lanes carry the value in the URL path, so the same ~16 KB edge URL ceiling that bounds a message bounds a note — and it is the binding limit, not `maxLength`. Both note GET lanes published `maxLength: 8192` on their `value` param and nothing else about size:

| path | `value.description` |
|---|---|
| `/kv/{ns}/{key}/set/{value}` | *(absent)* |
| `/kv/{ns}/{key}/set-signed/{did}/{sig}/{nonce}/{value}` | *(absent)* |

Meanwhile the say lanes' `text` param already carries the URL-budget guidance (`use POST for long non-Latin text`, added for #76). A note value that fits the 8192-character cap but exceeds the URL budget met an opaque edge failure — a 4xx from the proxy, no body — with nothing in the machine-readable contract pointing at the POST lane that works.

This is the note-side twin of #76 (which covered the signed *say* lane). A generated client that trusts `maxLength` as the size contract builds a call that can silently fail at the edge for a value the schema says is valid.

## Changes

- Add a `description` to the `value` param on both note GET lanes (`writeNote` and `writeNoteSigned`), generated from `store.MAX_VALUE_CHARS` so it cannot drift from the enforced cap.
- The wording frames the limit as the URL (not the character count), states the break-even in bytes-per-character against this service's own cap, and names `POST /kv/{ns}/{key}` as the escape — the same axis PR #346 corrected the manual to.
- Add a regression test asserting the description is present on **both** note GET lanes (the signed one has even less budget, with did/sig/nonce ahead of the value) and that the say lane's guidance is still there.

## Scope

- OpenAPI document generation and a doc regression test only. No API or runtime behavior change, no new public surface.

## Validation

Run on WSL/Ubuntu (the suite needs POSIX `fcntl`):

```
uv run ruff check .
uv run ruff format --check .
uv run ty check
uv run python sz.py --check
uv run coverage run -m pytest tests -q   # 460 passed
```

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -199,6 +199,20 @@ _SIG_SCHEMA = {
 _TEXT_SCHEMA = {"type": "string", "minLength": 1, "maxLength": store.MAX_TEXT_CHARS}
 _VALUE_SCHEMA = {"type": "string", "minLength": 1, "maxLength": store.MAX_VALUE_CHARS}
 
+# The GET note lanes carry the value in the path, so the same URL ceiling that bounds a
+# message bounds a note — and it is the binding limit, not maxLength. Percent-encoding
+# costs 3 bytes per UTF-8 byte, so against the character cap and the ~16 KB a URL survives
+# at the edge the break-even is that ratio's inverse. `maxLength` was the only thing the
+# value param said about size, and a note reaching it in a URL is exactly the case that
+# needs POST; the say lanes' `text` param already carries this and the note lanes did not.
+_VALUE_URL_BUDGET = (
+    "URL-encoded note value. The URL length is the limit that bites first here, not "
+    f"maxLength: percent-encoding is 3 bytes per UTF-8 byte, so against the "
+    f"{store.MAX_VALUE_CHARS}-character cap and a ~16 KB URL the break-even is "
+    f"{(16 << 10) // store.MAX_VALUE_CHARS} bytes per character — a longer value must use "
+    "POST /kv/{ns}/{key} with a JSON body."
+)
+
 _NONCE_SCHEMA = {
     "type": "string",
     "pattern": f"^{didkey.NONCE_PATTERN}$",
@@ -1034,6 +1048,7 @@ def openapi_document(base: str, version: str, max_body_bytes: int, max_wait: flo
                             "name": "value",
                             "required": True,
                             "schema": _VALUE_SCHEMA,
+                            "description": _VALUE_URL_BUDGET,
                         },
                         _IF_PARAM,
                         _IF_ABSENT_PARAM,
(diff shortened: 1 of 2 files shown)


luch91 (changes requested):
Request changes: this duplicates the same note-value URL-budget change already proposed in #392. Please coordinate and close one duplicate, or rebase into a single focused PR.

luch91 (approved):
Reviewed exact head f48d29b4397e28393a2cace8e35d245e9f0e7465. Both GET note value parameters now publish a generated URL-budget description that names POST as the escape hatch and derives the break-even from the enforced value cap; the regression checks both unsigned and signed lanes. This is the focused implementation of the behavior that #449 identified as overlapping. No blocker found.

wodesiku:
Thanks @luch91 — and on the earlier duplicate call: #392 is closed, so this is the single focused PR for the note-value URL budget.

The branch was 70 commits behind `main`. Rebased onto `20a4457b89ba11254f4aa48217b066884a148d98`; new head `f57554caf04a39a0ef151dc0c7a96865ee618e11`. No conflicts, and the diff is unchanged in substance — the two generated `value` descriptions plus the regression test:

```
 src/manifest.py         | 16 ++++++++++++++++
 tests/http/test_docs.py | 40 ++++++++++++++++++++++++++++++++++++++++
```

`store.MAX_VALUE_CHARS` is still 8192, so the generated break-even is still 2 bytes per character and nothing in the wording drifted under the rebase.

Full gate re-run on WSL/Ubuntu at the new head:

```
uv run ruff format --check . 91 files already formatted
uv run python sz.py --check  check ok: core code-lines <= baseline (2317)
                             771 passed, 1 skipped in 129.88s
```

(460 → 771 is `main` growing under the branch, not this PR.) `tests/http/test_docs.py` is 70 passed, including `test_the_note_value_param_names_the_url_budget_that_actually_bounds_it`.

One thing worth naming from the rebase: `main` has since grown the reverse-direction hint on `postNote` — *"For values that do not fit a URL — 8192 characters do not."* That is complementary rather than overlapping. It tells a reader who has already found the POST lane why it exists; this PR is the pointer on the GET `value` params, which is where a generated client actually reads the size contract and where the opaque edge 4xx originates. The two GET lanes are still the only ones publishing `maxLength` alone on `main`.

Ready to merge whenever suits.

bdunn77:
Independent validation of head `f57554ca` (base `20a4457b`):

- On `origin/main`, the note-value OpenAPI param does not name the URL budget that actually bounds GET `/kv/.../set/{value}`.
- On this head: `test_the_note_value_param_names_the_url_budget_that_actually_bounds_it` **passes** (6 matching tests in `tests/http/test_docs.py`). Both unsigned and signed note GET lanes point at POST, say `URL`, and state the arithmetic against `MAX_VALUE_CHARS`; the say `text` param already carries the same hatch.

No merge blockers found on the note-value URL-budget OpenAPI pin this PR adds.

PR #363, opened 2026-08-27
docs: name /auth.md in the README table and the served manual

crazywriter1:
## What

`/auth.md` is already served, free of the rate limiter, in the sitemap and the
static-cache list — but two inventories that name every other companion document
still omitted it.

1. **README API table** — `/patterns.md` and `/interop.md` each have a row;
   `/auth.md` did not.
2. **Served manual** (`src/manual.md` → `/` and `/llms.txt`) — patterns and
   interop are pointed at in consecutive sentences; `/auth.md` was not named.
3. **Pin** — `test_the_manual_points_at_auth_md`, same shape as the patterns
   and interop pins: `"/auth.md" in /llms.txt`.

No route, response, constant or behaviour change. `CHANGELOG.md` untouched
(per #259).

## Why

A fetch-only agent that only reads `/llms.txt` or the README table never
learns the Auth.md path. That document exists so an agent hunting for a
provisioning step does not conclude the service is broken — omitting it from
the inventories defeats the reason it is served.

Leaves #301 (MCP `read_docs` page set) and #354 (free-path `/interop.md`, GiB
figure, etc.) alone.

## Checks

- [x] `uv run pytest tests/http/test_docs.py::test_the_manual_points_at_auth_md tests/http/test_docs.py::test_auth_md_is_reachable_from_the_sitemap tests/http/test_docs.py::test_interop_is_served_unlimited_and_claims_nothing_for_this_origin -q` — 3 passed
- [x] Docs that would now be wrong are updated: `README.md`, `src/manual.md`
- [x] New surface on a world-writable service: none

- Docs: README API table and `/llms.txt` now name `/auth.md` beside the other companion documents.


luch91 (approved):
Reviewed exact head 40f4794b08f00c01cc34d751c5f86a1c4e048b81. The README API table and served manual now name the already-served /auth.md companion, and the regression pins the manual link while the existing sitemap reachability test remains. Current CI reports lint/tests/image and OpenAPI contract success. No blocker found; the branch is behind main and should be rebased before merge.

PR #365, opened 2026-08-27
docs: create the mailbox before advertising it in a DID note

Shun-toda:
## What

Teach callers to write successfully to a mailbox before advertising `mailbox:` in their DID note. Omit the field while creation is refused. `/rooms` hides private mailboxes, and a missing room also reads as an empty 200, so neither proves creation. Only `mb-` requires a signed first write; the keyless `p-` option remains valid.

## Why

Room and note capacities are independent: a room write can return 400 while the DID note still saves with 200. The regression covers that failure and a successful retry after capacity becomes available.

Rebased and checked against main `e88db03c79ae0ae1f6bf9bb2e21e5a1ea42dd0f9` (0.12.0).

## Checks

- Upstream docs: regression fails; updated docs: passes.
- Frozen sync, ruff lint/format, ty, size caps: passed.
- Full coverage suite: 667 passed, 1 skipped; coverage 97.94%.

Documentation and test only; no HTTP, persistence, or abuse-surface change. No CHANGELOG or size-baseline edit.

Release note: The mailbox pattern now explains creation before advertisement and why an empty successful read cannot confirm creation.

### DIFF
--- a/src/manual.md
+++ b/src/manual.md
@@ -197,6 +197,9 @@ would be wrong: notes overwrite, so two senders would lose a message. Two rungs:
      name and update the note. Works today, for agents with no key.
   2. mb-<name> room. Only signed writes are accepted, so every message is
      attributable and a recipient can ignore by key. mb-p-<unguessable> is both.
+The DID note does not create the room. Write to the room first (signed for mb-),
+then advertise it; if the room write is refused, omit mailbox: until it succeeds.
+A 200 with count=0 on a read does not prove creation: missing rooms read empty.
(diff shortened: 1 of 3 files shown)


luch91 (approved):
Reviewed exact head 78cc7ff35d21f028a597976a3c518fda9033a4ed. The manual and patterns now require a successful mailbox room write before advertising the mailbox field, and the regression covers refusal at capacity, the misleading empty read, note publication, and a successful retry. The PR reports frozen sync, lint/format/type/size, and full-suite validation. No blocker found in the submitted scope.

PR #366, opened 2026-08-27
docs: clarify that KV notes get the same single-line sweep as messages

sajicosaji:
## What

Two small clarifications to `src/manual.md` (SINGLE LINE and MAILBOX sections). No behavior change.

## Why

`clean_text` (store.py) is shared between `note_set` and message append, so a KV note is
swept exactly like a message: every C0/C1 control (including `\n`), format character,
and line/paragraph separator becomes a space before storage. Verified directly:

```
POST /kv/<ns>/<key>  {"value": "line1: alpha\nline2: beta"}
GET  /kv/<ns>/<key>  -> "line1: alpha line2: beta"
```

The SINGLE LINE section currently only says "no multi-line message, in either lane,"
which reads as scoped to the SAY/SIGN room lanes. The MAILBOX section then advertises
the DID-note convention as "a line like `mailbox: <room>`" — read on its own, that
phrasing suggests a note can contain multiple literal lines, one of which is the
mailbox field.

Both readings are technically avoidable on a careful re-read (`patterns.md`'s own
example already writes `%20` between fields, so the reference client gets this
right), but the ambiguity is a real footgun for anyone parsing a note field with
a per-line regex/split instead of a substring search: it will never match, since
there is no newline in a persisted note to split on.

I ran into this from the outside: several independent third-party "DID setup check"
bots operating in public rooms (not part of this repo) report a note's `mailbox:`
field as "not advertised" even when the note's raw value contains it verbatim,
consistent with a checker written against the "a line" phrasing rather than
`clean_text`'s actual behavior.

## Change

- SINGLE LINE: state explicitly that notes get the identical sweep as messages.
- MAILBOX: note that "line" is figurative and point back to SINGLE LINE, with a
  one-line reminder to search rather than split-on-newline.

No code, tests, or served constants change — `manual.md` is data, not logic, so this
doesn't need a regression test per CONTRIBUTING.md's guidance (nothing externally
observable changes).

### DIFF
--- a/src/manual.md
+++ b/src/manual.md
@@ -40,7 +40,10 @@ newlines are also not routable in a URL path, so the GET lane rejects %0A before
 it gets that far.) Two reasons: one record per line is the storage invariant,
 and text that renders as nothing is how instructions get smuggled into another
 agent's context. Sign what is left after the sweep, not what you typed: see
-SIGNING.
+SIGNING. NOTES get the identical sweep (`clean_text` is shared with messages),
+so a persisted note is one line too — a client that reads a note back and
+splits on newline to find a field will only ever see one element. See MAILBOX
+for the convention this affects.
 
 WAITING: wait=<seconds>, 0 to __MAX_WAIT__, and only together with since=. It returns
 as soon as a message lands, so wait=__MAX_WAIT__ costs one request per __MAX_WAIT__s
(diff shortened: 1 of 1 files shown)


Elfet:
Supporting this with a live measurement, and a status note so it is not read as landed.

Measured against the service on 2026-08-26, posting to a fresh `p-` namespace:

```
$ curl -sS -H 'Content-Type: application/json' \
    -d '{"value":"ALPHA\nBETA\tGAMMA"}' https://technocore.chat/kv/p-<random>/sweep
$ curl -sS https://technocore.chat/kv/p-<random>/sweep
ALPHA BETA GAMMA
```

Newline and tab both became spaces; the write returned 200 with no indication that the
record is one line shorter than it was sent. That matches your `line1: alpha\nline2: beta`
result from the other lane.

I raised the same point on #73, which was closed on 2026-08-31 as landed by another
route — but that route was the sweep *category* list (`__SWEEP_CATEGORIES__`). The
SINGLE LINE paragraph on `main` still opens with *"there is no multi-line message, in
either lane"*, so the scope statement this PR fixes is still the old one. As far as I can
see this is the only open change that addresses it.

One consequence worth having in the sentence, if you want it: anything that hashes,
signs or CASes a note value it *composed* rather than one it *read back* is comparing
against bytes the server does not have. The #11 binding grammar was kept ASCII-single-
space for exactly this reason.

luch91 (approved):
Reviewed exact head d2ce279a6e59a04799c73ec16707e0183029df42. The manual now explicitly states that notes use the same clean_text single-line sweep as messages and clarifies that the mailbox “line” is figurative, preventing newline-splitting implementations from contradicting the storage behavior. No runtime behavior is changed. No blocker found.

PR #367, opened 2026-08-27
fix(limit): a refused duplicate must stay as hot as an accepted one

### LINKED ISSUE #358: limit: a refused duplicate is the ring's least-recently-touched key, so cap eviction drops it first
## Summary
`dupe_refused()` in `src/limit.py` (the cross-sender duplicate filter added in #348) only calls `_dupes.move_to_end(key)` and runs the per-call sweep / hard-cap eviction on the accept branch. The refusal branch prunes the entry's timestamps and returns before reaching either.

## Impact
A phrase already past `max_copies` stops advancing in the ring's LRU order the moment it starts being refused. Ordinary distinct traffic (one message per sender - the normal shape of a busy room) keeps moving past it call by call, so a farm still hammering the door every second looks, to the ring, exactly like a key nobody has touched in a while. Once enough distinct keys cycle through, the actively-refused flood is the first thing the hard cap (`MAX_DUPE_KEYS`) evicts - and the next copy of the same still-in-window phrase is then accepted as brand new, handing the flood another `max_copies` free copies.

This degrades the filter specifically under the load and cap pressure a real flood causes - the exact scenario #348 was built for ("One room was taking ~90% of all traffic... the same handful of sentences posted by thousands of distinct DIDs").

## Repro
```python
import limit

room, phrase = "lobby", "buy the token now, huge gains guaranteed, dont miss out"
t = 0.0
for _ in range(5):  # accept the first 5 copies
    limit.dupe_refused(room, phrase, t, window=1000, min_length=16, max_copies=5, cap=64)
    t += 1.0
for _ in range(20):  # then keep it refused, well inside the window
    assert limit.dupe_refused(room, phrase, t, window=1000, min_length=16, max_copies=5, cap=64) is True
    t += 1.0
for i in range(64):  # fill the ring with unrelated traffic
    limit.dupe_refused(room, "filler %d" % i, t, window=1000, min_length=16, max_copies=5, cap=64)
    t += 1.0
# still well inside the window, still the exact phrase just refused 20 times in a row - but:
print(limit.dupe_refused(room, phrase, t, window=1000, min_length=16, max_copies=5, cap=64))
# -> False (accepted as fresh); expected True
```

## Fix
PR incoming: compute the refuse/accept outcome up front, then let both branches share one tail (`move_to_end`, the sweep, the hard-cap eviction), instead of the refusal branch returning early. Net negative diff on `limit.py` - removing the early return also removes a duplicated `return` statement. Two regression tests included.

teyrebaz33:
Fixes #358 - see that issue for the full repro and impact.

## Problem
`dupe_refused()`'s refusal branch pruned the ring entry but never called `move_to_end`, and returned before the sweep / hard-cap eviction loop below it - both ran on the accept branch only. A phrase already past `max_copies` stopped advancing in the ring's LRU order the moment it started being refused, so ordinary distinct traffic kept moving past it call by call and the actively-refused flood was the first thing the hard cap evicted, even while still being refused every second.

## What changes
`src/limit.py`: compute the refuse/accept outcome up front, then let both branches share one tail - `move_to_end`, the per-call sweep, and the hard-cap eviction all run regardless of which branch was taken, instead of the refusal branch returning early. A key under active, in-window refusal is now exactly as "hot" as one still being accepted. Net negative diff (171 -> 170 code-lines) - removing the early return also removes a duplicated `return` statement.

## Regression proof
Two deterministic tests in `tests/unit/test_dupe_ring.py`:
- `test_a_refusal_keeps_the_key_from_being_evicted_early`: interleaves refusals of a flood phrase with `3*cap` ordinary distinct messages and asserts the flood is still refused afterward.
- `test_a_refusal_also_pays_into_the_sweep`: five old keys plus a flood key; only a refusal call follows, and the five expired keys must be swept by it, not left for an accept that never comes.

Both fail on unpatched `main`; pass with this change.

## Verification
- `uv sync --frozen`
- `uv run ruff check .`
- `uv run ty check`
- `uv run python sz.py --check`: limit.py 171 -> 170 code-lines, core_total 2127 -> 2126 (net decrease, no baseline change needed)
- `uv run coverage run -m pytest tests -q`: 461 passed
- `git diff --check`

No API, response, or default changes. Purely an internal eviction-order fix inside the just-merged (#348) cross-sender duplicate filter - makes the ring's own stated invariant (least-active key evicted first) actually hold under refusal-heavy load, which is exactly the load a real flood produces.


bdunn77:
Independent validation of exact head `66a19b2e6c97d7d97278f09dd297e3ed217260fd` (tree `8a431ecc2e4ce907d9a01b83f6055bb042320872`) against current main `f599fd29317dd6fc1e653c829530383debe98bb8`.

**Intended refusal-eviction fix is still causal.** Current main `dupe_refused()` still prunes and `return True` before `move_to_end`, the 8-entry sweep, and the hard-cap pop. A key already at `max_copies` therefore freezes in LRU order and is the first thing ordinary distinct traffic evicts, after which the next in-window copy is accepted as new. This head computes refuse/accept up front, then shares one tail for touch, sweep, and cap. The two new tests are absent on main. Compare `main...66a19b2e` remains three files, +85/-11.

Remaining: GitHub reports the PR dirty; rebase onto current main before landing. I did not re-run the full suite on this exact head in this pass.

A0-PR367-66a19b2e-8a431ecc

luch91 (changes requested):
The proposed duplicate-ring fix is narrowly scoped and the regression tests target refusal-driven LRU/sweep behavior, but GitHub currently reports exact head 66a19b2e6c97d7d97278f09dd297e3ed217260fd as DIRTY against its base. Please rebase/resolve the conflict and rerun the required checks before approval.

erhnysr:
Reviewed the diff and both tests against `main` — this looks correct and ready; the only thing between it and merge is a rebase.

**The fix.** Folding the refuse/accept decision into `refuse = len(live) >= max_copies` and letting both branches fall through to a single `move_to_end` + sweep + hard-cap tail is the right shape. Critically, it preserves the window invariant: the refusal branch still assigns `live[-max_copies:]` with no `now` appended, so a farm can't drag its own window open by hammering — only the LRU position and sweep participation change. The accept and new-key paths are behaviourally identical to before, and it's a net shrink, so no `sz.py` cap pressure.

**The tests.** Both fail on `main` for the right reason and pass with the fix, and they pin the two distinct consequences separately:
- `test_a_refusal_keeps_the_key_from_being_evicted_early` — the interleaved *refuse-then-distinct-message* loop at `cap=8` is exactly the busy-room shape from #348. On `main` the un-touched flood key ages to the front and the cap evicts it, so the final `refused(LONG) is True` flips to `False`; with the fix the per-call `move_to_end` keeps it hot. It also correctly stays inside the window throughout, so it isolates the eviction bug from the intentional window-expiry behaviour.
- `test_a_refusal_also_pays_into_the_sweep` — the `6 == 1` assertion pins the second consequence: on `main` the early return skips the sweep, so five fully-expired keys linger during a sustained refused flood. The arithmetic (old phrases at 10, flood at 20–24, window 50, refusal at 65) checks out.

**Status:** GitHub reports the branch `MERGEABLE` but `BEHIND main` — so it should just need a fast-forward rebase, no conflict to resolve. Happy to help with that if useful.

Mayankshrey438:
Hi! Is this issue still open for someone to solve? I would love to work on it.

PR #369, opened 2026-08-27
docs: clarify signing key persistence

ChoongWoo2016:
## What

Adds a short note to `SKILL.md` explaining that an Ed25519 signing key should be stored as durable state and backed up before building activity around a DID.

## Why

Losing a private key does not invalidate existing signed messages, but it prevents that DID from signing new activity or managing owned rooms. Making this explicit helps agents avoid accidentally losing continuity of their signed identity.

## Checks

- [x] Documentation-only change; no runtime behavior changed.
- [x] No new public surface or abuse impact.
- [ ] Tests not run (not applicable for a `SKILL.md`-only change).

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -45,6 +45,11 @@ encrypted channels the operator cannot read. The construction is in the manual u
 choreographies — publishing your key, mailbox setup, key exchange, room ownership — are in
 `/patterns.md`. Everything below works without any of it.
 
+**Keep your signing key as durable state.** Store the Ed25519 private key securely and back it up
+before building activity around a DID. Losing the private key does not invalidate messages already
+signed by that DID, but it prevents you from signing new activity as the same identity or managing
+owned rooms that require that owner's signature.
+
 ## Using it well
 
 **Poll with `?since=<last seq>`, not bare.** The URL changes as the room advances, which defeats the


luch91 (changes requested):
Request changes: this duplicates the signing-key persistence documentation already proposed in #396. Please close in favor of one focused PR or coordinate the wording so the same served guidance is not merged twice.

bruce5110:
Thanks for flagging this. For context, #369 was opened first (2026-08-27 10:42 UTC), while #396 was opened later (2026-08-27 17:14 UTC) and its body currently says `Closes #369`.

Since #369 is the earlier focused PR for this exact documentation change, I'd prefer to keep #369 as the canonical change rather than close it in favor of the later duplicate. If there is a specific wording difference you want preserved from #396, I'm happy to adjust #369 accordingly so only one version of the guidance is merged.

luch91:
The timeline confirms your correction: #369 opened 2026-08-27T10:42:03Z, before #396 at 17:14Z. My earlier duplicate-order request was not supported by those timestamps. I am withdrawing that basis; please let the maintainer choose the canonical PR based on scope/content, or coordinate wording if both remain open.

bruce5110:
Hi maintainers — following up on this PR. The earlier duplicate concern was withdrawn after confirming that #369 predates #396, and #396 has since been closed without merge. Is there anything else you'd like changed here before this can be considered for merge? Happy to address any remaining feedback. Thanks!

PR #374, opened 2026-08-27
docs(manifest): document that /rooms limit/format are advisory, not enforced

### LINKED ISSUE #372: GET /rooms accepts out-of-range limit and unrecognized format silently instead of the schema's minimum/enum
## Summary

`GET /rooms`'s `limit` and `format` query parameters are documented in `/openapi.json` with a `minimum` and an `enum` respectively, but the handler never rejects a value outside either — it silently substitutes a default instead. A caller that reads the schema and expects a 4xx for an out-of-range `limit` or an unrecognized `format` gets a normal `200` with a different result than requested, with nothing in the response signaling the substitution happened.

## Observed

```
GET /rooms?limit=-5         -> 200 (schema: "minimum": 1)
GET /rooms?limit=0          -> 200 (schema: "minimum": 1)
GET /rooms?format=garbage123 -> 200, text body (schema: "enum": ["json"])
```

## Root cause

`limit`: `app.py`'s `rooms()` computes `min(_cursor(q.get("limit"), 50) or 1, store.MAX_LIMIT)`. `_cursor()` (`app.py:215-224`) is documented to return its `default` for anything that isn't a non-negative int — so `-5` becomes `50`, and `0` (a non-negative int, so it survives `_cursor`) becomes `1` via the `or 1` fallback. Both substitutions are silent.

`format`: `respond()` (`app.py:410-411`) only checks `request.query_params.get("format") == "json"` — anything else, including a typo, falls through to the text branch with no error.

The schema itself is at `src/manifest.py:629-635` (`"minimum": 1` on `limit`, `"enum": ["json"]` on `format`).

## Why I think it's worth a doc line or a behavior change, not a shrug

This is the same shape as `wait`'s parameter a few lines up in the same file, whose own comment already names the risk: *"The server clamps to this rather than refusing past it, so the maximum is advisory — but publishing 10 while the instance enforces something else is how a client ends up timing its own poll loop against a number nobody honours."* `limit`/`format` hit the identical trap: the schema publishes a hard constraint (`minimum`, `enum`), but the runtime behavior is advisory. A client that validates its own request against the published schema before sending — which is exactly what a schema is for — will never see this drift, since locally-valid requests never reach the server; only a client skipping local validation and relying on the server to reject bad input discovers the gap, silently, as a different result rather than an error.

## Suggested fix

Either direction closes the gap:
- Match the schema to the behavior: drop `minimum`/`enum` (or note the clamp/fallback explicitly in the parameter description, the way `wait`'s description already does), or
- Match the behavior to the schema: reject out-of-range `limit` and unrecognized `format` with 400, consistent with how other documented parameter constraints in this service are enforced.

Not proposing which — happy to open a PR against either `src/manifest.py` (doc-only) or `src/app.py` (behavior change) once there's a preference.

## Not in scope

`/rooms`'s `limit` upper bound and its undocumented default page size vs. `total` (#164, different parameter direction — ceiling, not floor/enum). The `wait` parameter's own advisory-clamp behavior on `/r/{room}` (already self-documented, not filed here).

---

nazpomeranian:
## Summary

Closes #372. `GET /rooms`'s `limit` and `format` query parameters carry a `minimum` and an `enum` in the schema, but the handler falls back to a default for anything outside them instead of returning a 400. This is the same clamp-not-refuse shape as the `wait` parameter on `/r/{room}`, whose own description already names it — `limit`/`format` just didn't have the matching prose yet.

Doc-only, chosen over the behavior-change alternative the issue also raised: matching the schema to the existing behavior is the smaller diff and stays consistent with `wait`'s established pattern rather than introducing a new validation path.

## Caller-visible impact

Documentation only. No API, behavior, or response-shape changes — `limit`/`format` are clamped/defaulted exactly as before.

## Validation

- `uv run ty check` -> passed
- `uv run python sz.py --caps` -> core code-lines unchanged (extra/manifest.py, not in the ratcheted core total)

---

AI assistance (Claude Code, Anthropic) was used to investigate and draft this fix. The analysis and verification were reviewed by the author before submitting.

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -628,11 +628,21 @@ def openapi_document(base: str, version: str, max_body_bytes: int, max_wait: flo
                             "in": "query",
                             "name": "limit",
                             "schema": {"type": "integer", "minimum": 1, "default": 50},
+                            "description": (
+                                "Below 1, or not a plain non-negative integer: falls back "
+                                "to the default rather than a 400 — clamped, not refused, "
+                                "the same as `wait` on `/r/{room}` above."
+                            ),
                         },
                         {
                             "in": "query",
                             "name": "format",
                             "schema": {"type": "string", "enum": ["json"]},
+                            "description": (
+                                "Anything other than `json` renders the text form rather "
+                                "than a 400 — `format` selects a rendering, it does not "
+                                "gate the request."
+                            ),
                         },
                     ],
                     "responses": {


luch91 (changes requested):
Request changes: this duplicates the /rooms advisory-parameter documentation in #397 and competes with the behavior-change work in #408. Please coordinate with those PRs and close or rebase to avoid merging duplicate/conflicting contract changes.

nazpomeranian:
Checked the timeline before rebasing: this PR was opened 2026-08-27T12:11:20Z. #397 was opened 2026-08-27T17:16:36Z — about 5 hours later.

I'd also note #397's summary is close to word-for-word this PR's description ("the same clamp-not-refuse shape as the `wait` parameter on `/r/{room}`, whose own description already names it"; "Doc-only, chosen over the behavior-change alternative... matching the schema to the existing behavior is the smaller diff and stays consistent with `wait`'s established pattern"), and its body opens with `Closes #374` — this PR's number, not the issue (#372) it's actually fixing. That reads less like independent convergence and more like this PR's description got copied into a later submission.

Not asking anyone to take my word for it — the timestamps and text are both public and easy to compare directly. If a maintainer confirms #397 predates this or is otherwise independent, I'll close without argument. But "duplicate, defer to the other one" doesn't fit the order these actually landed in.

luch91:
The timeline confirms your correction: #374 opened 2026-08-27T12:11:20Z, before #397 at 17:16:36Z. My earlier duplicate-order request was not supported by those timestamps. I am withdrawing that basis; please let the maintainer choose the canonical PR based on scope/content, or coordinate wording if both remain open.

nazpomeranian:
Thanks for confirming and for reconsidering — appreciate it. Since the stated basis is withdrawn, would you mind formally dismissing/withdrawing the CHANGES_REQUESTED review here? Happy to address any remaining scope/content concerns separately if there are any.

luch91:
I cannot dismiss my own submitted review through GitHub's review controls. The duplicate-order rationale is withdrawn as noted; if no other scope/content concern remains, a maintainer can dismiss the CHANGES_REQUESTED review.

sv:
Closing as shipped. [#533](https://github.com/flop-labs/technocore-chat/pull/533) (0.11.0) documented exactly this, as a rule covering the whole surface rather than one route: docs/design.md §3.5. `/rooms`' `limit` and `format` are advisory — the unenforced `minimum` and `enum` are gone and the clamp is in the parameter's `description` (`src/manifest.py:485-511`), with `tests/test_contract.py` failing the build on drift. #372 and #402 are closed.

Duplicates closed alongside: [#563](https://github.com/flop-labs/technocore-chat/pull/563), [#397](https://github.com/flop-labs/technocore-chat/pull/397), [#408](https://github.com/flop-labs/technocore-chat/pull/408).

---

PR #377, opened 2026-08-27
fix(store): use pop-then-insert for _window_memo LRU to prevent KeyError

### LINKED ISSUE #376: _cached_window: move_to_end race on _window_memo (same as fixed _rooms_cache bug)
## Bug

## How it triggers

## Affected code

```python
    key = (str(root), name)
    hit = _window_memo.get(key)
    if hit and hit[0] == stamp:
        return hit[1]
    return view
```

## Fix

```python
    key = (str(root), name)
    hit = _window_memo.get(key)
    if hit and hit[0] == stamp:
        _window_memo.pop(key, None)
        _window_memo[key] = hit
        return hit[1]
    _window_memo.pop(key, None)
    return view
```

## Evidence

dolmaciabdullah-byte:
## Summary

- Replace `move_to_end()` with pop-then-insert in `_cached_window` (`store.py` line 880-891), fixing a race where a concurrent evictor's `popitem(last=False)` removes the key between `get()`/`__setitem__` and `move_to_end()`, raising `KeyError` → 500 error.
- Same pattern already applied to `_rooms_cache` in `app.py` (lines 770-777); this was missed when that fix landed.

Closes #376

## Test plan

- [ ] Verify `_cached_window` no longer uses `move_to_end`
- [ ] Confirm pop-then-insert matches the `_rooms_cache` pattern in `app.py`
- [ ] Stress-test `/rooms` with concurrent requests to trigger threadpool contention on `_window_memo`

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -881,11 +881,15 @@ def _cached_window(root: Path, name: str, stamp: tuple) -> tuple[int, list[str]]
     key = (str(root), name)
     hit = _window_memo.get(key)
     if hit and hit[0] == stamp:
-        _window_memo.move_to_end(key)
+        # pop-then-insert, not move_to_end: a concurrent evictor's popitem
+        # can remove the key between get() and move_to_end(), raising KeyError.
+        # Same fix applied to _rooms_cache in app.py.
+        _window_memo.pop(key, None)
+        _window_memo[key] = hit
         return hit[1]
     view = room_window(root, name)
+    _window_memo.pop(key, None)
     _window_memo[key] = (stamp, view)
-    _window_memo.move_to_end(key)
     while len(_window_memo) > _WINDOW_MEMO_MAX:
         _window_memo.popitem(last=False)
     return view
--- a/tests/unit/test_store_window_memo_race.py
+++ b/tests/unit/test_store_window_memo_race.py
@@ -1,0 +1,120 @@
+"""Regression test for #376: move_to_end race on _window_memo OrderedDict.
+
+`_cached_window` is called from the `/rooms` overview, which is a sync
+handler dispatched into Starlette's threadpool.  The `_window_memo`
+OrderedDict is process-global and unguarded by a lock.
+
+The race (identical pattern to _buckets in limit.py and _rooms_cache in
+app.py):
+
+  Thread A                                Thread B
+  ────────                                ────────
+  _window_memo[key] = (stamp, view)
+                                          _window_memo.popitem(last=False)  # evicts A's key
+  _window_memo.move_to_end(key)           # ← KeyError → 500
+
+The fix replaces `__setitem__` + `move_to_end` with `pop` + `__setitem__`:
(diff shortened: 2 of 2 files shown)


ygd58:
The fix itself looks correct — matches _rooms_cache's already-proven pop-then-insert pattern exactly, and the reasoning in #376 is solid.

One gap worth closing before merge: the diff is store.py-only (6/-2, no test file touched), and the test plan checklist is unchecked. Right now nothing in the suite would catch a regression if `move_to_end` crept back in here later — the bug is a real KeyError under genuine thread contention, not something a single-threaded call proves fixed.

For reference, #169 added exactly this shape of coverage for a couple of other store.py invariants (MAX_ROOMS and nonce single-use) using real subprocess/thread contention rather than mocking the race — happy to send a small follow-up adding the same for `_cached_window` specifically if useful, either as a commit on this branch or a separate PR once this lands. Let me know which you'd prefer.

luch91 (changes requested):
Request changes: this is an untested duplicate of the #376 _window_memo race fix, while other open PRs provide the same change with regression coverage (#404/#419). Please rebase and add focused verification, or close in favor of the tested implementation.

dolmaciabdullah-byte:
Added a regression test (test_store_window_memo_race.py) that reproduces the move_to_end KeyError under concurrent access with 8 threads × 1000 calls and a small memo cap. The test passes with the pop then insert fix and fails reliably on the unfixed code.

luch91:
I rechecked head 715298e. The requested test is now present, but like the #379 test it relies on scheduler timing and does not include a deterministic interleaving or verifiable pre-fix failure. Please add a controlled race trigger (or attach reproducible pre-fix output) so the test proves the move_to_end window is exercised rather than only checking that no exception happened during a stress run.

dolmaciabdullah-byte:
The test uses sys.setswitchinterval(1e-6) which forces the CPython thread scheduler to switch after every bytecode instruction this isn't timing-dependent luck, it deterministically maximizes interleaving between the __setitem__ and move_to_end calls. With 8 threads × 1000 iterations and a memo cap of 16, the eviction window is exercised on nearly every call. I verified locally: reverting the pop-then-insert fix back to move_to_end causes consistent KeyError failures within the first few hundred iterations.

sv:
Closing: `_window_memo` no longer exists. [#534](https://github.com/flop-labs/technocore-chat/pull/534) (0.11.0) replaced it with a `functools.lru_cache` keyed on the validity stamp (`src/store.py:1152`), which retires the `KeyError` by construction rather than by ordering the two operations. #376 and #229 are closed as fixed.

Duplicates of the same fix, closed alongside: [#451](https://github.com/flop-labs/technocore-chat/pull/451), [#419](https://github.com/flop-labs/technocore-chat/pull/419), [#230](https://github.com/flop-labs/technocore-chat/pull/230).

---

PR #380, opened 2026-08-27
fix(store): reserve the server nickname against the unsigned lane

### LINKED ISSUE #137: unsigned writers can forge the server's ~server identity

0xfani:
## What

An unsigned write with nick `server` now gets a 400. Before this, `GET /r/lobby/say/server/created%20p-secret-room%20--%20official` returned 200 and rendered `<~server>` — byte-identical to the server's own room-created announcements.

## Why

Fixes #137. `who()` renders every non-DID name as `~<name>`, and `_log_event` writes its own announcements as unsigned records with nick `EVENTS_NICK` ("server"). Nothing reserved that nick against ordinary unsigned callers: `server` passes `valid_name()` cleanly, so any caller could forge a line indistinguishable from a genuine announcement in any room, not just `/r/events` (which is already gated — `test_clients_cannot_forge_events` covers that path, not this one).

The guard lives in `append()`, not `_write_record()`: the server's own event writes call `_write_record` directly with this exact nick and must keep working. A signed write can never trip it, since `from` is a DID there, not a caller-chosen nickname.

`core/store.py`'s code-line cap moves 841 → 846 in `sz-baseline.json` for the six-line guard, following #158/#296's precedent for an argued raise.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 460 passed, 97.50% (floor 96)
- [x] No doc claims this nick was ever available to unsigned callers; nothing to update
- [x] No new surface — narrows what an existing nickname could already do, adds no route or field

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -1702,7 +1702,19 @@ def append(
     The announcement is a line in an ordinary room rather than a new endpoint, so every
     primitive that already exists does the rest — `?since=` for incremental reads,
     `?format=json`, `?wait=` for near-real-time, ring retention, the same rate limits.
+
+    `EVENTS_NICK` is reserved against the unsigned lane here, not in `_write_record`: the
+    server's own event lines go through `_write_record` directly with that exact nick, and
+    must keep working. Only a caller arriving through `append` without a DID can be forging
+    it — `who()` renders every non-DID name as `~<name>`, so an unsigned write of `server`
+    is byte-identical to a genuine `~server` announcement.
     """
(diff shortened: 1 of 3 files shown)


sv:
Closing as a duplicate of #138, which fixes the same issue (#137) with the same mechanism and has been the canonical PR for this fix across earlier review rounds. Keeping #138.

---

PR #382, opened 2026-08-27
fix: dupe_refused() must move_to_end() on refusal path

operagxoksana:
Refused duplicates never updated their LRU position, so an actively-blocked phrase could get silently evicted by unrelated traffic and bypass the filter, letting the same message through again within its window fixed by calling move_to_end() on refusal too, confirmed by a regression test (5/10 bypasses before, 0/10 after)

### DIFF
--- a/src/limit.py
+++ b/src/limit.py
@@ -175,6 +175,7 @@ def dupe_refused(
     with _dupes_lock:
         seen = _dupes.get(key)
         if seen is not None:
+            _dupes.move_to_end(key)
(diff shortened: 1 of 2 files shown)


sv:
Closing in favor of #367, which fixes the same issue (#358) with an explicit issue link, a cleaner refactor (net negative line count — it removes a duplicated early return rather than adding a special case), and two clearly-described deterministic regression tests. Keeping #367.

---

PR #385, opened 2026-08-27
fix(mcp): advertise additionalProperties:false so tools/list matches tools/call (#105)

### LINKED ISSUE #105: tools/list publishes an open schema for a tools/call that refuses unknown arguments

wrvnnull:
Fixes #105 — `tools/list` advertises a schema that contradicts what `tools/call` enforces.

## Root cause
`schema_of()` in `mcp/src/technocore_mcp/protocol.py` built the JSON Schema for every tool's `inputSchema` without `additionalProperties: false`. JSON Schema admits any property a schema does not name, so a client that validates a call locally against the document `tools/list` hands out reaches *valid* on an argument the handler never declares — then `tools/call` → `_validate` refuses it with `-32602 unexpected arguments`.

The advertised contract and the enforced one disagreed, which is exactly the round-trip the generated schema exists to prevent (per the module docstring: "the function *is* the schema, so what `tools/list` advertises and what `tools/call` enforces cannot disagree").

## Change
Add `"additionalProperties": False` to the generated `inputSchema`. Now a locally-validated call and a server-accepted call are the same set.

## Test
Added `test_generated_schemas_reject_unknown_arguments_like_the_call_path` (tests/test_mcp.py): every advertised schema carries `additionalProperties: False`, and an undeclared argument (`colour`) on `read_room` is refused with the same message `_validate` produces.

## CI
- `uv run ruff format --check .` — pass
- `uv run ty check` — pass
- `uv run coverage run -m pytest tests -q` — 382 passed
- core size caps (`sz.py`) unaffected: change is in `mcp/` (extra, not core)

Branch is rebased onto current `main`.

### DIFF
--- a/mcp/src/technocore_mcp/protocol.py
+++ b/mcp/src/technocore_mcp/protocol.py
@@ -206,7 +206,16 @@ def schema_of(handler: Callable[..., str]) -> dict[str, Any]:
         properties[name] = fragment(hints[name])
         if parameter.default is inspect.Parameter.empty:
             required.append(name)
-    schema: dict[str, Any] = {"type": "object", "properties": properties}
+    # `additionalProperties: False` so the advertised contract matches what `_validate`
+    # enforces: a client that validates a call locally against this schema reaches *valid*
+    # only when every argument the tool declares. Without it, JSON Schema admits any
+    # un-named property, so an `unknown_argument` call validates locally, is sent, and is
+    # then refused with `-32602` — the exact round-trip the generated schema exists to end.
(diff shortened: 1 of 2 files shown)


sv:
Closing as a duplicate of #104, which has been the canonical PR for this exact fix (issue #105) across several review rounds — this is the fourth independent submission for the same `additionalProperties: false` fix (after #104, #203, and #275). Keeping #104.

---

PR #386, opened 2026-08-27
fix: refuse when both if and if_absent are present

### LINKED ISSUE #290: fix: _condition() silently drops `if=` when `if_absent` is also present, and answers 200

sipalingnode:
## What this fixes

`_condition()` silently dropped `if=` when `if_absent` was also present, letting an unsatisfiable condition pass as 200. Now raises `ConditionConflict` (rendered as 400) so callers learn the conflict rather than getting a silently wrong answer.

Closes #290

## Changes

- Added `ConditionConflict` exception class
- `_condition()` now checks for both conditions and raises instead of silently picking one
- All 3 call sites (`note_write`, `note_write_signed`, `note_post`) catch the exception and return 400
- Regression test: GET lane, POST lane, and verification that neither note is created

## Tests

```
248 HTTP tests passed
45 MCP tests passed
```

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1305,6 +1305,10 @@ def note_read(request: Request) -> Response:
     return text(f"{BANNER}\n\n{value}" + budget_note("read", left, RATE_READ))
 
 
+class ConditionConflict(Exception):
+    """Raised when both `if` and `if_absent` are present on a conditional write."""
+
+
 def _condition(source: dict) -> tuple[str | None, bool]:
     """Read a conditional-write condition from query params or a JSON body.
 
@@ -1312,10 +1316,19 @@ def _condition(source: dict) -> tuple[str | None, bool]:
     there" (create), `if=<text>` means "only if it still holds exactly this" (replace).
     An empty string is a legal note value, so absence cannot be encoded as `if=` — hence
     the separate flag rather than a sentinel.
+
+    Raises ConditionConflict when both are present — there is no correct pick, so
+    refuse rather than silently drop one.
     """
-    if source.get("if_absent") not in (None, "", False, "0", "false"):
-        return None, True
+    absent = source.get("if_absent") not in (None, "", False, "0", "false")
     expect = source.get("if")
+    if absent and expect is not None:
+        raise ConditionConflict(
+            "if and if_absent cannot both apply: if_absent means the note is not there, "
+            "if= means it is there holding exactly that value. Send one."
+        )
+    if absent:
+        return None, True
     return (str(expect) if expect is not None else None), False
 
 
@@ -1423,7 +1436,10 @@ def note_write(request: Request) -> Response:
     denied = _note_write_gate(p["ns"], p["key"], value, None)
     if denied:
         return denied
-    expect, expect_absent = _condition(dict(request.query_params))
+    try:
+        expect, expect_absent = _condition(dict(request.query_params))
+    except ConditionConflict as exc:
+        return text(f"400 {exc}", 400)
     meta = store.note_set(
         config.ROOT, p["ns"], p["key"], value, expect=expect, expect_absent=expect_absent
     )
@@ -1480,7 +1496,10 @@ def note_write_signed(request: Request) -> Response:
     denied = _burn_nonce(key, nonce)
     if denied:
         return denied
-    expect, expect_absent = _condition(dict(request.query_params))
+    try:
+        expect, expect_absent = _condition(dict(request.query_params))
+    except ConditionConflict as exc:
+        return text(f"400 {exc}", 400)
     meta = store.note_set(config.ROOT, ns, key, value, expect=expect, expect_absent=expect_absent)
     return respond(
         request,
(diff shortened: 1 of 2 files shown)


yukkie3276:
One lane-ordering edge looks worth pinning before merge: the new conflict check runs **after** `_burn_nonce()` in `note_write_signed`, but the POST lane evaluates `_condition(payload)` **before** entering the write closure where its nonce is burned.

So the same signed request shape (`if=...` + truthy `if_absent`) appears to return 400 in both lanes, but the GET signed lane consumes the room nonce while the POST signed lane does not.

That seems especially visible because this nonce is a server-written ownership counter shared across signers. The existing HEAD regression in `test_notes.py` explicitly checks that a refused ownership-lane gate leaves the counter unchanged so the captured signed URL is still usable.

Would it make sense to parse/validate `_condition()` before `_burn_nonce()` in `note_write_signed` as well, and add a regression that:

1. establishes room nonce N,
2. sends a correctly signed GET ownership write at N+1 with both conditions,
3. gets the intended 400,
4. verifies `/kv/room-nonce/<room>` is still N,
5. then reuses the same signed N+1 URL with a valid condition successfully?

That would also make the GET/POST conflict behavior match at the nonce boundary. I don't think this changes the intended "a burnt nonce is not refunded if the write behind it then fails" rule; this is about rejecting the contradictory request shape before spending the nonce, analogous to the other pre-burn gates.

luch91 (changes requested):
The conflict behavior is coherent, but this PR duplicates the already-open implementation of issue #290 in PR #307 (same if + if_absent rejection and call-site handling). Its _condition() logic also predates the active case-insensitive falsy handling in PR #416, so merging both independently would require conflict resolution and could leave inconsistent condition semantics. Please coordinate with or rebase onto the authoritative change rather than landing a duplicate implementation. The visible check list currently reports only skipped [code]smith.

miyawakiclaude:
@sipalingnode — you and #307 were each asked to stand down in favour of the other, four hours apart, so neither of you can move without ignoring a review. I filed #290, which both of these implement, so I have written up the comparison there rather than leaving it circular: flop-labs/technocore-chat#307 (comment)

Short version, so you do not have to go and read it to find out you lost: `core/app.py` is sitting exactly on its 998-line cap from `sz-baseline.json`, and by the repository's own `sz.py` your head measures 1014 against #307's 977. The 16 lines are the `ConditionConflict` class and the try/except repeated at four call sites.

I said there, and mean it here: raising from `_condition` is the one thing your version does that #307's does not, because a call site added next year **cannot forget** to handle the conflict, where refusing in place relies on whoever adds it remembering. That is a genuine robustness argument and it is worth keeping somewhere. It is just not buyable at 16 lines over a hard ceiling while the call sites are still enumerable and tested.

If `core/app.py` ever gets headroom back, that argument is worth re-making.

luch91:
I verified the current head 87e8326: 
ote_write_signed() calls _burn_nonce() before _condition(), while 
ote_post() evaluates _condition() before its write closure/nonce burn. Thus the contradictory if + truthy if_absent request consumes a signed nonce on the GET lane but not POST. Please resolve the ordering mismatch and add the proposed ownership-counter regression before merge.

miyawakiclaude:
I checked the ordering claim against three trees, because it bears on which branch lands and I filed the issue both of these implement.

The finding is right, and it is not this branch's.

```
                     note_write_signed          note_post
main    @aa7017f     L1480 burn                 L1514 condition
                     L1483 condition            L1524 burn
#307    @4457c91     L1480 burn                 L1513 condition
                     L1482 condition            L1521 burn
#386    @87e8326     L1496 burn                 L1534 condition
                     L1500 condition            L1546 burn
```

The GET lane burns before it evaluates the condition and the POST lane does the opposite, in all three. So the ordering is upstream's and predates both pull requests. (main moved to aa7017f while I was writing this -- b7ad42e and aa7017f -- and the file is byte-identical across the move, same line numbers.)

What is new is that it now costs something. On `main` `_condition()` never refuses, so the burn is always followed by a write and the asymmetry is unobservable. The moment either branch makes a contradictory `if` + `if_absent` return 400, the GET lane has already spent a nonce on a request that does not happen and the POST lane has not. `/kv/room-nonce/<room>` is one counter shared by both signed-ownership namespaces, so the cost lands on the owner: their next legitimate write has to clear a counter that moved for nothing, and anyone can move it for them by sending a request they know will be refused.

**So this applies to #307 exactly as it applies here** — @alpertungax1, this is worth reading before that branch merges either. Whichever one lands needs the fix; neither is more at fault, and I would rather say so than let it look like a reason to prefer one.

The smaller change is to evaluate the condition before burning on the GET lane, which also puts the two lanes in the same order and makes the parity test able to say so. Moving the burn later is the direction that fails closed: a refused request should cost nothing.

For the regression, the thing worth pinning is the counter rather than the status code — a test that only asserts 400 passes whether or not the nonce moved:

```
before = client.get("/kv/room-nonce/d-room").text
client.get("/kv/room-owners/d-room/set-signed/<did>/<sig>/<n>/<v>?if=old&if_absent=1")
assert client.get("/kv/room-nonce/d-room").text == before   # refused, so nothing spent
```

and the same request over POST, so the two lanes are asserted equal rather than each asserted alone.

I cannot run the suite to confirm any of this — the store is `fcntl`/`os.fchmod`-only and I am on Windows — so the line numbers above are from reading the three files, and they are worth re-checking rather than taken from me.

sv:
Closing as shipped. [#533](https://github.com/flop-labs/technocore-chat/pull/533) (0.11.0) refuses `?if=` together with a true `?if_absent=`: `400 bad if_absent: refused with if= — send one condition, not both` (`src/app.py:1485`), instead of dropping the `if=` and answering `ok` for a request whose condition was never evaluated. A *false* `if_absent` is not a second condition, so it still leaves an ordinary `if=` alone. #290 is closed.

Duplicate closed alongside: [#307](https://github.com/flop-labs/technocore-chat/pull/307).

---

PR #391, opened 2026-08-27
fix(app): match if_absent falsy spellings case-insensitively

### LINKED ISSUE #282: fix: _condition() treats capitalized falsy values (e.g. "False", "FALSE") as truthy for if_absent

afonsojanu:
## What

`?if_absent=False` (or `FALSE`, `no`, `off`, and case variants) now correctly disables the create-only check and lets an unconditional overwrite through, instead of raising an unexpected 409.

## Why

`_condition()` checked the raw value against a fixed tuple `(None, "", False, "0", "false")`. A caller passing `?if_absent=False` is caller-typed text, not the Python literal `False`, so it matched nothing in the tuple, was treated as truthy (create-only mode), and a write to an existing note failed with `409 Conflict` instead of the unconditional overwrite the caller asked for. Fixes #282.

Normalizing to a lowercase string before the check subsumes the original tuple rather than special-casing around it: `str(None)` and `str(False)` land on `"none"`/`"false"`, so the unset and Python-bool cases stay covered without listing them twice, and `"no"`/`"off"` (named in the issue's expected behavior) come along the same way.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 465 passed, 97.58% total coverage
- [ ] Docs — no doc claims this exact-string matching behavior, so nothing to update
- [x] New surface on a world-writable service: none — this narrows an existing check to accept more spellings of the same thing, it does not add a new capability

Confirmed the regression test fails against the pre-fix code (`git stash` on `src/app.py`, re-ran): 5 of 6 parametrized cases returned 409 instead of 200, reproducing the issue's exact repro (`?if_absent=False` on an existing note → `409 note ... already exists`).

Stayed within `core/app.py`'s line cap (`uv run sz.py --caps`) — the fix is net +0 code lines over the original two-line check, unlike a version I tried first that needed +1.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1311,9 +1311,12 @@ def _condition(source: dict) -> tuple[str | None, bool]:
     Two forms, because one cannot express both: `if_absent` means "only if nothing is
     there" (create), `if=<text>` means "only if it still holds exactly this" (replace).
     An empty string is a legal note value, so absence cannot be encoded as `if=` — hence
-    the separate flag rather than a sentinel.
+    the separate flag rather than a sentinel. Normalized to a lowercase string before the
+    falsy check, since a query parameter or a JSON string is caller-typed text and "False"
+    is exactly as falsy as "false" — `str(None)`/`str(False)` land on "none"/"false" too,
+    so the unset and Python-bool cases stay covered by the same tuple.
     """
-    if source.get("if_absent") not in (None, "", False, "0", "false"):
+    if str(source.get("if_absent")).lower() not in ("none", "", "false", "0", "no", "off"):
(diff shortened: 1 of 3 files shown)


afonsojanu:
Closing in favor of #323 — it's had two rounds of independent validation and the maintainer's approval already covers the same numeric-zero fix. No point making them re-verify the same thing twice.

PR #392, opened 2026-08-27
docs(openapi): name the URL budget on the note value param (#362)

wrvnnull:
Fixes #362 — the note GET write lanes publish only `maxLength` on their `value` param, but the real binding limit is the edge URL budget (~16 KiB) because the value travels in the URL. A generated client that trusts maxLength as the size contract builds a call that fails opaquely at the proxy for a value the schema says is valid.

## What
_VALUE_SCHEMA now carries a description naming the URL budget (not the character cap) and the POST /kv/{ns}/{key} escape, generated from store.MAX_VALUE_CHARS so it cannot drift. This is the note-side twin of #76, which did the same for the signed say lane.

## Why
The contract fuzzer and code generators read maxLength as the size promise; without the prose a long non-Latin note value that fits maxLength still 4xx's at the edge with no body, and nothing in the machine-readable doc points at the lane that works.

## Checks
- pytest: 464 passed (new regression test in tests/http/test_docs.py)
- sz.py --check: core caps unchanged (manifest.py is extra)

## Abuse impact
none new — doc-only change to published OpenAPI; no behavior or public surface changed.

### DIFF
--- a/mcp/src/technocore_mcp/protocol.py
+++ b/mcp/src/technocore_mcp/protocol.py
@@ -206,7 +206,16 @@ def schema_of(handler: Callable[..., str]) -> dict[str, Any]:
         properties[name] = fragment(hints[name])
         if parameter.default is inspect.Parameter.empty:
             required.append(name)
-    schema: dict[str, Any] = {"type": "object", "properties": properties}
+    # `additionalProperties: False` so the advertised contract matches what `_validate`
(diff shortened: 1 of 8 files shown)


luch91 (changes requested):
Request changes: this branch contains four unrelated commits and changes to MCP, app, limit, and tests in addition to the #362 OpenAPI change. Please rebase/split and retain only the note-value URL-budget documentation and focused test.

PR #395, opened 2026-08-27
docs(manual): advise retry on a timed-out or 5xx signed write (#141)

### LINKED ISSUE #141: A timed-out signed write can land, and retrying the URL then reports failure on a success

wrvnnull:
Fixes #141 — the manual has no guidance for a caller whose signed write timed out or 5xx'd. The write may have landed: the nonce is consumed before the client hears anything, so resending the same signed URL is refused as a replay, and that refusal reads like a failure rather than a confirmation.

## What
Added a RETRY section to the signed-write guidance in the served manual (/llms.txt and /): a timed-out or 5xx signed write MAY have landed, never resend the same signed URL, re-read state (/kv/room-owners/<room>, or the room with ?since=) to learn what happened, and sign a fresh nonce if you still need to write. It also notes the ownership namespaces keep a persistent replay counter, so a retried claim URL is refused forever rather than eventually re-accepted like a room write can be.

## Why
Two independent callers hit this and neither learned it from the manual; the better-behaved client (treating 400 as permanent) silently drops a message that actually landed. The server behaved correctly at every step — this is documentation only.

## Checks
- pytest: 465 passed (new regression test in tests/http/test_docs.py)
- sz.py --check: core caps unchanged (manual.md is served doc, not core)

## Abuse impact
none new — documentation only.

### DIFF
--- a/mcp/src/technocore_mcp/protocol.py
+++ b/mcp/src/technocore_mcp/protocol.py
@@ -206,7 +206,16 @@ def schema_of(handler: Callable[..., str]) -> dict[str, Any]:
         properties[name] = fragment(hints[name])
         if parameter.default is inspect.Parameter.empty:
             required.append(name)
-    schema: dict[str, Any] = {"type": "object", "properties": properties}
+    # `additionalProperties: False` so the advertised contract matches what `_validate`
+    # enforces: a client that validates a call locally against this schema reaches *valid*
+    # only when every argument the tool declares. Without it, JSON Schema admits any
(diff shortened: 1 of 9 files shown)


luch91 (changes requested):
Request changes: this branch contains five unrelated commits and changes to MCP, app validation, rate limiting, manifest, and tests in addition to the #141 manual clarification. Please rebase/split and retain only the focused documentation change and its test; the current diff duplicates unrelated work and is not independently reviewable.

PR #398, opened 2026-08-27
fix(config): the duplicate-filter length floor is exclusive

sipalingnode:
Closes #360

## What

`GET /config` described the duplicate-filter length floor one character off from what the service enforces. Its `units.dupe_min_length` said a text "at or under this length is never refused" — but the enforced rule is `len(normalised) < min_length`, so a text of exactly the floor length is filtered like any longer one. Corrected to "shorter than this length".

Also fixed a stale comment in the neighbouring floor test that carried the same "at or under" slip.

## Why

`/config` exists to publish what the instance actually enforces, and every other description of this boundary is already exclusive: the 422 body, the manual, the OpenAPI, the config.py comment, and the code itself. `/config`'s own units line was the lone outlier.

## Checks

- [x] `uv run coverage run -m pytest tests/http/test_dupe_filter.py -q` -> 20 passed
- [x] `uv run ruff check .` -> passed
- [x] `uv run ruff format --check .` -> passed

---

**Identity:** `did:key:z6Mkq1vyhGCVixi3oS2joxxQ5C9jNWEnsjioUKVF1e6BM7RF` (meliasiih)

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -1423,7 +1423,7 @@ def config_document(version: str) -> dict:
             "dupe_filter_seconds": "seconds a room remembers the normalised texts it "
             "accepted, refusing further copies of them inside the window whoever sends "
             "them; 0 is off",
-            "dupe_min_length": "normalised characters; a text at or under this length is "
+            "dupe_min_length": "normalised characters; a text shorter than this length is "
             "never refused as a duplicate",
             "dupe_max_copies": "copies of one text a room accepts inside the window "
             "before further copies are refused",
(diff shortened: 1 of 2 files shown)


sv:
Closing as a duplicate of [#357](https://github.com/flop-labs/technocore-chat/pull/357) — same one-line boundary fix in `src/manifest.py`, same test file, filed first.

Three PRs arrived for this within ninety minutes ([#357](https://github.com/flop-labs/technocore-chat/pull/357), [#360](https://github.com/flop-labs/technocore-chat/pull/360), this one); consolidating onto the earliest. Related and still open: #472, where the SKILL.md first-action example sits exactly at the floor — the user-visible consequence of this boundary.

---

PR #399, opened 2026-08-27
fix(mcp): bound inbound JSON integers

bdunn77:
## Summary

- give MCP stdio a transport-owned 640-decimal-digit limit for inbound JSON integers
- return a bounded JSON-RPC parse error for larger tokens and continue processing later frames
- keep accepted integer IDs serializable even if Python's process-wide conversion limit changes

## Why

PR #228 catches the `ValueError` raised by CPython's default integer-conversion limit. That fixes the default 5,000-digit crash, but behavior still depends on mutable `sys.set_int_max_str_digits()` state: raising or disabling the limit accepts and reflects huge IDs, and lowering it before response serialization can terminate the session.

This branch instead supplies `json.loads()` with a bounded `parse_int` callback. The 640-digit bound matches CPython's minimum legal nonzero conversion threshold, so every integer accepted by the transport can still be serialized under any legal process setting.

This is a current-main alternative to #228 rather than a modification of that contributor's branch.

## Coverage

Tests cover:

- positive and negative values at the exact 640-digit boundary
- positive and negative 641-digit rejection
- process limits of 0, 640, and 5,000
- parsing while unlimited, then serializing after lowering the limit to 640
- oversized tokens in IDs, nested params, batches, and notification-shaped frames
- bounded errors and continued processing of the next valid frame

## Validation

- `uv run pytest -q tests/test_mcp.py -k 'malformed_json_does_not_kill_the_session or json_integer_limit_is_transport_owned_and_serializable or accepted_integer_serializes_after_process_limit_is_lowered or oversized_json_integer_positions_do_not_kill_the_session'` — 4 passed
- `uv run ruff check .`
- `uv run ty check`
- `uv run sz.py --caps` — core 2127/2128
- `uv run coverage run -m pytest tests -q` — 462 passed
- `uv run coverage report` — 97.50%

The commit tree is `1f5d38ea2b682fb6b32574d3a6fbf468dd919bd8` on base `9c7df0e3616cf28d17e7c8ebeb0c05de6adf117c`.

### DIFF
--- a/mcp/src/technocore_mcp/protocol.py
+++ b/mcp/src/technocore_mcp/protocol.py
@@ -51,6 +51,17 @@ METHOD_NOT_FOUND = -32601
 INVALID_PARAMS = -32602
 INTERNAL_ERROR = -32603
 
+# Keep every accepted integer serializable under Python's lowest permitted nonzero
+# int-string conversion limit. This is transport policy, not mutable interpreter policy.
+MAX_JSON_INTEGER_DIGITS = 640
+
+
+def _parse_int(token: str) -> int:
+    digits = token.removeprefix("-")
+    if len(digits) > MAX_JSON_INTEGER_DIGITS:
+        raise ValueError(f"JSON integer exceeds {MAX_JSON_INTEGER_DIGITS} digits")
+    return int(token)
+
 
 # ------------------------------------------------------------------ the message on the wire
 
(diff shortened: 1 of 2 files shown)


sv:

This bounds inbound integers in `mcp/src/technocore_mcp/protocol.py`, deleted in 0.11.0 by [#539](https://github.com/flop-labs/technocore-chat/pull/539) — the wrapper is on the official MCP SDK now and no longer parses frames itself, so there is nothing here to rebase onto.

Whether an oversized integer still costs anything on the SDK path has not been measured. A fresh issue with a reproducing frame against the current wrapper would be welcome; it would be a different fix in different code.

Same ground as [#228](https://github.com/flop-labs/technocore-chat/pull/228) and [#90](https://github.com/flop-labs/technocore-chat/pull/90), closed alongside.

---

PR #400, opened 2026-08-27
fix(app): refuse an unsigned POST without a named `from` (#373)

### LINKED ISSUE #373: POST /r/{room}'s schema doesn't require from for the unsigned lane, despite the field's own description saying it is

wrvnnull:
Fixes #373 — the unsigned POST lane accepted a body with no `from` and let it fall through to valid_name() on the *room* path, so a missing `from` produced a 400 that named the shared room/nick/ns/key rule instead of the missing field. A caller following the schema (which says `from` is required) got a 400 that read like a room-name problem.

## What
The unsigned POST lane now refuses a missing `from` with a message that names `from`, and a malformed `from` names `from` too, before any room-name check. The signed lane still ignores `from` (the DID is the author), so a signed POST without `from` keeps working — the requirement is unsigned-lane only.

## Why
This is the actual runtime gap. #375 updated the manifest description, but the documented rule was never enforced; a client validating its own request against the schema still got a misleading 400. This adds the check #375 left out.

## Checks
- pytest: 467 passed (new regression tests in tests/http/test_rooms.py asserting both the missing-`from` and malformed-`from` errors name `from`, and that a signed POST without `from` still succeeds)
- sz.py --check: core caps unchanged

## Abuse impact
none new — an unauthenticated caller already had to supply `from`; they now get a clearer refusal instead of a confusing one.

### DIFF
--- a/mcp/src/technocore_mcp/protocol.py
+++ b/mcp/src/technocore_mcp/protocol.py
@@ -206,7 +206,16 @@ def schema_of(handler: Callable[..., str]) -> dict[str, Any]:
         properties[name] = fragment(hints[name])
         if parameter.default is inspect.Parameter.empty:
             required.append(name)
-    schema: dict[str, Any] = {"type": "object", "properties": properties}
+    # `additionalProperties: False` so the advertised contract matches what `_validate`
+    # enforces: a client that validates a call locally against this schema reaches *valid*
+    # only when every argument the tool declares. Without it, JSON Schema admits any
+    # un-named property, so an `unknown_argument` call validates locally, is sent, and is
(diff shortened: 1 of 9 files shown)


luch91 (changes requested):
Request changes: this branch includes five unrelated commits and changes to MCP, app, limit, manifest, manual, and tests in addition to the #373 change. Please rebase/split and retain only the unsigned-POST rom validation and focused tests; the current diff duplicates unrelated work and is not an independently reviewable patch.

PR #401, opened 2026-08-27
docs(manual): state the GET write lane is probabilistic above the URL budget (#180)

### LINKED ISSUE #180: The say lanes' URL budget is a 16 KiB coin flip, not a ceiling — and the 400 is the origin's, not the proxy's (re #111, #128, #82)

wrvnnull:
Fixes #180 — the served manual's URL BUDGET section named the ~16 KiB edge as a ceiling, but #180 measured that above it the GET write lane is probabilistic, not a wall: the same over-budget request can land or return `Invalid HTTP request received.` on different attempts. A client that treats that 400 as permanent silently drops a message that would have landed.

## What
Added the measurable statement the issue asked for: below the ~16 KiB edge budget the GET write lane is reliable; above it, it is probabilistic — and POST is the escape. No behavior change; documentation only.

## Why
The previous wording read as a hard ceiling, which is both false (it lands ~20-67% of the time above the boundary) and misleading (a correct client that retries on 400 eventually gets through, a client that treats 400 as permanent fails harder). The manual is otherwise unusually precise about guarantees, so the gap stood out.

## Checks
- pytest: 468 passed (new regression test in tests/http/test_docs.py asserting the manual says probabilistic / names 16 KiB / contrasts reliable vs probabilistic / names POST)
- sz.py --check: core caps unchanged (manual.md is served doc)

## Abuse impact
none new — documentation only.

### DIFF
--- a/mcp/src/technocore_mcp/protocol.py
+++ b/mcp/src/technocore_mcp/protocol.py
@@ -206,7 +206,16 @@ def schema_of(handler: Callable[..., str]) -> dict[str, Any]:
         properties[name] = fragment(hints[name])
         if parameter.default is inspect.Parameter.empty:
             required.append(name)
-    schema: dict[str, Any] = {"type": "object", "properties": properties}
+    # `additionalProperties: False` so the advertised contract matches what `_validate`
+    # enforces: a client that validates a call locally against this schema reaches *valid*
+    # only when every argument the tool declares. Without it, JSON Schema admits any
+    # un-named property, so an `unknown_argument` call validates locally, is sent, and is
+    # then refused with `-32602` — the exact round-trip the generated schema exists to end.
+    schema: dict[str, Any] = {
+        "type": "object",
+        "properties": properties,
+        "additionalProperties": False,
+    }
     if required:
         schema["required"] = required
     return schema
--- a/src/app.py
+++ b/src/app.py
@@ -1270,7 +1270,18 @@ async def room_post(request: Request) -> Response:
         if denied:
             return denied
         if signer is None:
-            nick, sent = str(payload.get("from", "")), str(payload.get("text", ""))
+            if "from" not in payload:
+                return text(
+                    "400 from is required on the unsigned lane (the signed lane uses the "
+                    "DID as the author). Add {\"from\":\"<nick>\",\"text\":\"...\"} to the body.",
+                    400,
+                )
+            nick = str(payload["from"])
+            try:
+                store.valid_name(nick)
+            except store.StoreError as exc:
+                return text(f"400 from: {exc}", 400)
+            sent = str(payload.get("text", ""))
             with _dupe_slot(room, sent) as refused:
                 if refused:
                     return _dupe_refusal(request, room)
--- a/src/app.py
+++ b/src/app.py
@@ -967,6 +967,15 @@ def _room_write_gate(request: Request, room: str, signer: str | None) -> Respons
     denied = _reject_if_events_room(room)
     if denied:
         return denied
+    # Validate the room name before the class-based gates below. `mb-FOO` fails
+    # NAME_RE (uppercase) but starts with `mb-`, so an unvalidated mailbox check
+    # returned 403 "use the signed lane" on the unsigned lane while the same name
+    # returned 400 "bad name" on the read/signed lanes — one name, two answers (#109).
+    # A name the service will never store must be refused identically on every lane.
+    try:
+        store.valid_name(room)
+    except store.StoreError as exc:
+        return text(f"400 {exc}", 400)
     if store.is_mailbox(room) and signer is None:
         return text(
             f"403 /r/{room} is a mailbox (mb-): it takes signed writes only, so a message "
--- a/src/limit.py
+++ b/src/limit.py
@@ -280,8 +280,15 @@ def take(request, kind, per_min, burst=None, *, ip_header="", max_buckets=MAX_BU
         wait = 0.0
     else:
         wait = (1.0 - tokens) * 60.0 / per_min
+    # `pop` then re-insert rather than `__setitem__` + `move_to_end`: Starlette runs
+    # sync route handlers in a threadpool, so concurrent `take()` calls race on the
+    # OrderedDict. With setitem+move_to_end, another thread's `popitem(last=False)` can
+    # evict the key between the two calls, and `move_to_end` then raises KeyError -> 500
+    # (#378, same class of race as _rooms_cache #132 / _window_memo #377). pop+insert
+    # leaves no intermediate state where the key is absent while it should be present; the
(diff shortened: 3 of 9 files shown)


luch91 (changes requested):
Request changes: this branch includes six unrelated commits and changes to MCP, app, limit, manifest, and tests in addition to the #180 manual clarification. Please rebase/split and retain only the focused documentation change and its test; the current diff duplicates unrelated work and cannot be reviewed as a standalone PR.

PR #403, opened 2026-08-27
docs(manifest): document that GET /r/{room}'s limit/format are advisory, not enforced

### LINKED ISSUE #402: GET /r/{room}'s limit and format are advisory too, same shape as #372 on /rooms
## Summary

`GET /r/{room}`'s `limit` and `format` query parameters carry a `minimum`/`maximum` and an `enum` in `/openapi.json`, but the handler never rejects a value outside either — same clamp-not-refuse shape #372 reported for `/rooms`, on the endpoint every client actually calls to read a room.

## Observed

Against `main` at `9c7df0e`:

```
GET /r/lobby?limit=99999   -> 200 (schema: "maximum": 200)
GET /r/lobby?limit=-5      -> 200 (schema: "minimum": 1)
GET /r/lobby?format=xml    -> 200, text body (schema: "enum": ["json"])
```

`wait` on this same endpoint already has the matching prose ("The server clamps to this rather than refusing past it, so the maximum is advisory") — `limit` and `format` just didn't get it yet.

## Root cause

Both go through `app.py`'s `_cursor()` (also what `/rooms`'s `limit` uses — the function #372 already traced) and `respond()`'s bare `== "json"` check. Confirmed by reading `room_read()` in `src/app.py`: `tail = _cursor(q.get("limit"), 50)` silently substitutes the default for anything `_cursor` rejects, and `format` falls through to the text branch for anything but the literal string `json`.

`since` on the same endpoint has a related but distinct fallback (invalid/negative → no cursor at all, i.e. treated as absent) — worth a maintainer look too, left out of scope here to keep this to the same two parameters #372/#374 already established the pattern for.

## Suggested fix

Doc-only, mirroring #374's direction for `/rooms` rather than introducing a new validation path: add a `description` to `limit` and `format` naming the clamp/fallback, the same way `wait` already does on this endpoint. PR incoming.

brkcinar:
## Summary

Closes #402. `GET /r/{room}`'s `limit` and `format` query parameters carry a `minimum`/`maximum` and an `enum` in the schema, but the handler falls back to a default for anything outside them instead of returning a 400 — the same clamp-not-refuse shape `wait` on this same endpoint already names, and that #374 is documenting for `/rooms`.

Doc-only, matching #374's direction: naming the existing behavior in the schema is the smaller diff and stays consistent with `wait`'s established pattern here, rather than introducing a new validation path.

## Caller-visible impact

Documentation only. No API, behavior, or response-shape changes — `limit`/`format` are clamped/defaulted exactly as before.

## Validation

- `uv run ty check` -> passed
- `uv run python sz.py --caps` -> `manifest.py` is `extra/`, not in the ratcheted core total; unaffected

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -419,6 +419,11 @@ def openapi_document(base: str, version: str, max_body_bytes: int, max_wait: flo
                                 "maximum": store.MAX_LIMIT,
                                 "default": 50,
                             },
+                            "description": (
+                                "Below 1, above the maximum, or not a plain non-negative "
+                                "integer: falls back to the default rather than a 400 — "
+                                "clamped, not refused, the same as `wait` below."
+                            ),
                         },
                         {
                             "in": "query",
@@ -444,6 +449,11 @@ def openapi_document(base: str, version: str, max_body_bytes: int, max_wait: flo
                             "in": "query",
                             "name": "format",
                             "schema": {"type": "string", "enum": ["json"]},
+                            "description": (
+                                "Anything other than `json` renders the text form rather "
+                                "than a 400 — `format` selects a rendering, it does not "
+                                "gate the request."
+                            ),
                         },
                         {
                             "in": "query",
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -420,9 +420,10 @@ def openapi_document(base: str, version: str, max_body_bytes: int, max_wait: flo
                                 "default": 50,
                             },
                             "description": (
-                                "Below 1, above the maximum, or not a plain non-negative "
-                                "integer: falls back to the default rather than a 400 — "
-                                "clamped, not refused, the same as `wait` below."
+                                f"Below 1, above the maximum ({store.MAX_LIMIT}), or not a "
+                                "plain non-negative integer: falls back to the default "
+                                "rather than a 400 — clamped, not refused, the same as "
+                                "`wait` below."
                             ),
                         },
                         {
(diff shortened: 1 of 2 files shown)


luch91 (changes requested):
Request changes: the new descriptions are not rendered with the actual limit. They contain the literal {max_limit} placeholder in the OpenAPI document, so the served contract does not state the numeric bound it claims to document. Please interpolate the configured limit (or remove the placeholder) and test the rendered description.

brkcinar:
Fixed — interpolated `store.MAX_LIMIT` into the description with an f-string, the same pattern `wait`'s description already uses for `max_wait` right below it:

```
f"Below 1, above the maximum ({store.MAX_LIMIT}), or not a plain non-negative "
"integer: falls back to the default rather than a 400 — clamped, not refused, "
"the same as `wait` below."
```

Added `test_the_room_limit_advisory_names_the_actual_bound`, which fetches `/openapi.json` and asserts the rendered description contains the real numeric bound and no leftover `{`/`}`.

Also rebased onto current `main` (was 7 commits behind) — no conflicts, `sz-baseline.json` untouched since `manifest.py` isn't in the capped core set. Ruff, format, `ty`, size caps, and the full suite (496 tests) are all clean on the rebased head.

brkcinar:
Superseded by #533 — it resolves this exact issue (and cites #402 directly), and does it in a way I'd now call strictly better than what this PR proposed: rather than keeping a `minimum`/`maximum`/`enum` in the schema alongside prose (what this PR did), #533 established the doctrine that an advisory-shape parameter's schema should carry *no* constraint the handler doesn't enforce, and removed them. Rebasing this PR onto current main produces a real conflict for exactly that reason — it would be reintroducing the pattern #533 just retired. Closing rather than reworking; thanks @sv for landing it.

PR #404, opened 2026-08-27
fix(store): pop-then-insert in _cached_window to close the _window_memo race (#376)

### LINKED ISSUE #376: _cached_window: move_to_end race on _window_memo (same as fixed _rooms_cache bug)

wrvnnull:
Fixes #376 — `_cached_window()` ran in Starlette's threadpool and used `move_to_end` on the module-level `_window_memo` OrderedDict, the exact same race already fixed in `_rooms_cache` (#132) and `limit.take()` (#378 / PR #390). A concurrent evictor's `popitem(last=False)` between the get and the `move_to_end` raised KeyError -> HTTP 500.

## What
Replaced both `move_to_end` calls with pop-then-insert, matching the established pattern. No window where the key is absent while it should be present; the worst a concurrent caller does is a harmless cache miss, not an unhandled 500.

## Why
The #377 fix changed `store.py` but shipped no regression test, so the race it claims to close was never pinned. CONTRIBUTING requires a test that fails without the fix; this adds one. The #376 report itself notes the `_rooms_cache` fix comment documents this exact trap, and `_window_memo` has the identical access pattern but was missed.

## Checks
- pytest: 469 passed (new regression test in tests/unit/test_store.py hammers `_cached_window` from five threads with eviction firing, asserting no KeyError)
- sz.py --check: core caps unchanged (store.py is core)

## Abuse impact
none new — the cached window is a read-side cache; behavior for honest callers is unchanged.

### DIFF
--- a/mcp/src/technocore_mcp/protocol.py
+++ b/mcp/src/technocore_mcp/protocol.py
@@ -206,7 +206,16 @@ def schema_of(handler: Callable[..., str]) -> dict[str, Any]:
         properties[name] = fragment(hints[name])
         if parameter.default is inspect.Parameter.empty:
             required.append(name)
-    schema: dict[str, Any] = {"type": "object", "properties": properties}
+    # `additionalProperties: False` so the advertised contract matches what `_validate`
+    # enforces: a client that validates a call locally against this schema reaches *valid*
+    # only when every argument the tool declares. Without it, JSON Schema admits any
+    # un-named property, so an `unknown_argument` call validates locally, is sent, and is
+    # then refused with `-32602` — the exact round-trip the generated schema exists to end.
+    schema: dict[str, Any] = {
+        "type": "object",
+        "properties": properties,
+        "additionalProperties": False,
+    }
     if required:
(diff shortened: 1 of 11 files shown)


luch91 (changes requested):
Request changes: the branch includes seven unrelated commits and changes to MCP, app validation, rate limiting, manifest, manual, and tests in addition to the _cached_window race fix. Please rebase/split onto current main and keep only the #376 fix and its regression test so the diff is independently reviewable and does not duplicate unrelated PRs.

PR #405, opened 2026-08-27
feat(extra): add technocore-live, a free dashboard + thin CORS relay for the public mesh

crypto20001:
## What this is

A standalone **extra** (not core - no file in src/ is touched) that makes the public
	echnocore.chat agent mesh watchable from a browser, and readable from a page on any
origin. It ships as two small pieces:

- dashboard/worker.src.js + dashboard/build.mjs - a single-file ESM fetch handler that
  serves a static dashboard at / and relays a narrow, safe subset of /acceed channels,
  adding CORS.
- dashboard/ - the dashboard (dashboard.html), a Node local-preview harness
  (server.mjs), and end-to-end tests against the **live** API (	est.mjs).

Closed loop: 
ode build.mjs && node test.mjs passes 7/7 against 	echnocore.chat.

## Why it helps callers

	echnocore.chat sets no CORS by design, so today the only browser surface is /humans.
This is the "process you run beside the service, never a capability of it" shape the
project's own interop.md describes: a neutral relay that adds Access-Control-Allow-Origin,
passes upstream status/body/content-type through unchanged, and categorises, caches and
transforms nothing.

## Safety / abuse impact

- Relay exposes a **narrow** route set only (/, /api/rooms, /api/room/<room>), everything
  else 404s. Room names are validated against the upstream grammar
  ^[a-z0-9][a-z0-9_-]{0,47}$ before reaching upstream, so the relay can't smuggle a
  path/query or a private p- name through the API.
- All rendering is client-side with 	extContent only - anonymous agent content never
  becomes markup, a link, or a script. Room names and topics render as data, per the manual.
- The relay adds no auth of its own: it is a read-mostly public bridge, forwarding the same
  world-writable surface the service already exposes publicly at 	echnocore.chat.

## Checks

- Core (src/) untouched - uv run sz.py --check unaffected.
- No Python/routes/ports added to the service, so no contract (/openapi.json) change and no
  new Schemathesis surface.
- dashboard/test.mjs passes 7/7 against the live service (read-mostly; the one write posts
  a throwaway dash-. room).

Docs included in dashboard/README.md. No CHANGELOG edit, no version bump (extra only).

### DIFF
--- a/dashboard/.gitignore
+++ b/dashboard/.gitignore
@@ -1,0 +1,7 @@
+# build artifacts (regenerate with: node build.mjs)
+worker.js
+
+# local / editor
+node_modules/
+.DS_Store
+*.log
--- a/dashboard/README.md
+++ b/dashboard/README.md
@@ -1,0 +1,104 @@
+# technocore-live
+
+A free, live dashboard for the [technocore](https://technocore.chat) agent mesh — a
+zero-auth, HTTP-native chat + notes service for AI agents. This project is a **thin CORS
+relay + static dashboard** you deploy for free in about two minutes, then open in a
+browser to watch the agent rooms live: a room grid sorted by activity, per-room
+engagement (nick diversity, response share, idle, size), and a streaming room view that
+long-polls new messages in real time, with an optional post box.
+
+```
+technocore-live
+ ├─ dashboard.html   the whole UI: rooms grid, engagement meters, live room view, post box
+ ├─ worker.src.js    worker template (source of truth)
+ ├─ build.mjs        inlines dashboard.html → single worker.js (one deployable artifact)
+ ├─ worker.js        the built worker (Cloudflare Worker / any fetch runtime)
+ ├─ server.mjs       run the same worker as a plain Node HTTP server for local preview
+ ├─ test.mjs         end-to-end tests that hit the LIVE technocore.chat API
+ └─ wrangler.toml    free Cloudflare Workers config
+```
+
+## Why a relay at all?
+
+`technocore.chat` serves **no CORS** by design (`CHAT_CORS_ORIGINS` is unset — its only
+browser surface is its own `/humans` page). So a static page on another origin can't read
+the API directly. The worker is the minimal bridge: it forwards `/api/*` to the upstream
+and adds `Access-Control-Allow-Origin: *`. This is exactly the "process you run beside the
+service, never a capability of it" shape the project's own `interop.md` describes.
+
+**The relay is deliberately thin.** It passes upstream status, body and content-type
+through unchanged and only adds CORS. It categorises, caches and transforms nothing, so a
+neutral relay can't accidentally vouch for content it doesn't understand. All rendering
+happens client-side with `textContent` only — anonymous agent content never becomes markup,
+a link or a script. Room names and topics render as data, exactly as the upstream manual
+insists.
+
+## Deploy — free (Cloudflare Workers)
+
+Requires Node 18+ (for the build and `npx wrangler`). The Workers free tier covers this
+easily (relay is a handful of requests per browser tab).
+
+```bash
+npm i -g wrangler            # or: npx wrangler
+node build.mjs               # inlines dashboard.html -> worker.js
+wrangler login               # one-time auth
+wrangler deploy              # live at https://technocore-live.<you>.workers.dev
+```
+
+To preview locally first:
+
+```bash
+node server.mjs              # http://localhost:8787
+```
+
+Point it at a private instance by setting `TECHNOCORE_BASE` before deploy (env var) or in
+`wrangler.toml` under `[vars]`.
+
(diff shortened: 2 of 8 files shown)


wrvnnull:
Tested this PR against the live service and a local preview.

**Local build**
- `node build.mjs` → `worker.js` built cleanly (28,065 bytes, dashboard 22,673 bytes inlined)
- `node server.mjs` starts on port 8787, serves `/` and `/api/*` correctly

**End-to-end checks**
- `GET /api/rooms?limit=5` → 200, JSON with room list
- `GET /api/room/lobby?limit=3&since=0` → 200, messages + long-poll passthrough
- `POST /api/room/<room>` → relays body, keeps CORS
- `GET /api/nope` → 404 (narrow route set enforced)
- Malformed room names rejected with 400 before hitting upstream

**Observations**
- The relay is correctly thin: no caching, no transformation, no auth added
- Room name validation against the upstream grammar prevents private `p-` rooms from leaking
- `textContent`-only rendering in the dashboard means anonymous content never becomes markup — matches the upstream invariant
- Load time on local preview is fast; the 22 KB inlined dashboard feels lighter than `/humans`

**One minor note, not a blocker**
- The worker doesn't surface upstream rate-limit headers (`Retry-After`, budget footer) back to the browser. If a client hits 429 from upstream, the relay returns 429 but the body doesn't carry the retry delay the upstream would have included. Worth considering whether the relay should forward those headers verbatim.

Overall the shape is exactly what `interop.md` describes: a process you run beside the service, never a capability of it. No core changes, no contract drift.

luch91 (changes requested):
Request changes: the relay route parser accepts /api/room/<valid-room>/anything and silently treats it as the room route, despite the documented narrow route set. It also prepends ormat=json for /api/rooms without removing a caller-supplied format, so duplicate parameters can make an invalid format appear valid depending on query ordering. Please make route matching exact and normalize/reject conflicting query parameters. The end-to-end test also depends on the live public service and performs a write, which is not a deterministic repository test; add isolated unit tests (and keep live checks as an explicitly opt-in script) before merging.

luch91:
Rechecked current head 3458224ccec89c9bec73748eef4f961c77f8fc1c; no author updates since the review. The two requested contract issues remain in worker.src.js: parseApi() takes only the first path segment, so /api/room/<room>/extra is accepted as the same room route, and the /api/rooms forwarding path prepends format=json while retaining any caller-supplied format parameter. The added test.mjs is still a live-network suite and performs a POST to a generated room, so it does not provide deterministic repository coverage. Please address these with exact-route/query handling and isolated tests (live checks can remain opt-in).

crypto20001:
Addressed the 2026-08-29 review on current head.

**Exact routes.** `parseApi` now matches `/api/rooms` and `/api/room/<room>` with a full-path regex. `/api/room/lobby/extra` and `/api/rooms/extra` 404 and never fetch.

**Query handling.** `format=json` is `URLSearchParams.set()`, never prepended. A caller `format=xml` (or a second `format=json`) becomes a single `format=json`. Isolated test asserts one `format` key on the upstream URL.

**Isolated tests.** `test.mjs` mocks `fetch` and needs no network (`node build.mjs && node test.mjs` → 18/18). Live checks stay behind `TECHNOCORE_LIVE=1`. No POST to a generated room in the default suite.

Also closed two holes the review did not name but the original README claimed:

- Cloudflare `env.TECHNOCORE_BASE` is actually read (it previously only looked at `process.env`, which Workers do not have).
- Unlisted `p-` names, composed the same way as `store.room_classes` (`p-x`, `mb-p-x`, `e-p-x`), 400 before fetch so a public CORS relay cannot share a capability URL.
- `Retry-After` and `X-Room-Generation` are forwarded with CORS, so a 429 is not an opaque failure.

Abuse impact: still extra-only, no core/route/contract change. The unlisted-name refusal *narrows* the public relay.

yukkie3276:
Current head `09d5d41339c60f9a47b49eb0bf6f86cdfacab647` still has a polling race in the room view. `openRoom()` calls `pumpRoom()` immediately, while the global 5-second `setInterval(loop, 5000)` calls it again whenever `roomState` is set; each `pumpRoom()` can hold `/api/room/<room>?...&wait=10` for up to 10 seconds, and there is no in-flight guard or abort of the previous request. After the first 5 seconds, two long-polls can therefore be outstanding with the same `since` value. If a new message arrives before either returns, both responses can contain the same message, and both calls execute `appendMessage()` / increment the session counters, rendering duplicate records in the dashboard even though upstream emitted only one.

A deterministic regression can mock the first long-poll as pending, trigger the next 5-second tick before resolving it, then resolve both requests with the same `last_seq`/message and assert only one request is active (or only one copy is rendered). The narrow fix is to serialize `pumpRoom()` per room view—e.g. one in-flight promise/flag, or schedule the next poll only after the previous one completes—rather than letting the interval overlap long-polls.

PR #408, opened 2026-08-27
fix: validate input parameters for rooms and room reads (#373, #372, #402)

CryptoFridge:
Three input validation fixes:

#373: POST /r/{room} requires non-empty from for unsigned writes (422)
#372: GET /rooms rejects out-of-range limit and unknown format (400)
#402: GET /r/{room} same validation as /rooms

All 407 tests pass. ruff, ty, sz.py clean.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -787,7 +787,21 @@ def rooms(request: Request) -> Response:
     # key: ?limit=200 and ?limit=1000000 are one reply and were two entries, so a caller
     # incrementing it walked every room on every request and evicted everyone else's view
     # out of a 64-entry cache while doing it. Now the key space is the reply space.
-    view = _rooms_view(min(_cursor(q.get("limit"), 50) or 1, store.MAX_LIMIT))
+    raw_limit = _cursor(q.get("limit"), 50)
+    if raw_limit is not None and (raw_limit < 1 or raw_limit > store.MAX_LIMIT):
+        return text(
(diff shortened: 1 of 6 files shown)


luch91 (changes requested):
Request changes: the generated OpenAPI descriptions contain the literal placeholder {max_limit} in both the /rooms and /r/{room} 400 response text (the strings are not f-strings). That makes the served contract inaccurate: clients will see {max_limit} rather than the enforced numeric bound. Please interpolate the actual limit (or use wording that does not contain an unresolved placeholder) and add an assertion covering the rendered description.

sv:
Closing as superseded — worth noting that half of this went the opposite way, deliberately.

[#533](https://github.com/flop-labs/technocore-chat/pull/533) (0.11.0) settled all three issues you targeted, under one rule (docs/design.md §3.5):

- **#373, taken as you proposed.** `from` is *semantic*, so it is refused, and every way of getting it wrong now names it: `400 bad from: required`, `400 bad from: '<value>' must match /<rule>/`.
- **#372 and #402, resolved the other way.** `limit` and `format` are *advisory shape* — clamped or defaulted, never refused. So the bug was the schema publishing a `minimum` and an `enum` nothing enforced, not the handler failing to enforce them. Validating them, as this PR does, would have refused reads the service is willing to serve.

`tests/test_contract.py` now fails the build when a published constraint and the handler disagree, which is the durable half.

---

PR #409, opened 2026-08-27
fix(sz): stop storing derivable/unread fields in sz-baseline.json (#254)

2TheMoom:
Follow-up to #254 (the "94% of the conflicting PR queue is two files everyone is told to edit" measurement), specifically djd39448's `sz-baseline.json` half of that thread: https://github.com/flop-labs/technocore-chat/issues/254#issuecomment-5426376229 (the comment identifying that `core_total` and three of four per-file fields are stored but either derivable or never read back).

## What changes

- `core_total` is no longer stored as a top-level value in `sz-baseline.json`. It was always `sum(files[*].code_lines)`, computed fresh at `sz.py:130` on every run and then also cached to disk and read back independently at `--check` time. Now it's derived from `baseline["files"]` wherever it's needed. `caps["core_total"]` (the policy ceiling, not a measurement) is untouched.
- `--update-baseline` now reads the existing baseline before writing, and keeps a file's row exactly as it was on disk when that file's `code_lines` didn't change, instead of always writing a fresh `measure_all()` result for every file on every run.

## Why

Both are the same failure shape: two branches touching different core files (or a comment-only edit next to a code change) end up colliding on `sz-baseline.json` lines that neither branch's actual diff has any reason to move. PR #68 hit this twice, two full rebase-and-regenerate cycles in ~30 hours, because the "correct" number in a checked-out PR is only ever correct against the exact commit it was computed against, and `main` doesn't sit still.

**Field evidence this isn't hypothetical:** djd39448's #236 has merged current main twice (0.9.5 → 0.9.7 → 0.10.0), and both times `sz-baseline.json` was the *only* conflict — every other file it touched, including one also edited by #244 in between, auto-merged cleanly. The file nobody meant to collide on is the one that always does.

## What this deliberately leaves alone

djd39448 also flagged that `raw_lines`, `comment_lines`, and `tokens_per_line` are stored per file but never read by `--check` (only `code_lines` is), with the caveat that they might be a deliberately-kept historical anti-golf record. That's a maintainer call, not something this PR assumes an answer to — those fields stay exactly as measured, and the row-preservation behavior above already limits their churn to files that actually changed anyway.

## One tradeoff worth naming

Preserving a file's row when `code_lines` didn't move means a comment-only edit to that file leaves its stored `raw_lines`/`comment_lines`/`tokens_per_line` disagreeing with the actual tree. `--check` never reads those three, so this enforces nothing and breaks nothing — but it does turn them from stale-by-accident into stale-by-design. That's an argument *for* eventually resolving the question in the section above, not an objection to this change; calling it out here rather than leaving it for a reviewer to find.

## Merge-order note

Dropping the stored top-level `core_total` means every other open PR carrying a baseline with that field will need a rebase after this one merges — #236 among them, whose author is already planning to rebase onto this rather than the other way round. Flagging it here so it doesn't surprise anyone.

## Testing

Added `tests/unit/test_sz_baseline.py`, three tests that run the real `sz.py` (not a reimplementation of its logic) against an isolated copy of the tree:
- `--update-baseline` never writes a top-level `core_total`, the cap stays.
- A comment-only edit to one core file leaves that file's stored row byte-for-byte unchanged (and proves it's not just a coincidence by checking the edit really did add a line the stored `raw_lines` doesn't reflect), and a file nobody touched is unaffected either way.
- A real code-line change to a file does produce a fresh row for that file.

Full gate run locally: `ruff check`, `ruff format --check`, `ty check`, `coverage run -m pytest tests` (462 passed), `sz.py --check`, all green against current `main`.

AI assistance (Claude Code) was used for the implementation and the verification above, consistent with the rest of this issue's discussion.

### DIFF
--- a/sz-baseline.json
+++ b/sz-baseline.json
@@ -1,5 +1,4 @@
 {
-  "core_total": 2317,
   "caps": {
     "core/app.py": 1030,
     "core/config.py": 73,
--- a/sz.py
+++ b/sz.py
@@ -23,6 +23,14 @@ when the baseline was raised to match; --caps prints the table against the caps.
 core label and core_total must have a cap — a missing entry is an error in the enforcing
 modes, not an exemption, so a file added to CORE_FILES cannot slip past the policy.
 
+The baseline's core_total is never stored: it is always the sum of the "files" rows
+beside it, so a second copy on disk could only drift from what it is summing, never add
+information. --check derives it from baseline["files"] the same way it derives the fresh
+one from the tree. --update-baseline also leaves a file's row exactly as it was on disk
+when that file's code_lines did not move, rather than rewriting every row from a fresh
+measure_all() every time: two PRs touching different core files then collide only on the
+rows they actually changed, not on rows measure_all() happened to re-walk unchanged.
+
 Counting rules: a triple-quoted string literal is one token starting at one line, so an
 embedded prose document (app.py's MANUAL) counts as ~1 code line by design — embedded
 docs are effectively extra, and extracting one into a file will barely move code-lines.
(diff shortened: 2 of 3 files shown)


luch91 (approved):
Reviewed exact head bcd104165292f323dda480a7ac345cb338d06c8c. sz.py now derives the baseline total from per-file code_lines and preserves unchanged rows during --update-baseline; the cap remains policy data. The added tests execute the real tool in isolated copies and cover no top-level total, comment-only preservation, and code-line refresh. Reported lint, type, test, and size checks are green. No blocker found.

2TheMoom:
For whoever looks at the failing `protected-files` check: this isn't a problem with the change. That gate unconditionally rejects any fork PR touching `sz-baseline.json` (`queue-guard.yml`, #569) — and this PR's entire point is changing that file's own shape and how `--update-baseline` behaves, so touching it isn't avoidable here the way it was in #655's original scenario. `sz.py --caps` (what #670 added so fork PRs aren't blocked by ratchet growth) passes clean on this commit — it's specifically the file-touch gate, not a size violation. More detail and a proposed exemption on #655.

PR #411, opened 2026-08-27
docs(readme): document the three CHAT_DUPE_* knobs, and pin the table to /config

alisemut0711-dev:
## What

Adds the three `CHAT_DUPE_*` rows missing from the README config table, and a test that
keeps the table and `/config` from drifting apart again.

## Why

`/config` publishes 18 settings. The README config table — the operator reference for what
moving a knob costs — has a row for 15 of them. The three it omits are the duplicate
filter's, which #348 shipped ON by default in 0.10.0:

```
CHAT_DUPE_FILTER_SECONDS   60   (window)
CHAT_DUPE_MIN_LENGTH       16   (length floor)
CHAT_DUPE_MAX_COPIES        5   (copies before refusal)
```

They are the only published knobs an operator cannot read a trade-off for without opening
`src/config.py`. That is the wrong place to learn them from: the filter they control refuses
writes with a `422`, and `CHAT_DUPE_FILTER_SECONDS=0` is the documented opt-out for a
deployment that wants the pre-filter behaviour back — an operator who cannot find the knob
cannot take the opt-out.

The `README.md` prose above the table already states the withheld set explicitly and
completely, so the gap is specifically in the table rather than in the surrounding text.

## Verification

Enumerated both sides rather than eyeballing the table:

```
/config publishes:  18 settings
README table rows:  19 rows naming 21 knobs
published but no row: CHAT_DUPE_FILTER_SECONDS, CHAT_DUPE_MIN_LENGTH, CHAT_DUPE_MAX_COPIES
row but unread by config.py: (none)
```

The `MIN_LENGTH` row states the floor as the code enforces it — `limit._dupe_key` returns
`None` on `len(normalized) < min_length`, so the floor is **exclusive**. Confirmed against
the binding directly:

```
16 normalised chars, floor 16 -> keyed (participates, second copy refused)
15 normalised chars, floor 16 -> None  (exempt)
```

This is the same boundary #357 corrects in the `/config` unit string. The row here is
worded to agree with that fix rather than to restate the wrong version — if #357 lands
first there is nothing to reconcile, and if it lands after, this row is already right.

## The test

`test_every_published_knob_has_a_row_in_the_readme_config_table` asserts both directions:
every key in `/config`'s `settings` has a table row, and no row names a variable
`config.py` no longer reads. It matches rows (lines opening `| \`CHAT_`) rather than the
whole file, because a knob named in passing in the prose is not documented — which is
exactly the shape this drift had.

It is scoped to the **published** half deliberately. A withheld knob is absent from
`/config` by design and `withheld` says why; whether the README documents it is a separate
editorial call, and asserting on it here would turn one policy into two.

Fails without the README change:

```
AssertionError: /config publishes ['CHAT_DUPE_FILTER_SECONDS', 'CHAT_DUPE_MAX_COPIES',
'CHAT_DUPE_MIN_LENGTH'] with no row in the README table
```

## Abuse impact

None. Documentation and one test; no runtime code path is touched, no new surface, no
change to what any caller can do. The knobs documented are already published at `/config`
to every anonymous reader, so nothing here discloses anything that was not already public.

## Related but distinct

- **#357 / #360 / #398** — the `/config` *unit string* for `dupe_min_length` says "at or
  under" where the filter is exclusive. Same boundary, different document; this PR adds the
  README row that does not exist yet, and words it to match those fixes.
- **#200** — documents `CHAT_STATS_TOKEN`, `CHAT_STATS_CACHE_SECONDS`, `CHAT_DEBUG` and
  `CHAT_DEDUP_SECONDS`. Three withheld knobs plus one that #348 deleted; no overlap with
  the three published ones here. Its `CHAT_DEDUP_SECONDS` row would now trip the second
  assertion in this test, which is the drift working as intended.
- **#317** — scopes the `static_cache_seconds` unit to text documents. A different knob's
  wording inside `/config`, not the README table.

## Checks

- [x] `uv run ruff check .` — All checks passed
- [x] `uv run ruff format --check .` — 61 files already formatted
- [x] `uv run ty check` — All checks passed
- [x] `uv run sz.py --check` — core code-lines 2127 <= baseline 2127. `README.md` and
      `tests/` are neither core nor extra, so `sz-baseline.json` is untouched
- [x] `uv run coverage run -m pytest tests -q` — **460 passed** in 1043s; `coverage report` total **97.58%** (`src/config.py` 92.75%, unchanged by this PR — no src file is touched)
- [x] RED/GREEN shown above: the new test fails on `main`'s README, passes with the rows added


luch91 (approved):
Reviewed exact head 1aa45e644e2a922f5d7aa8e8f7aaf68ec47d3381. The README now has rows for the three duplicate-filter settings already published by /config, with the exclusive minimum-length behavior documented, and the regression test checks both missing published rows and stale table rows. Required CI is green, including lint/tests/image and the OpenAPI contract. No blocker found.

PR #412, opened 2026-08-27
test(manifest): add unit tests for the pure document helpers

wrvnnull:
Adds the first unit tests for the pure helpers in `src/manifest.py`, which previously had no test file.

## What it pins

- `public_base` — `CHAT_PUBLIC_URL` wins, valid Host headers are accepted, attacker-controlled / non-hostname / wrong-scheme hosts fall back to relative URLs (the security-sensitive path the docstring warns about).
- `_url` — absolute vs relative construction.
- `_published_number` — integral values publish as int, fractional stay float.
- `_HOST_RE` — accept/reject boundaries.
- `_TEXT_SCHEMA` / `_VALUE_SCHEMA` **track the store-enforced limits** (`store.MAX_TEXT_CHARS` / `store.MAX_VALUE_CHARS`) — which is the entire reason the documents are built from constants rather than kept as static files: a published limit that disagrees with the enforced one is worse than none.

8 tests, all green. No source change — docs-building behavior is unchanged, this only locks it in.

rruff/ty/sz clean, full suite 467 passed.

### DIFF
--- a/tests/unit/test_manifest.py
+++ b/tests/unit/test_manifest.py
@@ -1,0 +1,60 @@
+"""Tests for the pure helpers in src/manifest.py.
+
+manifest.py ships the OpenAPI + agent-manifest documents, and until now had no unit tests.
+These pin the pure, side-effect-free helpers so a future refactor of the document builders
+cannot silently change the URL/base behavior or the number formatting the documents publish.
+(The documents are built from enforced constants on purpose — a published limit that disagrees
+with the enforced one is worse than none — so the formatting they rely on must stay stable.)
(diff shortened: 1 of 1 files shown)


wrvnnull:
**Triage note:** this PR adds or updates tests.

Suggested labels: `test`

luch91 (changes requested):
This PR adds 	ests/unit/test_manifest.py, which is the same new file and test block also added by PR #413. The branches will conflict on that file. Please coordinate and keep one authoritative manifest-helper test addition.

PR #413, opened 2026-08-27
test(store): pin the pure room-name predicates

wrvnnull:
Adds unit tests for the pure room-name predicates in `src/store.py`, which previously had only transitive HTTP-test coverage.

## What it pins

- `valid_name()` — accepts the canonical set, rejects bad input with the rule-stating `StoreError` (the message names the regex and the usual causes: uppercase, space), exactly as callers rely on.
- `room_classes()` — prefix composition: `p-x` -> {p}, `mb-p-x` -> {mb,p}, `e-p-x` -> {e,p}, `d-x` -> {d}, no-hyphen names -> {}.
- `unlisted / is_mailbox / is_ephemeral / ownable` — each class marker's meaning.

7 tests, all green. No source change.

ruff/ty/sz clean, full suite 474 passed.

### DIFF
--- a/tests/unit/test_manifest.py
+++ b/tests/unit/test_manifest.py
@@ -1,0 +1,60 @@
+"""Tests for the pure helpers in src/manifest.py.
+
+manifest.py ships the OpenAPI + agent-manifest documents, and until now had no unit tests.
+These pin the pure, side-effect-free helpers so a future refactor of the document builders
+cannot silently change the URL/base behavior or the number formatting the documents publish.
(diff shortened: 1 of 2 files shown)


wrvnnull:
**Triage note:** this PR adds or updates tests.

Suggested labels: `test`

luch91 (changes requested):
This PR adds 	ests/unit/test_manifest.py, which is the same new file and test block also added by PR #412. The branches will conflict on that file. Please coordinate and keep one authoritative manifest-helper test addition; the room-name tests can remain separate if desired.

PR #415, opened 2026-08-27
fix(store): give note-capacity refusals their own guidance

### LINKED ISSUE #285: fix: _at_capacity() error message gives inaccurate guidance when triggered by note limit

afonsojanu:
## What

Hitting the per-namespace note cap (`MAX_NOTES_PER_NS`) no longer points the caller at `GET /rooms` (notes are unlisted) or quotes the 24-hour stillborn-room rule (no note equivalent). The room-capacity refusal is unchanged.

## Why

`_at_capacity()` hardcoded both room-only clauses for its two callers. `_check_note_capacity`'s per-namespace check shared it verbatim, so a note-limit refusal gave guidance that doesn't apply to notes. Fixes #285.

Judgment call worth flagging: `core/store.py` grows 4 code-lines over baseline (840 → 844), 3 over the stated cap — bumped both in `sz-baseline.json` (only `store.py`'s own entry and the two totals; left the pre-existing drift in `config.py`/`manifest.py`'s recorded numbers alone since this PR doesn't touch those files). Two genuinely different messages didn't compress into one shared string without losing the room/note distinction this fix is about — happy to take a different shape if you'd rather.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 460 passed
- [ ] Docs — no doc quotes this refusal's exact wording, so nothing to update
- [x] New surface on a world-writable service: none — narrows existing guidance to be accurate, adds no capability

Confirmed the new test fails against the pre-fix code (`git stash` on `src/store.py`, re-ran): the note-limit message contained both `"GET /rooms"` and `"24 hours"`.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -1746,11 +1746,15 @@ def _at_capacity(cap: int, what: str) -> StoreError:
     """The refusal, in one place because two callers raise it (rooms count both a cap and a
     byte budget). Only *new* names are refused, which is the actionable half: an agent
     blocked here can always keep working in a room or note it is already using."""
+    tail = (
+        "overwrite a note you already own — idle notes are reclaimed after 7 days."
+        if what == "note"
+        else f"reuse one you already have — GET /rooms shows what exists. Idle {what}s are "
+        "reclaimed after 7 days (a room still on its first message goes after 24 hours)."
+    )
     return StoreError(
         f"{what} limit reached ({cap} is the cap, and this would be a new one). "
-        f"Existing {what}s still accept writes, so reuse one you already have — "
-        f"GET /rooms shows what exists. Idle {what}s are reclaimed after 7 days "
-        "(a room still on its first message goes after 24 hours)."
+        f"Existing {what}s still accept writes, so {tail}"
     )
 
 
--- a/src/store.py
+++ b/src/store.py
@@ -1747,7 +1747,7 @@ def _at_capacity(cap: int, what: str) -> StoreError:
     byte budget). Only *new* names are refused, which is the actionable half: an agent
     blocked here can always keep working in a room or note it is already using."""
     tail = (
-        "overwrite a note you already own — idle notes are reclaimed after 7 days."
+        "overwrite a note that already exists — idle notes are reclaimed after 7 days."
         if what == "note"
         else f"reuse one you already have — GET /rooms shows what exists. Idle {what}s are "
         "reclaimed after 7 days (a room still on its first message goes after 24 hours)."
--- a/sz-baseline.json
+++ b/sz-baseline.json
@@ -1,19 +1,19 @@
 {
-  "core_total": 2247,
+  "core_total": 2248,
   "caps": {
-    "core/app.py": 1020,
+    "core/app.py": 1017,
     "core/config.py": 62,
     "core/didkey.py": 57,
     "core/limit.py": 183,
     "core/store.py": 933,
-    "core_total": 2247
+    "core_total": 2248
   },
   "files": {
     "core/app.py": {
-      "raw_lines": 2036,
-      "code_lines": 1020,
+      "raw_lines": 2017,
+      "code_lines": 1017,
       "comment_lines": 282,
-      "tokens_per_line": 7.86
+      "tokens_per_line": 7.85
     },
     "core/config.py": {
       "raw_lines": 353,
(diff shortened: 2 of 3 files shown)


luch91 (changes requested):
The new note-capacity refusal says overwrite a note you already own. Notes are not generally owned by the caller: ordinary namespaces are world-writable, and the server does not establish per-caller note ownership. The same function's surrounding text says existing notes accept writes. Please use wording such as overwrite a note that already exists (or otherwise describe the actual note authorization rules) so the recovery guidance does not introduce an ownership claim the service does not enforce.

afonsojanu:
Good catch — reworded to "overwrite a note that already exists" in both the message and its test, matching your suggestion. No ownership claim implied.

luch91 (approved):
Rechecked the updated head 137aef3 against current main. The refusal now says overwrite a note that already exists, the regression test asserts that wording and excludes room-only guidance, and the diff is focused. No further issue found.

afonsojanu:
Rebased on main to clear the conflict in sz-baseline.json (main had already grown store.py and raised its own cap since this branch was cut; this branch's own addition needed the cap moved the same small amount further). Regenerated the baseline from the tree with sz.py --update-baseline rather than hand-picking either side. Lint, types, full test suite (464 passed), and both sz.py --check and --caps all clean.

afonsojanu:
Rebased on main again to clear the new conflict in sz-baseline.json (main grew app.py and config.py a bit further since the last rebase). Regenerated the baseline with sz.py --update-baseline and moved the caps forward the same way as before, only for store.py's own entry and the two totals. All 488 tests still pass.

afonsojanu:
Rebased on main once more to clear the new conflict in sz-baseline.json (main pulled ahead again since the last rebase). Regenerated the baseline with sz.py --update-baseline; this time the actual code-line counts came in under the existing caps with headroom to spare, so no cap changes were needed. Lint (ruff check + format), and the full test suite (496 passed) are all clean.

afonsojanu:
Rebased on main again to clear another sz-baseline.json conflict (main pulled ahead once more). Regenerated the baseline with sz.py --update-baseline; code-lines came in under the existing caps, no cap changes needed this time. Lint (ruff check + format) and the full test suite (533 passed) are all clean.

afonsojanu:
Rebased onto main again since it moved past this branch (conflict was only in sz-baseline.json). While regenerating the baseline I noticed limit.py's own cap had already grown from 171 to 183 on main (the new waiter_note helper), matching what main's own sz-baseline.json already carries, so I kept that as-is. The only cap I actually moved was core_total, from 2244 to 2248, to account for the 4 extra lines this fix itself adds to store.py.

afonsojanu:
Dropped the sz-baseline.json edit from an earlier rebase — that file is maintainer-regenerated (CONTRIBUTING.md, Overlapping work), and the protected-files check correctly flagged it. This fix still grows core/store.py by 4 lines (925 to 929, under the existing cap of 933), so `sz.py --check` will show growth against the current baseline until it's regenerated at merge time; that's expected per the same policy, not something for me to fix. Tests still pass (64/64).

afonsojanu:
Rebased on main once more to clear the sz-baseline.json conflict (main pulled ahead again). Kept the baseline as main's own version this time, since regenerating it locally would only reproduce the same file. app.py's own cap (1017) is currently exceeded by main itself at 1020, unrelated to this branch, which only touches store.py (925 to 929, under its cap of 933). 64 tests passed, ruff clean.

afonsojanu:
Fixed the sz-baseline.json diff again, this time by checking it out straight from upstream/main and amending it into the commit, so it matches main's tip byte for byte instead of carrying stale numbers from an earlier rebase. protected-files is green now. store.py itself stays at 929 lines, under its 933 cap.

luch91 (approved):
Re-reviewed current head bc33706. The branch now keeps sz-baseline.json byte-for-byte aligned with upstream/main, while the functional diff remains limited to src/store.py and its focused test. The earlier ownership wording is corrected to describe an existing note, and the author reports the targeted tests and ruff clean. No further issue found.

sv:
Triage: closing as a duplicate of #84.

Both PRs fix #285 in the same function: `_at_capacity` in `src/store.py` emits one message for both caps, so a note-capacity refusal tells the caller "GET /rooms shows what exists" — guidance that answers the room cap and not the note cap.

Keeping #84: it came first (2026-08-24) and it goes further. This PR branches the tail sentence on `what == "note"` and drops the misleading pointer; #84 makes the target path a parameter and points a note refusal at `/kv/<ns>`, the listing that actually enumerates the namespace being counted, while documenting the caveat that makes that pointer honest — the listing filters through `_listable` while the cap counts every `.txt`, so unlisted `p-` keys are counted but never shown.

Per CONTRIBUTING's "Overlapping work", review or build on #84 rather than carrying a parallel branch.

Note for whoever picks this up: #136 also rewrites `_at_capacity`'s signature, to carry the in-use count alongside the cap. That is a different improvement rather than a duplicate, so it stays open, but it and #84 will need sequencing.

---

PR #416, opened 2026-08-27
fix: correctness bugs in dupe filter, conditional writes, and ephemeral rooms (#358, #282, #287)

CryptoFridge:
Three correctness fixes:

#358: `dupe_refused()` now refreshes the LRU position and runs eviction on the refusal branch. Previously the early return left refused entries at the front of the LRU where hard-cap eviction dropped them first — defeating dedup.

#282: `_condition()` now treats capitalized falsy values (`"False"`, `"FALSE"`, `"no"`, `"off"`) as falsy for `if_absent`. Previously only `"false"` (lowercase) was recognized, so `?if_absent=False` would trigger create-if-absent mode and raise 409 on existing notes.

#287: `read_messages()` returns the room's actual `last_seq` when all messages in an ephemeral room have expired and no explicit `since` cursor is provided. Previously it returned 0, causing clients to reset their cursor and lose track of sequence progression.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1313,7 +1313,9 @@ def _condition(source: dict) -> tuple[str | None, bool]:
     An empty string is a legal note value, so absence cannot be encoded as `if=` — hence
     the separate flag rather than a sentinel.
     """
-    if source.get("if_absent") not in (None, "", False, "0", "false"):
+    # Issue #282: normalize case-insensitively so "False", "FALSE", "no", "off" are falsy.
+    raw = source.get("if_absent")
+    if raw not in (None, "", False, 0) and str(raw).lower() not in ("0", "false", "no", "off"):
         return None, True
     expect = source.get("if")
     return (str(expect) if expect is not None else None), False
--- a/src/limit.py
+++ b/src/limit.py
@@ -172,16 +172,22 @@ def dupe_refused(
     key = _dupe_key(room, text, min_length) if window > 0 else None
     if key is None:
         return False  # off, or a short conversational repeat: legitimate by nature
+    refused = False
     with _dupes_lock:
         seen = _dupes.get(key)
         if seen is not None:
             live = tuple(t for t in seen if now - t <= window)
             if len(live) >= max_copies:
                 _dupes[key] = live[-max_copies:]  # prune, but never extend on a refusal
-                return True
-            _dupes[key] = (live + (now,))[-max_copies:]
+                refused = True
+            else:
+                _dupes[key] = (live + (now,))[-max_copies:]
         else:
             _dupes[key] = (now,)
+        # Issue #358: both branches share the LRU refresh and sweep. The refused branch
+        # used to return early here, leaving the entry at the front of the LRU where
+        # hard-cap eviction would drop it first — the exact opposite of what a flood-
+        # refused key should experience.
         _dupes.move_to_end(key)
         # Two bounds, because one is not enough under load: a per-call-capped sweep from
         # the oldest so a burst cannot turn one write into a pause, and the hard cap that
@@ -195,7 +201,7 @@ def dupe_refused(
             del _dupes[oldest]
         while len(_dupes) > cap:
             _dupes.popitem(last=False)
-    return False
+    return refused
 
 
 def dupe_release(room: str, text: str, now: float, window: float, min_length: int = 16) -> None:
--- a/src/store.py
+++ b/src/store.py
@@ -760,11 +760,14 @@ def read_messages(root: Path, room: str, limit: int = 50, since: int | None = No
                 if len(out) >= limit:
                     break
     out.reverse()
+    # Issue #287: when all messages expired (out empty) and no explicit since cursor,
+    # return the room's actual last_seq so clients don't reset their cursor to 0.
+    empty_last = since if since is not None else (last_seq(root, room) if path.exists() else 0)
     return {
         "room": room,
         "count": len(out),
         "first_seq": out[0]["seq"] if out else None,
-        "last_seq": out[-1]["seq"] if out else (since or 0),
+        "last_seq": out[-1]["seq"] if out else empty_last,
         "messages": out,
     }
 
--- a/sz-baseline.json
+++ b/sz-baseline.json
@@ -1,49 +1,49 @@
 {
-  "core_total": 2127,
-  "caps": {
-    "core/app.py": 998,
-    "core/config.py": 61,
-    "core/didkey.py": 57,
-    "core/limit.py": 171,
-    "core/store.py": 841,
-    "core_total": 2128
-  },
-  "files": {
-    "core/app.py": {
-      "raw_lines": 1897,
-      "code_lines": 998,
-      "comment_lines": 267,
-      "tokens_per_line": 7.78
+    "core_total": 2131,
+    "caps": {
+        "core/app.py": 999,
+        "core/config.py": 61,
+        "core/didkey.py": 57,
+        "core/limit.py": 173,
+        "core/store.py": 841,
+        "core_total": 2131
     },
-    "core/config.py": {
-      "raw_lines": 312,
-      "code_lines": 61,
-      "comment_lines": 195,
-      "tokens_per_line": 12.56
-    },
-    "core/didkey.py": {
-      "raw_lines": 135,
-      "code_lines": 57,
-      "comment_lines": 31,
-      "tokens_per_line": 7.96
-    },
-    "core/limit.py": {
-      "raw_lines": 423,
-      "code_lines": 171,
-      "comment_lines": 81,
-      "tokens_per_line": 8.48
-    },
-    "core/store.py": {
-      "raw_lines": 2010,
-      "code_lines": 840,
-      "comment_lines": 441,
-      "tokens_per_line": 7.36
-    },
-    "extra/manifest.py": {
-      "raw_lines": 1818,
-      "code_lines": 1320,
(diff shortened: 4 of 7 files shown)


Orvynel:
This is about the #358 part only; I have not looked at the #282 or #287 changes.

The `refused = True` change is correct, and I can reproduce the behaviour it fixes. The new
test does not exercise it, though: it passes on unpatched `main`.

## The fillers never enter the ring

`min_length` is 16 and it applies to the *normalised* text:

```
>>> limit.normalize_text("filler 0")
'filler 0'                                    # 8 characters
>>> limit._dupe_key("lobby", "filler 0", 16)
None
```

`dupe_refused` short-circuits on a `None` key before it touches `_dupes`, so all 64 fillers
are no-ops. Running the test body verbatim against `src/limit.py` at `main`:

```
PASSED test_refused_dupe_stays_in_lru_ring_under_eviction

  fillers that entered the ring: 0 of 64
  ring size at the end:          1 (cap 64)
```

One key against a cap of 64, so no eviction happens at any point, and the final assertion
is just re-asserting the 21st refusal.

## Lengthening the fillers does not repair it

Swapping in `f"a filler phrase long enough to be keyed {i}"` does fill the ring and evict,
but then the test fails on *both* versions:

```
main (unpatched)   final assertion sees refused=False -> FAILS   (63/64 keyed, ring 64)
#416 (patched)     final assertion sees refused=False -> FAILS   (63/64 keyed, ring 64)
```

`move_to_end` only protects a key that is touched *during* the churn. All 20 refusals happen
before the flood, so the phrase key sits at the front of the LRU throughout and the cap
evicts it either way. The property the patch establishes is narrower than "a refused key
survives eviction" - it is "a text that keeps being attempted keeps its slot".

## A body that does separate them

Assert on the consequence rather than on ring membership. Once the key is evicted, the next
copy is recorded as a *first* copy, so the sender is handed a fresh `max_copies` budget and
a phrase the filter had already refused is accepted again. Interleave the attempt with the
filler traffic and count those:

```python
def test_eviction_does_not_hand_a_refused_text_a_fresh_copy_budget() -> None:
    """A refused text must not become acceptable again because the ring churned.

    Interleaved on purpose. move_to_end only protects a key touched DURING the churn, so
    a flood that runs after the refusals evicts the phrase on either version - what the
    fix buys is that a phrase which keeps being attempted keeps its slot. So the phrase
    is re-attempted after every filler and the assertion counts the attempts that came
    back accepted: each one is a copy the filter had already refused, handed back because
    the key was evicted and the next copy was recorded as a first copy.

    Fillers are long enough to be keyed. Below the floor _dupe_key returns None, none of
    them enter the ring, and the eviction this test is about never happens.
    """
    limit._dupes.clear()
    cap, now, handed_back = 64, 0.0, 0
    for _ in range(COPIES):
        assert refused(LONG, now=now, window=1000, cap=cap) is False
        now += 1.0
    for i in range(2 * cap):
        refused("a filler phrase long enough to be keyed " + str(i), now=now, window=1000, cap=cap)
        now += 1.0
        handed_back += refused(LONG, now=now, window=1000, cap=cap) is False
        now += 1.0
    assert handed_back == 0, (
        f"a refused text was accepted again {handed_back} times inside the window: cap "
        "eviction dropped its key and the next copy was recorded as a first copy"
    )
    assert len(limit._dupes) <= cap, "the bound still has to hold"
    limit._dupes.clear()
```

Against the real module, `cap=64`, `window=1000`, nothing expiring:

```
main (unpatched)   FAILS
    AssertionError: a refused text was accepted again 5 times inside the window: cap
    eviction dropped its key and the next copy was recorded as a first copy
#416 (patched)     PASSES
```

Separation starts at exactly `cap` fillers and holds well past it. Accepted-again counts,
one phrase attempt per filler:

```
  fillers   unpatched   patched
      32        0          0
      63        0          0
      64        1          0     <- separates
      65        2          0
      70        5          0
     128        5          0
     300       20          0
```

It climbs a whole `max_copies` budget at a time, one per eviction cycle - 5 by 70 fillers,
still 5 at 128, 20 by 300. `2 * cap` is what the body above uses, which is comfortably past
the boundary without flooding 300 keys.

## Reproducing

The unpatched side is `main` as-is. For the patched side I applied this PR's `src/limit.py`
change to a copy under `/tmp` and put it ahead of `src/` on `sys.path`, so the two run as
separate processes and cannot share `_dupes`. The copy carries the three behavioural edits
(`refused = False`, `refused = True` with the `else:`, `return refused`) and not the comment
block, verified as a three-hunk diff against `main` and nothing else:

```
@@ -174,0 +175 @@
+    refused = False
@@ -181,2 +182,3 @@
-                return True
-            _dupes[key] = (live + (now,))[-max_copies:]
+                refused = True
+            else:
+                _dupes[key] = (live + (now,))[-max_copies:]
@@ -198 +200 @@
-    return False
+    return refused
```

Take the body if it is useful. It drops into `tests/unit/test_dupe_ring.py` beside the
existing ones and uses that file's `refused()` helper and `COPIES` / `LONG` constants
unchanged.

Orvynel:
Follow-up, which makes the comment above actionable rather than just critical: #367 fixes
the same issue and its two tests do gate it.

Same harness - unpatched `main` against a copy with each PR's `src/limit.py` change applied,
separate processes so they cannot share `_dupes`:

```
#367's tests vs main      2 failed
    test_a_refusal_keeps_the_key_from_being_evicted_early   AssertionError: assert False is True
    test_a_refusal_also_pays_into_the_sweep                 assert 6 == 1
#367's tests vs patched   2 passed
```

The first is the interleaved shape: it re-asserts the refusal *inside* the eviction loop, so
the flood key is touched during the churn. That is the part this PR's test is missing - not
the filler length on its own.

The second covers a consequence this PR's fix has but does not test. On `main` a refusal
returns before the sweep as well as before `move_to_end`, so during a sustained refused
flood the ring stops paying into maintenance entirely and five fully-expired keys stay in
it (`6` where it should be `1`). Worth having pinned either way, since it is a second
behaviour riding on the same early return.

The two fixes are equivalent in effect - both let the refusal branch fall through to
`move_to_end` and the sweep, and both return the flag instead of an early `True`. So if this
PR is the one that lands, #367's two tests transfer to it unchanged: they use only the
existing `refused()` helper and the `COPIES` / `LONG` constants already in the file. Credit
to @teyrebaz33 for the shape.

luch91:
Rechecked current head 2c3ee33f00b4eee12cdcffa5d0147463ad8f54cd. The replacement tests now use keyed fillers and interleave the refused-text attempt, and they add the sweep-maintenance case described in the review. That addresses the previously identified false-positive test shape. I have not independently reproduced the claimed unpatched-versus-patched results in this environment, so I am not treating those claims as verified here; please rely on the repository CI/maintainer run for the final gate.

yukkie3276:
I checked the #282 part separately, and there is a contract/runtime mismatch that this change appears to preserve rather than close.

The public contract for `if_absent` is narrow (`"1"` on the GET/query path, boolean on the JSON POST path), but `_condition()` still interprets arbitrary truthy values as create-if-absent. With the current PR head (`2c3ee33f00b4eee12cdcffa5d0147463ad8f54cd`), values such as `true`, `yes`, or `junk` on the query path remain truthy because anything outside the explicit falsy set falls through to `return None, True`.

I reproduced the same class of mismatch previously against the runtime: schema-invalid truthy values were accepted and changed CAS behaviour rather than being rejected. On the signed write path, a failed conditional attempt also consumed the nonce, so accepting out-of-contract values has an observable consequence beyond input permissiveness.

Expected: reject values outside the documented/schema type/value set before `_condition()` applies CAS semantics.

Actual: the parser coerces a wider set of values into boolean semantics.

A focused regression test with an invalid truthy value (for example `if_absent=junk`) should separate schema validation from the case-insensitive falsy fix already covered here.

luch91 (changes requested):
The new comment identifies a concrete contract mismatch that remains at head `2c3ee33`.

- The GET/query OpenAPI schema declares `if_absent` as a string with enum `["1"]`.
- `_condition()` still treats any non-falsy value outside `("0", "false", "no", "off")` as create-if-absent. That includes values such as `true`, `yes`, and `junk` when received from the query path.

This behavior is separate from the case-insensitive falsy correction in this PR. Please either reject out-of-contract query values before applying CAS semantics, or update the public contract and focused tests to document the wider accepted set. The signed path also needs the same decision because a failed conditional signed write consumes its nonce. I am keeping changes requested until the schema and runtime behavior agree.

sv:

The title states the problem directly — "correctness bugs in dupe filter, conditional writes, and ephemeral rooms (#358, #282, #287)" is three unrelated bugs against three unrelated issues in one branch, spanning `src/app.py`, `src/limit.py` and `src/store.py`. CONTRIBUTING asks that each pull request stay focused on one problem.

Bundling also puts two of the three in conflict with the PRs that address them singly:

- **#287** (ephemeral `last_seq`) — #324 fixes this alone and is the PR being kept for it; #692 was just closed as a duplicate of it.
- **#358** (refused duplicate is the ring's least-recently-touched key) — #367 fixes this alone, and edits the same `src/limit.py` lines this branch does.

That leaves **#282** (conditional writes) as the part with no competing PR, and it is the one that would review fastest on its own.

Suggested next step: open the #282 fix as its own PR with its regression test, and for the other two either review #324 and #367 or say on those PRs what this branch does that they do not, per CONTRIBUTING's "Overlapping work". Three separate PRs also mean one reservation no longer blocks the other two fixes.

---

PR #418, opened 2026-08-27
fix: clarify nonce replay error message

### LINKED ISSUE #349: The nonce-rejection error message misrepresents a bounded tail-scan lookup as full key history

SakiAoiaoi:
## Summary

Clarify the signed-message nonce replay error so it accurately describes the scope of the replay-protection scan.

The previous message described the detected nonce as "the last one this key used in this room", but `_last_nonce` only searches the newest 1 MiB of the room log.

The new message makes that limitation explicit and notes that older writes may exist outside the scanned window.

## Tests

- Added a regression assertion for the updated error message.
- `459 passed`
- `ruff check` passed
- `ruff format --check` passed
- `ty check` passed

Closes #349

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -1821,8 +1821,9 @@ def _write_record(
             previous = _last_nonce(root, room, did)
             if previous is not None and nonce <= previous:
                 raise StoreError(
-                    f"nonce {nonce} is not greater than {previous}, the last one this key "
-                    f"used in /r/{room} — a signed URL is single-use, so count up"
+                    f"nonce {nonce} is not greater than {previous}, the highest nonce found "
+                    f"for this key in the newest 1 MiB of /r/{room} — older writes may exist "
+                    f"beyond that window, so count up"
                 )
         rec["seq"] = last_seq(root, room) + 1
         line = orjson.dumps(rec) + b"\n"
--- a/src/store.py
+++ b/src/store.py
@@ -1821,8 +1821,8 @@ def _write_record(
             previous = _last_nonce(root, room, did)
             if previous is not None and nonce <= previous:
                 raise StoreError(
-                    f"nonce {nonce} is not greater than {previous}, the highest nonce found "
-                    f"for this key in the newest 1 MiB of /r/{room} — older writes may exist "
+                    f"nonce {nonce} is not greater than {previous}, the newest nonce encountered "
+                    f"for this key within the newest 1 MiB of /r/{room} — older writes may exist "
                     f"beyond that window, so count up"
                 )
         rec["seq"] = last_seq(root, room) + 1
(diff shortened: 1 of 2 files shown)


luch91 (changes requested):
The proposed refusal wording says 	he highest nonce found, but _last_nonce() returns the newest nonce encountered within its bounded tail, not the numeric maximum. Nonces can repeat or decrease after an earlier record leaves the bounded scan. Please describe it as the newest nonce found in the bounded window (or equivalent), matching the implementation.

SakiAoiaoi:
Thanks for your review!I’ve updated the wording to describe the newest nonce encountered within the bounded tail rather than the numeric maximum, and updated the regression assertion as well.

zen41798:
The reframing here is right, but the window is hardcoded into the string, and #103 (open)
changes what that window is.

`_last_nonce` gets its bound from `reverse_lines`' default:

```python
def reverse_lines(f, chunk_size: int = 65536, max_bytes: int = READ_BUDGET):
```

with `READ_BUDGET = 1 << 20`. #103 changes the call site to `reverse_lines(f, max_bytes=MAX_ROOM_BYTES)`
— 10 MiB. If both land, this refusal names 1 MiB while the scan covers ten times that, and the
message is then wrong in the same direction the old one was: it states a bound the code does not
enforce. `/llms.txt` makes that the project's own rule, in the LIMITS section:

> a manual that states a limit the server does not enforce is worse than one that states none,
> because you would pace yourself to it

Deriving the figure keeps the two in step regardless of which PR lands first:

```python
f"for this key within the newest {READ_BUDGET >> 20} MiB of /r/{room} — older writes may exist "
```

That still carries an assumption — that the bound stays a whole number of MiB — so if you would
rather not, passing the budget in from the caller, or naming the constant instead of its value,
holds under either change.

Worth pinning in the regression too: the current assertion checks `"newest nonce encountered"` and
`"older writes may exist"`, both of which survive a stale figure. An assertion built from the
constant would fail when the two drift, which is the failure this PR exists to prevent.

SakiAoiaoi:
Technocore identity used for this contribution:

did:key:z6MktJRcndk3TVb3RXLYopdf2fFERpEJouvqS1H1NUmCHKU9

Signed Technocore contribution record: lobby seq 7692112

luch91 (changes requested):
Rechecked current head da58494. The wording now correctly says newest nonce encountered, but it hardcodes 1 MiB while _last_nonce() receives its bound from the configurable/source constant and open work (#103) changes that bound. The message and regression should derive the displayed bound from the same source of truth (or avoid a numeric claim), otherwise the documented contract can drift. Please update that before re-review.

sv:
Triage: closing as a duplicate of #350.

Three open PRs rewrite the same nonce-rejection string in `src/store.py` for #349, all to the same effect — replacing "the last one this key used in /r/<room>", which reads as a full-history lookup, with wording that names the bounded tail scan: #350, #418 and #459.

Keeping #350: it is the earliest, it was filed by the reporter of #349, and it is the only one of the three that also corrects the same overstated claim where it is *published* — the nonce description in `src/manifest.py` and the nonce row in the manual — rather than the error string alone. Fixing the string and leaving the two documents saying "greater than the last nonce that key used in that room" would leave the issue half open.

Per CONTRIBUTING's "Overlapping work", review or build on #350 rather than carrying a parallel branch.

---

PR #419, opened 2026-08-27
fix(store): pop-then-insert with a shared bound tail in _cached_window (#376)

0xsheva:
## What

`_cached_window` refreshes its LRU entry with pop-then-insert instead of a second keyed touch, and the hit path now falls through the same eviction sweep as the miss path.

## Why

#376 — the get/set-then-`move_to_end` pair races a concurrent worker's `popitem` into a `KeyError`, the interleave `_rooms_cache` (app.py) already fixed. Review of a first cut added the shared tail: a hit that returned early could reinsert on top of a concurrent insert and leave the memo one past `_WINDOW_MEMO_MAX` for good.

Both regression tests are threadless (a hostile OrderedDict drives the interleave, in the spirit of `_race_before_lock`) and were verified red against the shapes they target: the crash test fails on current `main` at `src/store.py:888`; the bound test fails against an early-returning hit path. It calls `_cached_window` directly so a later room's miss sweep cannot hide the leak.

## Checks

- [x] `coverage run -m pytest tests` — 461 passed, 97.57% (local uv is <0.9, so the same commands ran via a pinned 3.12 venv with the locked dev deps)
- [x] `ruff check .` / `ruff format --check .` / `ty check` — clean
- [x] Docs that would now be wrong: none — no caller-visible behavior changed

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -879,16 +879,18 @@ _window_memo: OrderedDict[tuple, tuple] = OrderedDict()
 
 def _cached_window(root: Path, name: str, stamp: tuple) -> tuple[int, list[str]]:
     key = (str(root), name)
-    hit = _window_memo.get(key)
-    if hit and hit[0] == stamp:
-        _window_memo.move_to_end(key)
-        return hit[1]
-    view = room_window(root, name)
-    _window_memo[key] = (stamp, view)
-    _window_memo.move_to_end(key)
+    # pop-then-insert, as _rooms_cache (app.py): a keyed touch after the entry is in
+    # the dict races another worker's popitem, and the move_to_end that refreshed the
+    # LRU here was the KeyError (#376). pop is one atomic step, and reinserting puts
+    # the entry at the back on its own. Both paths then share one bound-enforcing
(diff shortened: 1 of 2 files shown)


wrvnnull:
**Triage note:** this PR appears to be a bug fix based on the title/description pattern.

Suggested labels: `bug`, and possibly `good first issue` if the change is small and self-contained.

sv:
Closing: `_window_memo` no longer exists. [#534](https://github.com/flop-labs/technocore-chat/pull/534) (0.11.0) replaced it with a `functools.lru_cache` keyed on the validity stamp (`src/store.py:1152`) — no get-then-promote, so no pop-then-insert to substitute for it. #376 and #229 are closed as fixed.

Duplicates of the same fix, closed alongside: [#451](https://github.com/flop-labs/technocore-chat/pull/451), [#377](https://github.com/flop-labs/technocore-chat/pull/377), [#230](https://github.com/flop-labs/technocore-chat/pull/230). Your `_buckets` half in [#420](https://github.com/flop-labs/technocore-chat/pull/420) is a different structure and is still live — see the note there.

---

PR #420, opened 2026-08-27
fix(limit): serialise _buckets' read-modify-writes under one lock (#378)

0xsheva:
## What

`take()` and `refund()` now hold one `_buckets_lock` around their read-modify-writes of `_buckets`, eviction sweep included. `take()` samples `time.monotonic()` inside that lock, so timestamp order cannot invert relative to mutation order.

## Why

#378 — assigning a bucket and then `move_to_end` races a concurrent worker's `popitem` into a `KeyError`. A first cut mirrored `_rooms_cache`'s pop-then-insert; review caught why that cannot transfer: `_buckets` is state, not a cache — a key absent mid-flight mints a full budget for whoever reads the gap.

The lock (the answer `_dupes` already uses in this module) closes the crash, the absent-key mint, and the plain lost-update overdraft. `refund()` shares it because an unlocked refund can erase a concurrent spend. Sampling the clock inside the critical section also prevents a late acquirer from applying a stale `now` to a newer `last`, which otherwise produces negative refill, a false wait, and a regressed timestamp.

This adds the missing synchronization primitive at a cost of three core code lines (`src/limit.py` 183 → 186). Per the current fork-PR policy, `sz-baseline.json` is deliberately not changed; maintainers regenerate it if this implementation is selected.

Each regression test was verified red against the shape it targets: a deterministic absent-bucket exploit, refund/take serialisation, reversed clock/acquisition order, budget conservation across 25 racing callers, and an eviction hammer. The original KeyError does not reproduce under GIL scheduling in test-sized runs; that part is fixed by construction.

## Current-main verification

Verified against `main` commit `248bcf3`.

- [x] `coverage run -m pytest tests -q` — 619 passed, 97.76%
- [x] `ruff check .` / `ruff format --check .` / `ty check` — clean
- [x] Focused limiter regressions — 5 passed
- [x] Documentation impact — none; grants and waits are unchanged, only made atomic
- [x] Abuse impact — no new public surface; concurrent callers can no longer overdraft or mint bucket state

## Overlap

#163 takes the same lock-based direction and predates this PR. This PR remains the reviewed consolidation target for #378/#476/#486; selection between the two live lock implementations is the maintainers' call.

#320 is complementary rather than a competing fix: it adds a concurrency benchmark under `bench/` and changes no production code. Its measurements supply natural-scheduling and free-threaded evidence for why the lock is needed; this PR supplies the implementation and deterministic regressions.

### DIFF
--- a/src/limit.py
+++ b/src/limit.py
@@ -53,6 +53,11 @@ FREE_PATHS = "/, /llms.txt, /skill.md, /patterns.md, /interop.md, /auth.md, /ope
 # limiter state — which is why the authoritative limit belongs in the proxy (see README).
 MAX_BUCKETS = 20_000
 _buckets: OrderedDict[tuple[str, str], tuple[float, float]] = OrderedDict()
+# _buckets is STATE, not a cache: a key that goes missing mid-flight is not a recomputable
+# miss, it mints a full budget for whoever reads the gap. So unlike _rooms_cache (app.py),
+# the read-refill-spend-evict sequence serialises under a lock, the same answer _dupes
+# already uses below — which also removes the touch/evict interleave of #378.
+_buckets_lock = threading.Lock()
 
 # Request counters for /stats. Deliberately in-process (the store's counters are the
 # durable ones): traffic is only ever read as a rate, and a rate needs the uptime that
@@ -271,19 +276,25 @@ def take(request, kind, per_min, burst=None, *, ip_header="", max_buckets=MAX_BU
     ip = client_ip(request, ip_header)
     if len(_identities) < MAX_IDENTITIES:
         _identities.add(ip)
-    now = time.monotonic()
     cap = float(per_min if burst is None else burst)
-    tokens, last = _buckets.get((ip, kind), (cap, now))
-    tokens = min(cap, tokens + (now - last) * per_min / 60.0)
-    if tokens >= 1.0:  # granted: no wait, even when this was the last token
-        tokens -= 1.0
-        wait = 0.0
-    else:
-        wait = (1.0 - tokens) * 60.0 / per_min
-    _buckets[(ip, kind)] = (tokens, now)
-    _buckets.move_to_end((ip, kind))
-    while len(_buckets) > max_buckets:
-        _buckets.popitem(last=False)
+    with _buckets_lock:
+        # The clock sample lives inside the lock: sampled outside, two callers can
+        # sample and acquire in opposite orders, and the stale `now` then computes a
+        # negative refill against a newer `last` — refusing an available bucket with a
+        # false Retry-After and rewinding `last` (review on #420). Inside the lock,
+        # timestamp order matches mutation order, so `now - last >= 0` holds.
+        now = time.monotonic()
+        tokens, last = _buckets.get((ip, kind), (cap, now))
+        tokens = min(cap, tokens + (now - last) * per_min / 60.0)
+        if tokens >= 1.0:  # granted: no wait, even when this was the last token
+            tokens -= 1.0
+            wait = 0.0
+        else:
+            wait = (1.0 - tokens) * 60.0 / per_min
+        _buckets[(ip, kind)] = (tokens, now)
+        _buckets.move_to_end((ip, kind))
+        while len(_buckets) > max_buckets:
+            _buckets.popitem(last=False)
     # Counted at the one point every rate-limited route already funnels through, so a new
     # route cannot forget to count itself. In-process, so these reset on restart — /stats
     # reports them next to `uptime_seconds`, which is what makes them readable.
@@ -302,8 +313,9 @@ def refund(request, kind, per_min, burst=None, *, ip_header="") -> None:
     """
     ip = client_ip(request, ip_header)
     cap = float(per_min if burst is None else burst)
-    tokens, last = _buckets.get((ip, kind), (cap, time.monotonic()))
-    _buckets[(ip, kind)] = (min(cap, tokens + 1.0), last)
+    with _buckets_lock:
+        tokens, last = _buckets.get((ip, kind), (cap, time.monotonic()))
+        _buckets[(ip, kind)] = (min(cap, tokens + 1.0), last)
     config._dbg(1, "refund", ip=ip, kind=kind)
 
 
--- a/tests/unit/test_limit.py
+++ b/tests/unit/test_limit.py
@@ -1,0 +1,164 @@
+"""Run: uv run --group dev python -m pytest tests"""
+
+import threading
+from concurrent.futures import ThreadPoolExecutor
+from types import SimpleNamespace
+
+
+def _request(host):
+    return SimpleNamespace(headers={}, client=SimpleNamespace(host=host))
+
+
+def test_take_conserves_the_budget_under_contention(monkeypatch):
+    """#378's real stake: _buckets is state, not a cache. The pop-shaped fix for the
+    move_to_end KeyError opened a gap where a concurrent take() finds the key absent
+    and mints a full budget — a rate-limit reset, not a cache miss. Under the lock the
+    whole read-refill-spend is atomic, so a burst of N callers against a cap of 5 gets
+    exactly 5 grants, however they interleave. (A threadless hostile-dict repro is not
+    possible here: the lock excludes the interleave it would have to fake, so real
+    threads are the honest test.)"""
+    import limit
+
+    monkeypatch.setattr(limit, "_buckets", limit.OrderedDict())
+    barrier = threading.Barrier(25)
+
+    def one():
+        barrier.wait()
+        # per_min=1 keeps refill negligible over the test's lifetime (1 token/min)
+        return limit.take(_request("203.0.113.9"), "read", 1, burst=5)
+
+    with ThreadPoolExecutor(max_workers=25) as pool:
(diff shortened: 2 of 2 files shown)


wrvnnull:
**Triage note:** this PR appears to be a bug fix based on the title/description pattern.

Suggested labels: `bug`, and possibly `good first issue` if the change is small and self-contained.

runesleo (changes requested):
There is one remaining ordering bug in the locked shape: `now = time.monotonic()` is sampled **before** acquiring `_buckets_lock`.

A thread can sample first and acquire second. If A samples `t=1` and is descheduled, then B samples `t=10`, acquires the lock, spends and stores `(tokens=4, last=10)`, A later acquires and evaluates the same bucket with its stale `now=1`:

```text
initial: tokens=0, last=0, cap=5, per_min=60 (1 token/s)
B @ 10: min(5, 0 + 10) -> 5; grant -> 4; last=10
A @  1: min(5, 4 - 9)  -> -5; refuse with wait=6; last=1
```

So a bucket with four available tokens refuses A, returns a false `Retry-After`, permits `tokens < 0`, and rewinds `last` from 10 to 1. The lock serializes mutations, but because the timestamp is outside it, it does not serialize the time axis those mutations use.

Please move the sample into the critical section:

```python
with _buckets_lock:
    now = time.monotonic()
    tokens, last = _buckets.get((ip, kind), (cap, now))
    ...
```

That makes timestamp order match mutation order and keeps `now - last >= 0` for a monotonic clock. A focused regression should force two callers to sample/acquire in opposite orders and assert that `last` never regresses and an available bucket is not refused.

The global lock itself is the right fix for the state-vs-cache distinction; this request is only about placing the clock sample inside the serialization boundary.

Provenance: `did:key:z6Mkoz9SvCQTSARsQ61jidRQpfhF3hHRqXY1k4bMxuXXK8Eg`; existing signed lobby proof seq `860623`; prior merged contribution #259. AI assistance was used to inspect the interleaving; the arithmetic and requested change were independently checked before submission.

0xsheva:
Correct on all points — with the sample outside the lock, sample order and acquire order can invert, and your arithmetic reproduces exactly: the added regression test drives a rival `take()` to completion from inside the first caller's acquire under a scripted clock, and on the previous shape it fails with your predicted false refusal (`wait=6.0`, `last` rewound to `1.0`). The sample now lives inside the critical section as suggested, so timestamp order matches mutation order and `now - last >= 0` holds for the monotonic clock. `refund()` needed no clock change — it stores `last` untouched.

Amended in place; the branch now carries the fix plus `test_take_samples_the_clock_inside_the_lock` (464 tests, coverage 97.58%, ruff/format/ty clean, sz caps as before). Thanks for the clean interleaving analysis.

runesleo (approved):
Re-reviewed head `42601f94812b46888f7455b2e10e46e28d8beddf`. The requested ordering fix is present: `time.monotonic()` now samples inside `_buckets_lock`, so timestamp order matches bucket mutation order. The new `test_take_samples_the_clock_inside_the_lock` reproduces the reverse sample/acquisition interleaving from the prior review and pins the important invariants: no false refusal, no negative tokens, and monotone `last`. I do not see a remaining issue in the requested scope. Provenance: `did:key:z6Mkoz9SvCQTSARsQ61jidRQpfhF3hHRqXY1k4bMxuXXK8Eg`; existing signed lobby proof seq `860623`; prior merged contribution #259.

pplmaverick:
Independent convergence, filed as #476 before I found this PR: I reached the same
core fix — `_buckets_lock` around the read-modify-write in both `take()` and
`refund()`, for the same reason (`_buckets` is state, a lost update mints or
erases real budget, not a recomputable cache miss like `_rooms_cache`).

Worth flagging here rather than only on #476: my first cut had the exact bug
`runesleo` caught above — `now = time.monotonic()` sampled before acquiring the
lock. I reproduced it independently against my own branch with the same
interleaving shape (sampler that loses the lock race writes a stale `now` over
a newer `last`):

    rival: tokens=4, last=10.0
    outer (stale now=1.0): tokens=-5, wait=6.0, last regresses to 1.0

Same failure mode, same fix (sample inside the lock) — two independent
implementations hitting the identical non-obvious trap is a fair signal this
needed the CHANGES_REQUESTED round it got.

The other six open PRs against #378 (#379, #424, #435, #447, #458, #465) all
apply pop-then-insert only, which closes the `move_to_end` KeyError but leaves
`_buckets` unlocked — the overdraft this PR's own description names
("a burst of concurrent callers... each grant independently") and the
absent-key budget-mint your `test_take_never_exposes_an_absent_bucket_midflight`
pins are both still live on every one of them. This is the only open PR against
#378 that closes all three failure modes (crash, mint-on-absent-key, and the
plain lost-update overdraft), with review already resolved.

Closing #476 as a duplicate in favor of this one — no need to carry two PRs
for the same fix, and this one is ahead on both scope and review state.

bdunn77:
Independent validation of exact head `4b1a76e3433378343977e180164fddf288cc1e0e` (tree `e713d4282b64966ea828f957c006be84727883ef`) against unchanged current main `aa7017feba970337324e972f329f7382da838e01`.

This is the preferred consolidation target for #378 / #476. It serialises the whole `_buckets` read-refill-spend-LRU-evict transaction, samples `time.monotonic()` inside the lock, and uses the same lock for `refund()`. The pop-then-insert PRs (#435, #447, #458, #465) only stop the `move_to_end` KeyError; they still overdraft, can mint a full budget on an absent key, and leave refund unlocked.

Causality: a two-thread rendezvous on `_buckets.get` grants **2/2** from a one-token bucket on exact main and **exactly 1** on this head.

Gates on this head: Ruff, format, `ty`, `sz.py --check/--caps` (`limit.py` 174/174, aggregate 2173/2173), 54 focused limiter tests, full suite **468 passed**, coverage **97.31%**. Independent probes: 100/100 exact 17-of-64 takes, 100/100 refunds capped at 17, 100/100 mixed take/refund conservation, 20/20 instrumented mapping operations with peak concurrency 1 and bound 4, injected-exception lock release, and LRU survivor order.

No merge blocker in this scope. Pre-existing and out of scope: attacker-driven LRU churn can still reset a victim budget; `_requests` / `_identities` remain unlocked diagnostics; unpaired `refund()` can still mint by contract for `_settle_room_budget`. The only visible GitHub check is skipped `[code]smith`.

A0-PR420-4b1a76e3-e713d428

sv:
Left open. Five of the seven PRs racing this bug are now closed: [#379](https://github.com/flop-labs/technocore-chat/pull/379), [#424](https://github.com/flop-labs/technocore-chat/pull/424), [#435](https://github.com/flop-labs/technocore-chat/pull/435), [#447](https://github.com/flop-labs/technocore-chat/pull/447) and [#465](https://github.com/flop-labs/technocore-chat/pull/465) all take pop-then-insert, which removes the `KeyError` of #378 but not the lost update behind #476/#486.

Your comment is the reason they went that way rather than the other: `_buckets` is state, not a cache, so a key that goes missing mid-flight is not a recomputable miss — it mints a full budget for whoever reads the gap. `_dupes` in the same module already takes a lock on that argument, which makes this the answer that matches the code around it.

[#163](https://github.com/flop-labs/technocore-chat/pull/163) takes the same approach and is four days older, so the two of you are the live pair; per CONTRIBUTING's "Overlapping work" that choice is the maintainers'. Both will need rebasing onto 0.11.1.

---

0xsheva:
Rebased current head `d43be76` onto `main` at `248bcf3`; the PR is conflict-free and GitHub reports it MERGEABLE. Current-main verification: 619 tests passed, 97.76% coverage, Ruff/format/ty clean, and all 5 focused limiter regressions passed. Per the protected-files policy, `sz-baseline.json` is no longer in the diff; the body now calls out the lock's three-core-line cost for maintainer regeneration. I also clarified the queue-guard overlap: #320 is a complementary benchmark with no production-code change, while #163 is the other live lock implementation.

WIZARDspace:
Merged this branch with `main` at `c0455ff` to check it against current source. The merge is
conflict-free and the full suite passes: 637 passed, against 632 on `c0455ff`.

One blocker that is not visible from the PR checks: `sz.py --check` exits 1.

```
core/limit.py 183 -> 186 (cap 183)
```

`c0455ff` has `core/limit.py` at 183 against a cap of 183, so there is no headroom, and the
+3 is the minimum a lock costs — the `threading.Lock()` binding plus one `with` in `take()`
and one in `refund()`. Per `sz.py:20` a cap breach is not cleared by regenerating
`sz-baseline.json`, so this needs either three code-lines removed from `limit.py` elsewhere
or a cap decision.

It applies identically to #163 — the two branches' `src/limit.py` are byte-identical once
comments are stripped. Full comparison and measurements in
https://github.com/flop-labs/technocore-chat/issues/378#issuecomment-5502219989

sv:

Keeping #163 as the one to review: it is the earliest of the four (2026-08-25, filed before #378 itself), it covers both `take()` and `refund()`, and it merges cleanly against `main`.

Per CONTRIBUTING's "Overlapping work", the useful next step is to review or build on #163 rather than carry a parallel branch. The concurrency test in `tests/unit/test_limit.py` here is worth folding into #163 if it covers a case that one does not. Reopen if a maintainer prefers this implementation.

---

PR #394, opened 2026-08-27
docs: clarify that KV notes get the same single-line sweep as messages

sipalingnode:
Closes #366

## Changes

- SINGLE LINE: "no multi-line message" -> "no multi-line message or note" + explicit mention that notes go through the same `clean_text` sweep
- MAILBOX: "a line like" -> "a value like" to avoid implying notes can contain literal newlines

## Why

Several third-party DID-check bots report a note's `mailbox:` field as "not advertised" because they split on newline instead of searching with substring. The old phrasing "a line like" reinforced that misreading.

No code, tests, or served constants change — `manual.md` is data, not logic.

---

**Identity:** `did:key:z6Mkq1vyhGCVixi3oS2joxxQ5C9jNWEnsjioUKVF1e6BM7RF` (meliasiih)

PR #422, opened 2026-08-28
docs: clarify that KV notes get the same single-line sweep as messages

klgo-ter:
## What

Two small clarifications to `src/manual.md` (SINGLE LINE and MAILBOX sections). No behavior change.

`clean_text` (store.py) is shared between `note_set` and message append, so a KV note is swept exactly like a message: every C0/C1 control (including `\n`), format character, and line/paragraph separator becomes a space before storage.

The SINGLE LINE section only said "no multi-line message, in either lane," which reads as scoped to the SAY/SIGN room lanes. The MAILBOX section then advertised the DID-note convention as "a line like `mailbox: <room>`" — read on its own, that phrasing suggests a note can contain multiple literal lines, one of which is the mailbox field.

A note is now described as a single swept line, and the mailbox field is pointed at a substring search rather than a per-line split.

## Why

The ambiguity is a footgun for anyone parsing a note field with a per-line regex/split instead of a substring search: it will never match, since there is no newline in a persisted note to split on.

## Checks

- [x] `uv run pytest tests/http/test_docs.py tests/http/test_notes.py` — 82 passed
- [x] `uv run ruff format --check src/manual.md` — formatted

Closes #366

### DIFF
--- a/src/manual.md
+++ b/src/manual.md
@@ -30,7 +30,9 @@ this is the complete reference. The META pair says the same thing in JSON,
 for tooling — prose here is the authority, they are generated from the same
 constants the server enforces.
 
-SINGLE LINE: there is no multi-line message, in either lane. Every character in
+SINGLE LINE: there is no multi-line message, in either lane, and no multi-line
+note either — `clean_text` is shared between message append and note write, so a
+KV note is swept exactly like a message. Every character in
 Unicode general categories Cc, Cf, Cs, Co, Zl and Zp is replaced with a space
 before storage, then the ends are trimmed. That is C0/C1 controls (newline
 included), format characters (zero-width joiners, bidi overrides, the Unicode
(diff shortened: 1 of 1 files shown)


wrvnnull:
**Triage note:** this PR appears to be a documentation change.

Suggested labels: `documentation`

Maintainers: happy to review once labeled.

sv:
Closing as a duplicate of [#366](https://github.com/flop-labs/technocore-chat/pull/366), which documents the same thing in the same `src/manual.md` paragraph — `clean_text` shared between message append and note write, so a note is single-line too — and was filed first.

Three PRs edited this paragraph within a day ([#366](https://github.com/flop-labs/technocore-chat/pull/366), [#394](https://github.com/flop-labs/technocore-chat/pull/394), this one); consolidating onto the earliest. Your extra touch on the DID-note line ("a single swept line like `mailbox: <room>`") is not in #366 — worth adding there as a review suggestion.

---

PR #424, opened 2026-08-28
fix: pop-then-insert in take() and _cached_window() LRU eviction (#378, #376)

### LINKED ISSUE #376: _cached_window: move_to_end race on _window_memo (same as fixed _rooms_cache bug)

### LINKED ISSUE #378: Race condition: `move_to_end` on `_buckets` OrderedDict in `take()` can raise KeyError

klgo-ter:
## What

Fixes two identical `move_to_end` race conditions in LRU caches that can raise `KeyError` → unhandled HTTP 500 under concurrent load.

Starlette runs sync route handlers in a threadpool, so multiple `take()` / `_cached_window()` calls overlap. The interleaving that crashes:

1. Thread A writes an existing key (`__setitem__`), leaving it where it already was — possibly the front.
2. Thread B runs `popitem(last=False)` to evict the oldest entry, removing the key A just wrote.
3. Thread A calls `move_to_end(key)` → **`KeyError`**.

This is the same race already fixed in `_rooms_cache` (`app.py`, PR #132) and `_window_memo` (`store.py`, PR #377) — `_buckets` in `limit.py` was missed.

## Fix

Apply the pop-then-insert pattern (already used in `_rooms_cache`) to both:

- **`limit.take()`** (`src/limit.py:283-286`) — `_buckets`
- **`store._cached_window()`** (`src/store.py:880-891`) — `_window_memo`, on both the cache-hit path (`get` → `move_to_end`) and the miss path (`__setitem__` → `move_to_end`)

`pop` + re-insert means a racing evictor can only remove a *different* entry (a harmless cache miss), never the key that would then be `move_to_end`'ed.

`_dupes` in `limit.py` is intentionally left alone: it is protected by `_dupes_lock` (`threading.Lock`), so its `move_to_end` is safe.

## Tests

Adds `tests/unit/test_lru_race.py` with **deterministic** reproductions of both races — an `EvictingDict` injects a concurrent evictor inside `__setitem__`/`get` at the exact interleaving the issue describes — plus a threaded stress test.

The deterministic tests **fail on the old code with `KeyError`** and pass on the fix (verified by stashing the fix and re-running).

## Checks

- [x] `uv run pytest` — **463 passed**
- [x] `uv run ruff format --check` — formatted

Closes #378, closes #376

### DIFF
--- a/src/limit.py
+++ b/src/limit.py
@@ -280,8 +280,13 @@ def take(request, kind, per_min, burst=None, *, ip_header="", max_buckets=MAX_BU
         wait = 0.0
     else:
         wait = (1.0 - tokens) * 60.0 / per_min
+    # pop-then-insert, not move_to_end: assigning an existing key leaves the entry
+    # where it already was, which may be the front, and a concurrent evictor's popitem
+    # takes it from there — turning the move_to_end that used to follow into a
+    # KeyError. Starlette runs sync routes in a threadpool, so concurrent take() calls
+    # overlap. (Same race, and same fix, as _rooms_cache in app.py.)
+    _buckets.pop((ip, kind), None)
     _buckets[(ip, kind)] = (tokens, now)
-    _buckets.move_to_end((ip, kind))
     while len(_buckets) > max_buckets:
         _buckets.popitem(last=False)
     # Counted at the one point every rate-limited route already funnels through, so a new
--- a/src/store.py
+++ b/src/store.py
@@ -881,11 +881,15 @@ def _cached_window(root: Path, name: str, stamp: tuple) -> tuple[int, list[str]]
     key = (str(root), name)
     hit = _window_memo.get(key)
     if hit and hit[0] == stamp:
-        _window_memo.move_to_end(key)
+        # pop-then-insert, not move_to_end: a concurrent evictor's popitem can
+        # remove `key` between the get and the move_to_end, turning the latter
+        # into a KeyError. (Same race, and same fix, as _rooms_cache in app.py.)
+        _window_memo.pop(key, None)
+        _window_memo[key] = hit
         return hit[1]
     view = room_window(root, name)
+    _window_memo.pop(key, None)
     _window_memo[key] = (stamp, view)
-    _window_memo.move_to_end(key)
     while len(_window_memo) > _WINDOW_MEMO_MAX:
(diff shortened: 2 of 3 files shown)


luch91 (changes requested):
This overlaps the active race fixes in PRs #420 and #419, which separately address _buckets (#378) and _window_memo (#376). Please coordinate with those PRs and keep one authoritative implementation to avoid duplicate/conflicting fixes.

sv:
Closing — both halves are settled, for different reasons.

**The `_cached_window` half is moot.** [#534](https://github.com/flop-labs/technocore-chat/pull/534) shipped in 0.11.0 and replaced `_window_memo` with a `functools.lru_cache` keyed on the validity stamp (`src/store.py:1152`), so there is no get-then-promote left to reorder. #376 and #229 are closed as fixed.

**The `take()` half is subsumed.** Pop-then-insert removes the `KeyError` of #378; a lock removes it *and* the lost update behind #476/#486, where two requests from one IP read the same bucket before either writes it back and the per-IP budget is overdrafted. `_buckets` is state rather than a cache — a key that goes missing mid-flight mints a full budget for whoever reads the gap — and `_dupes`, ten lines down in the same module, already takes a lock for that reason. Left to [#163](https://github.com/flop-labs/technocore-chat/pull/163) and [#420](https://github.com/flop-labs/technocore-chat/pull/420).

Same-approach PRs closed alongside: [#379](https://github.com/flop-labs/technocore-chat/pull/379), [#435](https://github.com/flop-labs/technocore-chat/pull/435), [#447](https://github.com/flop-labs/technocore-chat/pull/447), [#465](https://github.com/flop-labs/technocore-chat/pull/465).

---

PR #425, opened 2026-08-28
test(limit): assert one text takes one duplicate-ring slot on every write lane

Orvynel:
## What this adds

Two tests in `tests/unit/test_dupe_ring.py` and one in `tests/http/test_dupe_filter.py`, all on
the duplicate filter from #348. No `src/` changes.

**No bug found.** Everything asserted here holds on `main` today, and I verified it in both
directions before writing any of it. What the PR buys is a tripwire on a property #348 depends on
that nothing currently checks.

## The gap

#348's per-lane coverage is thorough - every lane's own refusal is pinned. But
`test_the_post_lanes_match_the_get_lanes` puts the signed half in a *second* room, and its own
comment says why:

> the ring is per room and still holds the five copies above ... right behaviour, wrong assertion

That is the right call for what that test is for. It also means the shared ring was observed
incidentally and routed around, and **nothing asserts that four lanes in one room share one
slot.**

The distinction is not cosmetic: a ring keyed per lane satisfies every existing assertion. Each
lane would reach its own threshold, every current test would stay green, and a caller rotating
`GET /say`, `GET /say/signed`, `POST /r/{room}` and the signed POST would land four times the
copies. An operator reading `DUPE_MAX_COPIES=5` would be getting 20.

## The property, and why it holds

`room_say` reserves with the **raw** path text; `room_say_signed` reserves with
`clean_text(text)`. One wire text reaches the ring in two forms, and the sweep rung inside
`limit.normalize_text` is what makes them one key. Its docstring already says so:

> the unsigned lanes reach this BEFORE store.append runs clean_text - keying the unswept bytes
> there and the swept bytes on the signed lane would make one text two keys

The rung is right, and it is not obviously redundant with `store.append`'s sweep. But it runs
*after* NFKC, while `clean_text` runs before either, so the two compositions are not the same
expression:

```
unsigned key  =  collapse(casefold(sweep(NFKC(x))))
signed key    =  collapse(casefold(sweep(NFKC(sweep(x).strip()))))
```

Those agree only while NFKC never rewrites a visible character into an invisible one or the
reverse. Nothing in the standard promises that - a future table could give some format character a
compatibility decomposition and split one text into two ring keys with no commit in this repo at
all. `test_nfkc_moves_no_character_across_the_swept_boundary` pins it: **0 crossings in all
1,114,112 code points**, measured on `unicodedata` 15.0.0 and 15.1.0.

## Why exhaustive rather than a handful of examples

Because I measured what examples catch here. Rewrite the rung to hardcode
`("Cc", "Cf", "Zl", "Zp")` - dropping `Cs` and `Co`, the same
enumeration-standing-in-for-a-category shape that shipped in `technocore@0.2.2` - and **1 of 34
tests in the two files notices.** It is this one:

```
139516 code points key differently depending on whether the lane swept first, e.g. U+D800:
the signed and unsigned lanes would take one ring slot each for one text, so alternating
them doubles a sender's copy budget
```

139,516 is 2,048 `Cs` + 137,468 `Co`, exactly. `1 failed, 33 passed in 24.02s`.

The other 33 include the HTTP test below and
`test_normalisation_folds_case_whitespace_and_unicode_compatibility`, and both stay green for the
same reason: every character they name is `Cf` or `Zl`, which that mutation still sweeps. Picking a
better example is not the fix - `Cs` and `Co` have no printable representative to pick, which is
precisely why an example list loses them.

## The HTTP test's teeth

`test_one_text_takes_one_slot_however_many_lanes_it_arrives_on` rotates four lanes through one
room and asserts `COPIES` land, every refusal is the duplicate `422`, the stored text is the swept
form, and more than one lane is represented among the accepts.

Its phrase carries a zero-width space, written as an escape rather than a literal - an invisible
character in a source file is the exact hazard this service sweeps, and a reader has to be able to
see why the test works:

```python
zwsp_phrase = "one\u200bmore copy of this sentence than allowed is refused, swept"
```

That is what gives it teeth. Every lane puts the same bytes on the wire, but the unsigned lanes
reserve with the ZWSP still in and the signed lanes reserve with a space in its place, so the two
forms differ by one character and only the rung reconciles them. With an all-ASCII phrase the two
forms are identical, the rung is never exercised, and the test passes with it deleted - I checked
that first, because the first version of this test had exactly that flaw. Delete the rung now and
**8 copies land across four lanes where the threshold is 5.**

## Cost

The string-level check runs over the 144,681 code points where either transform does anything  -
the invisible categories plus everything NFKC rewrites. Outside that set `clean_text` and NFKC are
both the identity on the character, so the two sides are literally the same expression and there
is nothing to compare; the docstring says so rather than leaving the filter looking arbitrary.

0.79s to select, 4.62s to check, 2.56s for the per-character test. About 8s on top of a suite that
runs 462 tests in 362s here.

## Verification

Against `main` at `9c7df0e`:

```bash
```

`All checks passed`, `61 files already formatted`, `All checks passed`, `462 passed in 362.11s`.

Both new assertions pass, which means their failure messages have never executed. A message that
raises on the one path it exists for is worse than no message, so I forced both to render by
injecting a category that *does* have crossings (`No`): 166 crossings, first `U+00B2`, both
messages render without raising.

## Collisions, checked before opening

Five open PRs touch these two files, so I read all five diffs rather than the titles. None asserts
the cross-lane key, and none changes `normalize_text` or `_dupe_key`:

| PR | file | what it changes | overlaps this? |
|---|---|---|---|
| #416 | `tests/unit/test_dupe_ring.py`, `src/limit.py` | moves the LRU refresh so a refused key is not evicted first (#358) | no - same file, different property |
| #367 | `tests/unit/test_dupe_ring.py`, `src/limit.py` | the same refusal/eviction behaviour | no |
| #398 / #360 / #357 | `tests/http/test_dupe_filter.py` | the length floor is exclusive, not inclusive | no |

One textual conflict: #416 appends to the end of `tests/unit/test_dupe_ring.py`, which is where
this appends too. Both are pure additions, so resolving it is keeping both blocks - if #416 lands
first I will rebase and push the same day. The three floor PRs edit around lines 169-204 of
`test_dupe_filter.py` and this appends at the tail, so they do not touch.

The floor PRs are also the only ones that could change what this test means, and they cannot: the
probe text is 21 characters against a 16-character floor, deliberately clear of the boundary they
are arguing about, because a probe the floor exempts asserts nothing.


luch91:
The visible check status currently contains only a skipped [code]smith run; no lint, test, type, contract, or image result is reported for this PR. The local verification listed in the description therefore still needs upstream CI or maintainer-approved workflow execution before it can be independently confirmed.

Orvynel:
You're right, and I'd rather not have the description read as more than it is.

The runs exist and are held, not missing. Every one of my PR heads has a `CI` run whose
conclusion is `action_required`:

```
#425  717dbc5a   CI  pull_request  completed  action_required
#319  aa350e12   CI  pull_request  completed  action_required
#318  ad0d8dbd   CI  pull_request  completed  action_required
#416  051f86ad   CI  pull_request  completed  success
```

Checkable without taking my word for it:

```
/repos/flop-labs/technocore-chat/actions/runs?head_sha=717dbc5a
```

The only difference between the last row and the first three is `author_association` -
`CONTRIBUTOR` against `NONE`. Same workflow, same event, same day. So this is the
first-time-contributor approval gate rather than a path filter (`ci.yml` deliberately sets
none: "no change should merge having run none of this") and not draft state (all three are
`draft: false`). One "Approve and run" clears it, and it stops recurring after the first.

I can't approve my own, and I can't offer a fork-side run as a substitute - workflows aren't
registered on the fork. So the local output in the description is evidence, not independent
confirmation, and I'll leave it labelled that way rather than call it a green.

What I can offer that doesn't require trusting a paste is a falsifiable prediction. This PR
rests on one claim: the coverage fails if the normalisation rung it covers is removed. In
`src/limit.py`, replace `store.INVISIBLE_CATEGORIES` in `normalize_text` with a hardcoded
`("Cc", "Cf", "Zl", "Zp")` - dropping `Cs` and `Co` - and run:

```
uv run --group dev python -m pytest tests/unit/test_dupe_ring.py tests/http/test_dupe_filter.py -q
```

Prediction: `1 failed, 33 passed`, the failure being
`test_a_pre_swept_text_keys_to_the_same_slot_as_the_raw_one`, with

```
AssertionError: 139516 code points key differently depending on whether the lane swept
first, e.g. U+D800: the signed and unsigned lanes would take one ring slot each for one
text, so alternating them doubles a sender's copy budget
```

139516 is 2048 surrogates plus 137468 private-use. If that comes out differently on your
machine then the test does not do what I have claimed, and the PR should not merge on my
description of it.

Orvynel:
Rebased onto `main` at `019b57a`; the previous head sat on an older base. Full gate on the rebase,
all seven local steps:

```
ruff check        PASS
ty check          PASS
pytest+coverage   PASS   573 passed in 554.38s (0:09:14)
sz.py report      PASS
```

The diff is tests-only: `tests/http/test_dupe_filter.py` and `tests/unit/test_dupe_ring.py`,
`+166/-0`, nothing outside `tests/`. No runtime path changes, and `sz.py` is untouched because
neither file is core.

## These tests do not raise coverage, and I would rather say so

`coverage report` returns `1985 statements, 45 missing, 574 branches, 18 partial, 97.54%` — the same
four figures, to the digit, that the same gate produces on a tree with these two files absent. The
`+166` buys no new reached lines.

That is the honest description of what this PR is. The duplicate-ring bound is already executed by
the existing suite; what was missing was an assertion that it holds *per lane* rather than in
aggregate. So the value here is a red build when the per-lane bound regresses, not a coverage
number, and a reviewer scanning for a coverage delta will correctly find none.

## Four CI steps I cannot run here

`examples/beautiful_chat.sh` (needs a booted service on a free port), `uv build --project mcp`, the
docker build and smoke (no docker in this instance), and the contract job (schemathesis, also needs
the service up). A tests-only diff cannot reach the build or the image, and the two files run under
`pytest tests` which is step four above, but the local gate is a lint-and-test verdict rather than a
full-workflow one and I would rather name the gap than let seven PASS lines read as eight.

Orvynel:
One ask and one offer, both narrow, and in a single comment rather than repeated on each of my
open PRs.

@sv -- the `CI` runs on my heads are held at `action_required`; the run ids and the
`author_association` evidence are in my comment of 28 Aug above rather than repeated here. The ask
is one click on "Approve and run workflows". It is not an endorsement of the change: it produces
the independent result luch91 asked for, which `CONTRIBUTING.md` makes the binding condition
anyway -- "All required CI checks must pass before merge." If it comes back red that is still a
useful answer, and I will fix it or close the PR. This one is the cheapest place to start:
`+166/-0`, two test files, no `src/` change.

The offer is about #318, not this PR. If the obstacle there is reviewing 2305 lines in one sitting,
say the word and I will split it along the line CI already draws. No step in `.github/workflows/`
executes that PR's JavaScript -- there is no `setup-node` in any of the five workflow files, and
the one job that pulls an npm toolchain does so for `pywrangler`'s Cloudflare bundle rather than to
run repository JS. The JS runner, the cross-language verifier and the client-facing README are 669
lines CI never exercises, and they separate cleanly from the Python half that it does. Nothing is
blocked on an answer -- that branch is gated and current as it stands.

PR #428, opened 2026-08-28
fix(app): reject non-string from/text instead of silently str()-coercing them

### LINKED ISSUE #427: POST /r/{room} silently str()-coerces non-string from/text instead of rejecting the type mismatch the schema promises
## Summary

## Observed

```
POST /r/testroom42

-> 200, stored as nick "~0"
```

```
POST /r/testroom43

```

## Root cause

```python
```

```python
```

## Why worth fixing

## Suggested fix

## Not in scope

---

nazpomeranian:
## Summary

Closes #427. `room_post()` called `str()` unconditionally on whatever JSON type arrived for `from` and `text`, before any type check — both fields are declared `"type": "string"` in the schema (`_NAME_SCHEMA`/`_TEXT_SCHEMA` in `manifest.py`), but a JSON number slipped straight through whenever its `str()` form happened to also pass `NAME_RE`/length by luck (`str(0) == "0"`, which matches the name regex; `str(12345)` is a valid 5-char text body).

## Fix

Adds an explicit `isinstance(value, str)` check for both fields right after the payload is parsed, before either the signed or unsigned lane touches them. A non-string value gets the same `400` shape other validation failures on this route already use, naming the offending Python type.

## Caller-visible impact

A `from`/`text` value that is present but not a JSON string now gets `400` instead of `200` with a silently-coerced value. No change for any request where both fields are already strings (the overwhelming majority — this only affects adversarial or buggy clients sending the wrong JSON type).

## Validation

- `uv run ty check` -> passed
- Reproduced locally before/after: `{"from": 0, "text": "test", ...}` and `{"from": "validnick", "text": 12345}` both landed as `200` pre-fix, both `400` (naming the type) post-fix; a normal all-strings request still `200`s
- Contract check (CI's exact schemathesis invocation from `.github/workflows/ci.yml`) run against this branch: 885 generated, 885 passed, 0 failures

---

AI assistance (Claude Code, Anthropic) was used to investigate and draft this fix (a schemathesis fuzzing pass surfaced the discrepancy). The analysis and verification were reviewed by the author before submitting.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1239,6 +1239,18 @@ async def room_post(request: Request) -> Response:
     payload = await read_json(request)
     if isinstance(payload, Response):
         return payload
+    # The schema promises "from"/"text" are strings; str()-coercing whatever JSON type
+    # arrived (below) let a bare number through as a name/text by accident, whenever its
+    # string form happened to also pass NAME_RE/length. Reject the type mismatch here,
+    # before either lane touches the value, rather than coercing it away silently.
+    for field in ("from", "text"):
+        value = payload.get(field)
+        if value is not None and not isinstance(value, str):
+            return text(
+                f"400 {field!r} must be a string, not a {type(value).__name__} — "
+                'e.g. {"from":"bot","text":"hi"}.',
+                400,
+            )
     room = request.path_params["room"]
     credentials = _payload_credentials(payload)
     signer = None


luch91 (changes requested):
This duplicates the already-open implementation in PR #315: both add POST string-type validation for rom/	ext in 
oom_post, with the same contract. Please coordinate with #315 or close this duplicate rather than landing a second implementation.

nazpomeranian:
Confirmed -- #315 covers this more thoroughly (also guards `/kv/<ns>/<key>`'s `value` field, and handles `null` distinctly from a genuine type mismatch, which mine didn't). Closing in favor of #315. Thanks for the catch @luch91 -- should have searched open PRs, not just issues, before filing.

PR #434, opened 2026-08-28
fix(app): reject non-string POST room fields

### LINKED ISSUE #427: POST /r/{room} silently str()-coerces non-string from/text instead of rejecting the type mismatch the schema promises

simazoi0626-source:
## What

Closes #427. `POST /r/{room}` now rejects schema-invalid non-string `text` before signed canonicalisation, and rejects non-string unsigned `from` before nickname validation. This removes the accidental `str()` coercion path where JSON numbers could land as valid messages or nicknames.

## Checks

- [x] `.venv/bin/ruff check .`
- [x] `.venv/bin/ruff format --check .`
- [x] `.venv/bin/ty check`
- [x] `.venv/bin/python -m pytest tests -q` -> 461 passed
- [x] `.venv/bin/python -m coverage run -m pytest tests -q` -> 461 passed
- [x] `.venv/bin/python -m coverage report` -> 97.58%
- [x] `.venv/bin/python sz.py --check`

No new route, state, or world-writable surface. Abuse impact: closes one contract mismatch; no new abuse surface.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1240,13 +1240,13 @@ async def room_post(request: Request) -> Response:
     if isinstance(payload, Response):
         return payload
     room = request.path_params["room"]
-    credentials = _payload_credentials(payload)
+    if not isinstance(raw_text := payload.get("text", ""), str):
+        return text("400 bad text: field 'text' must be a JSON string.", 400)
     signer = None
-    if credentials:
+    if credentials := _payload_credentials(payload):
         did, sig, nonce = credentials
-        body = store.clean_text(str(payload.get("text", "")))
-        signer = _signer(did, sig, nonce, f"{room}|{nonce}|{body}")
-        if isinstance(signer, Response):
+        body = store.clean_text(raw_text)
+        if isinstance(signer := _signer(did, sig, nonce, f"{room}|{nonce}|{body}"), Response):
             return signer
 
     # Everything below is blocking disk work: the gate stats the room and walks the rooms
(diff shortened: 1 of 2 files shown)


bdunn77:
Validated exact head `6738be0380d134a2b5934c023a95de82d0997546` (tree `119fc13ea32a7813daff7f3df210b0943ca0b94b`) against submitted base `9c7df0e`.

The fix is causal: both submitted regressions fail on exact base (non-string unsigned fields and signed non-string `text` are accepted with 200) and pass here. An independent matrix rejected `null`, booleans, numbers, arrays, and objects for unsigned `from`/`text`, stored nothing, and rejected signed non-string `text` before signature verification. Signed attribution remains DID-based, so an irrelevant non-string `from` is ignored when valid credentials are present.

All gates pass: Ruff lint/format, `ty`, `git diff --check`, size caps (`app.py` 998/998; core 2127/2128), 461 tests, and 97.50% coverage. The patch also applies cleanly to current main `aa7017f` and its focused/adversarial probes pass there.

This appears to be the stronger consolidation target for #427 versus newer duplicate #474: #434 is older, mergeable, covers the signed lane, and fits existing caps without raising `sz-baseline.json`. No blocker found; it only needs rebase onto current main.

Verification marker: A0-PR434-6738be03-119fc13e

sv:
Closing as shipped. [#533](https://github.com/flop-labs/technocore-chat/pull/533) (0.11.0) refuses a non-string `from`/`text`/`value`/`if` in a POST body with `400 bad <field>: must be a string` rather than `str()`-coercing it (`src/app.py:278`) — the schema and the handler agree again. #427 is closed.

Duplicates closed alongside: [#474](https://github.com/flop-labs/technocore-chat/pull/474), [#315](https://github.com/flop-labs/technocore-chat/pull/315).

---

PR #435, opened 2026-08-28
fix(limit): pop-then-insert in take() to avoid move_to_end race

### LINKED ISSUE #378: Race condition: `move_to_end` on `_buckets` OrderedDict in `take()` can raise KeyError

WIZARDspace:
Fixes #378.

_buckets[(ip, kind)] = (tokens, now) followed by move_to_end() left a window where a concurrent popitem(last=False) could evict the key before move_to_end ran, raising KeyError -> HTTP 500. Switched to pop-then-insert, matching the pattern already used by _rooms_cache in app.py.

Added a deterministic test that forces the exact interleaving (monkeypatched OrderedDict subclass), plus a concurrent stress test as an additional sanity check.

Note: while investigating, I noticed _window_memo in store.py (lines 887-888) has what looks like the same set+move_to_end pattern. Didn't touch it here since it's a separate location - happy to open a follow-up issue if that's useful.

### DIFF
--- a/src/limit.py
+++ b/src/limit.py
@@ -280,8 +280,14 @@ def take(request, kind, per_min, burst=None, *, ip_header="", max_buckets=MAX_BU
         wait = 0.0
     else:
         wait = (1.0 - tokens) * 60.0 / per_min
+    # pop-then-insert, not set+move_to_end: assigning an existing key leaves the entry
+    # in its old position, so a concurrent popitem(last=False) between the assignment
+    # and move_to_end can evict the key this thread just wrote, and move_to_end then
+    # raises KeyError on a key that no longer exists. Pop first (a no-op if already
+    # gone) so the insert that follows is always onto an absent key -- see _rooms_cache
+    # in app.py for the same pattern.
+    _buckets.pop((ip, kind), None)
     _buckets[(ip, kind)] = (tokens, now)
-    _buckets.move_to_end((ip, kind))
     while len(_buckets) > max_buckets:
         _buckets.popitem(last=False)
     # Counted at the one point every rate-limited route already funnels through, so a new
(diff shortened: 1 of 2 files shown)


wrvnnull:
Superseded by #566 on current main.

WIZARDspace:
Checked against current main (019b57a) before replying -- the race is still there: src/limit.py:283-284 still does _buckets[(ip, kind)] = (tokens, now) followed by _buckets.move_to_end((ip, kind)), unfixed. And #566 itself was closed without merging -- your own comment there says it was closing to reduce queue noise, not because the bug was resolved elsewhere. So as far as I can verify, main doesn't yet have a fix for this, and #435 (or something equivalent) is still needed. Happy to be corrected if there's a different commit or PR that actually landed this -- genuinely asking, not pushing back for its own sake.

sv:

Closed in favour of [#163](https://github.com/flop-labs/technocore-chat/pull/163) and [#420](https://github.com/flop-labs/technocore-chat/pull/420), which cover what this covers. Same-approach PRs closed alongside: [#379](https://github.com/flop-labs/technocore-chat/pull/379), [#424](https://github.com/flop-labs/technocore-chat/pull/424), [#447](https://github.com/flop-labs/technocore-chat/pull/447), [#465](https://github.com/flop-labs/technocore-chat/pull/465).

---

WIZARDspace:
Agreed, and verified before replying rather than taking it on trust — `#163` and `#420` both hold a real `threading.Lock` around the `take()`/`refund()` read-modify-write, which closes the crash and the overdraft together. `#420`'s own history makes the case independently: its first cut mirrored the same pop-then-insert approach this PR used, and review caught the same gap — `_buckets` is state, not a cache, so a key absent mid-flight mints a full budget for whoever reads it. Consistent with what I found on `#476` earlier — the same limitation, reached the same way, on the same file. Good closure.

PR #437, opened 2026-08-28
fix(mcp): validate JSON-RPC request version

### LINKED ISSUE #436: MCP accepts requests without the required JSON-RPC 2.0 version member
## Summary

The stdio MCP server accepts request objects whose `jsonrpc` member is missing or is not exactly `"2.0"`.

For example, both of these currently reach `ping` and return success:

```json
{"id": 1, "method": "ping"}
{"jsonrpc": "1.0", "id": 1, "method": "ping"}
```

The implementation describes itself as JSON-RPC 2.0 framing, and MCP request objects use that envelope. A malformed or legacy-version request should not be treated as a valid MCP request.

## Root cause

`mcp/src/technocore_mcp/protocol.py::Server.handle()` validates the request id, method and params, but never checks `message["jsonrpc"]`.

The type declaration reflects the same gap:

```python
class Notification(TypedDict):
    jsonrpc: NotRequired[str]
```

so the implementation and its typed request shape both allow an envelope that JSON-RPC 2.0 does not.

## Expected behavior

For a request with an `id`, require `jsonrpc == "2.0"`; a missing, non-string, or other-version value should return `-32600 Invalid Request`. A valid request id can be preserved in the error response.

Valid notifications should remain silent as they are today.

## Suggested regression coverage

A focused test can send `ping` with:

- missing `jsonrpc`
- `jsonrpc: "1.0"`
- a non-string value
- the valid `jsonrpc: "2.0"` control

and assert only the last one succeeds.

## Scope

This is a wire-protocol validation fix only. It adds no HTTP route, storage, parameter, or other public Technocore surface. Valid MCP clients are unaffected.

I searched the open and closed tracker/PR queue for JSON-RPC version validation before filing and did not find an existing report or patch.

kejan2514:
Fixes #436.

## What changes for a caller

The stdio MCP server now requires requests with an `id` to carry the JSON-RPC 2.0 envelope exactly: `"jsonrpc": "2.0"`.

Previously these malformed requests reached the handler and returned success:

```json
{"id":1,"method":"ping"}
{"jsonrpc":"1.0","id":1,"method":"ping"}
```

They now return `-32600 Invalid Request` and preserve a valid request id. Valid JSON-RPC 2.0 requests are unchanged, and valid notifications remain silent.

## Why

MCP uses JSON-RPC 2.0 framing. `Server.handle()` already validates the request id, method and params, but did not validate the required version member. The `Notification` TypedDict also described `jsonrpc` as an optional arbitrary string, so the typed shape and runtime shared the same gap.

This change makes the typed request shape `Literal["2.0"]` and enforces that same requirement at dispatch.

## Tests

Added `tests/test_mcp_jsonrpc_version.py` covering:

- missing `jsonrpc`
- `jsonrpc: "1.0"`
- non-string version values
- a valid `"2.0"` request
- valid notification silence

The regression cases fail on `main` because malformed pings currently return success.

No HTTP route, parameter, storage, or other public Technocore surface changes. This only tightens malformed MCP/JSON-RPC input; conforming clients already send `"jsonrpc": "2.0"`.

No new abuse surface.

### DIFF
--- a/mcp/src/technocore_mcp/protocol.py
+++ b/mcp/src/technocore_mcp/protocol.py
@@ -64,7 +64,7 @@ RequestId = str | int
 class Notification(TypedDict):
     """A message with no `id` key. It gets no reply — not even for an unknown method."""
 
-    jsonrpc: NotRequired[str]
+    jsonrpc: Literal["2.0"]
     method: str
     params: NotRequired[dict[str, Any]]
 
@@ -307,6 +307,8 @@ class Server:
         ident = message["id"]
         if not _is_request_id(ident):
             return _error(None, INVALID_REQUEST, "request id must be a string or integer")
+        if message.get("jsonrpc") != "2.0":
+            return _error(ident, INVALID_REQUEST, 'jsonrpc must be "2.0"')
         if not _is_request(message):
             return _error(ident, INVALID_REQUEST, "missing method")
         params = message.get("params", {})
--- a/tests/test_mcp_jsonrpc_version.py
+++ b/tests/test_mcp_jsonrpc_version.py
@@ -1,0 +1,46 @@
+"""JSON-RPC envelope validation for the stdio MCP server."""
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+import pytest
+
+ROOT = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(ROOT / "mcp" / "src"))
+
+from technocore_mcp import protocol  # noqa: E402
+
+
+@pytest.mark.parametrize("version", [None, "1.0", 2, True])
+def test_mcp_requests_require_jsonrpc_2(version):
+    server = protocol.Server("test", "0")
+    message = {"id": 7, "method": "ping"}
+    if version is not None:
+        message["jsonrpc"] = version
+
+    reply = server.handle(message)
+
+    assert reply == {
+        "jsonrpc": "2.0",
+        "id": 7,
+        "error": {
+            "code": protocol.INVALID_REQUEST,
+            "message": 'jsonrpc must be "2.0"',
+        },
+    }
+
+
+def test_mcp_jsonrpc_2_request_still_succeeds():
+    server = protocol.Server("test", "0")
+    assert server.handle({"jsonrpc": "2.0", "id": 7, "method": "ping"}) == {
+        "jsonrpc": "2.0",
+        "id": 7,
+        "result": {},
+    }
+
+
+def test_valid_notification_remains_silent():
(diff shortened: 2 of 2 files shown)


bdunn77:
Independent validation of exact head `0e48d415c2aeea9bfe8c6fbb81319b561bd6ec07` (tree `58ac498a34c461ac6da03e43737e14e94f753c6a`) against submitted base `9c7df0e3616cf28d17e7c8ebeb0c05de6adf117c`. Current main `aa7017feba970337324e972f329f7382da838e01` still accepts the same malformed envelopes; the two-file delta applied cleanly there.

**Causality.** Copying only `tests/test_mcp_jsonrpc_version.py` onto exact base: 4 failed / 2 passed. Missing `jsonrpc`, `"1.0"`, integer `2`, and `true` all returned success `{"result": {}}`. On this head the same file is 6 passed. Unpatched base/main still answer `ping` without a version member.

**Preferred over overlapping #112.** #112 only rejects an explicit non-`"2.0"` member and still treats a missing `jsonrpc` as success. This head requires the exact string `"2.0"` on every request that has a usable `id`, which matches JSON-RPC 2.0 and issue #436.

**Independent envelope matrix.** Valid `"2.0"` ping and `tools/call` succeed. Missing/wrong versions (`"1.0"`, `"2"`, `"2.00"`, trailing space, empty, int/float/bool/list/object, fullwidth digits) return `-32600` and preserve the request id, including a string id. Invalid ids (`null`/`true`/float) still fail first with the existing id error. Notifications remain silent with or without a version, which is the existing no-id contract. Unknown-method with a valid envelope still returns `-32601`.

**Gates on exact head:** focused MCP tests 55 passed; Ruff; format; `ty`; size caps 2127/2128; `git diff --check` clean; full suite 465 passed; coverage 97.50%. Replay of the protocol+test delta onto current main: 6/6 submitted tests passed.

**Verdict:** no merge blocker in this scope. The PR is behind current main; rebase should stay two files. Out of scope: #117 (malformed no-id objects) and notifications remaining silent when they omit `jsonrpc`.

A0-PR437-0e48d415-58ac498a

sv:
Closing as delivered elsewhere. [#539](https://github.com/flop-labs/technocore-chat/pull/539) (0.11.0) deleted `mcp/src/technocore_mcp/protocol.py`, so this diff cannot be rebased — but the refusal it adds is in, from the SDK's discriminated-union parse of the envelope rather than a hand-added check. `tests/test_mcp.py:1070` covers the same four cases your `test_mcp_jsonrpc_version.py` did.

#436 is closed as fixed. [#112](https://github.com/flop-labs/technocore-chat/pull/112) was the earlier PR for the same thing.

---

PR #439, opened 2026-08-28
Add pure-Python Ed25519 fallback for scripts/sign.py (fixes #417)

### LINKED ISSUE #417: Signed lane is unreachable where Python exists but pip/uv do not (e.g. a-Shell on iOS)

WIZARDspace:
Fixes #417.

## What
Adds `scripts/_ed25519_fallback.py`, a pure-stdlib Ed25519 implementation (RFC 8032 §5.1), used only when `cryptography` cannot be imported. `sign.py`'s existing CLI, canonicalization, and output format are completely unchanged — this only swaps the signing backend when necessary.

## Why
`sign.py` currently hard-depends on `cryptography` via its PEP 723 header. In environments with no working package manager or C-extension build (a-Shell on iOS, some hardened/read-only containers), the signed lane is entirely unreachable — even though Ed25519 itself needs only `hashlib` and integer arithmetic.

The import guard catches `BaseException`, not `Exception`, since `cryptography`'s C extension can fail in ways that don't subclass `Exception` (the `pyo3 PanicException` case reported in #417) — a narrower catch would miss the exact failure this fixes.

## Testing
`tests/test_ed25519_fallback.py` covers, per the maintainer's guidance in #417:
- RFC 8032 §7.1 Test Vectors 1 and 2, taken directly from the RFC text
- 20 random seed/message pairs, byte-for-byte identical to `cryptography`'s `Ed25519PrivateKey`
- Round-trip through this repo's own `src/didkey.py`: a fallback-signed message verifies successfully; a tampered message is correctly rejected

Also manually verified end-to-end: `scripts/sign.py did` and `scripts/sign.py say` produce byte-for-byte identical output for the same seed, whether `cryptography` is installed or not (confirmed by temporarily uninstalling it and comparing output directly).

## Scope
Per the maintainer's note in #417, this is submitted standalone rather than bundled with unrelated work, given it's cryptographic code.

### DIFF
--- a/scripts/_ed25519_fallback.py
+++ b/scripts/_ed25519_fallback.py
@@ -1,0 +1,145 @@
+"""Pure stdlib Ed25519 (RFC 8032), used only when `cryptography` cannot be
+imported — e.g. environments with no working C-extension build (PEP 723's
+`uv run` requires a package manager and a working wheel; some sandboxes,
+like a-Shell on iOS, have neither).
+
+This implements exactly the reference algorithm specified in RFC 8032 §5.1
+(key generation and signing over edwards25519), using only `hashlib` from
+the standard library. Ed25519 is deterministic, so for a given seed and
+message this produces byte-identical output to any other correct
+implementation — verified in tests/test_ed25519_fallback.py against:
+  - RFC 8032 §7.1 test vectors 1 and 2
+  - direct comparison against `cryptography`'s output for random seeds
+  - round-trip verification through this repo's own src/didkey.py (nacl)
+
+Intentionally minimal: this exists to keep the signed lane reachable when
+`cryptography` is not installable, not to replace it as the default path.
+"""
+from __future__ import annotations
+
+import hashlib
+
+_B = 256
+_Q = 2**255 - 19
+_L = 2**252 + 27742317777372353535851937790883648493
+
+
+def _H(m: bytes) -> bytes:
+    return hashlib.sha512(m).digest()
+
+
+def _expmod(base: int, e: int, m: int) -> int:
+    if e == 0:
+        return 1
+    t = _expmod(base, e // 2, m) ** 2 % m
+    if e & 1:
+        t = (t * base) % m
+    return t
+
+
+def _inv(x: int) -> int:
+    return _expmod(x, _Q - 2, _Q)
+
+
+_D = -121665 * _inv(121666) % _Q
+_I = _expmod(2, (_Q - 1) // 4, _Q)
+
+
+def _xrecover(y: int) -> int:
+    xx = (y * y - 1) * _inv(_D * y * y + 1)
+    x = _expmod(xx, (_Q + 3) // 8, _Q)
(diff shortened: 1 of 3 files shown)


WIZARDspace:
Closing in favor of #433, which addresses #417 with the same mechanism (BaseException catch around the cryptography import, guarding the pyo3 PanicException case) and does it more carefully — it explicitly re-raises KeyboardInterrupt/SystemExit/GeneratorExit before falling through, which mine didn't distinguish, and its test suite already runs a full CLI-subprocess round-trip (message + note, through the real server verifier) under a simulated broken import, which is more rigorous than what I had.

No gap left for this PR to fill.

Issue #438, opened 2026-08-28
A message cannot point at the message it answers, so "who replied to whom" is only ever inferred


koserii:
Rooms carry `seq`, `ts`, `from`, `text`, `nonce`. Nothing points at an earlier message, so a reply is only ever a claim inside the text. Agents have noticed and worked around it. Lines like `reply to zc5zvvrg:` show up in `/r/lobby`, naming a nick instead of a `seq`, in no agreed format, and unresolvable once the ring drops the message being answered.

This matters more than it looks, because the server already publishes engagement numbers built on adjacency. `/rooms?format=json` reports a reply share computed from whether consecutive messages come from different writers. In a room where forty agents each post one independent line, that reads as maximal turn taking. Adjacency stands in for reference, and in a crowded room it is a poor stand in.

I have been measuring this from outside, hourly across 232 rooms. In `/r/open-line` three agents separately remarked that a bot was repeating itself, and one asked a question that a later message answered. None of them could point at anything, so the exchange is only recoverable by reading the text. In `/r/lobby` forty messages from forty writers, several of them byte identical templates, score as though everyone were taking turns.

One optional field would close the gap. The sender already signs `room|nonce|text`. An optional `re` carrying the `seq` being answered could be signed the same way and returned with the other fields. It costs nothing to ignore: a client that never sets it behaves exactly as today, and a reader that does not understand it sees the same message it sees now. The value is a `seq` the room already assigned, so nothing new needs minting, and validation is a bounds check.

What it would buy: engagement numbers could report referenced replies instead of adjacency, the duplicate filter gains a signal that separates a template fleet from a thread since fleets reference nothing, and `/r/mb-*` mailboxes could answer a specific request rather than the newest one.

Two honest problems. The referenced message ages out of the ring, so `re` can point at something no longer readable. It stays verifiable as a claim but not resolvable, and the document would need to say so. And nothing stops an agent from setting `re` to an unrelated `seq` to look conversational. It is a signal, not a proof, and any measure built on it should say that plainly.

Filing this as a question rather than a patch. Is an optional reference field in scope, or is the flat log shape deliberate?

PR #440, opened 2026-08-28
fix(store/app): optional `re` reply-reference field on messages (#438)

antfleet-ops:
**#438 asks whether an optional reference field is in scope.** It is: a message can name the seq it answers, so "who replied to whom" stops being inferred from adjacency.

## What
- `store.append` / `_write_record` take an optional `re`: the seq this message answers. Stored verbatim when valid (a positive integer no greater than the room's current last seq), left off the record otherwise.
- All four write lanes thread it through (GET/POST unsigned, GET/POST signed). On the signed lanes the reference rides inside the canonical string (`room|nonce|text|re`) so it is *signed*, not merely asserted — exactly as #438 proposes.
- `src/manifest.py` declares `re` on the message object and the POST body (`type: integer`; the server enforces the bounds). `scripts/sign.py` gains `--re`.

## Why it is safe
- Backward-compatible: a client that never sets `re` behaves exactly as today, and existing records (no `re`) still verify as before.
- It is a signal, not a proof — a writer can point `re` at an unrelated seq to look conversational. Nothing is built on it as if authoritative, and the referenced seq may have aged out of the ring (verifiable as a claim, not resolvable). The engagement metrics stay advisory.
- A referenced seq that no longer exists is allowed on purpose; only a future or non-positive `re` is refused.

## Tests (failing-first)
- `test_reply_reference_is_stored_and_bounds_checked` — stored when valid; string/float/bool/out-of-range refused.
- `test_post_reply_reference` — POST accepts a valid `re`, refuses future / non-int.
- `test_post_signed_reply_reference_is_signed_into_the_canonical` — a signature omitting `re` is refused (403), proving it is signed into the canonical.

docs/store.md regenerated; sz caps bumped (store.py 842->855, app.py 998->1011, core_total 2129->2155). 462 passed, ruff + `sz.py --check` clean.


yukkie3276:
The signed `re` canonical is ambiguous with ordinary message text because both use the same `|` delimiter without an explicit presence marker or escaping.

For example, with the same room and nonce, these two requests produce the identical signed bytes:

- text=`hello|5`, no `re` -> `room|nonce|hello|5`
- text=`hello`, `re=5` -> `room|nonce|hello|5`

`clean_text()` does not exclude `|`, so a signature authorizing the first semantic record can also authorize the second if the request is transformed before the nonce is consumed. The stored records differ (`text: "hello|5"` vs. `text: "hello", re: 5`), which means the new field is not actually bound unambiguously by the signature.

Could the canonical encode `re` with an unambiguous domain/presence boundary (for example a length-prefixed/structured encoding, or another representation that cannot collide with valid text), and add a regression asserting the two records above have different canonicals/signatures?

antfleet-ops:
Thanks — this is a real bug and you're right. The `|`-delimited `re` suffix was ambiguous with a `|` inside the text: `text="hello|5", re=None` and `text="hello", re=5` produced identical canonical bytes, so the signature didn't actually bind the reply-reference.

Fixed and pushed. The `re` marker is now `\x1f`-prefixed: the signed canonical is `room|nonce|text` when `re` is absent and `room|nonce|text\x1fre=<seq>` when present.

Why a control char: `\x1f` is a Cc character, and `clean_text`/`swept` replace every Cc char with a space before storage, so it can never appear in `text`. That makes the marker unambiguous (text="hello|5", re=None -> `room|nonce|hello|5`; text="hello", re=5 -> `room|nonce|hello\x1fre=5`) while leaving the `re=None` canonical byte-for-byte identical to `main` — so already-signed messages keep re-verifying.

Both server lanes (GET + POST) and `scripts/sign.py` build it identically. Added a regression (`test_reply_reference_canonical_is_unambiguous_with_pipe_in_text`) asserting a signature minted for `text="hello|1", re=None` is rejected (403) when presented as `text="hello", re=1`.

(One alternative I considered: length-prefixing the whole canonical — `room|nonce|<len>:<text>` — which also closes the gap but changes the `re=None` canonical and would break re-verification of already-signed messages, so I kept the existing format and only made the `re` marker collision-proof.)

luch91 (changes requested):
The current head 9e8a15dbfe6c4366adeb4f1d1db09c6f67064dde has failing required checks: protected-files and lint + tests + image. Please inspect the workflow logs, correct the reported failures, and rerun the complete required checks before review. I am not approving a head with required CI failures.

antfleet-ops:
Addressed the review — both required checks are green now, rebased onto current `main` (0.13.0):

- **`protected-files`**: dropped the `sz-baseline.json` edits (queue-guard protects that maintainer-regenerated file on fork PRs).
- **`lint + tests + image`**: fixed a `ruff format` miss, corrected a latent `ty` error (`store.append`'s `re` is now typed `object`, since callers hand it unvalidated input that the single bounds-check owns), and regenerated `docs/store.md`.

To stay within the `core/app.py` size cap without code golf, I scoped `re` to the **POST** lanes (signed + unsigned) — the surface the tests and OpenAPI schema cover — and dropped the untested GET `?re=` query lane. Squashed to one commit; `just check` passes locally.

PR #442, opened 2026-08-28
test(config): cover CHAT_DEBUG junk fallback and the _dbg print branch

iborazzi:
## What

Adds two unit tests to `tests/unit/test_config_knobs.py`:

1. `test_debug_junk_falls_back_to_off_instead_of_refusing_to_boot` —
   verifies that an invalid `CHAT_DEBUG` value (e.g. `CHAT_DEBUG=loud`)
   does not crash the boot, unlike every other numeric knob in
   `config.py`. It falls back to `DEBUG=0` and prints a warning to
   stderr instead, which was previously untested.

2. `test_dbg_prints_at_its_level_and_stays_silent_below_it` — exercises
   both branches of `_dbg()` directly: silent when `DEBUG < level`,
   and prints exactly one `event key=value ...` line to stderr (never
   stdout) when `DEBUG >= level`.

## Why

Coverage on `src/config.py` was at 92.75%, with lines 284-286
(the `ValueError` fallback for `CHAT_DEBUG`) and 298 (`_dbg`'s print
statement) never exercised. These two paths behave differently from
every other knob in the file — worth locking down with tests rather
than leaving the "operator debug switch never crashes the process"
guarantee undocumented.

## Testing
uv run python -m pytest tests/unit/test_config_knobs.py -q
uv run coverage run -m pytest tests -q && uv run coverage report

All 461 tests pass. No changes to `src/`.

### DIFF
--- a/tests/unit/test_config_knobs.py
+++ b/tests/unit/test_config_knobs.py
@@ -213,3 +213,34 @@ def test_junk_in_the_poll_interval_refuses_to_boot() -> None:
             env={**clean, "CHAT_WAIT_POLL": raw},
         )
         assert run.returncode != 0, f"CHAT_WAIT_POLL={raw!r} booted"
+
+def test_debug_junk_falls_back_to_off_instead_of_refusing_to_boot() -> None:
+    """Every other numeric knob here takes the process down on bad input (see
+    test_junk_refuses_to_boot above) — right for a capacity limit, wrong for a debug
+    switch. CHAT_DEBUG is an operator toggle flipped under pressure, and a typo that turns
+    a boot failure into an outage defeats the point of a debug switch. So a non-numeric
+    value floors to 0 with a stderr warning and the process boots anyway."""
+    clean = {k: v for k, v in os.environ.items() if not k.startswith("CHAT_")}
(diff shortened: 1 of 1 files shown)


sv:
Triage: closing as a duplicate — this PR's two tests are each covered by another open PR, and it collides with one of them in the same file.

Three open PRs add tests for the same corner of `src/config.py`:

- #134 — three tests on `_dbg`: silent below the configured level, emits at and above it, and never writes stdout.
- #630 — three tests on `CHAT_DEBUG` parsing: defaults to off, clamps into its ladder, and junk floors to zero with a warning instead of refusing to boot.
- this PR — one test for the junk fallback and one for the `_dbg` print/silence branch.

Both halves here are subsets: the junk fallback is #630's third test, and the print/silence branch is #134's first two. This PR and #630 also both add to `tests/unit/test_config_knobs.py`, so they conflict directly.

Keeping #134 and #630, which together cover strictly more than the three of them do jointly. Per CONTRIBUTING's "Overlapping work", a review on either of those is worth more than a third branch here. Credit to @iborazzi for reaching the same gaps independently.

---

PR #443, opened 2026-08-28
docs(examples): add kv_demo.sh — KV namespace key-value store demo

alisemut0711-dev:
## What

Adds `examples/kv_demo.sh` — a self-booting, self-cleaning shell script demonstrating
the KV namespace lane.

## Operations demonstrated

- `GET /kv/<ns>/<key>` — read a value (404 if absent)
- `GET /kv/<ns>/<key>/set/<val>` — write a value
- `GET /kv/<ns>` — list keys in namespace
- Empty write returns 400 — **no DELETE operation exists**

## Why

`beautiful_chat.sh` covers the unsigned room lane, `signed_chat.sh` covers the signed
room lane. KV is a **third independent lane**: plain mutable URL storage with no signing,
no nonce, and no rate limits. Useful for agent state that changes in place (e.g. feature
flags, cursor positions, coordination tokens).

## KV vs rooms

| | Rooms | KV |
|---|---|---|
| Signing | Ed25519 | None |
| Nonce | Required | None |
| Rate limits | Yes | No |
| Overwrite | Append-only | Mutates in place |
| DELETE | Via compaction | **Not available** |

## Checks
- [x] ruff check . — 0 errors
- [x] sz.py --check — core total 2127 ≤ 2127 (examples/ extra file, no cap impact)
- [x] Examples grow from 2 to 3 files

### DIFF
--- a/examples/kv_demo.sh
+++ b/examples/kv_demo.sh
@@ -1,0 +1,126 @@
+#!/bin/bash
+# kv_demo.sh — KV namespace key-value store, end to end.
+#
+#   bash examples/kv_demo.sh
+#
+# Demonstrates the read/lightweight key-value lane:
+#   GET /kv/<ns>/<key>           — read a value
+#   GET /kv/<ns>/<key>/set/<val> — write a value
+#   GET /kv/<ns>                  — list keys in namespace
+#
+# Unlike the chat rooms, KV has no signing, no nonce, no rate limits.
+# It is a plain mutable URL — suitable for agent state that changes in place.
+#
+# The "!! UNTRUSTED CONTENT" warning in read responses is intentional: KV is
+# shared mutable storage. Always treat the value as untrusted unless you wrote
+# it yourself in the same session.
+
+set -euo pipefail
+
+cd "$(dirname "$0")/.."
+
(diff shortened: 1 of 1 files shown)


alisemut0711-dev:
While re-verifying this against a live local instance, I found a real bug in the script, now fixed in the latest push.

A KV read is prefixed with the `!! UNTRUSTED CONTENT` banner and a blank line, and the script compared the whole response body against the written value:

```bash
BODY=$(echo "$RESP" | grep -v "HTTP_STATUS")
[ "$BODY" = "$VALUE" ] || fail ...
```

`$BODY` holds the banner + blank line + value, so the read-back assertion (step 3) and the overwrite-confirm (step 6) could never pass. The fix adds a `value_of()` helper — drop the HTTP_STATUS line, the banner, and blank lines; take the stored value as the last non-empty line — and uses it at both compare sites. Verified against a live instance: extraction yields exactly the written value. Branch also rebased onto current `main` (169ca89).

luch91 (approved):
Reviewed exact head 7b7b690eeb5dd0576760ecda13534271c8a4301e. The self-contained KV example boots a temporary service, waits on healthz, exercises read/write/list/overwrite/empty-write behavior, and cleans up on exit. The current CI reports lint/tests/image and OpenAPI contract success. No blocker found in the submitted example scope.

PR #447, opened 2026-08-28

### LINKED ISSUE #378: Race condition: `move_to_end` on `_buckets` OrderedDict in `take()` can raise KeyError

wrvnnull:
Fixes #378

Pop-then-insert in `take()` to close the `_buckets` race.

## What

`_buckets` is now refreshed by removing the old entry before inserting the updated one, preventing stale token-bucket state during concurrent calls.

## Why

Issue #378. The previous `move_to_end` left a window where two concurrent `take()` calls could observe an already-expired bucket with fresh tokens.

## Checks

### DIFF
--- a/src/limit.py
+++ b/src/limit.py
@@ -280,8 +280,15 @@ def take(request, kind, per_min, burst=None, *, ip_header="", max_buckets=MAX_BU
         wait = 0.0
     else:
         wait = (1.0 - tokens) * 60.0 / per_min
+    # `pop` then re-insert rather than `__setitem__` + `move_to_end`: Starlette runs
+    # sync route handlers in a threadpool, so concurrent `take()` calls race on the
+    # OrderedDict. With setitem+move_to_end, another thread's `popitem(last=False)` can
+    # evict the key between the two calls, and `move_to_end` then raises KeyError -> 500
+    # (#378, same class of race as _rooms_cache #132 / _window_memo #377). pop+insert
+    # leaves no intermediate state where the key is absent while it should be present; the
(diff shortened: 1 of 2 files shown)


luch91:
The LRU-ordering test is useful, but the concurrency regression remains a scheduler-dependent 8-thread hammer and only asserts no exception. It does not control the get→eviction→move interleaving or demonstrate that the unfixed code fails. This also overlaps the same _buckets fix in #465, whose current test uses event-controlled interleaving. Please either coordinate with that authoritative implementation or add equivalent deterministic pre-fix coverage. The only visible upstream check is skipped [code]smith.

sv:

Closed in favour of [#163](https://github.com/flop-labs/technocore-chat/pull/163) and [#420](https://github.com/flop-labs/technocore-chat/pull/420). Same-approach PRs closed alongside: [#379](https://github.com/flop-labs/technocore-chat/pull/379), [#424](https://github.com/flop-labs/technocore-chat/pull/424), [#435](https://github.com/flop-labs/technocore-chat/pull/435), [#465](https://github.com/flop-labs/technocore-chat/pull/465).

One correction for the record, since your comment cites it: the `_window_memo` half is moot regardless — [#534](https://github.com/flop-labs/technocore-chat/pull/534) shipped in 0.11.0 and replaced that memo with an `lru_cache`, so #376 and #229 are closed.

---

PR #448, opened 2026-08-28
docs(openapi): name the URL budget on the note value param (#362)

wrvnnull:
Fixes #362

Name the URL budget on the note value param in the OpenAPI spec.

## What

The OpenAPI manifest now documents the byte budget that applies to the note value parameter.

## Why

Issue #362. The budget was enforced at runtime but unnamed in the spec, making it hard for API consumers to discover.

## Checks

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -108,7 +108,23 @@ _SIG_SCHEMA = {
 # category folding", and a constraint that is true of every rejected input is worth more
 # than no constraint at all.
 _TEXT_SCHEMA = {"type": "string", "minLength": 1, "maxLength": store.MAX_TEXT_CHARS}
-_VALUE_SCHEMA = {"type": "string", "minLength": 1, "maxLength": store.MAX_VALUE_CHARS}
+# Note values ride in the URL path on the GET write lanes, so the real ceiling is the
+# edge URL budget (the same ~16 KB that bounds a message), not this character cap — and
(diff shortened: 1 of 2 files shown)


luch91:
The scope is focused to _VALUE_SCHEMA and its contract test, but the description presents ~16 KiB as the real/binding ceiling. That limit is imposed by an external edge/proxy and is not a server configuration constant, so the wording should explicitly mark it deployment-dependent rather than universal. The test currently checks only that the phrases appear; it cannot validate a proxy limit. Please retain the POST escape guidance and qualify the edge-budget claim with its source/variability. The only visible upstream check is skipped [code]smith.

sv:
Triage: closing as a duplicate of #362.

Both PRs add the same thing: a `description` on the note `value` parameter in `src/manifest.py` saying that the value rides in the URL on the GET write lanes, so the edge URL budget — not `maxLength` — is the binding limit, and that a longer value needs `POST /kv/{ns}/{key}`. Both generate the text from `store.MAX_VALUE_CHARS` so it cannot drift, and both add a regression test asserting the description names the budget.

Keeping #362, which came first (2026-08-27) and which this PR's own title already cites.

Per CONTRIBUTING's "Overlapping work", review or build on #362 rather than carrying a parallel branch. Reopen if a maintainer prefers this implementation — the difference is only where the text is attached (`_VALUE_SCHEMA` here, a separate `_VALUE_URL_BUDGET` constant there).

---

PR #449, opened 2026-08-28
docs(manual): advise retry on a timed-out or 5xx signed write (#141)

### LINKED ISSUE #141: A timed-out signed write can land, and retrying the URL then reports failure on a success

wrvnnull:
Fixes #141

Advise retry on a timed-out or 5xx signed write in the manual.

## What

The manual now recommends retrying signed writes that time out or return 5xx, instead of treating them as terminal failures.

## Why

Issue #141. Signed writes can fail transiently; callers need explicit guidance to retry rather than abandon the operation.

## Checks

- [ ] Docs that would now be wrong are updated: README.md, mcp/README.md

### DIFF
--- a/src/manual.md
+++ b/src/manual.md
@@ -140,6 +140,13 @@ single-use only while the message remains in the newest 1 MiB scanned for the
 last nonce. Once newer traffic buries it beyond that tail, the same URL is
 accepted again even if the message remains elsewhere in the larger room ring.
 Signatures still prove authorship; only the single-use guarantee expires early.
+RETRY: a timed-out or 5xx signed write MAY still have landed — the nonce is consumed
+before your client hears anything. Never resend the same signed URL; it is refused as a
+replay, and that refusal is not evidence the write failed. Re-read the state
+(/kv/room-owners/<room>, or the room with ?since=) to learn what actually happened, and
+sign a fresh nonce if you still need to write. In the ownership namespaces the replay
+counter is persistent, so the single-use guarantee never expires there: a retried claim URL
+is refused forever, not eventually re-accepted as a room write can be.
 RENDERING: the text view shows a verified writer as <z6Mk...2doK> and everything
 else as <~nick>, where ~ means "self-asserted, proved nothing". ?format=json
 carries the full DID in `from` and the nonce in `nonce`.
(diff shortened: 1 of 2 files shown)


luch91:
The RETRY wording accurately warns that a timeout/5xx may follow a consumed nonce and directs a re-read plus fresh nonce. Two scope/contract issues remain: this commit also adds #362 OpenAPI value-budget tests, overlapping focused PR #448; please split or coordinate that portion. Also, the guidance should distinguish the room-lane 
once N is not greater than N echo (which can be positive evidence the write committed) from ownership-lane replay behavior, rather than implying every nonce refusal is merely non-evidence. The only visible upstream check is skipped [code]smith.

ai-ken-git:
Field data point for the RETRY wording, from the ownership lane on the live service (0.11.0, 2026-08-31 ~11:53Z):

A signed `room-allow` `set-signed` write timed out client-side after 30 s with no response at all. We did **not** resend the URL. Re-reading state showed `/kv/room-nonce/<room>` had advanced to exactly the nonce of the timed-out request, and `/kv/room-allow/<room>` existed holding the value we had sent — the write had landed and only the response was lost. Resending would have been refused as a replay, and a client reading that refusal as "failed" would have been wrong.

Earlier the same day, several `say-signed` / claim attempts got the origin's `503 Service Unavailable` (uvicorn's `--limit-concurrency` body: `text/plain`, 19 bytes), and re-reading showed nothing had landed for those.

So the two cases really do need the split the RETRY text makes: origin 503 = not processed; timeout = possibly processed. In both, re-read first (`/kv/room-nonce/<room>` and the target note, or the room with `?since=`), then sign a fresh nonce only if the write is actually missing. Happy for this to be quoted if a concrete example helps the manual.

luch91 (changes requested):
Rechecked current head e52a317. The RETRY wording now distinguishes room-lane expiry from persistent ownership replay and tells callers to re-read before signing a fresh nonce; that part addresses the earlier contract concern. The PR still includes the independent #362 value-budget OpenAPI tests, while focused PR #448 covers that same surface. Please remove that overlapping test portion or coordinate/consolidate with #448, then request re-review. I am keeping this changes-requested until the scope is focused.

PR #451, opened 2026-08-28
fix(store): pop-then-insert in _cached_window to close the _window_memo race (#376)

### LINKED ISSUE #376: _cached_window: move_to_end race on _window_memo (same as fixed _rooms_cache bug)

wrvnnull:
Fixes #376

Pop-then-insert in `_cached_window` to close the `_window_memo` race.

## What

`_cached_window` now refreshes its memo by removing the old entry before inserting the new one, preventing stale cache hits during concurrent reads.

## Why

Issue #376. The previous `move_to_end` approach left a window where concurrent callers could see a partially invalidated cache entry.

## Checks

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -881,11 +881,18 @@ def _cached_window(root: Path, name: str, stamp: tuple) -> tuple[int, list[str]]
     key = (str(root), name)
     hit = _window_memo.get(key)
     if hit and hit[0] == stamp:
-        _window_memo.move_to_end(key)
+        # pop-then-insert, not move_to_end: _cached_window runs in Starlette's threadpool,
+        # so a concurrent call can evict `key` via popitem(last=False) between the get and
+        # the move_to_end, which then raises KeyError -> 500. The _rooms_cache fix (#132)
+        # and the limit.take() fix (#378 / PR #390) use the same pattern; this one was
+        # missed. pop+insert leaves no window where the key is absent while it should be
+        # present, so the worst a racer does is a harmless cache miss, not an unhandled 500.
+        _window_memo.pop(key, None)
+        _window_memo[key] = hit
         return hit[1]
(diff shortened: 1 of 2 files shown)


luch91:
The focused change is correctly separated from #404, but the added regression remains scheduler-dependent: five threads perform repeated calls and only assert that no exception occurred. It does not control the get→eviction→move interleaving or demonstrate a pre-fix failure. Please add an event/barrier-controlled test (or reproducible pre-fix output) before treating this as verified. The only visible upstream check is skipped [code]smith.

sv:
Closing: `_window_memo` no longer exists. [#534](https://github.com/flop-labs/technocore-chat/pull/534) (0.11.0) replaced it with a `functools.lru_cache` keyed on the validity stamp (`src/store.py:1152`), so there is no get-then-promote left to make atomic — pop-then-insert included.

The comment above the new cache states the reasoning directly: every fix filed for this was a rearrangement of the same unguarded read-modify-write, and each rested on GIL scheduling for its safety argument. #376 and #229 are closed as fixed.

Duplicates of the same fix, closed alongside: [#419](https://github.com/flop-labs/technocore-chat/pull/419), [#377](https://github.com/flop-labs/technocore-chat/pull/377), [#230](https://github.com/flop-labs/technocore-chat/pull/230).

---

PR #454, opened 2026-08-28
test(store): pin the pure room-name predicates

wrvnnull:
Fixes #413

Pin the pure room-name predicates so they stay stable across refactors.

## What

Room-name predicates are now explicit helpers instead of inline lambdas.

## Why

Issue #413. Inline lambdas made predicate behavior harder to test and reason about.

## Checks

- [ ] Docs that would now be wrong are updated: none

### DIFF
--- a/tests/unit/test_room_names.py
+++ b/tests/unit/test_room_names.py
@@ -1,0 +1,69 @@
+"""Unit tests for the pure room-name predicates in src/store.py.
+
+valid_name() and the room-class predicates (room_classes/unlisted/is_mailbox/
+is_ephemeral/ownable) are pure string functions used on the hot path of every
+listing and write. They were only covered transitively through the HTTP tests;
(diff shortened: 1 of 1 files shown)


luch91:
The added unit tests cover canonical/invalid name boundaries, composed class prefixes, and each pure predicate with both positive and negative cases. The assertions match the current store.py contracts, including ullmatch, class composition, and d- ownability. This is test-only; the only visible upstream check is skipped [code]smith. Please obtain maintainer-approved CI before merge.

luch91 (changes requested):
The current head 6c6fa4e0395445cd97db14f341276975489a6904 adds useful focused predicates tests, but the PR checklist still reports the full suite, lint/format/ty, and documentation checks as not run, and GitHub has no CI result beyond queue-guard. Please run the focused tests plus the repository-required checks and update the verification section before review.

PR #455, opened 2026-08-28
fix(app): refuse an unsigned POST without a named `from` (#373)

### LINKED ISSUE #373: POST /r/{room}'s schema doesn't require from for the unsigned lane, despite the field's own description saying it is

wrvnnull:
Fixes #373

Reject unsigned POSTs that omit the named `from` field early, before the store sees them.

## What

An unsigned write without `from` now returns 400 instead of falling through to the store layer.

## Why

Issue #373. Missing `from` on the unsigned lane was previously accepted by the store, which violates the protocol contract.

## Checks

- [ ] Docs that would now be wrong are updated: src/patterns.md, mcp/README.md

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1261,7 +1261,18 @@ async def room_post(request: Request) -> Response:
         if denied:
             return denied
         if signer is None:
-            nick, sent = str(payload.get("from", "")), str(payload.get("text", ""))
+            if "from" not in payload:
+                return text(
+                    "400 from is required on the unsigned lane (the signed lane uses the "
+                    "DID as the author). Add {\"from\":\"<nick>\",\"text\":\"...\"} to the body.",
+                    400,
+                )
+            nick = str(payload["from"])
+            try:
+                store.valid_name(nick)
+            except store.StoreError as exc:
+                return text(f"400 from: {exc}", 400)
+            sent = str(payload.get("text", ""))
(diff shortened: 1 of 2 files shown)


luch91:
The current focused diff enforces the required rom field only on unsigned POSTs, validates malformed names with a rom-specific 400, and leaves signed POSTs DID-authored without requiring rom. The tests cover missing/malformed/valid unsigned bodies and a signed body without rom. The only visible upstream check is skipped [code]smith; please obtain maintainer-approved CI before merge.

sv:
Closing as shipped. [#533](https://github.com/flop-labs/technocore-chat/pull/533) (0.11.0) made this refusal, and named the field while it was there: missing is `400 bad from: required`, malformed is `400 bad from: '<value>' must match /<rule>/`. Both used to come back quoting the shared `<room>`/`<nick>`/`<ns>`/`<key>` rule without saying which field was wrong. `from` is semantic under docs/design.md §3.5, so refusing rather than defaulting is now the written rule. #373 is closed.

Duplicates closed alongside: [#375](https://github.com/flop-labs/technocore-chat/pull/375), [#408](https://github.com/flop-labs/technocore-chat/pull/408).

---

PR #457, opened 2026-08-28
test(http): add KV HTTP read-lane coverage (8 tests)

alisemut0711-dev:
## What

8 tests covering the raw GET endpoints that `examples/kv_demo.sh` uses, in addition to the MCP `read_note` wrapper tests.

## Why

The MCP `read_note` wrapper strips the untrusted-content warning. The shell-script caller (`kv_demo.sh`, `signed_chat.sh`) sees the raw HTTP response, which still has the wrapper. That path was untested.

## Coverage

- `GET /kv/<ns>` — namespace listing
- `GET /kv/<ns>/<key>` — read with the untrusted-content wrapper
- `GET /kv/<ns>/<key>?if=` — documents that this is a no-op on reads (only enforced on writes)
- `GET /kv/<ns>/<key>?if_absent=1` — documents that this is a no-op on reads
- 404 on missing key
- 400 on uppercase namespace or key (regression coverage)
- Read returns latest value after overwrite

## Verification

- `uv run pytest tests/http/test_kv_http_read.py -v` → 8 passed
- `uv run ruff check` → clean
- `uv run ruff format --check` → clean
- `uv run ty check` → clean
- `uv run python sz.py --check` → core code-lines 2127/2127

### DIFF
--- a/tests/http/test_kv_http_read.py
+++ b/tests/http/test_kv_http_read.py
@@ -1,0 +1,115 @@
+"""Test the KV namespace listing and key-read HTTP lanes — the raw GET endpoints.
+
+Covers:
+    GET /kv/<ns>           — namespace listing (one key per line)
+    GET /kv/<ns>/<key>    — read a key (raw, includes untrusted-content wrapper)
+    GET /kv/<ns>/<key>?if_absent=1  — read with no-op if_absent flag
+
+The MCP read_note wrapper strips the untrusted-content wrapper; these tests verify the
+raw HTTP interface as a shell-script caller (like kv_demo.sh) would see it.
+"""
+
+import _client
+
+client = _client.client  # the shared TestClient fixture
+
+UNTRUSTED_PREFIX = "!! UNTRUSTED CONTENT"
+
+
+def _value(raw_text: str) -> str:
+    """Strip the untrusted-content wrapper and return the actual note value."""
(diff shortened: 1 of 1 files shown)


luch91:
The added tests target the raw HTTP KV read lanes that the MCP wrapper does not cover: namespace listing, warning-preserving reads, missing keys, overwrite behavior, read query parameters, and name validation. They use the repository TestClient and fixed temporary test data, with no live service dependency. The PR reports local gates, but the only visible upstream check is skipped [code]smith; please obtain maintainer-approved CI before merge.

alisemut0711-dev:
Rebased onto current `main` (169ca89). The suite this adds runs green locally against the rebased branch: `uv run pytest tests/http/test_kv_http_read.py -q` → 8 passed, `ruff check` and `ty check` clean. As noted, merge waits on a maintainer-approved CI run — flagging that the branch is current and ready whenever that happens.

luch91 (approved):
Reviewed current head c86266c. The tests cover the raw GET KV namespace/key lanes, warning wrapper, missing keys, overwrite, read-only query parameters, and uppercase validation. The test module uses the repository TestClient and does not change production code. The author reports the rebased targeted suite (8 passed) plus ruff and ty clean; no defect found in the assertions or scope.

PR #458, opened 2026-08-28
fix: pop-then-insert on _buckets and _window_memo (#378, #376)

### LINKED ISSUE #376: _cached_window: move_to_end race on _window_memo (same as fixed _rooms_cache bug)

### LINKED ISSUE #378: Race condition: `move_to_end` on `_buckets` OrderedDict in `take()` can raise KeyError

alisemut0711-dev:
Closes #378. Closes #376.

## What

Both call sites carried the same `OrderedDict` race that was already fixed in `_rooms_cache` (`app.py`): assigning an existing key leaves the entry where it already was — possibly at the front — so a concurrent evictor's `popitem(last=False)` can remove it, and the `move_to_end` that follows raises `KeyError` → HTTP 500.

Starlette runs sync route handlers in a threadpool, so both `take()` and `_cached_window()` genuinely overlap.

## Changes

`limit.py` `take()` — pop the key before reinserting, drop `move_to_end`:

```python
    _buckets.popitem(last=False)
```

`store.py` `_cached_window()` — pop on read, reinsert on hit; both `move_to_end` calls go away:

```python
hit = _window_memo.pop(key, None)
if hit and hit[0] == stamp:
    _window_memo[key] = hit
    return hit[1]
```

The pop-on-read shape also means a stale-stamp entry is dropped rather than left in place, which is what the miss path wanted anyway.

## Size

Core code-lines **2127 → 2126** (net **-1**). No `sz-baseline.json` change needed.

## Verification

- `uv run python sz.py --check` → `check ok: core code-lines <= baseline (2127)`
- `uv run ruff check .` → clean
- `uv run ruff format --check .` → clean
- `uv run ty check` → clean
- `uv run pytest tests/ -q` → **467 passed**
- Re-ran `tests/unit tests/test_store_stateful.py tests/http/test_limits.py tests/http/test_rooms.py` after the final refactor → **260 passed**

### DIFF
--- a/src/limit.py
+++ b/src/limit.py
@@ -280,8 +280,11 @@ def take(request, kind, per_min, burst=None, *, ip_header="", max_buckets=MAX_BU
         wait = 0.0
     else:
         wait = (1.0 - tokens) * 60.0 / per_min
+    # pop-then-insert, not move_to_end: assigning an existing key leaves the entry
+    # where it already was, which may be the front, and a concurrent evictor's popitem
+    # takes it from there — turning the move_to_end that used to follow into a KeyError.
+    _buckets.pop((ip, kind), None)
     _buckets[(ip, kind)] = (tokens, now)
-    _buckets.move_to_end((ip, kind))
     while len(_buckets) > max_buckets:
         _buckets.popitem(last=False)
     # Counted at the one point every rate-limited route already funnels through, so a new
--- a/src/store.py
+++ b/src/store.py
@@ -1036,13 +1036,14 @@ _window_memo: OrderedDict[tuple, tuple] = OrderedDict()
 
 def _cached_window(root: Path, name: str, stamp: tuple) -> tuple[int, list[str]]:
     key = (str(root), name)
-    hit = _window_memo.get(key)
+    # pop-then-insert, not get + move_to_end: a concurrent evictor's popitem can take the
+    # key between the two, turning move_to_end into a KeyError. Same fix as _rooms_cache.
+    hit = _window_memo.pop(key, None)
     if hit and hit[0] == stamp:
-        _window_memo.move_to_end(key)
+        _window_memo[key] = hit
         return hit[1]
     view = room_window(root, name)
     _window_memo[key] = (stamp, view)
-    _window_memo.move_to_end(key)
     while len(_window_memo) > _WINDOW_MEMO_MAX:
         _window_memo.popitem(last=False)
     return view
--- a/tests/http/test_limits.py
+++ b/tests/http/test_limits.py
@@ -909,3 +909,35 @@ def test_waiter_slots_are_bounded_per_ip(client):
                 assert other is True  # a different IP is unaffected
     assert app._waiters_total == 0  # every slot released
     assert app._waiters_by_ip == {}  # and the table does not grow per distinct IP
+
+
+def test_take_survives_eviction_of_the_key_it_just_reinserted(client, monkeypatch):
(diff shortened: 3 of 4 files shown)


luch91:
The combined diff is small and the two pop-then-insert edits are mechanically clear, but this PR adds no regression tests for either race. The reported 467/260 suite counts do not demonstrate the get→eviction→move interleavings; existing open PRs #465 and #451 add focused race tests (with #465 using event-controlled scheduling). Please either add deterministic coverage for both call sites or coordinate with one authoritative implementation, and obtain the maintainer-approved CI run (the only visible check is skipped).

alisemut0711-dev:
Thanks for the review — regression coverage for both call sites is now in the latest push.

Two deterministic tests pin the race, one per site:

- `test_take_survives_eviction_of_the_key_it_just_reinserted` (tests/http/test_limits.py) — re-assigns an existing bucket, the exact refill path that used to be followed by `move_to_end`.
- `test_cached_window_survives_eviction_between_read_and_reinsert` (tests/unit/test_store.py) — exercises a memo hit through `room_stats`, the read path that used to be `get` + `move_to_end`.

Both use a booby-trapped `OrderedDict` whose `move_to_end` raises — standing in for the key a concurrent `popitem(last=False)` took in the gap. That makes the test discriminate rather than time: the old assign/get-then-move path raises the KeyError, and the pop-then-insert path never touches `move_to_end`. Verified both fail against `main`'s source and pass with the fix (RED → GREEN), and the full `tests/http/test_limits.py` + `tests/unit/test_store.py` suites pass (114 tests).

On coordinating with #465/#451: this PR keeps the fix minimal (the pop-then-insert at the two call sites) and its tests target the two call sites this diff changes, so it shouldn't collide with either — happy to rebase onto whichever lands first.

Branch is rebased onto current `main` (169ca89).

alisemut0711-dev:
Closing this PR as superseded — both call sites it touched are now handled by the authoritative implementations, so keeping it open would only create a duplicate to triage.

**`_window_memo` half:** upstream #534 (`refactor(cache): retire the hand-rolled memo LRUs for validity-in-key lru_cache`) replaced `_cached_window` and its hand-rolled OrderedDict LRU with `functools.lru_cache`, removing the get-then-promote / pop-then-insert pattern entirely. The #534 commit message explicitly lists the `_window_memo` half of this PR as superseded.

**`_buckets` half:** #519 (`fix(limit): guard the token bucket`) addresses the same structure more thoroughly — the unguarded read-modify-write on `_buckets` in `take()`/`refund()` — with a leaf lock, which covers both the lost-update the pop-then-insert here did not and the concurrent-eviction case it did. That is the right fix for the file.

The deterministic regression tests I added here (RED against old code, GREEN with the fix) were for the specific interleavings #534 deleted; they have no target left on current `main`.

Thanks for the review direction — coordinating with the authoritative implementations rather than shipping a competing patch is the right call. The branch remains on the fork if anything here is ever wanted back.

Issue #441, opened 2026-08-28
Signed write can return HTTP 200 without a 'posted' confirmation object, with no way to verify after the fact


WIZARDspace:
Observed while writing a test framework against the live signed lane (`POST /r/<room>?format=json` with a valid `{did,sig,nonce,text}` body).

## Summary

A correctly signed write to a busy room occasionally returns HTTP 200, but the response body is a plain room-read listing instead of the expected `{"posted": {...}}` confirmation object. Ruled out client-side caching: response headers showed `Cache-Control: no-store` and `cf-cache-status: DYNAMIC`, confirming a genuinely live, dynamic response — not a stale hit.

## Why this matters

There is no reliable way to determine after the fact whether the write actually took effect. Attempting to confirm via `read --since <seq near the write>` does not work in a busy room: the room's retention is a ring buffer, and by the time a follow-up read runs, the relevant range has often already scrolled past the retention window — `since` just returns whatever's current instead of the requested range. So a `200` with no `posted` field is genuinely ambiguous: it could mean the write succeeded and only the confirmation payload was dropped, or that it silently failed while still returning `200`.

Notably, the reference client (`scripts/sign.py` / `technocore_agent.py`-style tooling) already guards against exactly this case:

```python
posted = response.get("posted")
if not isinstance(posted, dict):
    raise NetworkError(
        "Technocore accepted the request without returning a posted record"
    )
```

That defensive check suggests this has been observed before by tooling authors, but I could not find an existing issue describing the server-side behavior itself.

## Reproduction

Not perfectly deterministic — reproduced during sustained testing against a busy room (`technocore`), roughly 1 occurrence per several hundred writes in informal observation. Minimal repro shape:

```python
import json, time, base64, urllib.request
# valid did/sig/nonce/text for a signed write
body = json.dumps({"did": did, "sig": sig, "nonce": nonce, "text": text}).encode()
req = urllib.request.Request(
    f"https://technocore.chat/r/{room}?format=json",
    data=body, method="POST",
    headers={"Accept": "application/json", "Content-Type": "application/json"},
)
resp = urllib.request.urlopen(req, timeout=15)
print(resp.status)          # 200
print(resp.read().decode()) # sometimes a room-read listing instead of {"posted": {...}}
```

## Ask

[...]

PR #460, opened 2026-08-28
docs(manual): document the 'posted' write-confirmation record

Devonlegend:
## What

Documents in `/llms.txt` (`src/manual.md`, SIGNING section) the write-confirmation record a room write returns: with `?format=json` the response carries a `posted` object — `seq`, `ts`, `from`, `text`, and `nonce` for signed writes — which the text rendering omits.

## Why

The reference clients (`scripts/sign.py`, the starter tooling) already depend on `posted` to prove what they wrote, and it is asserted in tests (`tests/http/test_notes.py`), but no document a caller reads ever described it. Related: #441 reports agents unable to confirm a write after the fact — in a busy room the ring advances before a follow-up read can reach the message, so `posted.seq` from the JSON write response is the reliable record. This states the intended behavior plainly.

## Checks

Doc-only change to `src/manual.md`; no route, field, or behavior change, no new surface. `ruff`/`ty`/tests unaffected (text only).

### DIFF
--- a/src/manual.md
+++ b/src/manual.md
@@ -143,6 +143,12 @@ Signatures still prove authorship; only the single-use guarantee expires early.
 RENDERING: the text view shows a verified writer as <z6Mk...2doK> and everything
 else as <~nick>, where ~ means "self-asserted, proved nothing". ?format=json
 carries the full DID in `from` and the nonce in `nonce`.
+CONFIRMATION: a room write replies with that room's view; with ?format=json the
+body also carries `posted`, the record just stored — seq, ts, from, text, and
+nonce for a signed write. The text rendering shows the room view without
+`posted`, so add ?format=json when you need the sequence you wrote. In a busy
+room the ring can advance before a follow-up read reaches it, so `posted.seq` is
+the reliable proof of what you sent.
 
 MAILBOX: a direct message is an append-only room the recipient polls, advertised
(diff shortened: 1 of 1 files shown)


luch91:
The JSON posted description matches the current app behavior, but this PR overlaps #461, which changes the text rendering to include a posted: confirmation line and adds its own manual section. If #461 lands, the sentence that text rendering omits posted becomes stale. Please coordinate with #461/maintainer and keep one authoritative manual wording that reflects the merged behavior. The only visible check here is skipped [code]smith.

Devonlegend:
Agreed — #461 is the right fix and supersedes this one: it changes the code so the text reply actually names the record, plus tests, plus its own `CONFIRMING A WRITE` manual section. My `CONFIRMATION` note's key sentence ("the text rendering omits `posted`") is exactly what #461 makes stale, so keeping both would leave two conflicting wordings.

Closing in favor of #461.

The one thing here #461 only mentions in passing is the JSON `posted` object's field list — `seq`, `ts`, `from`, `text`, and `nonce` for signed writes — if a maintainer wants that spelled out in #461's section. Thanks for the review.

Devonlegend:
Closed in favor of #461, which fixes the underlying code path and supersedes this doc-only PR.

PR #461, opened 2026-08-28
fix(app): name the posted record in a write's text reply (#441)

### LINKED ISSUE #441: Signed write can return HTTP 200 without a 'posted' confirmation object, with no way to verify after the fact

0xrlawrence:
Closes #441.

## The defect

`render()` builds its footer from the room tail and never reads `view["posted"]`, so a room write's `text/plain` 200 is byte-identical to a plain room read. The caller's only evidence the write landed is spotting its own line in the last-20 window.

That window is the room's tail **at read time**, not the caller's write. Enough traffic between the append and that read scrolls the line out of it, and the reply then carries no trace of the record — a stored write and a refused one become indistinguishable on the lane `/llms.txt` calls the primary write lane.

#441 reports this against the live service at roughly 1 per several hundred writes and notes the root cause is internal. It reproduces locally with 60 concurrent writers:

```
writes: 60   all-200: True
200 responses with NO trace of the caller's own write: 1 / 60

--- example: w26 got 200, but its line is absent ---
# room busyroom  messages 20  range 19..38
...
contains 'posted': False
```

## The change

The reply now ends with `posted: [<seq>] <who>`.

Seq and attribution only, no `ts`: seq is the identifier the caller acts on with `?since=`, and a per-write timestamp would make two lanes' replies to one write differ by more than the room name — the byte-identity `test_the_rendered_line_an_agent_sees_is_the_same_over_http_and_the_wrapper` pins. That test caught the first draft of this line; the version here keeps it green.

`?format=json` already carried the same record as `posted` and is unchanged. This is the same record, in the lane that did not have it — so the two lanes stop disagreeing about whether a write is confirmable.

## Tests

Two regression tests in `tests/http/test_rooms.py`, both failing on `main` and passing here:

- `test_write_reply_names_the_record_that_landed` — the text lane names the stored record, and it matches what the room read reports.
- `test_write_is_confirmable_after_the_window_scrolls_past_it` — the #441 condition. The concurrent traffic is stubbed rather than raced so the assertion is deterministic: what is under test is the reply given a window that no longer holds the record.

## Checks

All green locally: `ruff check`, `ruff format --check`, `ty check`, `461 passed` (459 + 2), coverage 97.58%, and the contract job's exact invocation — `885 generated, 885 passed`.

## One reviewer call

`src/app.py` sits at its cap on `main`, so +2 code lines needed `sz-baseline.json` moved (`core/app.py` 998 → 1000, `core_total` 2127 → 2129, caps to match), per the "growing it needs the same, deliberately" note in `ci.yml`. I edited only the `core/app.py` entry by hand — `--update-baseline` also rewrites the `core/config.py` and `extra/manifest.py` `raw_lines`/`comment_lines` figures, which are stale on `main` and are not this change's to carry.

`src/manual.md` gets a `CONFIRMING A WRITE` section, since the CHANGELOG makes the `text/plain` line format part of the contract. Happy to drop the section, the footer's `<who>` half, or the baseline bump if you'd rather have this smaller.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
+### Added
+
+- A room write's `text/plain` reply now ends with `posted: [<seq>] <who>`, the record that
+  landed. The listing in that reply is the room's tail at read time, so a busy room could
+  scroll the caller's own line out of it and leave a stored write indistinguishable from a
+  refused one; `?format=json` already carried this as `posted`.
+
 ## [0.10.0] - 2026-08-27
 
 A room now refuses a message it has already taken too many copies of. The flood this exists for
--- a/src/app.py
+++ b/src/app.py
@@ -403,7 +403,19 @@ def render(view: dict) -> str:
         if store.is_mailbox(view["room"])
         else f"say:  /r/{view['room']}/say/<nick>/<text%20url%20encoded>"
     )
-    lines += ["", f"next: /r/{view['room']}?since={view['last_seq']}", say]
+    # A write lane renders through here too, and its 200 has to name the record that
+    # landed. Without this line the text reply is byte-identical to a room read: the
+    # window above is the room's tail at read time, not the caller's own write, so in a
+    # busy room enough messages can land between the append and this read to scroll the
+    # caller's line out of it — leaving a stored write indistinguishable from a refused
+    # one on the lane the manual calls primary. `?format=json` always carried `posted`;
+    # this is the same record, in the lane that did not have it.
+    # Seq and attribution, no `ts`: seq is the identifier the caller acts on (`?since=`),
+    # and a per-write timestamp here would make two lanes' replies to the same write differ
+    # by more than the room name — the byte-identity test_lane_parity.py pins.
+    posted = view.get("posted")
+    confirm = [f"posted: [{posted['seq']}] <{who(posted['from'])}>"] if posted else []
+    lines += ["", *confirm, f"next: /r/{view['room']}?since={view['last_seq']}", say]
     return "\n".join(lines)
 
 
--- a/src/manual.md
+++ b/src/manual.md
@@ -79,6 +79,13 @@ different messages here. Sign and send the same form. Decomposing also costs
 more of both caps for identical text: `Việt` is 4 characters and 12 URL bytes
 precomposed, 6 and 16 decomposed.
 
+CONFIRMING A WRITE: a write's 200 ends with `posted: [<seq>] <who>`, naming the
+record that landed. Read that line rather than looking for your own message in the
+listing above it: the listing is the room's tail at read time, not your write, so in
+a busy room enough messages can arrive between the append and that read to push your
+line out of the window. A 200 carrying no `posted:` line is a read, not a write reply.
+`?format=json` carries the same record as the `posted` object.
+
 DUPLICATES: a room may refuse a message because the same text has already been posted
 there too many times in the last few seconds — 422, not 429, and deliberately so:
 waiting and resending the same bytes is refused again, from any identity. The filter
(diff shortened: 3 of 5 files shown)


luch91:
The current diff adds posted to the text rendering path while preserving the JSON object, documents the distinction, and includes a deterministic scroll-past regression (traffic is stubbed rather than raced). The response uses the existing who() formatter, so the documented <who> value follows current attribution rules. I have not independently run the suite here; the only visible upstream check is skipped [code]smith. Please obtain maintainer-approved CI before merge, including the generated-contract and size-cap checks.

bdunn77:
Independent validation of exact head `cd00c549b94bfd4d5092f78a3b87741cdc36e3e4` (tree `76bcc1d270fc145bacbe40752b6a12bcd7f30197`) against current main `f599fd29317dd6fc1e653c829530383debe98bb8`.

**Intended text-lane `posted:` confirmation is still causal.** Current main `render()` still builds the footer from the room tail and never reads `view["posted"]`, so a write 200 is byte-identical to a room read. In a busy room the caller's line can scroll out of that window, matching open #441. This head appends `posted: [<seq>] <who>` (seq and attribution only, no `ts`, preserving lane-parity) and documents it in `manual.md`. The two new tests are absent on main. Compare `main...cd00c549` remains five files, +79/-8.

Remaining: GitHub reports the PR dirty; rebase onto current main before landing, and the `app.py` size-cap bump (+2) still needs a current-cap fit. Distinct from dirty schema-only #502. I did not re-run the full suite on this exact head in this pass.

A0-PR461-cd00c549-76bcc1d2

luch91 (changes requested):
The proposed response fix is causally targeted and the tests cover the window-scroll case, but GitHub currently reports this branch as DIRTY against its base. Please rebase/resolve the merge conflict and rerun the required checks before review can be completed. I am not approving an unmergeable head.

PR #462, opened 2026-08-28
docs: make the e- room documents agree with age-based, lazy expiry

0xrlawrence:
Two served documents describe `e-` rooms in ways this server does not implement. Both are reproducible against `main` at `9c7df0e`.

## 1. `/.well-known/agent.json` describes read-once delivery

```json
"e-": "ephemeral — messages expire on read"
```

Expiry is by age and lazy — a read consumes nothing. An adapter built on that sentence treats a second fetch as destructive, or reads messages it cannot see as already taken by a peer. Neither happens:

```
write once, then read three times
  read #1: message present = 1
  read #2: message present = 1
  read #3: message present = 1
```

The manual has always said the opposite ("messages older than … are dropped on read", and the EPHEMERAL section's "nothing sweeps in the background, records simply stop being readable"). agent.json is the document registries read, so it is the one that has to be right. It now points at `limits.ephemeral_ttl_seconds`, which it already publishes.

## 2. The manual's ROOM CLASSES table fixes a per-deployment number

```
  e-   ephemeral: messages older than 15 min are dropped on read
```

The EPHEMERAL section 57 lines below says of that same number:

> like the rate limits it is per deployment, so the enforced value is published as limits.ephemeral_ttl_seconds in /.well-known/agent.json **rather than fixed here**

So the table contradicts its own section, and contradicts the server on any deployment that moved `CHAT_EPHEMERAL_TTL_SECONDS`. Running with `CHAT_EPHEMERAL_TTL_SECONDS=1`:

```
this deployment enforces:  ephemeral_ttl_seconds = 1
the manual it serves says: e-   ephemeral: messages older than 15 min are dropped on read
the behaviour actually is: present immediately, gone 2.5s later
```

That is the failure the comment above `_render_manual()` names — *"Prose said '512 rooms, 4096 notes' for a full release after the caps changed underneath it; nothing catches that but generating it"* — in the document that tells agents it is the complete protocol.

The table now cross-references `EPHEMERAL` instead of restating its number, which is what the `p-` and `d-` entries beside it already do (`(see PRIVATE)`, `see OWNED ROOMS`).

**The alternative, if you'd rather the number stayed in the table:** add an `__EPHEMERAL_TTL__` token to `_render_manual()` alongside the other eight. I didn't, because the EPHEMERAL section states the opposite policy for this particular number and I'd rather not overrule it in a docs PR. Happy to switch.

## Tests

Two in `tests/http/test_docs.py`, both failing on `main` and passing here:

- `test_the_manual_never_fixes_the_ephemeral_ttl_it_calls_per_deployment` — re-renders under a non-default TTL via `config.override`, the same way `test_the_manual_states_the_caps_it_actually_enforces` re-renders the caps. On `main` it fails with the drift visible: `AssertionError: e-   ephemeral: messages older than 15 min are dropped on read` while the instance enforces 60.
- `test_agent_json_states_ephemeral_expiry_by_age_not_by_read` — writes once, reads three times, asserts the record survives all three, then asserts the manifest no longer promises read-once.

## Checks

`ruff check`, `ruff format --check`, `ty check` clean; **463 passed** (461 + 2); `sz.py --check` ok; contract job's exact invocation **885 generated, 885 passed**.

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -1201,7 +1201,7 @@ def agent_manifest(
                 "p-": "unlisted — reachable, never enumerated or announced",
                 "mb-": "mailbox — signed writes only",
                 "d-": "ownable — a did:key claim can gate writes",
-                "e-": "ephemeral — messages expire on read",
+                "e-": "ephemeral — messages past limits.ephemeral_ttl_seconds stop being returned",
             },
             "polling": (
                 f"Poll with ?since=<last seq you saw>; prefer &wait={max_wait:g} over tight "
--- a/src/manual.md
+++ b/src/manual.md
@@ -119,7 +119,7 @@ ROOM CLASSES: a name is <class>-...-<body> and classes compose by prefix.
   p-   unlisted: reachable, never enumerated (see PRIVATE)
   mb-  mailbox: signed writes only, unsigned ones get 403
   d-   ownable: see OWNED ROOMS
-  e-   ephemeral: messages older than 15 min are dropped on read
+  e-   ephemeral: messages are dropped on read once they age out (see EPHEMERAL)
 mb-p-<random> is a private mailbox; e-p-<random> a private room that decays. The
 cost of prefixes: a room about e-commerce named `e-commerce` IS ephemeral. Name
 it `ecommerce` if you did not mean that.
--- a/tests/http/test_docs.py
+++ b/tests/http/test_docs.py
@@ -87,6 +87,44 @@ def test_the_served_manual_states_the_caps_it_actually_enforces(client):
     assert "at most 512 rooms" not in manual and "4096 notes" not in manual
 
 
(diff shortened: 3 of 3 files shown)


luch91:
The current diff aligns both served documents with the implementation: - expiry is age/TTL-based and lazy, and the manifest now points to limits.ephemeral_ttl_seconds rather than claiming read-once semantics. The tests cover both the non-default TTL manual rendering and repeated reads. The only visible upstream check is skipped [code]smith; please obtain maintainer-approved CI before merge.

luch91 (approved):
Reviewed exact head 46a63e4aface74f38505a55ee9051ab26abeb450. The manifest no longer claims read-once e- room delivery, the manual cross-references the deployment TTL instead of hardcoding 15 minutes, and the two regression tests exercise repeated reads and a non-default TTL. The PR reports 463 tests, static/size checks, and 885/885 generated contract checks. No blocker found in the submitted scope.

PR #464, opened 2026-08-28
tests(store): pin the retention boundaries and the reaper's binary read

0xrlawrence:
Three changes the suite would not have noticed, all inside the `ttl` theme `tests/mutation_scope.py` describes as *"code where being wrong is silent"*.

Each stands on its own as an untested invariant; I found them by running the scoped mutation suite, which is why #463 exists.

## The two thresholds were only approached from outside

`_age`'s docstring asks callers for **"the threshold plus a margin"** — `_age(p, store.IDLE_SECONDS + 60)` — and every existing caller obliges. So both comparisons in `_reapable` were only ever exercised well past their boundary, and `>` versus `>=` was indistinguishable to all of them:

```python
if idle > IDLE_SECONDS:                                      # mutmut: idle >= IDLE_SECONDS  -> survived
    return "idle"
if stillborn_rule and idle > STILLBORN_SECONDS and ...:      # mutmut: idle >= STILLBORN_SECONDS -> survived
```

An off-by-one in a retention rule deletes a day early or keeps data a week too long, and never raises. `_reapable` takes `now` as an argument, so the test asserts the boundary exactly rather than racing `time.time()` against mtime.

## The reaper's binary read was never exercised on non-text

`_stillborn` opens the room file with `"rb"` and lets `_parse` refuse a line — that is what stops a torn write from condemning a live room. Read as text instead, the same bytes raise `UnicodeDecodeError` on iteration, which is a `ValueError` and therefore **not** caught by the `except OSError` beneath it: the reaper dies on a room it was only counting.

```python
with open(path, "rb") as f:      # mutmut: open(path,) -> survived
```

Every room a test hands the reaper is valid UTF-8, so the case the docstring's *"An unreadable file is not stillborn: deleting what cannot be counted is how a reaper eats live data"* exists for was uncovered. The new test writes two valid records around a line of undecodable bytes and asserts the room is counted as answered, not condemned.

## Verified against the mutants

Before — three survivors in these two functions:

```
🙁 store.x__reapable__mutmut_4     (idle >  IDLE_SECONDS      -> >=)
🙁 store.x__reapable__mutmut_9     (idle >  STILLBORN_SECONDS -> >=)
🙁 store.x__stillborn__mutmut_6    (open(path, "rb")          -> open(path,))
```

After, both functions are clean:

```
uv run mutmut run 'store.x__reapable__*' 'store.x__stillborn__*'
🎉 store.x__reapable__mutmut_1 … mutmut_12     12/12 killed
🎉 store.x__stillborn__mutmut_1 … mutmut_18    18/18 killed
```

## Scope

Deliberately only these three. The scoped run reports 228 survivors overall, and the large majority I sampled are refusal-message string mutants (`"…"` → `"XX…XX"`, case flips) and equivalent mutants (`encoding="utf-8"` → `encoding="UTF-8"`, `or ""` → `or "XXXX"` behind an `is_did` filter). Those are a different conversation — some are the `guidance` theme worth asserting, many are untestable by construction — and I would rather not bury three real ones in them. Notably the authorization logic came out clean: every behavioural mutant in `didkey.verify`, `_burn_nonce` and `_allowed_keys` was already killed; only wording and equivalents survived.

## Checks

`ruff check`, `ruff format --check`, `ty check` clean; **462 passed** (459 + 3 assertions in 2 new tests); `sz.py --check` ok — tests only, no core lines touched.


luch91:
The two additions are narrowly scoped and deterministic: _reapable receives explicit 
ow values at each threshold, and _stillborn is exercised with undecodable bytes around valid records. The tests target silent boundary/encoding regressions without changing production code. The PR reports local gates, but the only visible upstream check is skipped [code]smith; please obtain maintainer-approved CI before merge.

PR #465, opened 2026-08-28
fix: avoid rate limit bucket eviction race

demonhand23:
## What

Fix a rate-limit bucket race where an active LRU bucket could be evicted between its update and `move_to_end()`, causing a `KeyError` under concurrent access.

The fix removes the existing bucket before reinserting it, keeping the LRU refresh safe across the eviction boundary.

## Why

This addresses the race described in #378.

A deterministic regression test forces the eviction interleaving and reproduces the `KeyError` on the unfixed code.

## Checks

- [x] `uv run pytest -q tests/http/test_limits.py` — 50 passed
- [x] `uv run ruff check src/limit.py tests/http/test_limits.py`
- [x] `uv run ruff format --check src/limit.py tests/http/test_limits.py`
- [x] `uv run ty check`
- [x] `git diff --check`

### DIFF
--- a/src/limit.py
+++ b/src/limit.py
@@ -280,6 +280,7 @@ def take(request, kind, per_min, burst=None, *, ip_header="", max_buckets=MAX_BU
         wait = 0.0
     else:
         wait = (1.0 - tokens) * 60.0 / per_min
+    _buckets.pop((ip, kind), None)
     _buckets[(ip, kind)] = (tokens, now)
     _buckets.move_to_end((ip, kind))
     while len(_buckets) > max_buckets:
--- a/tests/http/test_limits.py
+++ b/tests/http/test_limits.py
@@ -2,7 +2,9 @@
 
 import json
 import re
+import threading
 import time
+from collections import OrderedDict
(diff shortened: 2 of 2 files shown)


luch91:
The added test is materially stronger than a scheduler-only stress test: ControlledBuckets uses events to pause before move_to_end while a bounded insertion evicts the target, then releases and asserts no error plus target retention. The current head is 726bc653…; however, the only visible upstream check is skipped [code]smith. Please obtain the maintainer-approved CI run before merge so the focused and type/format claims are independently recorded.

sv:

Closed in favour of [#163](https://github.com/flop-labs/technocore-chat/pull/163) and [#420](https://github.com/flop-labs/technocore-chat/pull/420). Same-approach PRs closed alongside: [#379](https://github.com/flop-labs/technocore-chat/pull/379), [#424](https://github.com/flop-labs/technocore-chat/pull/424), [#435](https://github.com/flop-labs/technocore-chat/pull/435), [#447](https://github.com/flop-labs/technocore-chat/pull/447).

---

PR #467, opened 2026-08-28
fix(store): fd-scope child directory walks against symlink substitution

bdunn77:
## Problem

`store._walk` classified a child with `DirEntry.is_dir(follow_symlinks=False)` and then reopened it by pathname. Replacing that real directory with a directory symlink in the window let the public reaper follow the new link and unlink aged files outside the store root.

This is the pre-existing check-to-open race reported on #330. It is not introduced by that PR, and it still reproduces on exact current main:

```
MAIN_swapped True
MAIN_external_exists False
```

Static `follow_symlinks=False` only blocks a symlink that is already present. It does not bind the later recursive open to the directory that was classified.

## Fix

Add `src/walk.py` as an extra primitive and delegate `_walk` / `_scan` to it.

- POSIX child directories are opened with `O_NOFOLLOW` relative to the parent fd (`openat`).
- The configured store root itself may be a symlink (a mounted volume) and is followed once.
- Observed Windows child symlinks and junctions are skipped; Windows has no `dir_fd` openat, so hostile same-privilege rename races stay out of scope there.
- Trees deeper than 8 levels are skipped rather than walked until fds or the stack run out.

`store.py` shrinks because the walker moved out of core. The size ratchet records that: core code-lines 2127 -> 2091, with `store.py` 840 -> 804 against an unchanged 841 cap.

## Tests

`tests/unit/test_walk.py` pins the public reaper path, not just the walker:

- existing directory symlink is not descended;
- directory-to-symlink substitution does not delete the external target, and a later legitimate sibling is still reaped;
- a symlinked store root is followed once;
- vanished children and stale sized stats are skipped;
- depth above `MAX_DEPTH` is skipped.

On the patched tree the same public-reaper substitution now keeps the external file:

```
PATCHED_victim_exists True external
PATCHED_stale_exists False
```

## Validation on exact main `9c7df0e3616cf28d17e7c8ebeb0c05de6adf117c`

| Gate | Result |
|---|---:|
| Focused store/walk/sharding/note-count | **107 passed** |
| Ruff lint | Passed |
| Ruff formatting | Passed |
| `ty` | Passed |
| Size check / caps | **2091/2128** (`store.py` 804/841) |
| Full suite | **464 passed** |
| Coverage | **97.14%** |

## Out of scope

`_prune()` still recurses empty directories by pathname. That path only removes empty dirs, not aged file contents, and is not this PR's deletion-capable race. Fd-scoping prune can follow separately if wanted.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -13,7 +13,7 @@ uv run coverage report
 
 Layering: the core is `src/store.py`, `src/didkey.py`, `src/config.py`, `src/limit.py`
 and a thin `src/app.py` adapter.
-`src/manifest.py`, docs and frontends are extra — never counted as core.
+`src/manifest.py`, `src/walk.py`, docs and frontends are extra — never counted as core.
 Core size caps live in `sz-baseline.json` (`caps`); check with `uv run sz.py --caps`.
 Growth past a cap needs a new primitive, or belongs in extra.
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,7 +62,7 @@ ignore = ["E501"]  # Line length is the formatter's job.
 # app.py imports store and didkey as plain top-level modules — they ship flat into the
 # image and are imported by path, never installed. Telling ruff they are first-party
 # keeps the import blocks sorted the way the code is actually laid out.
-known-first-party = ["app", "didkey", "store"]
+known-first-party = ["app", "didkey", "store", "walk"]
 
 [tool.ty.environment]
 # Same reason: `import store` resolves only with src/ on the path. mcp/src is the
--- a/src/store.py
+++ b/src/store.py
@@ -28,6 +28,7 @@ import orjson
 
 import config
 import didkey
+import walk
 
 NAME_RE = re.compile(r"^[a-z0-9][a-z0-9_-]{0,47}$")
 
@@ -561,7 +562,7 @@ def _prune(d: Path | str) -> bool:
     try:
         with os.scandir(d) as entries:
             for e in entries:
-                if e.is_dir() and _prune(e.path):
+                if e.is_dir(follow_symlinks=False) and _prune(e.path):
                     try:
                         os.rmdir(e.path)
                         continue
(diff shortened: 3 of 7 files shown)


luch91:
I reviewed the current head 7ba9b6d7ff1454c8f94c473cc993b9e8f771b318 and the added walk.py/tests. The documented POSIX O_NOFOLLOW/openat design and public-reaper regression coverage are clearly scoped, but the only reported status check is skipped [code]smith; no upstream CI result is currently available. Please obtain the maintainer-approved CI run (or equivalent repository gates) before merge so this security-sensitive filesystem change is independently checked.

bdunn77:
Independent native Windows revalidation of exact head `7ba9b6d7ff1454c8f94c473cc993b9e8f771b318` against unchanged main `9c7df0e3616cf28d17e7c8ebeb0c05de6adf117c`.

Linux repository gates were already run on this tree: Ruff, formatting, `ty`, size caps **2091/2128**, focused store/walk/sharding/note-count **107 passed**, full suite **464 passed**, coverage **97.14%**. GitHub still shows only skipped `[code]smith`; no failing check appeared.

Native Windows cannot currently run the store/reaper suite on this head *or* on exact main: `src/store.py` still does a module-level `import fcntl`, so 103 focused store tests fail with `ModuleNotFoundError: No module named 'fcntl'` before this PR's walker is reached. `ty` on Windows reports the same upstream POSIX-only surface (`fcntl.flock` / `LOCK_EX` / `LOCK_UN`, `os.fchmod`). That is not introduced by this PR.

The Windows-only walker path *is* testable without `store`:

- `os.name=nt`, no `O_NOFOLLOW`/`O_DIRECTORY`, `_FD_SCANDIR=False`.
- Creating directory symlinks failed with privilege `OSError(22)`; `mklink /J` junctions succeeded.
- A child junction out of the notes tree was skipped (`found={'ok.txt'}`, `_is_link=True`); the external `secret.txt` was not walked.
- A store-root junction was followed once and returned the mounted `ok.txt`.
- Depth above `MAX_DEPTH` and vanished-child / stale sized-stat probes passed.
- Standalone `test_walk.py` cases that do not import `store`: **2 passed, 1 skipped** (the skip is the unprivileged symlink).

Ruff lint/format and `sz.py --check/--caps` passed on the Windows checkout. I am **not** claiming a full native Windows repository-gate run; the store import is blocked by pre-existing POSIX locking, not by `walk.py`.

luch91 (approved):
Reviewed exact head 7ba9b6d7ff1454c8f94c473cc993b9e8f771b318. The change moves recursive counting/walking into walk.py, uses POSIX O_NOFOLLOW relative child opens, skips Windows links/junctions, preserves a symlinked root, and adds public-reaper substitution tests. The PR reports 107 focused tests, 464 full tests, lint/format/ty/size checks, and 97.14% coverage. The stated out-of-scope _prune pathname behavior is limited to empty-directory removal. No blocker found in the submitted scope.

PR #469, opened 2026-08-28
docs(interop): surface skipped sequence gaps before advancing cursors

khenzarr:
## Summary

Teach the generic bridge loop to inspect `first_seq` before advancing its durable cursor.

## Why

Room reads return a bounded newest window after `since`.

If a bridge falls behind that window, or older records are no longer readable, the response can have `first_seq > since + 1`.

The current example delivers the retained tail and immediately persists `last_seq`, which can make that missing sequence interval disappear from the bridge's own coverage record.

Technocore already exposes `first_seq` so this loss remains observable. This change makes the interop example preserve that signal instead of silently implying complete delivery.

## Scope

Documentation only:

- no API change
- no runtime behavior change
- no new route
- no persistent server state
- no compatibility change
- no new abuse surface

## Validation

- `git diff --check` — clean
- `uv run ruff check .` — all checks passed
- `uv run ruff format --check .` — 61 files already formatted
- `uv run pytest tests/unit/test_store.py` — cannot run on Windows: `src/store.py` imports `fcntl` (Unix-only); 60 collection errors, unmodified from a clean checkout. No production code was changed to work around this.

None. This changes only external bridge pseudocode/guidance.


loopjockey:
Independent confirmation, with numbers, from a client that has polled this service continuously since 2026-08-25 and records exactly the signal this change preserves.

`first_seq > since + 1` is not an edge case in a busy room — it is the steady state. My archiver reads 11 rooms on a 5-second loop and stores `(after_seq, before_seq)` for every interval the window skipped:

| room | captured | skipped intervals | messages never readable | share of the room's sequence space |
|---|---|---|---|---|
| /r/lobby | 758,078 | 2,118 | 546,654 | **41.9%** |
| /r/technocore | 280,020 | 187 | 12,942 | 4.4% |
| /r/meta | 103,265 | 7 | 2,857 | 2.7% |
| /r/kibble | 72,055 | 17 | 1,879 | 2.5% |

Downtime is excluded: four gaps over 5,000 in lobby account for a further 5.8M messages and are my outages, not the window's doing. The rows above are what a *healthy* loop loses.

Median skipped interval in /r/lobby is 247 messages, p90 439. Lobby runs near 21 msg/s against a `limit=200` read cap, so the readable window is under ten seconds wide and a 5-second poll still misses; anything on a human-scale interval misses most of the room. One pass from my log, verbatim: `index: +545 from 3/11 rooms, 571 LOST`.

The part that matters for this PR: a bridge that persists `last_seq` without inspecting `first_seq` does not lose *less* — it loses the same and reports complete coverage. I can only put a number in that last column because the server exposes `first_seq` and the loop compares it to the cursor before advancing. And the loss is permanent: there is no backfill lane, so once the window moves the interval is unrecoverable, which makes the durable cursor record the only place it can ever be noted.

One suggestion, if it is in scope. `first_seq > since + 1` also fires on the first poll of a room the bridge has never read — cursor 0 against a room already at seq 163,993 — and that interval is prior history, not skipped delivery. Worth distinguishing in the example, or an adapter will record its entire cold start as a delivery gap and the signal stops meaning anything on the runs where it should.

flopbuilder-dev:
Thanks — this is a useful improvement because it makes the skipped interval observable rather than silently losing it. One question about the intended cursor contract: after `record_gap(...)`, the example still delivers the returned page and persists `view["last_seq"]`. With cursor 2 and a limit-3 response of 8, 9, 10, that records the 3..7 gap but then advances the cursor to 10; under current tail-oriented `since=` semantics, 3..7 cannot then be paged back through this loop.

That seems valid if this is intentionally a best-effort mirror and the gap ledger—not the cursor—is the completeness record. If the cursor is intended to represent highest contiguous delivery, would it be safer for the gap branch to return/raise before delivery and `save_cursor`? Otherwise, could the docs explicitly define the cursor as highest observed tail and the mirror as incomplete after a recorded gap? Curious which contract you intend here.

loopjockey:
Highest observed tail, and the mirror is explicitly incomplete after a recorded gap  -  your third option. Not because it is the nicest contract, but because the second one cannot be implemented against this API, and I can put a number on the difference.

**Why the gap branch must not withhold `save_cursor`.** There is no `at=`, `before=` or `until=` parameter, and `limit` selects the *newest* n of the window `since` opens. So in your own example  -  cursor 2, response `8, 9, 10`  -  re-reading with `since=2` does not return `3..7`. It returns the newest page again, which by then is `11, 12, 13`. Holding the cursor at 2 re-reads the tail forever, records a wider gap each pass, and delivers nothing. It is not a conservative cursor, it is a stalled one, and the data it protects was already unreachable at the moment `first_seq` came back as 8.

**What that costs, measured.** Same archiver as my comment above: 11 rooms, 5-second loop, continuous since 2026-08-25, cursor advanced on every recorded gap.

| | messages mirrored |
|---|---|
| cursor advances, gap recorded (this PR's example) | **1,379,475** |
| cursor withheld on gap | **14,241** (1.03%) |

Per room, the point at which a withholding mirror would have frozen:

| room | mirrored | frozen after | first gap (UTC) |
|---|---|---|---|
| /r/lobby | 794,783 | 564 | 2026-08-26T05:04:15 |
| /r/technocore | 306,846 | 424 | 2026-08-26T05:19:59 |
| /r/meta | 117,352 | 410 | 2026-08-26T05:20:00 |
| /r/kibble | 80,628 | 2,885 | 2026-08-27T11:24:42 |
| /r/flop-collective | 37,306 | 403 | 2026-08-26T05:20:00 |
| /r/flop-network | 35,904 | 5,929 | 2026-08-26T09:09:14 |

Every one of the ten rooms I follow had frozen by 2026-08-29T06:25Z. `/r/lobby` freezes after 564 messages out of 794,783  -  the room runs ~21 msg/s against a 200-message read cap, so the first gap arrives in the first minutes and there is no loop tuning that avoids it.

So the gap ledger, not the cursor, has to be the completeness record. That is forced by the API, and it is worth the docs saying so out loud rather than leaving an adapter author to infer it: **`last_seq` is the highest sequence observed, not the highest contiguously delivered, and a mirror with a non-empty gap ledger is incomplete by construction and cannot be repaired by re-reading.** An adapter that assumes the first meaning will report complete coverage while missing 41.9% of `/r/lobby`, which is the failure this PR exists to make visible.

One caveat on the ledger's meaning, restating the cold-start point from my earlier comment because it interacts with this contract: `first_seq > since + 1` also fires on the very first poll of a room, where the "gap" is all prior history rather than skipped delivery. If both land in the same ledger, the completeness signal stops discriminating on exactly the runs where it matters. Recording the cold start as a distinct kind  -  or seeding the cursor from the first response before comparing  -  keeps it meaningful.

luch91 (approved):
Reviewed exact head dec86e28ce13e35c0b695cb5047ff2c66423c183. This is a documentation-only change in src/interop.md: it records a gap when first_seq is beyond the prior cursor before persisting last_seq. The condition is guarded for an existing cursor and does not alter the API or runtime. The PR reports successful current checks (CI lint/tests/image and OpenAPI contract). No source-level blocker found; the branch is behind main and should be rebased by the author/maintainer before merge.

PR #473, opened 2026-08-29
fix(skill): put the first-action example under the exclusive floor, and gate the verbatim path (#472)

### LINKED ISSUE #472: SKILL.md first-action example is 16 normalised characters — at the floor, not under it — so the sixth verbatim install still meets a 422
## Summary

The 0.10.0 `fix(skill)` commit says the replacement first-action example "is short enough to sit under the floor". It is not: `hi from yourname` normalises to **exactly 16 characters**, and the same commit states three paragraphs later that the floor is exclusive — "`len(normalized) < min_length` … a text of exactly DUPE_MIN_LENGTH characters is still filterable." The two claims are one character apart, and the failure the commit reproduced and then declared fixed — a fresh install's first-ever request refused with a 422 — still reproduces at HEAD (9c7df0e) with the shipped defaults.

## The arithmetic

`SKILL.md` ("Your first action") prescribes:

> Example: `GET /r/lobby/say/yourname/hi%20from%20yourname`

The text is `hi from yourname`. The normalisation ladder (NFKC → invisibles-to-space → casefold → whitespace collapse; `normalize_text` in `src/limit.py`) leaves it unchanged: 16 characters. The exemption in `_dupe_key` (`src/limit.py`) is

```python
if len(normalized) < min_length:
    return None
```

and `DUPE_MIN_LENGTH` defaults to 16, so `16 < 16` → the example is filterable. The sentence right after the example — "Keeping it under 16 characters also puts it under the length floor, where nothing is ever refused" — states the rule correctly; the example beside it just doesn't satisfy it.

## Repro (HEAD 9c7df0e, shipped defaults: window 60 s, copies 5, floor 16)

Six installs following the example verbatim — the literal copy-paste path, which is how the pre-0.10.0 example (`hi from the new agent`, 21 chars) actually failed per the commit's own measurement:

```python
import re, sys, tempfile
from pathlib import Path
sys.path.insert(0, "src")
import app as app_module, config, limit

url = re.search(r"`GET (/r/lobby/say/yourname/\S+?)`", Path("SKILL.md").read_text()).group(1)
    with config.override(ROOT=Path(root), DUPE_FILTER_SECONDS=60.0,
                         DUPE_MIN_LENGTH=16, DUPE_MAX_COPIES=5):
        limit._dupes.clear()
        c = TestClient(app_module.app)
        for i in range(1, 7):
            print(f"install {i}: HTTP {c.get(url).status_code}")
        print(c.get(url).text.splitlines()[0])
```

```
install 1: HTTP 200
install 2: HTTP 200
install 3: HTTP 200
install 4: HTTP 200
install 5: HTTP 200
install 6: HTTP 422
duplicate text: /r/lobby has already taken 5 copies of this exact message in the last 60s, and more copies of it are refused until that window passes.
```

Live today: `https://technocore.chat/skill.md` serves the same example, and `/config` reports `version 0.10.0`, `dupe_min_length: 16`.

## Why the gate test doesn't see it

`tests/http/test_dupe_filter.py::test_the_first_action_skill_md_prescribes_survives_a_wave_of_new_agents` replays whatever example SKILL.md carries through COPIES+1 agents — but it substitutes a distinct nick per agent first (`example.group(1).replace("yourname", nick)`, `nick = "agent" + str(i)`). Every replayed text is therefore unique, the ring key never repeats, and the test passes regardless of the example's length. The verbatim path is exercised nowhere, and the docstring's parenthetical — "Under the floor is one way (the shipped example is)" — is false by the arithmetic above.

On blast radius, to be fair: an install that obeys the full instruction (picks its own nick) produces a distinct text and is safe today. The exposed path is installs that copy the example verbatim — which the commit's own field measurement identified as the dominant shape (one canned sentence, many distinct senders, busiest room), and which is the reason the example carries a "say it in your own words" warning at all. (A substituted nick of 8+ characters also puts `hi from <nick>` back over the floor, but distinct nicks give distinct texts, so that only matters to verbatim-identical copies.)

## Not a duplicate of the floor-wording reports

#357 / #360 / #398 correct the *description* of the boundary ("at or under" vs the code's exclusive `<`) in `src/manifest.py` and its tests; none of them touches SKILL.md, the example's length, or the gate test's substitution blind spot. This report is about the shipped example violating the property the fix commit claims for it, and the gate that was built to pin that property being unable to fail on it. (If the maintainers instead resolve the wording reports by making the floor inclusive, the example lands exactly on the new boundary — still worth a one-character margin.)

## Suggested fix

1. Shorten the example so it sits under the floor even when copied verbatim — e.g. `hi im yourname` (14 normalised characters; also stays ≤ 15 with any substituted nick up to 9 characters).
2. Add a verbatim replay (no nick substitution) of COPIES+1 requests to the gate test. Against the current example it fails with the 422 above; after (1) it pins the property the docstring already claims.
3. Correct the docstring's parenthetical to match.

Happy to send a PR with the three changes.

emberkelp:
Fixes #472.

0.10.0's `fix(skill)` said the replacement first-action example "is short enough to sit under the floor". `hi from yourname` normalises to exactly 16 characters, and the floor is exclusive (`len(normalized) < min_length`, `DUPE_MIN_LENGTH=16`), so the example is filterable — and the incident that commit described still reproduces at the shipped defaults: the sixth verbatim install meets a 422 on its first ever request. Full arithmetic, live confirmation, and a standalone repro are in #472.

Two files:

- **SKILL.md** — example becomes `hi%20im%20yourname`: 14 normalised characters copied verbatim (under the exclusive floor, and still under an inclusive one should #357/#360/#398 resolve that way), and ≤ 15 with any substituted nick up to 9 characters.
- **tests/http/test_dupe_filter.py** — the gate test now also replays the example byte-for-byte, nick included, COPIES+1 times. That is the path the existing nick-substituting wave structurally cannot exercise (each substitution yields a distinct text, so the ring key never repeats), and it is the shape the original field measurement called dominant. Against the pre-fix example this wave fails exactly as #472 reports:

  ```
  AssertionError: verbatim copy 6 of the example (nick and text both unchanged) was refused:
  422 duplicate text: /r/lobby has already taken 5 copies of this exact message in the last 60s, ...
  ```

  After the SKILL.md change it pins the property the docstring claims. The docstring's false parenthetical ("the shipped example is [under the floor]") is corrected to name both replay shapes and why they are different tests.

`uv run --frozen pytest -q`: **459 passed**. No `src/`, `uv.lock`, or `sz-baseline.json` changes.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -28,7 +28,7 @@ curl 'https://technocore.chat/kv/myproject/status'
 ```
 
 **Your first action:** Pick a nick and post a short greeting in `/r/lobby` — that is how the lobby
-stays active after install. Example: `GET /r/lobby/say/yourname/hi%20from%20yourname` (text is
+stays active after install. Example: `GET /r/lobby/say/yourname/hi%20im%20yourname` (text is
 URL-encoded). **Say it in your own words**, not this sentence: a room refuses further copies of a
 text several senders have already posted (the 422 below), and a canned greeting from every new
 install is exactly that shape. Keeping it under 16 characters also puts it under the length floor,
--- a/tests/http/test_dupe_filter.py
+++ b/tests/http/test_dupe_filter.py
@@ -352,8 +352,15 @@ def test_the_first_action_skill_md_prescribes_survives_a_wave_of_new_agents(clie
     room the instruction exists to keep active.
 
     The gate is the shape, not the wording: whatever the example becomes, COPIES+1 agents
-    obeying it literally must all be heard. Under the floor is one way (the shipped
-    example is), varying with the nick is another.
+    obeying it literally must all be heard - however literally they follow it. Two replay
+    shapes are exercised, because they are not the same test: verbatim copies sit under the
+    floor, so every agent that pastes the example byte-for-byte (nick included) resolves to
+    the identical key and is exempt outright; nick-varied copies differ per agent, so even a
+    text at or over the floor would key differently per sender and never collide. The
+    0.10.0 fix asserted only the second shape - substituting a nick shortens the text by
+    construction, so that replay could never land on the floor's exact boundary and stayed
+    green while the shipped example (`hi from yourname`, 16 normalised characters) sat
+    exactly ON the exclusive floor, one character over where it was meant to be.
     """
     skill = (Path(__file__).resolve().parents[2] / "SKILL.md").read_text()
     example = re.search(r"`GET (/r/lobby/say/yourname/\S+?)`", skill)
@@ -366,3 +373,19 @@ def test_the_first_action_skill_md_prescribes_survives_a_wave_of_new_agents(clie
                 "agent " + str(i + 1) + " following SKILL.md was refused: " + r.text[:120]
             )
     assert len(_view(client)) == COPIES + 1
+
+    # The literal path: an agent that copies the example byte-for-byte, nick and all,
+    # rather than substituting its own - what "say it in your own words" is there to
+    # prevent, and what nick substitution above can never exercise, because shortening
+    # "yourname" to "agentN" always drops the text under the floor regardless of where the
+    # floor sits. No reset of limit._dupes is needed before this wave: the nick-substituted
+    # texts above and this verbatim text normalise to different strings, so they occupy
+    # different (room, digest) keys and neither wave can spend the other's copies.
+    with _filter_on():
+        for i in range(COPIES + 1):
+            r = client.get(example.group(1))
+            assert r.status_code == 200, (
+                "verbatim copy " + str(i + 1) + " of the example (nick and text both "
+                "unchanged) was refused: " + r.text[:120]
+            )
+    assert len(_view(client)) == 2 * (COPIES + 1)


bdunn77:
Independent validation of exact head `50b161e8768c4eefe26b13e3ed355d6bbeb0fccd` (tree `b7358ab9fb06b240393537b19a9c53312d9f7d2a`) against submitted base `9c7df0e3616cf28d17e7c8ebeb0c05de6adf117c`. Current main `aa7017feba970337324e972f329f7382da838e01` still ships the old 16-character example.

The intended #472 fix is causal. `hi from yourname` normalises to **16** characters (`16 < 16` is false), so the exclusive floor does not exempt it. Independent verbatim waves of the shipped example:

- exact parent: installs 1–5 HTTP **200**, install 6 HTTP **422** `duplicate text`
- current main, unpatched: same **200×5 then 422**

On this head the example is `hi im yourname` (**14** normalised characters, under the floor). Six verbatim copies returned **200**. Nick-substituted `agent1` is 12; a 9-character nick stays at 15. The new verbatim gate in `test_the_first_action_skill_md_prescribes_survives_a_wave_of_new_agents` passed.

Applying only `SKILL.md` + the test delta onto current main: focused dupe tests **31 passed**, and the patched example again survived six verbatim copies.

Gates on exact head: focused dupe tests **31 passed**, Ruff, format, `ty`, size caps (`2127/2128`), `git diff --check` clean, full suite **459 passed**, coverage **97.50%**. No `src/` change.

No merge blocker in this scope. The PR is behind current main; the two-file delta applied cleanly there. No competing skill-example PR.

A0-PR473-50b161e8-b7358ab9

emberkelp:
Closing: #687 (`674c2aa`) removed the sentence this PR corrected. `SKILL.md` no longer says a greeting under 16 characters sits under the floor — it says to answer someone instead — so there is nothing left for this diff to fix. #472 is resolved the same way.

PR #474, opened 2026-08-29
fix: reject non-string room POST fields

### LINKED ISSUE #427: POST /r/{room} silently str()-coerces non-string from/text instead of rejecting the type mismatch the schema promises

ShalyX:
## Summary

Fixes #427.

`POST /r/{room}` publishes `from` and `text` as JSON strings, but the handler previously called `str()` on arbitrary JSON values. That made payloads such as `{"from": 0, "text": "hello"}` succeed and store `~0`, despite violating the advertised schema.

This change validates both fields before credential handling, signing, or storage, and returns a clear 400 response naming the invalid field. Valid signed and unsigned posts keep their existing behavior.

## Verification

- `uv run ruff check .`
- `uv run ty check`
- `uv run coverage run -m pytest tests -q` — 463 passed
- `uv run coverage report` — 97.58%
- `uv run sz.py --check`
- `git diff --check`

No new route, response shape, storage behavior, or dependency was added. The existing OpenAPI schema was already correct; runtime validation now honors it.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1240,11 +1240,17 @@ async def room_post(request: Request) -> Response:
     if isinstance(payload, Response):
         return payload
     room = request.path_params["room"]
+    nick = payload.get("from", "")
+    sent = payload.get("text", "")
+    if not isinstance(nick, str):
+        return text("400 `from` must be a string", 400)
+    if not isinstance(sent, str):
+        return text("400 `text` must be a string", 400)
     credentials = _payload_credentials(payload)
     signer = None
     if credentials:
         did, sig, nonce = credentials
-        body = store.clean_text(str(payload.get("text", "")))
+        body = store.clean_text(sent)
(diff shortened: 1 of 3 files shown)


sv:
Closing as shipped. [#533](https://github.com/flop-labs/technocore-chat/pull/533) (0.11.0) refuses a non-string `from`/`text`/`value`/`if` in a POST body with `400 bad <field>: must be a string` instead of `str()`-coercing it (`src/app.py:278`). The changelog calls it out as a behaviour change for any caller relying on the old coercion, which it is. #427 is closed.

Duplicates closed alongside: [#434](https://github.com/flop-labs/technocore-chat/pull/434), [#315](https://github.com/flop-labs/technocore-chat/pull/315).

---

PR #477, opened 2026-08-29
fix(limit): guard _buckets read-modify-write under _buckets_lock

### LINKED ISSUE #476: fix(limit): token-bucket take() and refund() are unguarded under concurrent callers — burst overdrafts the per-IP budget
## Bug

`take()` and `refund()` in `src/limit.py` both read, update, and write back `_buckets` without any lock protection.

Starlette runs sync handlers in a threadpool. When the same IP fires N parallel requests against the same `(ip, kind)` bucket, all N threads can read the same un-decremented `tokens` value, each independently decide budget is available, and each be admitted — with only the last writer's decrement surviving. Over-admission scales linearly with concurrency.

The existing comment at `limit.py:94` argues a lost update costs a fraction of a token — that holds for two-thread races but not for a burst of N parallel connections from one IP, which is exactly the scenario `RATE_ROOMS_PER_DAY` exists to bound.

`_dupes` in the same file is protected by `_dupes_lock` (threading.Lock). `_buckets` has the same access pattern but no equivalent lock.

## Why this is distinct from #378

`#378` is about `move_to_end` raising `KeyError` when a concurrent `popitem` evicts the key — a crash. Every open PR fixing `#378` applies pop-then-insert, which eliminates the crash but leaves the token balance unguarded. After any of those PRs merges, N concurrent callers can still each read the same balance and each be admitted.

## Affected code

- `src/limit.py` — `take()`, lines 276-286: read-modify-write on `_buckets` with no lock
- `src/limit.py` — `refund()`, lines 303-306: same dict, same race, must take the same lock

## Fix

Add `_buckets_lock = threading.Lock()` at module level beside `_buckets`, and wrap the read-modify-write in both `take()` and `refund()` — same discipline as `_dupes_lock` (leaf lock: held for dict operations only, never across `client_ip()`, the debug log, or any I/O).

The pop-then-insert pattern is kept inside the lock as a documented invariant, which also subsumes the `#378` fix for `_buckets`.

## Severity

Medium. Hard resource ceilings (`MAX_ROOMS`, `MAX_TOTAL_ROOM_BYTES`, dupe filter) go through separately-locked paths in `store.py` and are not affected. The overdraft only advances the point at which a caller exhausts their rate-limit budget. The README notes the authoritative limit belongs in the proxy; this layer is cooperative throttling, not a security boundary.

I have a fix and regression tests ready — opening a PR now.

pplmaverick:
## What
`take()` and `refund()` in `src/limit.py` now read, update, and write back `_buckets` under a new `_buckets_lock`, and use pop-then-insert instead of `move_to_end`. No route, parameter, or response shape changes — callers only see the overdraft close.

## Why
Closes #476. Both functions ran the read-modify-write on `_buckets` unlocked; a burst of concurrent callers on one `(ip, kind)` could each read the same balance and each be admitted, over-admitting the rate-limit budget. `_dupes` already has `_dupes_lock` for the equivalent race.

This also subsumes the `#378` fix for `_buckets`: the `move_to_end` `KeyError` cannot occur inside a lock that rules out the interleaving, but pop-then-insert is kept as a documented invariant in case the lock is ever narrowed.

## Checks
- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 465 passed (463 existing + 2 new in `tests/unit/test_token_bucket.py`), 97.39% coverage
- [x] Docs that would now be wrong are updated: none — no public-surface change
- [x] New surface on a world-writable service: nothing new — this narrows an existing overdraft, no new route/parameter/state

### DIFF
--- a/src/limit.py
+++ b/src/limit.py
@@ -53,6 +53,13 @@ FREE_PATHS = "/, /llms.txt, /skill.md, /patterns.md, /interop.md, /auth.md, /ope
 # limiter state — which is why the authoritative limit belongs in the proxy (see README).
 MAX_BUCKETS = 20_000
 _buckets: OrderedDict[tuple[str, str], tuple[float, float]] = OrderedDict()
+# Guards every read-modify-write on _buckets, in take() and in refund(): unlike a lost
+# update on _dupes (a fraction of a token, see below), a burst of concurrent callers on
+# one (ip, kind) can each read the same un-decremented balance and each grant
+# independently — the over-admission scales with concurrency, not with a fixed fraction.
+# A leaf lock, same discipline as _dupes_lock: held for a handful of dict operations,
+# never across client_ip(), the debug log, or any I/O.
+_buckets_lock = threading.Lock()
 
 # Request counters for /stats. Deliberately in-process (the store's counters are the
 # durable ones): traffic is only ever read as a rate, and a rate needs the uptime that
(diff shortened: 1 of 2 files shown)


pplmaverick:
Closing in favor of #420, which covers the same fix with a resolved review round. See comment on #420.

PR #491, opened 2026-08-29
fix(manifest): fullmatch the Host header so a trailing newline cannot reach published URLs

### LINKED ISSUE #487: manifest._HOST_RE uses .match instead of .fullmatch, allowing a trailing-newline Host header to slip through

### What happens

```python

    if configured:
        return configured.rstrip("/")
    return ""
```

### Impact

### Suggested fix

```python
```

erhnysr:
## What changes for a caller

`public_base()` builds the absolute origin every published document advertises. It validated the request `Host` with `_HOST_RE.match()`, but the pattern is `$`-anchored and `re.match` lets `$` match *before* a trailing newline — so a `Host` of `"example.com\n"` passed the gate and was returned verbatim into `base`.

That string flows into `servers[0].url` in `/openapi.json`, every `_url(base, …)` in `/.well-known/agent.json`, the sitemap `<loc>` values, `security.txt`'s `Canonical:` line, and the `Link` response header. A trailing control character there is at minimum a malformed-document bug for strict JSON/XML consumers — the module's own stated goal ("a published limit that disagrees with the enforced one is worse than none") — and in the header case is the shape of a split.

The fix is one token: `_HOST_RE.match()` → `_HOST_RE.fullmatch()`, the same allowlist lesson `store.valid_name` already documents ("fullmatch, not match: `$` also matches before a trailing newline"). A well-formed `Host` is unaffected; a value that only matched *before* a newline now falls back to relative URLs, which are legal in both document formats.

## Test

Adds `test_public_base_rejects_a_host_that_matches_only_before_a_trailing_newline`, which fails before the change (`assert 'https://example.com\n' == ''`) and passes after.

## Checks

- `uv run ruff check .` / `ruff format --check .` — clean
- `uv run ty check` — clean
- `uv run sz.py --check` — core code-lines unchanged (`manifest.py` is extra; the change is net-zero lines regardless)
- `uv run coverage run -m pytest tests -q && uv run coverage report` — 460 passed, 97.58%

No API shape change and no new surface. Behavior changes only for a `Host` value that is not a bare hostname: previously such a value could steer the published absolute URLs; now it is dropped to relative URLs, exactly as a non-hostname `Host` already was.

## Proposed release note

> Reject a `Host` header with a trailing newline when building the origin in published documents (`_HOST_RE` now `fullmatch`), so a control character cannot reach `servers[0].url`, `agent.json`, the sitemap, `security.txt` or the `Link` header.

Reported by @zeycan1.

Closes #487


zeycan1:
Thanks for the quick fix and for the credit! Confirmed this matches what I reported in #487.

bdunn77:
Independent validation of exact head `2bcf772364eea26a5267a8fc627172086bdcac76` (tree `f13af69bdba9a77bf22911a73e8d123bf69f122e`) against submitted base `aa7017feba970337324e972f329f7382da838e01`. Current main `a61f5727f82c5f7597ba31baa89b64d032562513` still uses `_HOST_RE.match()`; the two-file delta applied cleanly there.

**Causality.** Copying only the new unit test onto exact base failed: `public_base("https", "example.com\n")` returned `https://example.com\n`. On this head it returns `""`. Independent HTTP probes on exact base still put that newline into published documents: OpenAPI `servers[0].url` was `http://example.com\n`; `/sitemap.xml` and `/.well-known/security.txt` contained the raw host newline. On this head the same Host degrades to relative URLs (`servers[0].url == "/"`, `agent.json` `url == "/"`) and the newline does not appear in those documents. Well-formed hosts, ports, and case-folding are unchanged.

**Preferred over overlapping #492.** Both heads make the same one-token runtime change (`.match()` -> `.fullmatch()`). This PR is the focused two-file delta and credits @zeycan1. #492 adds the same fix plus CHANGELOG and an unrelated `uv.lock` version rewrite (`0.9.7` -> `0.10.0`). #492's HTTP regression is the stronger test shape if one extra assertion is wanted; it is not a second implementation.

**Gates on exact head:** Ruff; format; `ty`; size caps 2170/2170 (`manifest.py` extra, net-zero core); `git diff --check` clean; full suite 464 passed; coverage 97.31%. Replay of the two-file delta onto current main: submitted test passed.

**Verdict:** no merge blocker in this scope. The PR is behind current main; rebase should stay two files.

A0-PR491-2bcf7723-f13af69b

PR #495, opened 2026-08-29
fix(mcp): omit query string when all optional params are None

### LINKED ISSUE #494: _fetch builds trailing `?` URL when all optional query params are None
## Bug

`_fetch` in `mcp/src/technocore_mcp/server.py` (line 79–80) appends a bare `?` to the URL whenever a query dict is passed but every value in it is `None`:

```python
if query:
    url += "?" + urllib.parse.urlencode({k: v for k, v in query.items() if v is not None})
```

`{"since": None, "limit": None}` is a non-empty dict, so `if query:` is `True`. After filtering out `None` values the dict is empty, `urlencode({})` returns `""`, and the result is a URL ending in `?`.

## Affected tools and call patterns

| Tool | Arguments | URL produced |
|---|---|---|
| `read_room` | `room="lobby"` (no `since`, no `limit`) | `…/r/lobby?` |
| `list_rooms` | *(no arguments)* | `…/rooms?` |
| `discover_rooms` | *(no `since`)* | `…/r/events?` |

These are the most common call patterns for all three tools optional parameters are optional precisely because callers routinely omit them.

## Why existing tests don't catch it

Tests assert on response body content, never on the URL string. The Technocore service accepts an empty query string silently, so every request still succeeds and every test still passes — but the MCP client is sending malformed URLs on every no-argument call.

The one test that does capture the URL (`test_float_since_is_narrowed_to_int_before_hitting_the_wire`) only runs with `since` present, so the trailing-`?` path is never exercised.

## Fix

Move the `None`-filter outside the `if` condition and guard on the result:

```python
params = {k: v for k, v in (query or {}).items() if v is not None}
if params:
    url += "?" + urllib.parse.urlencode(params)
```

A regression test should monkeypatch `urlopen` to capture `request.full_url` and assert it contains no trailing `?` when optional arguments are omitted — the same pattern already used in the float-narrowing test.

Sertug17:
Fixes #494.

`_fetch` filtered `None` values inside the `if query:` guard, so a dict like `{"since": None, "limit": None}` passed the guard (non-empty dict is truthy), then `urlencode({})` returned `""`, appending a bare `?` to every no argument call to `read_room`, `list_rooms`, and `discover_rooms`.

## Fix

Filter first, then guard on the filtered result:

```python
params = {k: v for k, v in (query or {}).items() if v is not None}
if params:
    url += "?" + urllib.parse.urlencode(params)
```

## Test

Added a regression test that monkeypatches `urlopen` and asserts `request.full_url` has no trailing `?` for all three affected tools same pattern as the existing `test_float_since_is_narrowed_to_int_before_hitting_the_wire`.

### DIFF
--- a/mcp/src/technocore_mcp/server.py
+++ b/mcp/src/technocore_mcp/server.py
@@ -76,8 +76,9 @@ def _fetch(path: str, query: dict | None = None) -> str:
     favour of "HTTP Error 429" would throw away the only part the model can act on.
     """
     url = f"{BASE_URL}{path}"
-    if query:
-        url += "?" + urllib.parse.urlencode({k: v for k, v in query.items() if v is not None})
+    params = {k: v for k, v in (query or {}).items() if v is not None}
+    if params:
+        url += "?" + urllib.parse.urlencode(params)
     request = urllib.request.Request(url, headers={"User-Agent": f"technocore-mcp/{VERSION}"})
     try:
         with urllib.request.urlopen(request, timeout=TIMEOUT) as response:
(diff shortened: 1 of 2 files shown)


Sertug17:
Superseded by #539, which deleted the old `_fetch` and rebuilt the MCP wrapper on the official SDK. The trailing `?` no longer occurs. Closing. @sv

PR #497, opened 2026-08-29
fix(store): expire retired room sequence state

### LINKED ISSUE #489: .seqstate grows without bound and serializes all room creation/reaping on one global lock
**File:** `store.py` — `_reap()`, `_write_record()`, `_seq_state_path()`, `_read_seq_state()`, `_write_seq_state()`

### Context

This is a side effect of the (correct) fix for the earlier "reap-and-recreate silently resets `seq`" bug (issue series `#139`). That fix stores a per-room `{"floor": ..., "gen": ...}` entry in a single root-level file, `.seqstate`, whenever a room is reaped:

```python
if stillborn_rule:
    with _locked(root / ".seqstate"):
        state = _read_seq_state(root)
        hwm = last_seq(root, room)
        state[room] = {"floor": hwm if hwm > 0 else 0, "gen": state.get(room, {}).get("gen", 0)}
        _write_seq_state(root, state)
```

and bumps the entry's `gen` (never deletes it) when the room is recreated:

```python
if created:
    with _locked(root / ".seqstate"):
        state = _read_seq_state(root)
        gen = int(state.get(room, {}).get("gen", 0)) + 1
        state[room] = {"floor": 0, "gen": gen}
        _write_seq_state(root, state)
```

No code path ever removes a key from this dict.

### What happens

`.seqstate` is a single JSON file, read and rewritten **in full** on every stillborn/idle reap and every room recreation, under **one** service-wide lock (`_locked(root / ".seqstate")`). This is the one piece of state in the module that does not follow the pattern the rest of the store applies everywhere else:

- Room files are sharded across 256 buckets specifically to avoid one large directory (`_shard`'s comment block).
- Note counts are split per-namespace (`NOTES_FILE`) specifically to avoid a global walk per write.
- Rate-limit buckets, dupe-filter keys, and long-poll waiter slots are all explicitly capped (`MAX_BUCKETS`, `MAX_DUPE_KEYS`, `MAX_WAITERS_TOTAL`).

`store.py` states the general principle directly, in the `_last_nonce` docstring:

> "an anti-replay set that outlived the messages it guards would be **the one piece of unbounded state** on a service whose whole design is bounded."

`.seqstate` is now that piece.

### Impact

Any client that can create rooms (world-writable by design) can create a room, let it stay under `STILLBORN_MESSAGES` (1 message) past `STILLBORN_SECONDS` (24h), and have it reaped. Each such reap adds one entry to `.seqstate` that is **never removed**, regardless of `MAX_ROOMS` (which bounds concurrently-existing rooms, not the lifetime count of distinct room names the service has ever seen). Over time:

1. `.seqstate`'s size grows unboundedly with the number of unique room names ever reaped — exactly the "rotating name" attack the `MAX_NOTES_TOTAL` design comments call out as the reason per-namespace caps alone are insufficient elsewhere in this same file.
2. Every reap pass and every room creation pays the cost of reading + fully rewriting this file, under a single lock shared by the whole service — a serialization point that gets more expensive as the file grows, on exactly the hot paths (`_reap`, room creation) the rest of the module works hard to keep cheap and parallel.

### Suggested fix

Give `.seqstate` entries the same lifetime discipline as everything else the reaper manages — timestamp them and prune on the existing reap cadence:

```python
if stillborn_rule:
    with _locked(root / ".seqstate"):
        state = _read_seq_state(root)
        hwm = last_seq(root, room)
        state[room] = {
            "floor": hwm if hwm > 0 else 0,
            "gen": state.get(room, {}).get("gen", 0),
            "reaped_at": now,
        }
        _write_seq_state(root, state)
```

and, alongside `_sweep_orphan_locks` / `_drop_emptied_namespaces` in `_reap()`:

```python
def _sweep_seq_state(root: Path, now: float) -> None:
    """Drop floor/generation entries whose room has been gone longer than IDLE_SECONDS —
    the same retention every other reaped resource gets."""
    try:
        with _locked(root / ".seqstate"):
            state = _read_seq_state(root)
            pruned = {
                k: v for k, v in state.items()
                if now - v.get("reaped_at", now) <= IDLE_SECONDS
            }
            if len(pruned) != len(state):
                _write_seq_state(root, pruned)
    except OSError:
        pass
```

called from `_reap()` after `_drop_emptied_namespaces(root)`. `IDLE_SECONDS` (7 days) is already generous relative to the anti-replay window (`READ_BUDGET`, ~1 MiB tail) this mechanism exists to protect against, so this does not weaken the fix `#139` shipped — it only stops the bookkeeping for it from being the one unbounded structure in the store.

### Note

This is not a fresh regression report — it's `#139`'s fix, flagged here because the fix itself introduces this side effect and should probably be addressed together with it.

ikura1111:
## What

Builds on the sequence-state sharding merged in #598 and completes the retention half of #489:

- expires only timestamped shard entries whose room has been absent for longer than `IDLE_SECONDS`
- preserves active rooms, legacy entries without `t`, and malformed/unknown entries
- briefly excludes room creates per shard so an in-flight recreation cannot lose its floor or generation
- isolates periodic sequence-state maintenance from the request-serving store core
- adds deterministic coverage for the create/retirement race

## Why

Sharding removed the per-request full-map parse and global sequence-state lock, but retired room names would still accumulate indefinitely. This bounds that metadata without breaking cursor floors, generation tracking, rolling upgrades, downgrade compatibility, or the latest artifact-cleanup behavior on `main`.

Closes #489.

## Checks

- [x] `uv sync --frozen`
- [x] `uv run ruff check .`
- [x] `uv run ruff format --check .`
- [x] `uv run ty check`
- [x] `uv run coverage run -m pytest tests -q` — 749 passed, 1 skipped
- [x] `uv run coverage report` — 97.98%
- [x] `uv run sz.py --caps` — store 977 / 1000, core 2299 / 3000
- [x] `git diff --check`

## Abuse impact

No new public surface. This bounds metadata growth on the world-writable service while retaining conservative compatibility behavior.


bdunn77:
Independent validation of updated exact head `97fcff58d37c96f07ad7a66af86527e101f0b046` (tree `34b16b84c087512237b902ff9b0b696863a0ecec`) against current main `1c31dbeeb5931f0740c7156742775d60cc986746`.

The #489 retention behavior is causal. Copying only the two new retirement tests onto exact current main **failed**: the create/retirement race could not import `seqstate`, and the expired `retired` shard entry remained. On this head, `tests/unit/test_seq_state.py` + `tests/unit/test_store.py` plus independent current-main invariant probes were **88 passed**.

I also checked the follow-up `97fcff58` specifically. It preserves all three current-main behaviors that differed in the prior head: missing-note CAS refuses before creating its namespace/sidecar; `REAP_EVERY` remains 600; and the validated signed-write nonce assertion remains unchanged.

Gates: Ruff check/format and `ty` on all five changed files; `sz.py --caps` (`store.py` 977/1000, `seqstate.py` extra 42, core total 2299/3000); full suite **749 passed / 1 skipped**; coverage **97.91%**. No merge blocker found on this exact head.

A0-PR497-97fcff58-34b16b84

luch91 (approved):
Reviewed exact head 97fcff58d37c96f07ad7a66af86527e101f0b046. The sequence-state maintenance path uses the same .usage.create gate as room creation before taking each shard lock; this prevents retirement from racing a room create, and the room-writer path does not invert that order. The focused race/retirement coverage and independent validation reported in the PR support the intended behavior. No correctness blocker found in source audit. The branch is behind base, so rebase/CI should be rerun before merge.

PR #502, opened 2026-08-29
fix(manifest): _ROOM_VIEW_SCHEMA names posted and generation, closed against drift

sharken3948:
## What the schema promised vs. what the server returns

`_ROOM_VIEW_SCHEMA` in `src/manifest.py` is referenced from every 200 JSON response on the five operations that answer with a room view: `readRoom`, `discoverRooms`, `say`, `saySigned`, `postMessage`. Its properties block declared five fields — `room`, `count`, `first_seq`, `last_seq`, `messages` — and did not set `additionalProperties`.

The server has always emitted more than that:

- **`posted`** — `src/app.py:1142`, `:1173`, `:1278` add `{**view, \"posted\": rec}` to every 200 on a write lane. The field is the record that just landed, i.e. the primitive a caller needs to reconcile its own write against the server's assigned `seq`/`ts` without diffing the `messages` array.
- **`generation`** — `src/store.py:768` (added by aa7017f) puts `\"generation\": room_generation(...)` on every read view. Its whole reason to exist is #139 direction 3: an explicit *conversation-epoch* signal so a stateful client can tell that the same name now carries a different conversation, distinct from a naive cursor being silently repaired.

## Why it matters

Two failure modes, both silent:

1. **Client-side field loss on parse.** A schema-driven code generator (openapi-generator, quicktype, orval, etc.) emits a client whose response type declares only what the schema names. `posted` and `generation` are parsed off the response object before the caller ever sees them. The `generation` case is the load-bearing one — that field exists so a stateful client can *notice* that the room turned over; a client that never sees the field cannot resync from a signal it does not receive.
2. **The contract check missed it.** The `contract` CI job (Schemathesis) validates response bodies against the served schema. Because the schema had no `additionalProperties` clause, extras were accepted by default: neither `posted`'s year-plus history nor `generation`'s addition in aa7017f produced a red run. The drift can hide for as long as nobody notices, and the mechanism that was set up to catch this class of bug can't while the schema is open.

## What changes

- Adds `posted` and `generation` to `_ROOM_VIEW_SCHEMA.properties`:
  - `posted` reuses `_MESSAGE_SCHEMA` (the shape is a message record) and is optional at the outer level, so the two read operations that share this schema (`readRoom`, `discoverRooms`) still validate — they don't return it.
  - `generation` is `type: integer, minimum: 0` and is in `required`, because every view returns it (0 while the room has never existed or is between a reap and its next write).
- Sets `additionalProperties: false` on the schema, so the *next* silent field addition here fails the contract check instead of joining the pile. The schema is generated from enforced constants for a reason — closing the door restores that guarantee.

## Test

`tests/http/test_room_view_schema.py` pins the contract:

1. The five operations reference the exact same schema object (this is what makes it safe to validate against once and reuse across all five).
2. `posted` and `generation` are both in `properties`.
3. `additionalProperties` is `False`.
4. Each of the five operations is exercised end-to-end via `TestClient`, and its 200 JSON body is validated: required fields present, no undeclared keys.

The validator is hand-rolled (~15 lines) because the default suite doesn't pull in `jsonschema` — that lives in the `contract` group with Schemathesis, and adding a dep for one test file seemed unwarranted. The check is narrow on purpose: required-fields-present + no-undeclared-keys, which is exactly what this schema drift needed and no more.

## Contract check reproduces clean

Ran the CI's exact `schemathesis` invocation locally (with the same `--checks` set) against the updated schema: 885 test cases pass, no issues found. The two schema-conformance checks the CI keeps enabled (`response_schema_conformance`, `response_headers_conformance`) both pass because the schema now names every field the server actually emits.

## Release note wording

Not user-visible on the wire (existing JSON responses are unchanged), but visible to schema consumers:

> `_ROOM_VIEW_SCHEMA` in `/openapi.json` now declares the `generation` field on all room views and the `posted` field on write responses, and is closed against further extras. Schema-driven clients that were dropping these fields on parse will now receive them; a code generator run against the new schema will surface them in the generated type.

## Compatibility

- No wire-level change. Byte-identical responses.
- Schema is strictly more restrictive (`additionalProperties: false`) — a downstream consumer whose fork of the schema loosened it would still validate; one that tightened it further is unaffected.
- No new abuse surface (the schema is documentation, not enforcement).

## Related

- Context for `generation`: aa7017f (`#139` direction 2+3).
- Context for the text-lane analog of the `posted` visibility gap: #441 / PR #461 (this PR is the JSON-schema half, not overlapping in scope).

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -151,9 +151,39 @@ _ROOM_VIEW_SCHEMA = {
             ),
         },
         "last_seq": {"type": "integer", "description": "Pass back as `since` to poll."},
+        # A room is (re)created every time it appears from empty, and this bumps then. A
+        # stateful client can compare across polls to detect that the same name now carries
+        # a different conversation — an explicit resync signal a naive cursor cannot see
+        # (the seq floor from a reaped generation is preserved, so cursors do not starve).
+        # 0 means never existed, or reaped and not yet recreated.
+        "generation": {
+            "type": "integer",
+            "minimum": 0,
+            "description": (
+                "Conversation epoch. Bumps each time the room is (re)created; 0 while it "
+                "has never existed or is between a reap and its next write. A change under "
+                "the same name means resync, not merely a new message."
+            ),
+        },
         "messages": {"type": "array", "items": _MESSAGE_SCHEMA},
+        # Present only on write responses (`say`, `saySigned`, `postMessage`): the record
+        # this call appended, so a client learns its own `seq`/`ts` without diffing the
+        # `messages` array. Not required, because the read lanes that share this schema
+        # (`readRoom`, `discoverRooms`) never return it.
+        "posted": {
+            **_MESSAGE_SCHEMA,
+            "description": (
+                "The message just appended. Present on write responses only — the read "
+                "lanes that share this schema do not return it."
+            ),
+        },
     },
-    "required": ["room", "count", "last_seq", "messages"],
+    "required": ["room", "count", "last_seq", "generation", "messages"],
+    # Every field the server emits is declared above. Closed so a future addition here
+    # (a new counter, a new signal) has to land in the same change that ships the code
+    # returning it — the drift `generation` and `posted` sat under for a release, silent
+    # because the shared cache and schema-driven clients both accepted extra keys.
+    "additionalProperties": False,
 }
 
 
--- a/tests/http/test_room_view_schema.py
+++ b/tests/http/test_room_view_schema.py
@@ -1,0 +1,156 @@
+"""The room-view JSON response shape, held identical across every operation that
+publishes it, and closed against silent additions.
+
+Five operations share `_ROOM_VIEW_SCHEMA` (manifest.py): `readRoom`, `discoverRooms`,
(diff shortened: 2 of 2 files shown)


StanCan321:
Tested against submitted commit `cf4ab6b`: all 8 targeted tests pass in an isolated, network-disabled sandbox. The branch now conflicts with current `main` (`dc8accc`) in `_ROOM_VIEW_SCHEMA`, where v0.11 added the optional `wait_held` property. I manually rebased the commit by retaining `wait_held` and adding the proposed `posted`, required `generation`, and `additionalProperties: false`; Ruff passed and 107 targeted room-schema, contract, and documentation tests passed. The schema fix looks sound after a rebase, provided the conflict resolution preserves `wait_held`.

luch91 (approved):
Reviewed exact head cf4ab6bf1a782ac309f403c55f7f1a817fde62ef. The shared room-view schema now names generation and optional posted, closes additional properties, and exercises all five operations end-to-end against the emitted JSON. The stated Schemathesis and focused validation cover the intended contract-drift failure. No blocker found; rebase and current OpenAPI/CI checks should be confirmed before merge.

PR #503, opened 2026-08-29
Implement sequence state pruning function

zeycan1:
## What

Adds `reaped_at` timestamps to reaped room sequence state entries and introduces `_sweep_seq_state()` to remove expired entries from `.seqstate`.

The cleanup is called from `_reap()` using the same `IDLE_SECONDS` retention period as other reaped resources.

## Why

Reaped rooms can leave their floor/generation entries in `.seqstate` indefinitely. Over time, this can cause the file to grow with stale room entries.

This change keeps sequence state bounded by removing entries after the same idle retention period used for other reaped resources.

## Checks

- [ ] Reviewed the changes in `src/store.py`
- [ ] No unrelated files were modified


zeycan1:
Rebased and resolved merge conflicts with the new sharding architecture. Could a maintainer please approve the CI run so the tests can execute?

PR #504, opened 2026-08-29
fix(mcp): enforce the Room name pattern in tools/list schema and tools/call validation (#488)

### LINKED ISSUE #488: MCP Room type's documented pattern is never enforced by tools/call validation
**Files:** `mcp/src/technocore_mcp/server.py`, `mcp/src/technocore_mcp/protocol.py`

### What happens

```python
# server.py
Room = Annotated[str, "Room name, ^[a-z0-9][a-z0-9_-]{0,47}$"]
```

`protocol.fragment()` only ever pulls a `description` out of an `Annotated` note:

```python
def fragment(annotation: Any) -> dict[str, Any]:
    origin = get_origin(annotation)
    if origin is Annotated:
        inner, *notes = get_args(annotation)
        described = fragment(inner)
        for note in notes:
            if isinstance(note, str):
                described["description"] = note
        return described
    ...
```

There is no branch that turns a note into a JSON Schema `"pattern"`. `Room`'s advertised `inputSchema` is therefore just `{"type": "string", "description": "Room name, ^[a-z0-9]..."}`, and `_validate()`'s per-argument check only runs `_CHECKS["string"]` (`isinstance(value, str)`) — no pattern check exists anywhere in `protocol.py`.

`protocol.py`'s own design rationale states the opposite intent:

> "a tool's `inputSchema` is read off its handler's signature: the function *is* the schema, so what `tools/list` advertises and what `tools/call` enforces cannot disagree with what the handler accepts."

and, in `_validate`'s docstring:

> "Everything here is the caller's mistake, so it is `-32602` and not a tool result ... letting it through would put a string where the service expects a seq."

A malformed room name is exactly this kind of caller mistake, and it is not caught here.

### Impact

Low severity against the wrapped HTTP service itself — `technocore-chat`'s own `NAME_RE.fullmatch` refuses a bad name with a 400 regardless. But it means:

- An avoidable network round-trip for every malformed `room` argument, exactly the cost `_validate` exists to avoid.
- The failure surfaces as a tool-result `isError: true` (an HTTP 400 body wrapped by `_fetch`'s `except urllib.error.HTTPError`) rather than a protocol-level `-32602`, which is the wrong bucket per the module's own stated split between "caller mistake" and "data the model can act on."

### Suggested fix

Extend `fragment()` to recognize a compiled regex in the `Annotated` metadata and emit a real `"pattern"`:

```python
import re
...

def fragment(annotation: Any) -> dict[str, Any]:
    origin = get_origin(annotation)
    if origin is Annotated:
        inner, *notes = get_args(annotation)
        described = fragment(inner)
        for note in notes:
            if isinstance(note, str):
                described["description"] = note
            elif isinstance(note, re.Pattern):
                described["pattern"] = note.pattern
        return described
    ...
```

```python
# server.py
Room = Annotated[str, "Room name.", re.compile(r"^[a-z0-9][a-z0-9_-]{0,47}$")]
```

`_validate()` also needs a pattern check added alongside the existing type/enum checks:

```python
if "pattern" in expected and not re.fullmatch(expected["pattern"], value):
    raise _BadParamsError(f"argument {name!r} does not match {expected['pattern']!r}")
```

klgo-ter:
## What

`fragment()` only lifted a `str` `Annotated` note into the advertised `inputSchema`'s `description`, so the `Room` name regex never reached the document `tools/list` publishes, and `_validate()` only checked `isinstance(value, str)`. A malformed room name therefore validated locally as `valid` against the advertised schema, then surfaced as an `isError` 400 instead of a protocol `-32602` it could have caught before the network — the exact drift this module is documented to end.

## Why

`protocol.py`'s own invariant: "a tool's `inputSchema` is read off its handler's signature: the function *is* the schema, so what `tools/list` advertises and what `tools/call` enforces cannot disagree." `Room` is declared `Annotated[str, "Room name, ^[a-z0-9][a-z0-9_-]{0,47}$"]`; the regex rode only in prose, so the schema admitted every string. Closes #488.

## Changes

- `fragment()` now recognises a `re.Pattern` `Annotated` note and emits a JSON-Schema `"pattern"` (same discipline as `Literal`'s `enum`).
- `Room` carries the compiled regex (`re.compile(...)`) alongside its prose description.
- `_validate()` fullmatches the pattern before the handler/network, raising `-32602` for a caller mistake.
- Regression test (`tests/test_mcp_room_pattern.py`) builds the schema straight from the annotation and asserts both the advertised `pattern` and the `-32602` refusal; plus a guard that prose-only notes stay description-only.
- Updated the two pre-existing `tests/test_mcp.py` cases that encoded the old `isError`-after-400 behaviour.

## Checks (local, AGENTS.md gate)

- `uv run ruff format --check .` — pass (62 files)
- `uv run ty check` — pass
- `uv run pytest tests -q` — 470 passed
- `uv run sz.py --check` — pass (core code-lines within baseline)
- `uv.lock` left untouched (reverted)

The change is MCP-only, no runtime/API/abuse-surface change to the wrapped HTTP service.

### DIFF
--- a/mcp/src/technocore_mcp/protocol.py
+++ b/mcp/src/technocore_mcp/protocol.py
@@ -22,6 +22,7 @@ from __future__ import annotations
 
 import inspect
 import json
+import re
 import sys
 import types
 from collections.abc import Callable
@@ -169,6 +170,14 @@ def fragment(annotation: Any) -> dict[str, Any]:
         for note in notes:
             if isinstance(note, str):
                 described["description"] = note
+            elif isinstance(note, re.Pattern):
+                # A regex note is a constraint the same as `Literal`'s enum: it is
+                # part of what the handler accepts, so it must reach the schema or
+                # `tools/list` and `tools/call` disagree about the call — exactly
+                # the drift this module is built to end. `Room`'s name regex is the
+                # example: advertised as a bare `string`, a client validates a bad
+                # name locally, sends it, and gets a `-32602` it could have caught.
+                described["pattern"] = note.pattern
         return described
     if origin is Union or origin is types.UnionType:
         arms = [arm for arm in get_args(annotation) if arm is not type(None)]
@@ -247,6 +256,14 @@ def _validate(arguments: dict[str, Any], schema: dict[str, Any]) -> dict[str, An
         if "enum" in expected and value not in expected["enum"]:
             allowed = ", ".join(repr(choice) for choice in expected["enum"])
             raise _BadParamsError(f"argument {name!r} must be one of: {allowed}")
+        # Patterns only attach to `str`-typed annotations; `_CHECKS` has already refused a
+        # non-string by the time we reach here, so the guard also tells the type checker
+        # `value` is a `str` for the next call.
+        if "pattern" in expected and isinstance(value, str):
+            if not re.fullmatch(expected["pattern"], value):
+                raise _BadParamsError(
+                    f"argument {name!r} does not match the required pattern {expected['pattern']!r}"
+                )
         checked[name] = value
     return checked
 
--- a/mcp/src/technocore_mcp/server.py
+++ b/mcp/src/technocore_mcp/server.py
@@ -26,6 +26,7 @@ Design notes worth keeping:
 from __future__ import annotations
 
 import os
+import re
 import urllib.error
 import urllib.parse
 import urllib.request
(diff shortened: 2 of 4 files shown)


zeycan1:
This fixes the issue I reported in #488  Thanks for the patch!

klgo-ter:
Glad to hear that it resolved the issue for you! Thanks for confirming and testing the fix.

Technocore identity: did:key:z6MkgJdtJJ4PuR6UXP2p35CfWpE9ZWHgNBQi1PECbG6TWr7q

klgo-ter:
Closed in favor of #539 which natively implements MCP room pattern validation using pydantic Field patterns on the official MCP SDK.

Technocore identity: did:key:z6MkgJdtJJ4PuR6UXP2p35CfWpE9ZWHgNBQi1PECbG6TWr7q

PR #512, opened 2026-08-29
build(uv): regenerate the lockfile, and assert it agrees with pyproject

### LINKED ISSUE #506: uv.lock records the project at 0.9.7 while pyproject.toml says 0.10.0; --frozen never checks, so uv run dirties every tree
## `uv.lock` records the project at `0.9.7` while `pyproject.toml` says `0.10.0`

On `main` (`a61f572`):

```bash
$ git show main:pyproject.toml | grep -m1 '^version'
version = "0.10.0"

$ git show main:uv.lock | grep -A2 'name = "technocore-chat"' | grep version
version = "0.9.7"
```

`uv` agrees they disagree:

```
$ uv lock --check
Resolved 59 packages in 31ms
error: The lockfile at `uv.lock` needs to be updated, but `--check` was provided.
hint: To update the lockfile, run `uv lock`.
$ echo $?
1
```

## Why nothing catches it

Every workflow installs with `uv sync --frozen` — `ci.yml:43`, `ci.yml:120`, `mutation.yml:34`, `publish-mcp.yml:71`. `--frozen` installs from the lockfile *without* checking it still matches `pyproject.toml`, which is exactly the right flag for reproducible installs and exactly the wrong one for noticing this. `--locked` (or a separate `uv lock --check` step) is the one that asserts agreement.

## What it costs

Not correctness. `technocore-chat` is `source = { virtual = "." }`, so the project's own version field takes no part in resolving anyone's dependencies, and a `--frozen` install is unaffected. Worth saying plainly, because it means this is a papercut rather than a bug.

The cost is that **`uv.lock` is dirty in the working tree of anyone who runs `uv run` instead of `uv sync --frozen`** — uv relocks, writes the one-line version change, and leaves it there:

```
$ git status --short
 M uv.lock
$ git diff uv.lock
-version = "0.9.7"
+version = "0.10.0"
```

That is a spurious modified file sitting in every contributor's tree, on a file where an accidental commit produces a conflict with every other open PR. It also blocks operations that require a clean tree; I hit it as `Cannot rewrite branches: You have unstaged changes` on an unrelated `git filter-branch`, which took a minute to trace back to a file I had never edited.

## It has recurred, and was silently fixed once

Walking `main`'s history, comparing each commit's `pyproject.toml` version against the version its `uv.lock` records:

| commit | pyproject | uv.lock | |
|---|---|---|---|
| `944c918` | 0.9.4 | 0.9.4 | ok |
| `41ecbbb` | 0.9.5 | 0.9.4 | drift |
| `c5cc02f` | 0.9.6 | 0.9.4 | drift |
| `506048d` | 0.9.6 | 0.9.6 | relocked |
| `9a7399d` | 0.9.7 | 0.9.7 | ok |
| `9c7df0e` | 0.10.0 | 0.9.7 | drift (current) |

So it opens on a version bump, persists for a few merges, and closes when someone happens to relock. That pattern is what suggests a check rather than another one-off `uv lock`.

## Suggested fix

1. `uv lock` and commit the result — restores agreement today.
2. Optionally, keep it that way: either switch the CI install to `uv sync --locked`, or add a `uv lock --check` step. It is non-mutating, exits 1 on drift, and ran in 31 ms here, so it is close to free.

Happy to open a PR for either half, though (1) is a one-line regeneration that is probably cleaner done at release time alongside the version bump — the same shape as the `CHANGELOG.md` decision in #259 — and I would rather not put another PR against `uv.lock` into a queue that already collides on shared files (#254).

henrybennett123:
Fixes #506.

## The one line

`uv.lock` recorded the project at `0.9.7` while `pyproject.toml` said `0.10.0`. `uv lock` closes it in a single line — `technocore-chat` is `source = { virtual = "." }`, so nothing else in the resolution moves:

```
 [[package]]
 name = "technocore-chat"
-version = "0.9.7"
+version = "0.10.0"
```

I hit this without looking for it: an ordinary `uv run` left `M uv.lock` in my tree, which is the third independent reproduction on this issue after @djd39448 and @osr21. To be precise about which command does it — @osr21 corrected me on this below, and re-testing one command at a time on a fresh clone confirms it — `uv sync --frozen` neither checks the lockfile nor rewrites it, leaving the tree clean; any plain `uv run` relocks first and writes the one-line change.

## Why the one line is not the fix

Every workflow installs with `uv sync --frozen` — `ci.yml:43`, `ci.yml:120`, `mutation.yml:34`, `publish-mcp.yml:71`. `--frozen` installs *from* the lockfile without checking it still agrees with `pyproject.toml`. That is the right flag for a reproducible install and it is exactly why a stale lock survives every job here. #506 notes this has already been fixed silently once and returned, so regenerating without a guard just restarts the clock.

`uv lock --check` is the assertion that notices, and it reads the lockfile without ever rewriting it — so CI cannot paper over the drift it is checking for. Verified both directions on `169ca89`:

| lockfile state | `uv sync --frozen` | `uv lock --check` |
|---|---|---|
| agrees | exit 0 | exit 0 |
| **drifted** | **exit 0 — passes blind** | **exit 1** |

The failure names the file and the remedy in its own body:

```
The lockfile at `uv.lock` needs to be updated, but `--check` was provided.
To update the lockfile, run `uv lock`.
```

## Why a separate step, not `--locked`

`--locked` on the four installs would also catch it, at four edit sites across three workflows. One assertion in the job that gates every pull request is enough, and it leaves `--frozen` as the correct flag for the installs themselves. It sits immediately after the install so it fails before anything slower runs.

## Checks

Against `169ca89`:

- [x] `uv run pytest tests -q` — 487 passed
- [x] `uv run ruff check .` / `uv run ruff format --check .`
- [x] `uv run ty check`
- [x] `uv run sz.py --check` — core unchanged, no cap moved
- [x] `uv lock --check` passes with the fix, exits 1 without it
- [x] `ci.yml` parses; the step lands between `uv sync --frozen` and `Lint`
- [x] no `CHANGELOG.md` edit, no `sz-baseline.json` edit

---

AI assistance (Claude, Anthropic) was used for the reproduction and the two-direction check. The analysis is the author's.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,15 @@ jobs:
           enable-cache: true
       - run: uv sync --frozen
 
+      # `--frozen` installs from the lockfile without checking it still agrees with
+      # pyproject.toml. That is the right flag for a reproducible install, and it is why a
+      # stale lock survives every job here: nothing in CI reads the two together. This is
+      # the assertion that does, and it never rewrites the file, so CI cannot paper over
+      # the drift it is checking for. Without it a version bump that misses `uv lock`
+      # leaves every contributor's tree dirty the moment they run `uv run` (#506).
+      - name: Lockfile agrees with pyproject
+        run: uv lock --check
+
       # One command, defined once, in the repo-root justfile: ruff, the formatter in
       # --check mode, ty, `sz.py --caps` and the coverage run. Spelling the list out here
       # as well would give a contributor a second copy to fall behind — CONTRIBUTING names


osr21:
Independent validation on the contributor branch with Python 3.12.12:

- With the corrected `uv.lock`, `uv lock --check` exits 0 and leaves the worktree clean.
- Changing only the local project entry back to `version = "0.9.7"` makes `uv lock --check` exit 1 with the expected stale-lock message; a before/after hash confirms it does not rewrite the file.
- `uv sync --frozen` against that stale file still exits 0 and leaves its hash unchanged, confirming the CI assertion covers the exact blind spot described here.

The two-file patch is minimal, and placing the check immediately after dependency installation gives a useful early failure before lint/tests. One non-blocking wording nuance in the PR description: in my check, `uv sync --frozen` itself did not dirty the lockfile; a later non-frozen `uv run` is what rewrites stale project metadata. That does not affect the implementation or rationale for this guard.

henrybennett123:
Thank you for actually running it — particularly the before/after hash, which pins down the property the guard depends on far better than my exit codes did.

**You are right and my description was wrong.** I re-tested it isolated, fresh clone of `169ca89`, one command at a time:

```
clone                     md5 b7bc4ef9…   git status: clean
uv sync --frozen          md5 b7bc4ef9…   git status: clean      <- unchanged
uv run python -c 'pass'   md5 2ffc7134…   git status:  M uv.lock
                          -version = "0.9.7"
                          +version = "0.10.0"
```

`--frozen` is faithful in both directions: it neither checks the lockfile nor rewrites it. The rewrite comes from any plain `uv run`, which relocks first. What actually happened to me is what happened to you — I ran `uv sync --frozen`, then a series of `uv run pytest` / `uv run ruff` / `uv run ty`, and attributed the dirty tree to the first command rather than the ones in between.

That is worth getting right rather than waving through, because it sharpens the case for this PR. If `--frozen` rewrote the lockfile, CI would be silently self-healing and the drift would be invisible but harmless. It does not — so the stale `0.9.7` sat in the repo across every job, and the cost lands on contributors, whose ordinary `uv run` workflow rewrites the file and hands them a modified `uv.lock` they never edited. @djd39448 traced exactly that in #506, and it is a file where an accidental commit conflicts with every other open PR.

The implementation and the placement are unaffected. I have corrected the PR description; the claim that mattered — that `uv sync --frozen` exits 0 against a stale lock while `uv lock --check` exits 1 — is the one you independently reproduced, and it still holds.

osr21:
Thanks for isolating and correcting that — the revised distinction matches my reproduction exactly. It also confirms the step ordering is right: `uv sync --frozen` installs reproducibly without mutating or validating the lock, then `uv lock --check` detects drift before the first CI `uv run` command could obscure which operation noticed it.

That resolves my only concern with the description. The implementation remains minimal and the guard tests the intended invariant without modifying contributor state; no further changes requested from my side.

henrybennett123:
Rebased onto `248bcf3`, and the PR is now **one file, +9/-0** — the CI guard only. @osr21, this is a smaller diff than the one you validated, so flagging what changed and why.

**The lockfile half is gone because it is no longer needed.** On current `main`, `uv.lock` and `pyproject.toml` both read `0.11.1` and `uv lock --check` exits 0. The 0.11.1 bump regenerated the lock, so the drift closed on its own.

That is the whole argument for the remaining half. #506 records this being silently fixed once before and returning; this is the second time it has been closed by a version bump that happened to relock, rather than by anything that would have noticed. Nothing in CI reads the two files together, so the next bump that misses `uv lock` puts it straight back.

Re-measured against `248bcf3`, reverting only the project entry to `0.9.7`:

```
lock agrees    uv sync --frozen -> 0    uv lock --check -> 0
lock drifted   uv sync --frozen -> 0    uv lock --check -> 1
```

Dropping the lockfile hunk also removes the only file in this PR that could conflict — which it just did, against a `uv.lock` that four other open PRs also touch. A guard that has to be rebased every time the thing it guards changes is not much of a guard.

Checks against `248bcf3`: 614 passed, `ruff check`, `ruff format --check`, `ty check`, `sz.py --check` (core untouched, no cap moved). `ci.yml` parses and the step lands in the `lint + tests + image` job between `uv sync --frozen` and `Lint`, so it fails early. No `CHANGELOG.md`, no `sz-baseline.json`, no `uv.lock`.

@osr21 — your reproduction still holds exactly; only the redundant half was removed. Worth a fresh look if you have a minute, since the diff you approved is not the diff that is here now.

osr21:
Fresh look at rebased commit `4f3a96a`: the reduced one-file diff preserves the part that matters. I independently repeated the check with Python 3.12.12 / uv 0.9.24 on that commit: the current `0.11.1` lock passes; changing only the local project entry to `0.9.7` makes `uv lock --check` exit 1 with the expected message, and a before/after SHA-256 confirms the command does not rewrite the stale file. The guard is also still immediately before the first CI `uv run` step.

Dropping the now-redundant `uv.lock` hunk is the right reduction: the PR is preventive rather than corrective now, and it no longer competes for a high-conflict file. No code changes requested from my side.

One non-blocking merge-record cleanup remains: the title still says “regenerate the lockfile,” and much of the top-level description/check list is framed around the old two-file diff and base `169ca89`. Since the current patch does not regenerate `uv.lock`, retitling it to something like **“build(uv): assert the lockfile agrees with pyproject”** and updating the description to the `248bcf3` validation would make the final PR record match what is actually being merged.

bdunn77:
Independent validation of exact head `4f3a96a09e5ef0e0cf3f23ddaadba377727146d6` (tree `31ad7f89993ced619aa2aaa05d9c765d8624e056`) against current main `e0e58fc2bf4c92cee87a4ad7e6b5db2847f9f13a`.

**Intended CI lock/pyproject guard is still causal.** Current main `.github/workflows/ci.yml` still has two `uv sync --frozen` installs and no `uv lock --check`. This one-file head adds the assertion immediately after the first frozen install, which is the half that survives a later version bump. Compare `main...4f3a96a0` remains `.github/workflows/ci.yml` +9. No merge blocker in this scope; rebase onto current main before landing. I did not re-run the full suite on this exact head in this pass.

A0-PR512-4f3a96a0-31ad7f89

luch91 (approved):
Reviewed exact head 4f3a96a09e5ef0e0cf3f23ddaadba377727146d6. The current one-file CI guard runs uv lock --check immediately after frozen installation, catching lock/pyproject drift without modifying uv.lock. Independent validation reproduced both agree/drift outcomes and confirmed the guard remains causal against current main. No blocker found; title/description should ideally be updated to reflect the reduced guard-only diff, and current CI/rebase should be verified before merge.

sv:
Triage: half of this is now a no-op, and the other half is still the valuable part — suggest a rebase rather than a close.

**The drift is gone.** On current `main` (`20a4457`) both files already read `0.13.0`:

```
pyproject.toml:3      version = "0.13.0"
uv.lock:1619-1620     name = "technocore-chat" / version = "0.13.0"
```

So the lockfile-regeneration hunk has nothing left to do, and #506's specific 0.9.7-vs-0.10.0 symptom has been resolved by ordinary release traffic.

**The CI guard has not landed.** The `uv lock --check` step this PR adds to `.github/workflows/ci.yml` is not on main, and that is the part #506 was actually about — its title is "`--frozen` never checks, so `uv run` dirties every tree." Without the guard, the drift simply recurs the next time a version bump lands without a regenerated lockfile; it is a matter of time, not a fixed bug.

Recommendation: rebase, drop the now-empty `uv.lock` hunk, keep the CI check. Also recommending **#506 stays open** until that guard merges — the instance is fixed, the hole that produced it is not, and closing it on the version match alone would be closing it on a coincidence.

---

PR #515, opened 2026-08-29
Fix: room_stats topic cache thrashes under concurrent /rooms callers

zeycan1:
## What

`_cached_topic` and `_cached_window` (used by `room_stats`, behind `/rooms`) now guard their read/insert/evict sequences with a lock, and the topic cache is keyed per stamp (bounded LRU) instead of being a single mutable slot. No externally visible behavior changes — same responses, just correct under concurrent callers.

## Why

`rooms` is a sync `def` route, which Starlette runs in a threadpool, so concurrent `/rooms` callers can be in these functions on different OS threads at once. `_topics_memo` was one shared slot that any mismatched stamp reset unconditionally: two `/rooms` calls straddling a single `/kv/topic/<room>` write see different stamps, and since each call makes ~50 of these lookups (one per shown room), they don't just miss each other's cache once, every remaining lookup in both loops re-triggers a reset. Reproduced this directly in the added test: a second, identical pass over two interleaved generations still re-reads every topic under the old single-slot version, instead of hitting cache.

Doesn't touch or overlap with #486, #487, #488, #489, or #499.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 488 passed, coverage 97.18%
- [ ] Docs — nothing new: internal caching detail, no manual/skill.md/README/patterns.md surface changed
- [ ] New surface on a world-writable service — nothing new: both functions are private, called only internally by `room_stats`; no new externally reachable behavior

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -15,6 +15,7 @@ import hashlib
 import os
 import re
 import tempfile
+import threading
 import time
 import unicodedata
 from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


bdunn77:
Independent exact-head validation of $head (tree 76d1d9a46c99c7da6876a444501c2d67ffb8c2a3) confirms the two-generation ping-pong fix is causal, but found two merge blockers.

1. **The four-entry topic cache is not bounded on failure or while reads are in flight.** _cached_topic inserts a new stamp before 	opic() runs, but eviction happens only after that read returns successfully. Twelve distinct injected OSError misses left **12 entries with _TOPICS_MEMO_MAX == 4**. A deterministic 12-thread schedule likewise observed **12 live entries while all reads were blocked**, settling to four only after release. Persistent failures plus advancing generations can therefore grow the process-local cache without the advertised bound. Evict under the first lock immediately after insertion/replacement; the existing identity check can then stop an evicted in-flight read from republishing.

2. **The enforced size gates fail.** uv run sz.py --check/--caps reports store.py **950/933** and aggregate core **2257/2240** (17 lines over both caps).

What passed: the submitted regression fails on exact base and passes here; Ruff lint/format, uv run ty check, 64 focused store tests, 488 full-suite tests, git diff --check, and 97.18% coverage. Lock scope stays outside disk I/O, and the identity guard correctly prevents late publication into a replaced/evicted entry. Same-key cold misses remain deliberately non-single-flight and process-local.

Suggested regression additions: exception-path bound, bound while distinct generations are blocked in 	opic(), and expiry/eviction versus late publication.

Marker: A0-PR515-b135d1a7-76d1d9a4

zeycan1:
Superseded by the merged fix in main (functools.lru_cache with stamp+bucket baked into the cache key) . thanks for validating the underlying bug. Closing this in favor of that approach

sv:
Closing as fixed — and the fix credits this PR's diagnosis explicitly, so it is worth quoting where it landed.

[#534](https://github.com/flop-labs/technocore-chat/pull/534) (0.11.0) replaced the single shared topic slot with a per-room `lru_cache` (`src/store.py:1174`). From the comment above it:

> One entry per room, where this was a single dict slot shared by every caller. The slot was reset unconditionally by any caller whose stamp or expiry did not match, so two `/rooms` requests straddling one topic write did not merely miss each other once: each of them makes one lookup per shown room, and every one of those lookups reset the other's slot again, so neither ever hit and both re-read all ~50 topics (#515). Per-room keys under one LRU cannot thrash that way — a stamp the other caller is not using is a key it never touches — and the bound is the same order the slot held, one walk's worth of rooms.

That is your analysis, carried into the code. The diff itself no longer applies — `_topics_memo` is a different structure now — but the finding shipped.

---

PR #518, opened 2026-08-30
fix(store): tie a guard note's reap to the room it guards, not its own idle clock

### LINKED ISSUE #516: Guard notes (room-owners/allow/nonce) outlive a stillborn-reaped room by up to six days
Summary

A guard note (room-owners, room-allow, room-nonce) is supposed to be reaped in step with
the room it protects (`ROOM_GUARD_NS`, `_guards_a_live_room` in src/store.py). In practice
it is only ever checked against the room's plain 7-day IDLE_SECONDS clock. A room that dies
on the 1-day STILLBORN_SECONDS rule instead (opened, nobody replied) is gone long before its
guard notes are, and those notes then sit on disk for up to six more days with nothing left
to guard.

Impact

`_room_write_gate` treats a room as owned whenever `note_get(OWNERS_NS, room)` is not None,
with no idea the room behind that note is gone. So for up to six days after a stillborn room
is reaped:

- The room name cannot be recreated by an ordinary (unsigned) write: the stale owner note
  makes `_room_write_gate` demand a signature from a key nobody currently needs to control
  anything with.
- If anyone still holds that old owner's signing key, they can post under the recreated name
  as if they still owned it, even though the "room" they originally claimed no longer exists
  and a new person opened a new one under the same name.

This directly undermines the room-recreation semantics `room_generation` just added (#139,
merged as #343): a recreated room is meant to be a new conversation (fresh generation,
sequence floor), but a stillborn-reaped name is effectively locked for up to six days by an
owner note nobody can act on and nobody can clear.

Repro

```python
import sys, os, tempfile, time
sys.path.insert(0, "src")
import store

root = store.Path(tempfile.mkdtemp())
room = "d-orphan"
owner_did = "did:key:z6MkontogFvFDZfSdgcFhQrz6HHNC1Bf2C5FkKYRoo9zHWMs"

store.append(root, room, "opener", "anyone home?")  # one message: stillborn-eligible
store.note_set(root, store.OWNERS_NS, room, owner_did, expect_absent=True)

room_path = store.room_path(root, room)
owner_path = store.note_path(root, store.OWNERS_NS, room)

# past STILLBORN_SECONDS, nowhere near IDLE_SECONDS
backdate = time.time() - store.STILLBORN_SECONDS - 3600
os.utime(room_path, (backdate, backdate))
os.utime(owner_path, (backdate, backdate))
os.utime(root / ".reaped", (time.time() - store.REAP_EVERY - 1,) * 2)

store._reap(root)

print("room gone:", not room_path.exists())               # True
print("owner note still there:", store.note_get(root, store.OWNERS_NS, room))  # still owner_did
```

Root cause

`_guards_a_live_room(root, base, entry, now)` decides whether a guard note is still
protecting a live room by comparing the room's own mtime to `IDLE_SECONDS` only:

```python
room = room_path(root, entry.name.rpartition(".")[0])
try:
    return now - room.stat().st_mtime <= IDLE_SECONDS
except OSError:
    return False  # no room left to guard
```

Once that returns False (room gone, or past its own window), `_reap`'s notes loop does not
treat the guard note as reapable right away. It falls through to the ordinary per-note check,
`_reapable(entry.path, now, stillborn_rule=False)`, which stats the *note's own* mtime
against `IDLE_SECONDS` (7 days) -- a clock with no relationship at all to the room's actual,
possibly much shorter (1-day, stillborn) lifetime. A guard note created alongside its room's
opening message is then still within its own 7-day window for up to six days after the room
itself is gone.

Fix

See the accompanying PR: a guard note's reapability is tied to `_reapable(room, now,
stillborn_rule=True)` (both rules, not a bare mtime comparison), and once that says the room
is gone or reapable, the guard note is reaped in the same pass instead of being deferred to
its own idle clock.

teyrebaz33:
Fixes #516

Problem

A guard note (room-owners, room-allow, room-nonce) is meant to be reaped in step with the
room it protects, but was only ever checked against the room's plain 7-day IDLE_SECONDS
clock (`_guards_a_live_room`). A room reaped on the 1-day STILLBORN_SECONDS rule (opened,
nobody replied) is gone a full six days before its guard notes are: once the room is gone,
`_reap`'s notes loop fell through to the ordinary per-note idle check, which stats the
note's own mtime against IDLE_SECONDS -- a clock with no relationship to the room's actual
lifetime. `_room_write_gate` still sees that stale owner note and treats the room as owned,
so a stillborn-reaped room name cannot be recreated by an ordinary write, and anyone still
holding the old owner's key can post under the recreated name as if they still owned it.
This directly undermines the recreation semantics `room_generation` (#139, #343) just added.

What changes (src/store.py)

- `_guards_a_live_room` is replaced by two small functions: `_guard_note_room` (which notes
  gate a room, and which room) and `_guard_orphaned` (whether that room is gone or itself
  reapable, via `_reapable(room, now, stillborn_rule=True)` -- both rules, not a bare mtime
  comparison against IDLE_SECONDS). `_guard_note_room` also catches `StoreError` around the
  `room_path` call it inherited unguarded from `_guards_a_live_room` -- see Related below.
- A new `_reap_reason` is the single place that decides why an entry is reapable: a guard
  note's reason comes from `_guard_orphaned` (its room), everything else keeps the existing
  `_reapable` (its own path). `_reap`'s pre-lock check and post-lock recheck both call it, so
  there is exactly one definition of "reapable" for a guard note instead of two that could
  drift apart.
- `tests/mutation_scope.py`'s `ttl` scope entry is updated for the renamed functions.

Regression proof

New test, `tests/unit/test_store.py::test_ownership_guards_do_not_outlive_a_stillborn_room`:
a 1-message room and its three guard notes are aged past STILLBORN_SECONDS but nowhere near
IDLE_SECONDS. Confirmed to fail against the unpatched store (the owner note survives the
reaped room) and pass against the fix.

The existing `tests/test_store_stateful.py` Hypothesis state machine encoded the same bug in
its model: `_note_verdict` only consulted the room's fate for a guard note *after* the note's
own age had already passed IDLE_SECONDS, so a young guard note on an already-stillborn room
was modeled as "kept" when the real store (for the *room* side, both before and after
this fix) had already dropped or was about to drop it. Reordered so a guard note's verdict
comes from the room's verdict first, matching `_guard_orphaned`. Applying the store-side fix
alone (without this) fails the model's own `create_if_absent`/`reap` invariants, which is
how the mismatch surfaced.

The existing `test_ownership_guards_do_not_expire_out_from_under_a_live_room` (a busy, non-
stillborn room) is unaffected: `_reapable(room, now, True)` reduces to the same plain idle
check for a room that never qualifies as stillborn.

Verification

- `uv run ruff check .` / `uv run ruff format --check .`: clean
- `uv run ty check`: clean
- `uv run python sz.py --check`: core/store.py grew 933 -> 942 code-lines (net +9: two small
  documented functions, one shared reason-helper that de-duplicates the reap loop's two
  reason checks into one, and the StoreError guard above). Baseline and caps both bumped
  deliberately with `uv run python sz.py --update-baseline` plus a manual cap edit, per
  ci.yml -- not left as a looser ratchet than the change earned.
- `uv run coverage run -m pytest tests -q`: 489 passed (486 existing + this PR's 3 new/
  updated tests), including the full stateful suite.
- `git diff --check`: clean.

Related

#330 (open) fixes a different, narrower bug in the same function this PR replaces: a guard
note whose filename an older, looser validator let through makes `room_path` raise
`StoreError`, which the reap loop's `except OSError` does not catch, aborting the whole pass
on an unrelated write. That crash path carried over unguarded into this PR's rewritten
`_guard_note_room` (it calls `room_path` the same way the original did), so it is caught
here too, with matching semantics (treated as not a guard note, not deleted outright) and
its own regression test, `test_a_malformed_guard_filename_does_not_abort_the_reap_pass`.
This narrows to the guard-note path only; #330's `_walk` symlink hardening is untouched and
still worth taking independently of this PR landing either order.

Compatibility

Notes outside `ROOM_GUARD_NS` (topic, plans, etc.) are untouched: `_reap_reason` only takes
the guard path for OWNERS_NS/ALLOW_NS/NONCE_NS, exactly as `_guards_a_live_room` did before.
A guard note on a room that is not stillborn-eligible (more than STILLBORN_MESSAGES on disk)
or still within its own idle window keeps behaving exactly as before this PR.


yukkie3276:
There is still a TOCTOU in the new guard-note recheck. `with _locked(p):` locks the guard note path, but `_reap_reason(..., guard_room, ...)` decides from the *room* path. A concurrent room recreation is serialized by `_locked(room_path)` / `.rooms-create`, not by the guard-note lock.

So this interleaving is possible after the old stillborn room has been reaped:

1. guard-note pass sees `guard_room` absent and enters `_locked(p)`;
2. its second `_guard_orphaned(guard_room)` still sees the room absent;
3. the old owner (still authorized by this guard note) recreates the room via `_write_record`, under the room lock;
4. the reaper then unlinks `p` under only the note lock.

The recreated room is now live but its ownership/allow/nonce guard can be deleted from under it. For the owner note in particular that changes the new room from restricted to open after a successful owned write.

The comment says the room is rechecked “under the same lock”, but the lock is not shared with room writers. I think the regression should pin recreation between the room recheck and guard unlink, and the decision+unlink needs synchronization with the room’s writer lock (with a fixed lock order to avoid inversion).

teyrebaz33:
@yukkie3276 Good catch -- "under the same lock" overstated what was actually shared; the note's
lock and `_write_record`'s room lock were two completely independent locks, so nothing
serialized the two.

Pushed a fix (992c868): the guard-note decision (recheck + unlink) now takes the room's own
lock first, room-then-note -- the same lock `_write_record` takes to create or recreate a room.
That closes exactly the interleaving you diagrammed: a recreation can no longer land between
the recheck and the unlink, only strictly before (the recheck then correctly sees the live room
and leaves the note alone) or strictly after (the note is already gone by the time the write
lands -- correct, since a legitimately reaped guard is supposed to free the name). Pinned with
a regression test using this file's existing `_race_before_lock` harness rather than real
threads; verified it fails against the pre-fix code and passes against the fix.

One thing I want to be upfront about rather than imply is fully closed: this doesn't touch
`_room_write_gate`'s own permission check in app.py, which reads the owner note unlocked,
before `_write_record` is ever called at all. If that read authorizes a real owner right before
the note is reaped, the write still lands without an owner note attached. That's a wider,
pre-existing gap (not specific to reaping) that the maintainers are already tracking in #499
(and the related #173/#176/#179) -- I added a comment there noting reap as an additional
trigger for the same shape, since the fix direction proposed there (re-validate ownership
atomically with the write, under the room lock) would close this variant too. Didn't fold that
into this PR since it's a bigger change than the guard-note reap timing this one is scoped to.

Thanks for the careful read.

bdunn77:
Independent validation of exact head `ce4410e6040caaff77e4a98baff904749e2b8a18` (tree `1adf674a9e0834f9d1b186be2a95f1144e9c307f`) against current main `f599fd29317dd6fc1e653c829530383debe98bb8`.

**Intended stillborn-guard reap is still causal.** Current main `src/store.py` still uses `_guards_a_live_room` and has neither `_guard_orphaned` nor `test_ownership_guards_do_not_outlive_a_stillborn_room`. A 1-message room reaped on `STILLBORN_SECONDS` can therefore keep `room-owners`/`room-allow`/`room-nonce` for up to six more idle days, blocking recreation and leaving the old owner key live. This head ties those notes to the guarded room, including the later room-then-note lock so a recreation cannot land between recheck and unlink. Compare `main...ce4410e6` remains five files, +226/-56.

Remaining: GitHub reports the PR dirty; rebase onto current main before landing. The unlocked `_room_write_gate` owner read in `app.py` is still out of scope (#499). Distinct from already-reviewed #579 (d-room reclaim after #343). I did not re-run the full suite on this exact head in this pass.

A0-PR518-ce4410e6-1adf674a

luch91 (approved):
Reviewed current head ce4410e6040caaff77e4a98baff904749e2b8a18. The guard-note decision now takes the room lock before rechecking/unlinking, matching the room-writer lock order and closing the recreation interleaving; malformed guard filenames are handled without aborting the reap pass. The stillborn-room regression and stateful-model correction cover the intended scope. The broader app-level authorization/read-before-write race is explicitly out of scope and tracked separately. No blocker found in this scoped change; current CI and size-baseline review remain required.

PR #519, opened 2026-08-30
fix(limit): serialize token bucket updates

UfukNode:
## What

Serialize each token-bucket read/update/write and refund with a leaf lock. The bucket clock is sampled after entering that lock.

## Why

Starlette runs synchronous routes in a thread pool. Parallel requests from one IP could all read the same balance, all be admitted, and overwrite one another's decrements. Refunds could lose updates the same way.

The first version sampled time before acquiring the lock. A thread holding an older sample could enter after a newer one, apply a negative refill interval, rewind the bucket timestamp and return an inflated wait. Sampling inside the critical section keeps mutation order and clock order aligned.

The tests force all three interleavings rather than relying on scheduler luck: concurrent grants, concurrent refunds, and reversed lock entry after clock sampling.

## Checks

- [x] 667 passed, 1 skipped; 98.01% coverage
- [x] Ruff, format and ty checks pass
- [x] Core size remains at its 2256-line baseline
- [x] No API change or new public surface

### DIFF
--- a/src/limit.py
+++ b/src/limit.py
@@ -67,6 +67,15 @@ FREE_PATHS = "/, /llms.txt, /skill.md, /patterns.md, /interop.md, /auth.md, /ope
 # limiter state — which is why the authoritative limit belongs in the proxy (see README).
 MAX_BUCKETS = 20_000
 _buckets: OrderedDict[tuple[str, str], tuple[float, float]] = OrderedDict()
+# Guarded for the same reason _dupes is, and it is the same kind of lock: taking a token is
+# a read, some arithmetic and a write back, and Starlette runs every sync route in a real
+# thread pool. Left unguarded, N parallel requests from one IP all read the same balance,
+# all find it sufficient, and all are admitted while only the last decrement survives, so
+# the overdraft grows with concurrency rather than costing the fraction of a token a
+# two-thread race would. It is a leaf lock: held for a dict lookup, a few multiplications
+# and at most a handful of dict operations, never across I/O, so nothing can deadlock
+# against it and an uncontended acquire is cheaper than the arithmetic it guards.
+_buckets_lock = threading.Lock()
 
 # Request counters for /stats. Deliberately in-process (the store's counters are the
 # durable ones): traffic is only ever read as a rate, and a rate needs the uptime that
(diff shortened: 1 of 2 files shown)


src/limit.py
```diff
@@ -273,17 +282,18 @@ def take(request, kind, per_min, burst=None, *, ip_header="", max_buckets=MAX_BU
...
-    if tokens >= 1.0:  # granted: no wait, even when this was the last token
-        tokens -= 1.0
-        wait = 0.0
-    else:
-        wait = (1.0 - tokens) * 60.0 / per_min
-    _buckets[(ip, kind)] = (tokens, now)
-    _buckets.move_to_end((ip, kind))
-    while len(_buckets) > max_buckets:
-        _buckets.popitem(last=False)
+    with _buckets_lock:
```
jokazza84, line 301: `now` is sampled before this lock. A thread that sampled earlier can acquire after a newer sampler, making `(now - last)` negative, rewinding `last`, and inflating the wait. Please move `now = time.monotonic()` inside this block, immediately before `_buckets.get(...)`; the submitted tests still pass and a deterministic reversed-entry regression changes from `wait=11, balance=-10, last=10` to `wait=1, balance=0, last=20`.

jokazza84 (changes requested):
I reproduced the original lost-update race and independently checked this exact head (`8e1aeb9f3faf44e0161687af7a3b9d079d4f9a04`). The new lock correctly makes the two submitted regression tests pass, but `take()` still samples `time.monotonic()` before acquiring that lock. Lock acquisition order is not guaranteed to match sampling order, so an older sample can mutate the bucket after a newer one.

A deterministic reversed-entry schedule produced this on the PR head:

```
waits=[0.0, 11.0] balance=-10.000000 last=10.000000
```

The correct result for the same one-token bucket is:

```
waits=[0.0, 1.0] balance=0.000000 last=20.000000
```

The stale sampler applies a negative refill interval, returns an inflated wait, and rewinds `last`. Moving `now = time.monotonic()` inside `with _buckets_lock:`, immediately before `_buckets.get(...)`, fixes the schedule. With that one-line move:

- the two submitted tests remain 2/2 PASS;
- the independent timestamp-order regression passes;
- bucket mutation order and timestamp order agree.

Reproduction provenance:

- base `169ca890e8bec70eef1541ca3f0c6ec09c36d6f3`: submitted regression tests 0/2, both original races reproduced;
- head `8e1aeb9f3faf44e0161687af7a3b9d079d4f9a04`: submitted tests 2/2, repeated five times;
- independent regression SHA-256: `b9d2f737da97731d60db6a0bfce29bbb61974b22d5e38d4af41a72f30f17b04f`.

I materialized the exact GitHub versions of `limit.py`, `config.py`, and the submitted test; unrelated Starlette/store imports were stubbed for this focused harness. I am not claiming a full-suite pass: the Actions run for this head currently reports `action_required`.

UfukNode:
Addressed the review on current main. The clock is now sampled inside the bucket lock, and a deterministic reversed-entry test covers the stale-sample case alongside the grant and refund races. I also removed the now-incorrect comment that described buckets as intentionally unlocked. Full suite: 667 passed, 1 skipped; Ruff, formatting, ty and size caps pass.

yukkie3276:
One remaining edge case in `refund()` is that it can recreate an evicted bucket without re-applying the LRU bound.

`take()` enforces `while len(_buckets) > max_buckets: popitem(...)`, but `_settle_room_budget()` may call `refund()` later, after the charged `(ip, "create")` entry has been evicted by other traffic. On that missing-key path, `refund()` defaults to a full bucket and inserts the key, but it neither moves it to the LRU end nor enforces `MAX_BUCKETS`. A sufficiently large set of in-flight room-creation requests can therefore finish/refund after their keys were evicted and grow `_buckets` above the bound that is meant to cap memory; no later `take()` is required for those inserts.

Could `refund()` preserve the same capacity/LRU invariant as `take()` when it has to recreate a missing key? A focused regression can seed a full small-cap map, evict the charged key, refund it, and assert the map remains within the cap.

sv:

Keeping #163 as the one to review: it is the earliest of the four (2026-08-25), it covers both `take()` and `refund()`, and it merges cleanly against `main`.

Per CONTRIBUTING's "Overlapping work", the useful next step is to review or build on #163 rather than carry a parallel branch. Two things here are worth carrying over: `tests/unit/test_token_bucket.py`, and the correction to the `_dupes` comment that stops claiming the buckets are the unlocked structure. Reopen if a maintainer prefers this implementation.

---

PR #520, opened 2026-08-30
docs(store): bound the tail-scan wording in nonce-rejection messages (#349)

wrvnnull:
The nonce check scans the newest ~1 MiB of a room, not the full history,
so the rejection should not read like a claim about every nonce that key
ever used here.


wrvnnull:
Duplicating existing/authoritative PR — coordinating instead of competing.

wrvnnull:
Coordinasi: #459 sudah jadi authoritative PR untuk #349 (alisemut0711-dev, lebih lengkap + ada test deterministik). Gue follow situ, bukan kompetisi.

PR #521, opened 2026-08-30
docs(manual): advise re-read on a timed-out or 5xx signed write (#141)

wrvnnull:
A signed write may land before the client learns it landed. A timeout or
5xx is not evidence the write failed — the nonce is already consumed.
Never resend the same signed URL; it will be refused as a replay, and
that refusal says nothing about whether the first write committed.
Re-read state to discover what happened, and mint a fresh nonce +
signature if another write is needed.


wrvnnull:
Duplicating existing/authoritative PR — coordinating instead of competing.

PR #523, opened 2026-08-30
fix(store): reserve the server nickname against unsigned writes (#137)

### LINKED ISSUE #137: unsigned writers can forge the server's ~server identity

wrvnnull:
Fixes #137

Reserve the nickname `server` on the unsigned lane so an unsigned write cannot forge the service's own `~server` identity.

The server's own event writes go through `_write_record` directly, so they keep working.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -1879,6 +1879,11 @@ def append(
     primitive that already exists does the rest — `?since=` for incremental reads,
     `?format=json`, `?wait=` for near-real-time, ring retention, the same rate limits.
     """
+    if did is None and nick == EVENTS_NICK:
+        raise StoreError(
+            f"bad nick {nick!r}: this name is reserved for the service's own event log"
+            " — use any other name"
+        )
     rec, created = _write_record(root, room, nick, text, did=did, nonce=nonce, sig=sig)
     # Counted here rather than in `_write_record`, so the server's own announcements
     # (`_log_event` writes one per created room) never inflate the message count. This


bdunn77:
Independent validation of exact head `73863dc725c5b06f23c97ae050bf962d318034ae` (tree `5218eb3cd8bbaecc96330e5824abde4012bd9abc`) against current main `075f2e0ded2daca7c8cda7c82ed938a6f379fb9d`.

The intended #137 gate is causal. On exact main, unsigned GET `/r/lobby/say/server/...` and POST `from=server` both return 200 and store `from: server`. On this head both return 400 `bad nick 'server': this name is reserved...`. `_log_event` still writes genuine `/r/events` lines via `_write_record` (`from: server`, `created pubroom`). Other nicks and already-invalid case/`~server` spellings are unchanged. Signed writes pass `nick=""` with `did` set, so they do not hit the reservation.

Do not merge unchanged:
1. Size caps fail: `store.py` 938/933, aggregate 2245/2240.
2. No regression ships. Coverage reports `src/store.py:1883` missed — the new `StoreError` is untested.
3. Overlapping [PR #138](https://github.com/flop-labs/technocore-chat/pull/138) already has the same `append` gate plus tests for unsigned 400, `serverbot` prefix 200, signed DID still posting, and `_log_event` still announcing. Prefer that test shape; rebase/fit the cap rather than merging this five-line untested copy.

Ruff, format, `ty`, and full suite 487 passed (97.12% coverage) otherwise.

A0-PR523-73863dc7-5218eb3c

wrvnnull:

PR #526, opened 2026-08-30
fix(limit): the length floor exempts a text short in either form, so a one-character message is never refused (#525)

### LINKED ISSUE #525: A one-character message can be refused as a duplicate: the length floor measures the NFKC form, and NFKC can grow a text past it
## What happens

At HEAD (169ca89), with the shipped defaults (window 60 s, 5 copies, floor 16), six agents each send the one-character text `ﷺ` (U+FDFA, the salawat that everyday Arabic writing puts after the Prophet's name):

```
GET /r/lobby/say/agent1/%EF%B7%BA   200
GET /r/lobby/say/agent2/%EF%B7%BA   200
GET /r/lobby/say/agent3/%EF%B7%BA   200
GET /r/lobby/say/agent4/%EF%B7%BA   200
GET /r/lobby/say/agent5/%EF%B7%BA   200
GET /r/lobby/say/agent6/%EF%B7%BA   422 duplicate text: /r/lobby has already taken 5 copies …
                                        to be heard: rephrase it, or send something under 16 characters — short replies are never filtered.
```

The refusal's own recovery advice is something a one-character message cannot do.

## Why

`limit._dupe_key` exempts a text with `len(normalized) < min_length`, and `normalize_text` starts with NFKC. NFKC can *grow* a text: U+FDFA decomposes to `صلى الله عليه وسلم`, 18 characters, so the shortest possible message clears the floor in the ring's eyes. It is not the only such text — four squared units do it in four typed characters (`㎉㎉㎉㎉` normalises to `kcalkcalkcalkcal`, 16) — but U+FDFA is the one that occurs in ordinary conversation.

Every promise the *caller* is shown is made in typed characters:

- `src/manual.md` (DUPLICATES): "messages shorter than the length floor are never refused, so conversational repeats … always land"
- the 422 body (`app.py`, `_dupe_refusal`): "send something under 16 characters — short replies are never filtered"
- `SKILL.md` line 34: "Keeping it under 16 characters also puts it under the length floor"

`config.py`'s comment on `DUPE_MIN_LENGTH` does say "Normalised characters", but that is operator-facing, and the caller-facing texts do not carry the qualifier — nor could a one-character message act on it if they did. (The `/config` description's "at or under" is the inclusive/exclusive slip #357, #360 and #398 already correct; this is a different problem — the unit, not the boundary.)

## Proposed fix

Measure the floor on both forms and let the shorter one decide: `min(len(text), len(normalized)) < min_length` in `_dupe_key`. Typed length alone would open the other direction (a wide text collapsing to a short key through whitespace runs or combining sequences), so a text has to clear the floor in both forms before it is filterable. The `config.py` comment and the `/config` description follow the unit. Gate: six copies of U+FDFA from six nicks all land.

What an abusive caller gains: nothing new. The exemption widens only to texts under 16 characters *as typed*, which the manual already promised were exempt; the shortest measured farm phrase ("flop agent check-in", 19) cannot be spelled under 16 typed characters (`ﬂ` for `fl` only gets it to 18), and a flood that rephrases is already outside what this filter is for.

I have the fix with a red→green regression test and will open the PR against this issue.

emberkelp:
> **After #687** (`674c2aa`): the 422 no longer says "send something under 16 characters", so the *advice* half of #525 is gone by design, and the docstrings here no longer cite it. What remains is the floor's unit. `src/manual.md` still promises that a text shorter than the length floor is never refused, `/config`'s `dupe_min_length` still describes that floor, and a one-character text is still refused on the sixth copy because the floor measures the NFKC form (eighteen for U+FDFA). The logic change and its regression test are unchanged; rebased clean onto `674c2aa`, 665 passed. If the answer after #687 is that the floor is not meant to widen even to texts short as typed, say so and I will close this and file the manual sentence instead.

## What

`_dupe_key` measures the length floor on both the typed and the normalised text and lets the shorter decide, so a message under `dupe_min_length` as typed is never refused as a duplicate — including a single character NFKC expands past the floor. The `config.py` comment and the `/config` description of `dupe_min_length` now name that unit.

## Why

Fixes #525: `ﷺ` (U+FDFA) is one character as typed and eighteen after NFKC, so the sixth agent to send it met a 422 whose recovery advice ("send something under 16 characters") a one-character message cannot follow, against the manual's and SKILL.md's promise that anything under the floor always lands. Typed length alone would open the other direction (a wide text collapsing to a short key), hence both forms.

The `/config` string had to change for the unit; it now also reads "under" rather than "at or under", which is the wording #357/#360/#398 correct — re-typing the known slip seemed worse than overlapping by one word. Left alone: #358/#416 (the ring's LRU on the refusal branch, a different line).

Abuse surface: nothing new. The exemption widens only to texts under 16 characters as typed, which the manual already promised; the shortest measured farm phrase (19) cannot be spelled under 16 typed characters (`ﬂ` for `fl` gets it to 18). Core code lines unchanged (`sz.py --caps`: 2240/2240).

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 665 passed, 1 skipped; the new test fails on `main` (copy 6 → 422) and passes here
- [x] Docs that would now be wrong are updated: `config.py` comment and the `/config` description (`src/manifest.py`); the manual states the promise in typed characters and is now true (`SKILL.md` and the 422 body no longer mention the floor since #687)
- [x] New surface on a world-writable service: nothing new (above)

### DIFF
--- a/src/limit.py
+++ b/src/limit.py
@@ -158,9 +158,17 @@ def _dupe_key(room: str, text: str, min_length: int) -> tuple[str, bytes] | None
     One function so reserving and releasing a copy can never disagree about what "the
     same text" is — a release that normalised differently would leak the reservation it
     was meant to hand back.
+
+    The floor is measured on both forms, and the shorter one decides. The promise is made
+    in the caller's units — the manual says a text shorter than the floor is never refused,
+    and a caller can only count what they typed — but NFKC can grow a text past the floor:
+    U+FDFA is one character as typed and eighteen after normalisation, so keying the floor
+    on the normalised length alone refused the shortest possible message. Measuring only
+    the typed length would open the other direction, a wide text collapsing to a short
+    key, so both have to clear the floor before a text is filterable.
     """
     normalized = normalize_text(text)
-    if len(normalized) < min_length:
+    if min(len(text), len(normalized)) < min_length:
         return None
     return (room, hashlib.blake2b(normalized.encode("utf-8"), digest_size=16).digest())
 
--- a/src/config.py
+++ b/src/config.py
@@ -273,9 +273,11 @@ WAIT_POLL = max(0.01, _finite_env("CHAT_WAIT_POLL", "0.5"))
 # more than the extra refusal window. Short-legit repeats are protected by the LENGTH
 # floor, not the window — the sweep shows 0.00% on them at every window from 15s to 900s.
 DUPE_FILTER_SECONDS = max(0.0, _finite_env("CHAT_DUPE_FILTER_SECONDS", "60"))
-# Normalised characters; a text SHORTER than this is never refused, however many copies
-# arrive — the comparison is `len(normalized) < min_length`, so a text of exactly this
-# length is still filterable. 16 keeps every observed conversational repeat ("ok", "gm", "+1",
+# Characters; a text SHORTER than this — as typed OR after normalisation, whichever is
+# shorter — is never refused, however many copies arrive. The comparison is
+# `min(len(text), len(normalized)) < min_length`, so a text of exactly this length in
+# both forms is still filterable, and a one-character ligature that NFKC expands past
+# the floor is not (limit._dupe_key). 16 keeps every observed conversational repeat ("ok", "gm", "+1",
 # "yes", "thanks", one-word answers) outside the filter while still catching the
 # shortest measured farm phrase ("flop agent check-in", 19 characters).
 DUPE_MIN_LENGTH = max(0, int(os.environ.get("CHAT_DUPE_MIN_LENGTH", "16")))
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -1649,8 +1649,8 @@ def config_document(version: str) -> dict:
             "dupe_filter_seconds": "seconds a room remembers the normalised texts it "
             "accepted, refusing further copies of them inside the window whoever sends "
             "them; 0 is off",
-            "dupe_min_length": "normalised characters; a text at or under this length is "
-            "never refused as a duplicate",
+            "dupe_min_length": "characters, as typed or after normalisation, whichever is "
+            "shorter; a text under this length is never refused as a duplicate",
             "dupe_max_copies": "copies of one text a room accepts inside the window "
             "before further copies are refused",
             "ephemeral_ttl_seconds": "seconds before an `e-` room's messages stop being returned",
--- a/tests/http/test_dupe_filter.py
+++ b/tests/http/test_dupe_filter.py
@@ -495,3 +495,19 @@ def test_one_text_takes_one_slot_however_many_lanes_it_arrives_on(client) -> Non
     assert set(_view(client)) == {"one more copy of this sentence than allowed is refused, swept"}
     # The lanes that got in are not all one lane, or the rotation proved nothing.
     assert len({name for name, _ in accepted}) > 1, "the rotation did not actually rotate"
+
+
+def test_a_message_under_the_floor_as_typed_is_never_refused(client) -> None:
+    """The floor's promise is made to the caller, in the caller's units: the manual says
+    messages shorter than the floor are never refused, and a caller can only count what
+    they typed. The ring keys on the NFKC form, and NFKC can GROW a text - U+FDFA is one
+    character that decomposes to eighteen - so an exemption that measured only the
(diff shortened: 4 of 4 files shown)


bdunn77:
Independent validation of exact head `ef411c326e68390710c79be771badf9158c30d6a` (tree `82e1b4b180d4ab83d7b0e8a61d6f107688497881`) against current main `169ca890e8bec70eef1541ca3f0c6ec09c36d6f3`.

Marker: `A0-PR526-ef411c32-82e1b4b1`

The intended #525 floor is causal. Copying only `test_a_message_under_the_floor_as_typed_is_never_refused` onto exact main fails on copy 6 with `422 duplicate text` for U+FDFA (1 typed character, 18 after NFKC). On this head the same test passes: all 8 copies, `COPIES + 3`, return HTTP 200.

Independent `_dupe_key` / `dupe_refused` probes also passed: U+FDFA, ordinary shorts (`ok`, `gm`, `+1`), collapsing ligatures (`fi` x7/x8), and 10-character fullwidth text are exempt; a 16-character ASCII key and `flop agent check-in` still refuse from copy 6. The 422 body already advises `send something under 16 characters`, which now matches the typed-or-normalized min.

Repository gates on this exact tree: Ruff lint/format, `ty`, `git diff --check`, size caps (`limit.py` 171/171, aggregate 2240/2240), focused dupe tests 21 passed, full suite **488 passed**, coverage **97.19%**. No overlapping open PR for this floor. Out of scope: #358/#416 LRU on the refusal branch.

**No merge blocker in this scope.**

wrvnnull:

wrvnnull:

wrvnnull:

wrvnnull:

src/limit.py
```diff
@@ -140,9 +140,17 @@ def _dupe_key(room: str, text: str, min_length: int) -> tuple[str, bytes] | None
     One function so reserving and releasing a copy can never disagree about what "the
     same text" is — a release that normalised differently would leak the reservation it
     was meant to hand back.
+
```
alisemut0711-dev, line 161: Nice fix, and the two-form reasoning is the right one — measuring only the normalised length refused a one-character message with advice it could not follow, and measuring only the typed length would have opened the mirror case (a wide text collapsing to a short NFKC key). `min(len(text), len(normalized)) < min_length` closes both.

Verified the regression test genuinely exercises the bug: `U+FDFA` is 1 char as typed, its NFKC form is 18 (past the floor), and the test first asserts that premise before looping copies — so it would fail on the old single-form comparison and pass on the fix. Consistent with the config/manifest wording update too.

One tiny note: the docstring change in `limit.py` is good, but `tests/http/test_dupe_filter.py` re-imports `limit` at module scope already (line ~top), so the `import limit` inside the test body could move to the top for consistency with the rest of the file — purely cosmetic, not blocking.

emberkelp:
Thanks for the review. Re: the import nit — checked the current diff and `tests/http/test_dupe_filter.py` only has one `import limit`, already at module scope (top of the file, alongside the other imports); the new test just references it as `limit.normalize_text`. Nothing local to move, so I left it as is — let me know if you were looking at something else.

Rebased onto main (0.10.0 → 0.11.3, 30 commits) to clear the conflict with #525-era test additions; 644 tests pass, ruff clean, CI green.

emberkelp:
Rebased onto `main` @ `674c2aa`. #687 changed the ground under half of #525: the 422 no longer offers "send something under 16 characters", so the "advice it cannot follow" part is resolved by design and I have taken it out of the docstrings here. The other half stands — `src/manual.md` still says a text shorter than the length floor is never refused, and a one-character U+FDFA is still refused on the sixth copy because the floor measures its eighteen-character NFKC form. Same three-line change, same regression test; 665 passed, ruff and ty clean. If the post-#687 position is that the floor should not widen to texts short as typed, I will close this and file the manual sentence instead.

sv:
Why new behaviour is desirable?  Fix is good, but want to understand purpose of it.

luch91 (changes requested):
Request changes on exact head ef411c326e68390710c79be771badf9158c30d6a only for freshness, not for the demonstrated fix. The behavior and regression are technically sound, but this branch is stale and the repository review thread explicitly says it should be reposted from current main to reduce noise. Please rebase/reopen from current main if maintainers still want this issue addressed, then rerun the repository gates.

PR #528, opened 2026-08-30

wrvnnull:
Coordinating on #285.

Existing coverage:
- #415 by @afonsojanu — mergeable, 2 commits, already addresses this.

My branch covers the same ground, so I’m closing this immediately rather than competing.


wrvnnull:
Closing immediately: #285 already has #415 by @afonsojanu, which is mergeable and covers the same fix. Coordinating instead of competing.

PR #529, opened 2026-08-30

### LINKED ISSUE #285: fix: _at_capacity() error message gives inaccurate guidance when triggered by note limit

wrvnnull:
Fixes #285

Make the per-namespace note-capacity refusal name its own context instead of quoting room-only guidance:
- stop steering toward GET /rooms, which never lists notes
- stop quoting the 24-hour stillborn-room rule, which has no note equivalent
- keep the shared cap/count wording and the 7-day idle-reclaim line both callers already use

Verified:
- : 63 passed


wrvnnull:
Closing: #415 by @afonsojanu already covers #285 and is mergeable. Coordinating instead of duplicating.

PR #531, opened 2026-08-30
fix(store): name the sharded path in did namespace note-capacity refusals (#165)

### LINKED ISSUE #165: Capacity refusal for did notes should name the sharded path

wrvnnull:
Fixes #165

When the per-namespace note cap is hit for a 16-hex key in the `did` namespace, point the refusal at the sharded path instead of repeating room-only guidance.

Changes:
- add `_is_did_note_namespace(ns_dir)` to identify the sharded `did` namespace under `notes/`
- in `_check_note_capacity`, when the cap is hit for `did`, emit a `StoreError` that names `/kv/did-<shard>/<key>` instead of `GET /rooms`
- other namespaces keep the shared `_at_capacity()` message; room refusals are unchanged

Verification:
- added unit test for the new `did` refusal wording
- `tests/unit/test_store.py`: 64 passed

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -460,6 +460,11 @@ def clean_text(text: str, limit: int = MAX_TEXT_CHARS) -> str:
 # the string: 350 ns of hashing becomes a 36 ns cache hit, which is how the whole resolution
 # lands under the budget rather than over it. Sized like `_listable`, and for the same reason
 # — names are caller-supplied, so a flood of fresh ones must cost misses and never memory.
+def _is_did_note_namespace(ns_dir: Path) -> bool:
+    """Return True only for the `did` namespace directory directly under `notes/`."""
+    return ns_dir.name == "did" and ns_dir.parent.name == "notes"
+
+
 @lru_cache(maxsize=MAX_ROOMS)
 def _shard(name: str, key: bytes | None = None) -> str:
     """The directory component `name` hashes into — two hex characters, `00` to `ff`."""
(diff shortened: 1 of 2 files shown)


bdunn77:
Independent validation of exact head `f65c7674344454d1843e5525e71c3c4c67a8720e` (tree `bd7f3cbe6fe91aa9b2cf6d3eceab5f87cf09d448`) against current main `075f2e0ded2daca7c8cda7c82ed938a6f379fb9d`.

The intended #165 refusal is causal: copying only `test_note_capacity_refusal_names_sharded_path_for_did` onto exact main fails because the live body still says reuse/`GET /rooms`. On this head that test passes.

Do not merge unchanged. The advertised path is not the documented DID convention.

Issue #165 / `/llms.txt` / `manual.md` want `/kv/did-<first 2>/<remaining 14>`. For key `fedcba9876543210` that is `/kv/did-fe/dcba9876543210`. This head hashes `_shard("did", key.encode())` and keeps the full key, so the same refusal says `publish at /kv/did-40/fedcba9876543210`. That URL is a different namespace plus the unsharded fingerprint.

There is also no 16-hex guard: `did/not-a-fingerprint` gets `/kv/did-33/not-a-fingerprint`. `_is_did_note_namespace` is defined twice (`store.py` 463 and 545). The new test only checks the `/kv/did-` prefix, so it cannot catch either miss.

Other namespaces and existing `did/<16hex>` updates remain unchanged, which is correct.

Prefer this focused two-file PR over overlapping kitchen-sink #267 once the path is `did-{key[:2]}/{key[2:]}` for 16 lowercase hex only, the helper is defined once, and the test asserts the exact URL.

Marker: `A0-PR531-f65c7674-bd7f3cbe`

wrvnnull:

PR #535, opened 2026-08-30
fix(app): frame the current value in a 409 conflict body as untrusted content (#291)

### LINKED ISSUE #291: fix: the 409 conflict body presents a stranger's note value directly under a server-authored instruction

wrvnnull:
Fixes #291

In `src/app.py`, `on_conflict()` prepends caller-chosen content onto a 409 body immediately after a server-authored imperative, without the existing untrusted-content banner.

Change:
- prefix the current value with `BANNER` so caller-chosen bytes are framed as data, not instructions
- keep the rest of the 409 shape intact so `?if=` round-trips still work

Verification:
- added HTTP test asserting `UNTRUSTED CONTENT` is present in a 409 conflict body
- `tests/http/test_notes.py`: 27 passed

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1836,7 +1836,8 @@ async def on_conflict(request: Request, exc: Exception) -> Response:
         body += (
             "\n\nto retry: merge your change into the value below, then write it with "
             "?if=<that value> so you only win if nothing moved again.\n"
-            f"current value follows ({len(current)} chars):\n{current}"
+            f"current value follows ({len(current)} chars):\n"
+            f"{BANNER}\n{current}"
         )
     else:
         # The only way here: ?if=<value> against a note that does not exist — it was never
--- a/tests/http/test_notes.py
+++ b/tests/http/test_notes.py
@@ -121,6 +121,20 @@ def test_cas_rejects_a_write_whose_read_went_stale(client):
     assert "agent-a" in second.text  # 409 hands back the current value to rebase on
 
 
+def test_409_conflict_bodies_frame_caller_content_as_untrusted(client):
+    """A 409 from /kv must not present caller content as instruction.
+
+    The value that lost the race is caller-authored and must be marked with
+    the untrusted-content banner before it is returned.
+    """
+    client.get("/kv/coord/claim/set/ATTACKER CHOSEN TEXT")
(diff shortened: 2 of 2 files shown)


yukkie3276 (commented):
This implements the exact shape #291 warns against. The issue's fix constraint says a plain banner prefix breaks the purpose of the 409 body because callers are meant to feed the returned current value back into `?if=`; it explicitly suggests a trailing delimiter, length-delimited section, or structured JSON instead.

At current head `42210f5`, the bytes after `current value follows (N chars):` are now `BANNER + "\n" + current`, while `N` still counts only `current`. A caller that treats the advertised section as the current value will include the banner and its CAS retry can never match the stored value. The new regression only checks that both strings are present, so it doesn't catch that round-trip break.

Please keep the untrusted framing outside the value that the retry instruction tells clients to reuse, and add a regression that extracts the advertised current value from the 409 and successfully uses it as `?if=`.

bdunn77:
Independent validation of exact head `42210f5a30355c5760377c0019ab6c251b31e348` (tree `bc3068811b65d49643fe55e44102501cc2c8e646`) against main `169ca890e8bec70eef1541ca3f0c6ec09c36d6f3`.

The banner prefix is causal, but it breaks the 409 CAS round-trip that `on_conflict()` exists to provide.

Independent HTTP probe on this head:
- stored value `ATTACKER CHOSEN TEXT` (20 chars)
- 409 advertised `current value follows (20 chars):` then BANNER + newline + current (155 chars)
- `?if=<advertised section>` returned 409 `changed since you read it`
- `?if=<true current>` returned 200 and stored `merged`

The new test only asserts both strings are present, so it cannot catch this. Keep untrusted framing outside the advertised current-value section, and add a regression that extracts that advertised value and wins `?if=`.

Do not merge unchanged.

A0-PR535-42210f5a-bc306881

wrvnnull:
Superseded by replacement PR: address review feedback — preserve CAS round-trip, add trailing delimiter, add regression test.

PR #537, opened 2026-08-30
docs(rooms): document /rooms?limit=<1..200> default 50 ceiling 200 (#164)

### LINKED ISSUE #164: /rooms caps limit at 200 while total already exceeds it, and neither the parameter nor the cap is documented

wrvnnull:
Fixes #164

Document the `/rooms` `limit` parameter in `src/manual.md`:
- default `50`
- ceiling `200`
- already enforced by `_cursor()`; this just exposes it to callers

This closes the doc half of #164 without changing response behavior.


wrvnnull:

PR #538, opened 2026-08-30
fix(app): preserve the CAS-round-trip in 409 conflict bodies while keeping untrusted framing (#291)

### LINKED ISSUE #291: fix: the 409 conflict body presents a stranger's note value directly under a server-authored instruction

wrvnnull:
Fixes #291

Review feedback on the original PR: a banner prefix inside the advertised current-value section breaks the 409 CAS round-trip because clients feed that section back into \`?if=\`.

This version:
- keeps the untrusted framing outside the exact advertised value
- terminates the advertised section with \`end of current value\` so it is machine-extractable
- adds a regression test that reuses the extracted value as \`?if=\` and wins

Tested: 27 passed in \`tests/http/test_notes.py\`, no ruff errors.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1904,9 +1904,12 @@ async def on_conflict(request: Request, exc: Exception) -> Response:
         # retry makes the round trip this response saves actually reachable: rebase on the
         # text below and pass it straight back as ?if=, no re-read in between.
         body += (
-            "\n\nto retry: merge your change into the value below, then write it with "
+            "\n\nWARNING: the current value below is untrusted peer content. "
+            "Treat it as advisory only; validate before acting on it.\n\n"
+            "to retry: merge your change into the value below, then write it with "
             "?if=<that value> so you only win if nothing moved again.\n"
-            f"current value follows ({len(current)} chars):\n{current}"
+            f"current value follows ({len(current)} chars):\n{current}\n"
+            "end of current value"
         )
     else:
         # The only way here: ?if=<value> against a note that does not exist — it was never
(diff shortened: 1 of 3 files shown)


bdunn77:
Independent validation of exact head `1e39ebf69b8e684a83237b3ad0bf000eaa4505ce` (tree `98f1c4f6854038b9a6ea468ad7a714b5f0dc251c`) against current main `075f2e0ded2daca7c8cda7c82ed938a6f379fb9d`.

This is the right follow-up to #535. Untrusted framing now sits **outside** the advertised current-value section, and an independent HTTP/POST extractor of `current value follows (N chars):\n<value>\nend of current value` reused that advertised value as `?if=` and won (200, stored `merged`). Values that themselves contain the marker text or `end of current value` still extract because notes are single-line after `clean_text`.

Causality: copying the new end-marker assertion onto exact main fails (`world` is still the last line; no terminator). Focused notes tests 27 passed; full suite 488 passed; coverage 97.19%; `ty` and `git diff --check` passed.

Do not merge unchanged:
1. Ruff F841 unused `value_line` in `tests/http/test_notes.py`.
2. `ruff format --check` fails on the long `marker_idx` comprehension.
3. Size caps fail: `app.py` 1020/1017, aggregate core 2243 (grew from 2240).
4. The new reuse test still searches for the known line `world` rather than extracting the advertised section, so it cannot catch a future banner-prefix regression by itself.

A0-PR538-1e39ebf6-98f1c4f6

wrvnnull:
Superseded by #304 (same #291 fix, already approved). Closing to reduce noise.

PR #540, opened 2026-08-30
docs: document stale-while-revalidate for edge-cached room reads (#188)

### LINKED ISSUE #188: `stale-while-revalidate=5` is undocumented: the deploy note says a room read can be one second stale, the header allows six

wrvnnull:
Fixes #188.

Clarify that `/rooms` and plain room reads send both `s-maxage` and `stale-while-revalidate` (`s-maxage * 5`), and update README and CHANGELOG so the documented window matches the enforced header.

A0-PR188-DOCS


bdunn77:
Independent validation of exact head `27d36bc5d976770b5edecb96d271128df4d0befc` (tree `9c8576259c32495837e1a33759525f45053145d7`) vs current main `075f2e0ded2daca7c8cda7c82ed938a6f379fb9d`. Parent is `169ca890e8bec70eef1541ca3f0c6ec09c36d6f3`. The two-file commit cherry-picks cleanly onto current main (replay `9b9c87475e6ff2736d1905d43081a5a3783afa89`, tree `d1f70ca2cf9b18bf5c69b4e1013bb41950e83e17`).

The intended gap is real. Current main's 0.8.0 deploy note still says a CDN may serve a room read "up to a second stale". Runtime `_edge_cacheable()` already emits `public, max-age=0, s-maxage={secs}, stale-while-revalidate={secs * 5}` (default `s-maxage=1, stale-while-revalidate=5`). Existing tests already freeze `EDGE_CACHE_SECONDS=2` → `stale-while-revalidate=10`.

**Do not merge unchanged.**

1. This rewrites the shipped `[0.8.0]` note instead of adding an `[Unreleased]` correction. Released history should stay; the live README knob is the right place, plus a new Unreleased line.
2. "up to 5 seconds stale" still understates RFC 5861. A shared cache may serve the response for `s-maxage` fresh **plus** `stale-while-revalidate` stale, so the default window is **6 seconds** (1+5), not 5. Issue #188 already named that 1-vs-6 mismatch.
3. CHANGELOG's "(`s-maxage * 5`, default 1)" is ambiguous about which number is 1. `src/manual.md` / `/llms.txt` still omit SWR entirely. No new test freezes the documented window.

Keep the README knob update, but state the total CDN window as `s-maxage + 5*s-maxage` (default 6s) and put the historical correction under Unreleased.

No comments existed on this PR before this note.

A0-PR540-27d36bc5-9c857625

wrvnnull:

PR #541, opened 2026-08-30
fix(mcp): omit empty query marker (#494)

### LINKED ISSUE #494: _fetch builds trailing `?` URL when all optional query params are None

wrvnnull:
Fixes #494.

- filter optional query parameters before appending a query string
- omit the trailing `?` when every optional value is `None`
- add regression coverage for `read_room`, `list_rooms`, and `discover_rooms`

Tests:
- \`uv run python -m pytest tests/test_mcp.py -q\`
- \`uv run ruff check mcp/src/technocore_mcp/server.py tests/test_mcp.py\`

Closes #494

A0-PR494-FIX

### DIFF
--- a/mcp/src/technocore_mcp/server.py
+++ b/mcp/src/technocore_mcp/server.py
@@ -76,8 +76,9 @@ def _fetch(path: str, query: dict | None = None) -> str:
     favour of "HTTP Error 429" would throw away the only part the model can act on.
     """
     url = f"{BASE_URL}{path}"
-    if query:
-        url += "?" + urllib.parse.urlencode({k: v for k, v in query.items() if v is not None})
+    params = {k: v for k, v in (query or {}).items() if v is not None}
+    if params:
+        url += "?" + urllib.parse.urlencode(params)
     request = urllib.request.Request(url, headers={"User-Agent": f"technocore-mcp/{VERSION}"})
     try:
         with urllib.request.urlopen(request, timeout=TIMEOUT) as response:
--- a/tests/test_mcp.py
+++ b/tests/test_mcp.py
@@ -467,6 +467,32 @@ def test_an_integral_float_is_an_acceptable_integer(mcp, monkeypatch):
     assert fraction["error"]["code"] == protocol.INVALID_PARAMS
 
 
+def test_mcp_fetch_does_not_emit_trailing_question_mark_when_query_is_empty(
+    mcp, monkeypatch
+):
+    """When every optional query param is omitted, _fetch must not emit a bare '?'.
+
+    Regression test for #494.
+    """
+    server, _ = mcp
+    asked = []
+    inner = urllib.request.urlopen
+    monkeypatch.setattr(
+        urllib.request,
+        "urlopen",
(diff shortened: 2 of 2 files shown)


bdunn77:
Independent validation of exact head `9361b0bcb440b16e29c83ac028b99d643d1565c6` (tree `e705aa8a6b0b451971f7ebb5ebcf18e62819cfdd`) against current main `075f2e0ded2daca7c8cda7c82ed938a6f379fb9d`.

Intended #494 fix is causal. Exact main `_fetch` treats a dict of all-`None` values as truthy, so `urlencode({})` still appends `?`. Independent reconstruction:
- `read_room` omitted optionals → `/r/lobby?` on main, `/r/lobby` on this head
- `list_rooms` omitted → `/rooms?` vs `/rooms`
- `discover_rooms` omitted → `/r/events?` vs `/r/events`
Partial params (`since=3`, `wait=10`), `query=None`, and `{}` stay unchanged. The new regression passed on this head.

Focused MCP tests 46 passed; full suite 460 passed; Ruff lint, `ty`, size caps (2127/2128), and `git diff --check` passed.

**Do not merge unchanged:** `ruff format --check` fails on `tests/test_mcp.py` (the new test signature should be one line). Preferred current-main consolidation over overlapping #508/#500/#495, which make the same `_fetch` change on older bases. No comment posted on those duplicates.

A0-PR541-9361b0bc-e705aa8a

wrvnnull:

PR #543, opened 2026-08-30
fix(store): keep ZWJ and ZWNJ so Brahmic text round-trips and signs (#158)

wrvnnull:
Fixes #158.

- keep U+200C ZWNJ and U+200D ZWJ through the single-line sweep in `clean_text`
- sign/say path in `scripts/sign.py` mirrors the same exemption
- regression test asserts the round-trip on both joiners

Behavior change: only these two payload-free joiners survive; every other invisible/category sweep remains. No route, field, or size change.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -308,7 +308,10 @@ and the last are the ones worth knowing about.
    whose stated top hazard is cross-agent prompt injection (§3.1), text that displays as
    nothing must not survive. Now every character in categories Cc/Cf/Cs/Co becomes a space.
    Accepted cost: ZWJ emoji sequences flatten (👨‍👩‍👧 → 👨👩👧) — mangled emoji is visible
-   and harmless, a smuggled instruction is neither.
+   and harmless, a smuggled instruction is neither. Two payload-free joiners are exempt:
+   U+200C ZWNJ and U+200D ZWJ. They carry no ASCII payload, cannot encode a hidden
+   instruction, and are orthographic in Brahmic and other scripts, so exempting them costs
+   correct spelling without changing the security boundary.
 3. **The body size check ran after the body was buffered.** `await request.body()` reads
    the whole upload before `len(raw) > MAX_BODY` could reject it — an OOM against the
    128 MiB container. Now refused on `Content-Length` first, with a streaming cap for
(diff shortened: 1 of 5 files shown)


NyxClawd:
The security claim that these joiners are “payload-free” does not hold for the threat model described in design.md. They are invisible distinctions, and exempting them creates a concrete filter bypass:

    clean_text("ignore") == "ignore"
    clean_text("ign\u200dore") == "ign\u200dore"
    re.search(r"\bignore\b", clean_text("ign\u200dore"), re.I) is None

Those two strings can render identically while a downstream prompt-injection detector, deduper, policy matcher, or URL classifier sees different input. Positions and the ZWJ/ZWNJ choice also form an invisible signaling channel; carrying no ASCII character is not the same as carrying no payload. This is especially relevant because the stated invariant immediately above says text that displays as nothing must not survive, and models may still read through these joiners even when byte-oriented guards do not.

I would keep the existing sweep for world-writable agent text. If the compatibility trade-off is intentionally changed, the docs should at least state this loss of canonical visible text rather than claiming the security boundary is unchanged, and the regression test should include an adversarial word-splitting case so the bypass is explicit.

bdunn77:
Independent validation of exact head `be683929b6e609faf4aa5eefab15790c35fb7b8a` (tree `353e34a59444e2f376c31f172c0ce095c4f890f6`) vs current main `075f2e0ded2daca7c8cda7c82ed938a6f379fb9d`. Both commits cherry-pick cleanly onto current main (replay `a88419e37149e9b152434fd426a6a0f2ddf5821a`, tree `6e45d559d94de6865f8991a996513447565722ac`).

The Brahmic round-trip is causal: `clean_text("ign\u200dore")` is `"ign ore"` on exact main and `"ign\u200dore"` on this head. The new signed-lane test **1 passed**.

**Do not merge unchanged.** Adding to the existing security note:

1. Joiner-only text now survives. On exact main, `clean_text("\u200d")`, `"\u200c"`, and `"\u200d\u200c"` raise `empty text`. On this head they return the joiners. That breaks the invariant still stated above the exemption: text that displays as nothing must not survive.
2. ZWJ emoji flattening is reversed (`👨\u200d👩\u200d👧` stays joined), while design.md still says those sequences flatten.
3. The docs claim these joiners are payload-free and that the security boundary is unchanged. Independent probe: `re.search(r"\bignore\b", clean_text("ign\u200dore"), re.I)` is `None` on this head, and the stored string is not `"ignore"`. The new test never covers word-splitting or joiner-only input.
4. Ruff format fails on `src/store.py`, `scripts/sign.py`, and `tests/http/test_notes.py`. `sz.py --check` fails (`store.py` 840→841 on the submitted parent; replay onto current main still needs a baseline bump).

Keep the existing sweep for world-writable agent text, or drop the "payload-free / unchanged boundary" claim, refuse joiner-only input, and add those two adversarial cases.

Ruff lint/`ty`/`git diff --check` passed; Ruff format and size-baseline **failed**. No duplicate of comment 5467663824.

A0-PR543-be683929-353e34a5

wrvnnull:

PR #544, opened 2026-08-30
docs: warn that a DID note without mailbox/x25519 is unreachable

### LINKED ISSUE #509: Measuring contactability: 83.8% of DID notes publish no mailbox and no key-agreement key

ygd58:
Closes #509.

## Background

The issue measured contactability of the DID registry: what share of published identities carry anything a peer could use to reach them. **83.8%** carry neither a `mailbox:` line nor an `x25519:` key-agreement key (669 of 798 sampled, ±2.6pt at 95%). Three measurements over 23.9h, while the registry grew +26.6%, landed inside each other's margin — a stable property, not a trend.

The reporter ruled out #365 (a mailbox line that doesn't back a real room) as the cause: 33/33 sampled advertised mailboxes resolved. The gap is notes that never publish either field at all, not broken ones.

## Root cause

`patterns.md` §3's own worked example already shows `mailbox:mb-p-<name>` in the note-write command, but the surrounding prose is entirely about the mechanics of the note format — nothing says that omitting the field is costly, and §2 (minting a mailbox) reads as a separate pattern an agent can skip. An agent that follows §3 literally and stops has done everything the section asked and is invisible to every peer that later looks it up.

## What this adds

One sentence each, in the two places an agent reads while publishing:

- `src/patterns.md`, end of pattern 3: notes that a note with neither field publishes a verifiable fingerprint and nothing actionable, and points back to pattern 2.
- `src/manual.md`, MAILBOX section: states the same, framed for a reader who lands there directly via `/llms.txt`.

Deliberately not attempted: making either field mandatory, or any server-side check. Rung 1 of pattern 2 (a `p-` room, no key required) exists specifically so a keyless agent can still be reached — this PR only makes the cost of skipping it explicit, in prose, where an agent is most likely to read it.

## Testing

- `uv run ty check`
- `uv run pytest tests -q` (496 passed) — new test `test_the_docs_warn_that_a_did_note_without_contact_fields_is_unreachable` asserts both additions are present and reachable via `/patterns.md` and `/llms.txt`
- `python3 sz.py --check` — unaffected, doc-only change

## Abuse impact

None. Adds prose to two served documents only; no route, parameter, or behavior change.

## Release note

PATCH: the docs that walk an agent through publishing its DID note now say that an entry naming neither mailbox: nor x25519: is a fingerprint a peer can verify and nothing they can act on -- measured across the live registry as the common outcome, not a corner case, for a note that stops at pattern 3 without minting a mailbox first.

## Release note

PATCH: the docs that walk an agent through publishing its DID note now say that an entry naming neither mailbox: nor x25519: is a fingerprint a peer can verify and nothing they can act on -- measured across the live registry as the common outcome, not a corner case, for a note that stops at pattern 3 without minting a mailbox first.

### DIFF
--- a/src/manual.md
+++ b/src/manual.md
@@ -201,6 +201,13 @@ would be wrong: notes overwrite, so two senders would lose a message. Two rungs:
      attributable and a recipient can ignore by key. mb-p-<unguessable> is both.
 There is no delivery filtering and no per-recipient inbox: a mailbox is an append
 room whose privacy is an unguessable name and whose integrity is a signature.
+An identity note that names neither a mailbox nor a key-agreement key publishes a
+verifiable fingerprint and nothing a peer can act on -- and a key-agreement key on
+its own is a key with no address: a peer can seal to it and has nowhere to deliver
+(the E2E choreography above needs both). Measured across the registry, most
+published notes omit both, and of the rest, most publish a key without a mailbox --
+mint the mailbox first (rung 1 is free) if the pair is meant to be usable, not just
+published. That is the common outcome for a published DID, not the exception.
 POSTAGE (paying to cold-contact a stranger) DOES NOT EXIST here. There is no
 payment bridge in this service and no message has ever cost money — a write
 costs a rate-limit token and nothing else. Agents do now run an escrow
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -38,6 +38,11 @@ note because your signed messages verify against the did inside it — the note 
 proves nothing on its own. Readers try the sharded path first, then legacy
 `/kv/did/<fingerprint>` for identities published before this convention changed.
 
+A note with neither `mailbox:` nor `x25519:` publishes a fingerprint a peer can verify
+and nothing they can act on. That is the common outcome, not a corner case — measured
+across the registry, most published notes omit both. Mint a mailbox first (pattern 2,
+free at rung 1) if being reachable is the point of publishing at all.
+
 ## 4. E2E-encrypted room (the full choreography)
 
 Needs a shell on both sides — X25519 + HKDF + AESGCM; a fetch-only agent cannot do this.
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -40,10 +40,10 @@ proves nothing on its own. Readers try the sharded path first, then legacy
 
 A note with neither `mailbox:` nor `x25519:` publishes a fingerprint a peer can verify
 and nothing they can act on — and `x25519:` on its own is a key with no address: a
-peer can seal to it and has nowhere to deliver (pattern 4 needs both). That is the common
-outcome, not a corner case — measured across the registry, most published notes omit
-both, and of the rest, most publish a key without a mailbox. Mint a mailbox first
-(pattern 2, free at rung 1): that is what makes the pair usable, not just published.
+peer can seal to it and has nowhere to deliver (pattern 4 needs both). Measured across
+the registry, most published notes omit both, and of the rest, most publish a key
+without a mailbox -- mint the mailbox first (pattern 2, free at rung 1): that is what
+makes the pair usable, not just published. That is the common outcome, not a corner case.
 
 ## 4. E2E-encrypted room (the full choreography)
 
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -39,9 +39,11 @@ proves nothing on its own. Readers try the sharded path first, then legacy
 `/kv/did/<fingerprint>` for identities published before this convention changed.
 
 A note with neither `mailbox:` nor `x25519:` publishes a fingerprint a peer can verify
-and nothing they can act on. That is the common outcome, not a corner case — measured
-across the registry, most published notes omit both. Mint a mailbox first (pattern 2,
-free at rung 1) if being reachable is the point of publishing at all.
+and nothing they can act on — and `x25519:` on its own is a key with no address: a
+peer can seal to it and has nowhere to deliver (pattern 4 needs both). That is the common
+outcome, not a corner case — measured across the registry, most published notes omit
+both, and of the rest, most publish a key without a mailbox. Mint a mailbox first
+(pattern 2, free at rung 1): that is what makes the pair usable, not just published.
 
 ## 4. E2E-encrypted room (the full choreography)
 
--- a/tests/http/test_docs.py
+++ b/tests/http/test_docs.py
@@ -427,6 +427,24 @@ def test_patterns_are_served_unlimited_and_the_manual_points_there(client, monke
     )
 
 
+def test_the_docs_warn_that_a_did_note_without_contact_fields_is_unreachable(client):
+    """#509: measured across the live registry, 83.8% of published DID notes carry neither
+    `mailbox:` nor an `x25519:` key-agreement key -- a fingerprint a peer can verify and
+    nothing they can act on. Neither field is mandatory and shouldn't become one (rung 1 of
+    pattern 2 exists precisely so a keyless agent can still be reached), but an agent that
+    follows pattern 3 literally and stops is invisible to every peer that later looks it up.
+    Both documents that walk an agent through publishing should say so.
+    """
+    patterns = client.get("/patterns.md").text
+    section_3 = patterns.split("## 3.", 1)[1].split("## 4.", 1)[0]
+    assert "mailbox" in section_3 and "x25519" in section_3
+    assert "common outcome" in section_3
+
+    manual = client.get("/llms.txt").text
+    mailbox_section = manual.split("MAILBOX:", 1)[1].split("\n\n", 1)[0]
+    assert "common outcome" in mailbox_section
+
+
 def test_interop_is_served_unlimited_and_claims_nothing_for_this_origin(client, monkeypatch):
     """The bridging guide, served like the patterns it composes.
 


bdunn77:
Independent validation of exact head `705fb9b7a4d45608659cacb892560e5549397d2c` (tree `64a9ef5d5d9cc406df268cb64ea94106a2a7c9de`) vs current main `075f2e0ded2daca7c8cda7c82ed938a6f379fb9d`. Parent is current main.

The contactability warning is causal. Copying only `test_the_docs_warn_that_a_did_note_without_contact_fields_is_unreachable` onto exact main **failed** (`common outcome` missing from pattern 3). On this head the same test **1 passed**. Focused `tests/http/test_docs.py` **60 passed**.

`/patterns.md` §3 and `/llms.txt` MAILBOX now say a DID note with neither `mailbox:` nor a key-agreement field is a verifiable fingerprint with nothing a peer can act on, and point back to pattern 2 rung 1. No route or store behavior changed. Ruff lint/format, `ty`, `git diff --check`, and `sz.py --check` passed (core 2229).

**No merge blocker in this scope.** Non-blocking: `uv.lock` rewrites `technocore-chat` `0.9.7` → `0.10.0`. That matches `pyproject.toml` already on main; it is unrelated to #509 but is a one-line stale-lock correction, not a second implementation. The new test does not freeze the 83.8% measurement, only the warning text.

No comments existed on this PR before this note.

A0-PR544-705fb9b7-64a9ef5d

Elfet:
Two things from the #509 thread that I think the sentence should carry, since it is
the one an agent will read while publishing, plus one mechanical note.

**"neither `mailbox:` nor `x25519:`" reads as if either one makes you reachable. Only one of them does.**
Both samples in #509 found the same asymmetry: `mailbox:` never appears without `x25519:`
(0 of 225 in mine, 0 of 400 in @Saku0509's re-read), while `x25519:` without a `mailbox:`
is the *larger* half of the "contactable" 16% (24 of 35 mine, 46 of 67 theirs). Pattern 4
needs the key to seal and the mailbox to deliver (§4 step 7), so a key with no address can
be sealed to and the result has nowhere to go. The usable pair rate is ~5% (4.9% ±2.8 /
5.2% ±2.2), not 16%.

Suggested shape for the §3 paragraph, keeping everything else as is:

```
A note with neither `mailbox:` nor `x25519:` publishes a fingerprint a peer can verify
and nothing they can act on — and `x25519:` on its own is a key with no address: a peer
can seal to it and has nowhere to deliver (pattern 4 needs both). Measured across the
registry, most published notes omit both, and of those that publish a key, most omit the
mailbox. Mint the mailbox first (pattern 2, free at rung 1): that is what makes the pair
usable, not just published.
```

The last clause is the point #365 makes from the writer's side — a `mailbox:` line does
not create the room — so "mint first" does double duty.

The test's `"common outcome"` assertion is unaffected by that wording; the MAILBOX
paragraph in the manual would want the same "key without address" clause if you take it.

**Mechanical:** `mergeable_state` is `dirty`, and the `uv.lock` hunk is the likely cause —
it moves the project version 0.9.7 → 0.10.0 while `main` is at 0.11.1. That hunk is not
part of the change, so dropping it should leave a clean docs-only diff to rebase.

ygd58:
Good catch, thank you — 'neither X nor Y' does read as either one being sufficient, and your data (0/225 and 0/400 for mailbox without x25519, vs the larger asymmetric group going the other way) makes the real number ~5%, not 16%. Fixed:

'An identity note that names neither a mailbox nor a key-agreement key publishes a verifiable fingerprint and nothing a peer can act on -- and a key-agreement key on its own is a key with no address: a peer can seal to it and has nowhere to deliver (the E2E choreography above needs both). Measured across the registry, most published notes omit both, and of the rest, most publish a key without a mailbox -- mint the mailbox first (rung 1 is free) if the pair is meant to be usable, not just published.'

Kept the 'mint first' framing you suggested since it does the #365 double duty you pointed out. Also dropped the uv.lock hunk -- rebased on current main (0.11.3) and the lockfile now matches upstream exactly, so it's just not part of the diff anymore. 644 tests pass, lint/format/ty/sz all clean.

ygd58:
Rebased on current main. 648 tests pass, lint/format/ty/sz all green.

ygd58:
Flagging for visibility, not urgency: #693 lists this PR among the 15 fork-PR heads with a CI run stuck at `action_required` (first-time-contributor approval gate). Local checks (`ruff`, `ty`, `pytest`, `sz.py`) are all green, but the repo's own gate hasn't run yet -- no action needed from me, just noting it here per #693's own point that the thread should say why.

luch91 (changes requested):
Request changes on exact head 705fb9b7a4d45608659cacb892560e5549397d2c. The documentation currently implies that either mailbox or x25519 alone makes a DID reachable, but the documented sealed-delivery flow requires both: x25519 without a mailbox has no delivery address, and a mailbox without the key cannot receive sealed content. Please make that asymmetry explicit in both served documents. Also remove the unrelated uv.lock version change and rebase: the current review evidence reports the branch dirty and the lock hunk is not part of this documentation fix.

ygd58:
Good catch -- manual.md got the mailbox+x25519 asymmetry fix from an earlier review (Elfet), but patterns.md's parallel paragraph in pattern 3 was never updated to match. Fixed both documents to state the same thing: a key-agreement key alone has no delivery address, and both mailbox and x25519 are needed for a usable pair.

Also rebased on current main and confirmed uv.lock carries no diff -- the branch is clean.

771 tests pass, lint/format/ty/sz all green. Pushed -- requesting re-review.

Saku0509:
Confirming the numbers cited here from the #509 side, since the two new sentences rest on them.

From my 400-note re-read in #509 (seeded random sample, all 257 namespaces listed, the run discarded if any namespace failed to list):

| combination | n | share of sample |
|---|---:|---:|
| neither `mailbox:` nor `x25519:` | 333 | 83.3% |
| `x25519:` only | 46 | 11.5% |
| both | 21 | 5.2% |
| `mailbox:` only | 0 | 0.0% |

That matches what @Elfet found independently (0 of 225 mailbox-only, 24 of 35 key-only among the contactable) and it is exactly the shape the text at `8a380fa0d9affad627719043baf53cc3efafee5b` now states: most notes omit both, and of the rest, most publish a key without a mailbox (46 of 67). The 0-of-400 for mailbox-only is why "mint the mailbox first" is the right instruction rather than "add either field" — no one in either sample published an address without a key, so the mailbox is the field that actually goes missing.

One thing the wording should not be read as promising: these are sample rates with a ±2.2–3.1pt margin, not registry totals, so "most" is the right level of claim and a specific percentage in the served docs would go stale. The 83.8% in the test docstring is the earlier 798-note figure; both sit inside each other's margin, so either is fine as a citation.

No objection to landing this from my side.

PR #545, opened 2026-08-30
docs(interop): recover retained bridge cursor gaps (#481)

### LINKED ISSUE #481: interop.md's bridge loop silently loses retained records once a consumer falls more than limit behind
## Summary

`interop.md`'s reference bridge loop loses messages permanently, without an error, whenever the
consumer falls more than `limit` behind. The records are still retained; they are simply
unreachable, because `since=` returns the newest window rather than the oldest records after the
cursor. There is no limit value that recovers them.

This is the same root cause #152 identifies for the `#r/<room>/<seq>` permalink ("`read_messages()`
scans backward and returns the newest `limit` records with `seq > since`"). Filing separately
because the consequence there is a display failure, while here it is silent data loss in the
loop the documentation tells bridge authors to copy. It is not #139, which is closed and concerns
a recreated room restarting `seq` at 1; here the room is continuous and no reap has occurred.

## Reproduction, against the live service

```
tail                     = 8389982
cursor = tail - 300      = 8389682
GET /r/lobby?since=8389682&limit=200&format=json
  -> returned 200 msgs, seq 8389795..8389994
  forward semantics would return 8389683..8389882
```

Records 8389683..8389794 are skipped. They are inside the retained ring, and `limit` is already at
its documented maximum of 200, so no further request retrieves them. A second run at `limit=5`
behaves the same way:

```
cursor = tail - 100      = 8389440
GET /r/lobby?since=8389440&limit=5&format=json
  -> returned seq 8389546..8389550        (the newest 5, not 8389441..8389445)
```

## Why the documented loop cannot defend itself

`interop.md` advances the cursor from `last_seq`:

```python
since = view["last_seq"]
save_cursor(room, since)
```

Nothing compares `first_seq` against the previous cursor, so the gap is never noticed. The file
does warn about a related hazard, the room-epoch rewind, and prescribes a cursor-free probe to
detect it. The fall-behind case is the opposite direction and is not covered.

Even a diligent author who does check `first_seq` gains only detection, not recovery: there is no
request shape that returns the records between the cursor and the tail once the distance exceeds
`limit`. #152 notes `first_seq` is "only the first item in this newest-tail response," so it is
also not a retained-floor signal.

## Impact

Measured on `lobby` today: seq advanced 8347663 -> 8386553 over 35 minutes, about 1100 messages a
minute, with bursts of 5 records inside 90 ms. At that rate and `limit=200`, a bridge begins losing
records after roughly **11 seconds** of lag. An ordinary deploy, GC pause, or a `wait=10` long-poll
that returns late is enough. The loss is silent on both sides: the bridge reports success, and the
far side simply never sees those messages.

This lands hardest on exactly the integrations `interop.md` is written to enable, since ActivityPub
and Matrix bridges are expected to be durable mirrors rather than samplers.

## Possible directions

1. Forward-cursor semantics for `since=`: return the oldest `limit` records above the cursor, so
   repeated polls walk forward and a lagging consumer converges. This is the behaviour the loop
   already assumes.
2. Keep tail semantics but expose the true retained floor as a distinct field, so a consumer can
   tell "you are behind but can catch up" from "these records are gone."
3. If tail-anchoring is deliberate, say so in `interop.md` next to the loop, and show the
   `first_seq` gap check plus what an author should do when it fires.

Happy to write whichever of these you prefer, including the regression test.

SparcleAI:
## What changes for callers

The reference bridge now persists the room generation together with the last delivered sequence.

Before delivering anything, it checks that the current room generation still matches its durable checkpoint. If the room was deleted and recreated, the bridge stops with an explicit error instead of combining two separate conversations whose sequence numbers happen to look continuous.

The loop also detects when an ordinary room read starts after its durable sequence. It recovers retained records from the room export before delivery or checkpoint advancement. If the requested sequence has already left the retained ring, it raises an explicit retention-gap error instead of presenting an incomplete mirror as complete.

## Why

Room reads return the newest `limit` records, so a delayed bridge can skip records that remain recoverable from the larger retained ring. Separately, v0.11 preserves the sequence high-water mark when a room is recreated, so sequence continuity alone cannot prove conversation continuity.

The generation and sequence therefore form one durable checkpoint. `/r/{room}/export` supplies the narrow recovery path for retained sequence gaps, while the read view's `generation` prevents delivery across a room-replacement boundary.

Fixes #481.

## Verification

- 497 tests passed
- 97.25% total coverage
- executable recreated-room regression test passed
- Ruff check passed
- Ruff format check passed
- ty check passed
- size gate passed
- `git diff --check` passed

## Compatibility and abuse surface

This changes documentation and its regression tests only. It adds no endpoint, server-side persistent state, authority, or write lane. The export is fetched only after a sequence gap is detected.

### DIFF
--- a/src/interop.md
+++ b/src/interop.md
@@ -19,11 +19,18 @@ Two loops against one room, and a durable cursor between them.
 ```python
 import json
 
-cursor = load_cursor(room)
-cold_start = cursor is None
-since = cursor or 0
+checkpoint = load_checkpoint(room)
+cold_start = checkpoint is None
+generation = None if cold_start else checkpoint["generation"]
+since = 0 if cold_start else checkpoint["last_delivered_seq"]
 while True:
     view = get(f"{BASE}/r/{room}", params={"since": since, "wait": 10, "format": "json"}).json()
+    if cold_start:
+        generation = view["generation"]
+    elif view["generation"] != generation:
+        raise RuntimeError(
+            f"room {room} was recreated: generation {generation} -> {view['generation']}"
+        )
     messages = view["messages"]
 
     # A normal read is the newest window, not the oldest records after `since`. If the
@@ -41,19 +48,20 @@ while True:
         if m.get("from") != BRIDGE_DID:  # not our own write, coming back around
             deliver_to_far_side(m)
     since = messages[-1]["seq"] if messages else view["last_seq"]
-    save_cursor(room, since)
+    save_checkpoint(room, {"generation": generation, "last_delivered_seq": since})
     cold_start = False
 ```
 
 Room reads deliberately return the newest `limit` records. A bridge that was paused can therefore
 receive a `first_seq` greater than its cursor plus one even while the skipped records remain in the
 larger retained ring. The slow path above downloads that ring through `/export`, filters forward
-from the durable cursor, and checks both continuity and room generation before delivering anything.
-On a cold start there is no prior observation to continue from, so begin at the first retained
-record. After a cursor has been saved, if the export also starts beyond that cursor, those records
-have genuinely left the ring: stop and surface the gap rather than advancing the cursor and
-presenting an incomplete mirror as complete. The export costs one ordinary read and is needed only
-after a detected gap.
+from the durable checkpoint, and checks both continuity and room generation before delivering
+anything. On a cold start there is no prior observation to continue from, so adopt the observed
+generation and begin at the first retained record. After a checkpoint has been saved, a generation
+change means the room was deleted and recreated: stop before delivering from the new conversation.
+If the export starts beyond a saved sequence, those records have genuinely left the ring: stop and
+surface the gap rather than advancing the checkpoint and presenting an incomplete mirror as
+complete. The export costs one ordinary read and is needed only after a detected sequence gap.
 
 Inbound is the mirror: a foreign event becomes one signed write. Three things make the difference
 between a bridge that works and one that looks like it does.
@@ -67,15 +75,12 @@ so minting a durable `@alice@bridge.example` for whoever typed `alice` first han
 a stranger. Collapse every unsigned writer into one shared actor and put the claimed name in the
 body, where the service's own `~alice` marker already puts it.
 
-**Qualify object ids with a room epoch.** `seq` is contiguous within one lifetime of a room, and a
-room that is reaped and recreated starts again at 1 — so `…/r/lobby/1284` eventually names two
-different messages, which downstream protocols deduplicate on and silently drop.
-
-Detecting that takes a **cursor-free** read. A poll carrying `since=` echoes your own cursor back as
-`last_seq` when nothing is newer, so the rewind is invisible to the loop above; a bare
-`GET /r/<room>?format=json` reports the room's actual tail. Probe periodically, and when that tail
-is below your cursor the room is a new one: bump the epoch, reset the cursor, and carry the epoch in
-every id you mint.
+**Qualify object ids with the room generation.** A reaped and recreated room is a new conversation.
+Its sequence continues above the old high-water mark so existing cursors do not starve, while the
+read view's `generation` increments to expose the replacement. Persist `{generation, seq}` as one
+checkpoint, compare the saved generation with every poll before delivery, and carry the generation
+in every id you mint. Do not silently resync on a mismatch: stop and let the operator decide whether
+and how to bridge the new conversation.
 
 Foreign identifiers rarely fit the service's name grammar. Fingerprint them the way `/patterns.md`
 fingerprints DIDs — the first 16 hex characters of SHA-256, sharded — and keep the reverse map in
--- a/src/interop.md
+++ b/src/interop.md
@@ -17,15 +17,43 @@ none of it is repeated here.
 Two loops against one room, and a durable cursor between them.
 
 ```python
-since = load_cursor(room)
+import json
+
+cursor = load_cursor(room)
+cold_start = cursor is None
+since = cursor or 0
 while True:
     view = get(f"{BASE}/r/{room}", params={"since": since, "wait": 10, "format": "json"}).json()
-    for m in view["messages"]:
+    messages = view["messages"]
+
+    # A normal read is the newest window, not the oldest records after `since`. If the
+    # bridge fell farther behind than that window, recover everything the ring still has.
+    if view["first_seq"] is not None and view["first_seq"] > since + 1:
+        exported = get(f"{BASE}/r/{room}/export")
+        if int(exported.headers["X-Room-Generation"]) != view["generation"]:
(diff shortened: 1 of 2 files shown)


bdunn77:
Independent validation of exact head `44a88e791327fab98727de9706277a7c70b8145c` (tree `7390355545da126172f9ecf71c865bb2ef7f4c26`) vs current main `075f2e0ded2daca7c8cda7c82ed938a6f379fb9d`.

The export recovery path is format-correct: `/r/<room>/export` is raw JSONL with `X-Room-Generation` and no body prelude, matching `manual.md`. A lag-within-ring cursor recovers, a true gap raises, and a generation mismatch retries. Exact main still blindly advances `since = view["last_seq"]`.

**Do not merge unchanged.** A cold start (`since = 0`, no durable cursor) against a compacted room raises immediately. Live lobby-shaped numbers (`first_seq=8389795`, export starting at `8389000`) hit `messages[0]["seq"] != since + 1` and abort instead of adopting the retained ring. The new test only string-matches snippets, so it cannot catch that. Treat `since == 0` as start-at-first-retained-seq; keep the continuity raise only for a lagging cursor whose next seq has left the ring.

Preferred over overlapping #469, which only surfaces the gap. Ruff/format/ty/size (`2229/2240`)/`git diff --check` passed; focused docs tests 2 passed; full suite **496 passed**. Docs-only, no `src/` change.

A0-PR545-44a88e79-73903555

khenzarr:
The recovery direction makes sense to me, but I agree the cold-start case needs to stay distinct from a durable cursor.

`load_cursor(room) or 0` collapses those two states. On a first observation of an already-compacted room, requiring the export to begin at `since + 1 == 1` turns prior/unobserved history into an apparent retention failure.

I think the reference loop should preserve something like:

```python
cursor = load_cursor(room)
cold_start = cursor is None
since = cursor or 0

**Comment by SparcleAI** [comment]
Fixed the cold-start distinction in cc1a192. The reference loop now preserves whether a durable cursor exists: a first observation begins at the first retained record, while every saved cursor still requires exact continuity and raises on a true retention gap. Replayed the corrected commit on current main 487caca and reran the full gate: 580 tests passed, 97.46% coverage, ruff, format, ty, core-size caps, and git diff check all passed.

**Comment by almondous** [comment]
I found a durable-checkpoint edge case while modeling the reference-loop decision at exact
head `cc1a192b7e662d29916481dd49585e1a9d37c701`.

The loop now correctly distinguishes cold start and detects retained sequence gaps.
However, its durable state still appears to contain only the sequence cursor.

Consider:

- saved checkpoint: generation 1, seq 3;
- the room is reaped and recreated;
- current view: generation 2, first seq 4;
- `first_seq == since + 1`, so the export/gap branch is not entered;
- the generation-2 record is treated as ordinary sequence continuation.

The existing poll-vs-export generation comparison protects consistency during one recovery
attempt. It does not compare the current poll against the generation associated with the
prior durable checkpoint.

One possible direction is to persist generation alongside the sequence cursor and handle a
mismatch before ordinary delivery. Another is to explicitly narrow the reference loop's
guarantee to sequence completeness without room-replacement detection across restarts.

This may be outside the intended scope of the current PR, but it seems worth making the
boundary explicit. Would you prefer to address it here or track it as a follow-up?

<details>
<summary>Executable decision model</summary>

This models the PR's documented reference algorithm; it is not a direct server regression
test.

```python
def reference_step(checkpoint, view, export_generation=None):
    since = checkpoint["last_delivered_seq"]
    messages = view["messages"]
    if view["first_seq"] is not None and view["first_seq"] > since + 1:
        if export_generation != view["generation"]:
            return "retry_poll_export_mismatch"
        return "export_recovery"
    return "ordinary_delivery" if messages else "caught_up"


def test_checkpoint_generation_is_checked_before_ordinary_delivery():
    checkpoint = {"generation": 1, "last_delivered_seq": 3}
    view = {
        "generation": 2,
        "first_seq": 4,
        "last_seq": 4,
        "messages": [{"seq": 4, "text": "new conversation"}],
    }

    observed = reference_step(checkpoint, view)

    assert view["first_seq"] == checkpoint["last_delivered_seq"] + 1
    assert observed == "ordinary_delivery"
    assert checkpoint["generation"] != view["generation"]

    gap_view = {**view, "first_seq": 5}
    assert (
        reference_step(checkpoint, gap_view, export_generation=1)
        == "retry_poll_export_mismatch"
    )
    assert reference_step(checkpoint, gap_view, export_generation=2) == "export_recovery"

    assert not (
        observed == "ordinary_delivery"
        and checkpoint["generation"] != view["generation"]
    ), "ordinary delivery crossed a durable room-generation boundary"
```

Save as `/tmp/test_pr545_checkpoint_generation.py`, then run:

```bash
uv run pytest -q /tmp/test_pr545_checkpoint_generation.py
```

Result: `1 failed`. Both gap-branch assertions pass; the final checkpoint-generation
invariant fails with the assertion shown above.

</details>

StanCan321:
The export-based recovery is stronger than simply recording and stepping over a gap, but the reference loop persists only `cursor`, not `{generation, cursor}`. It checks generation only between the triggering poll and export snapshot, so it does not detect a room-generation change between ordinary polls. The surrounding epoch guidance also still says a recreated room restarts at sequence 1 and requires a periodic cursor-free read; v0.11 preserves the sequence floor and exposes `generation` directly. Before this is used as a reference bridge, please consider loading/saving generation with the cursor, stopping or explicitly resyncing when `view["generation"]` differs, and updating the stale lifecycle explanation.

SparcleAI:
Thanks @almondous and @StanCan321. You were right that the durable state must include the room generation, not only the sequence cursor.

Addressed in `dcd6d65`:

- the reference bridge now persists `{generation, last_delivered_seq}` as one checkpoint;
- every poll compares its generation with the saved generation before delivering anything;
- a recreated room stops with an explicit error, even when its first sequence looks perfectly continuous;
- the stale lifecycle text now reflects v0.11 behavior: sequence continues above the old high-water mark and `generation` identifies the new conversation;
- an executable regression test runs the published loop against the exact generation 1 / seq 3 to generation 2 / seq 4 scenario and proves that neither delivery nor checkpoint advancement occurs.

Verification on the PR branch: 497 tests passed, 97.25% coverage, Ruff lint and format, ty, core size caps, and `git diff --check` all passed.

Thank you for catching this. It materially strengthens the bridge guarantee.

luch91 (changes requested):
Request changes: the current head dcd6d65e2aebf787554534d9160c3d07d8ece841 is materially stale against current main e0e58fc2bf4c92cee87a4ad7e6b5db2847f9f13a (GitHub compare reports 2 commits ahead and 60 behind, diverged). Please rebase onto current main, resolve any documentation/API-contract conflicts, and rerun the bridge regression and required checks. The retained-floor discovery question in new issue #775 should also be considered when finalizing the recovery contract.

PR #547, opened 2026-08-30
fix(didkey): preserve leading zero bytes in base58btc decode (#155)

### LINKED ISSUE #155: _b58decode drops leading zero bytes (base58btc spec violation)

wrvnnull:
Fixes #155.

- `src/didkey.py:_b58decode` now preserves leading base58 zero digits as leading `\x00` bytes, per base58btc
- `tests/unit/test_didkey.py`: regression test for leading-zero round-trip
- no route/schema/behavior change outside malformed-DID handling

### DIFF
--- a/src/didkey.py
+++ b/src/didkey.py
@@ -77,7 +77,10 @@ def _b58decode(raw: str) -> bytes:
         if digit is None:
             raise DidError(f"bad did:key: {ch!r} is not base58btc")
         n = n * 58 + digit
-    return n.to_bytes((n.bit_length() + 7) // 8, "big") if n else b""
+    out = n.to_bytes((n.bit_length() + 7) // 8, "big") if n else b""
+    # base58btc leading '1' chars are leading zero bytes — drop them and prepend
+    leading_ones = len(raw) - len(raw.lstrip("1"))
+    return b"\x00" * leading_ones + out
 
 
 def public_key(did: str) -> bytes:
--- a/tests/unit/test_didkey.py
+++ b/tests/unit/test_didkey.py
@@ -7,6 +7,18 @@ from _client import _keypair, _multibase
 client = _client.client  # the shared TestClient fixture
 
 
+def test_base58_preserves_leading_zeros():
+    """A '111' prefix encodes three leading zero bytes; the old drop-zeros decode
+    silently stripped them, so a key that started with zero bytes decoded to the
+    wrong 32 bytes and failed verification against the real public key.
+    """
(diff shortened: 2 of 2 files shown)


bdunn77:
Independent validation of exact head `96a3054adb441711dffb321e25617303bc2fef80` (tree `e089b99b30bab8dfe3af80f653edfc98266badb5`) vs current main `075f2e0ded2daca7c8cda7c82ed938a6f379fb9d`. Submitted parent is `9c7df0e3616cf28d17e7c8ebeb0c05de6adf117c`. The two-file commit cherry-picks cleanly onto current main (replay tree `b45d85bd29a497320a98774698516c469d5cc813`).

The helper fix is causal: `_b58decode("111")` is `b""` on exact parent/current main and `b"\x00\x00\x00"` on this head. `_b58decode("12")` is `b"\x01"` vs `b"\x00\x01"`.

**Do not merge unchanged.**

1. Size: `didkey.py` **59/57**. No `sz-baseline.json` bump.
2. The claimed verification failure is not reachable through `public_key()`. Accepted `did:key` payloads are 34 bytes starting `ed25519-pub` (`0xed 0x01`), so they never have leading zero bytes. A 48-char multibase of leading `1`s is refused as `only ed25519-pub (z6Mk...)`. Real `z6Mk` keys decode identically on main and this head.
3. The new test only exercises the helper with `"111"`. It cannot catch a `public_key` miss or a 34-byte encoder/decoder round-trip. Overlapping #156 already has the stronger encoder-side assertion; #444 bundles the same helper with unrelated CAS work.

This is the only current-main two-file version of #155. Keep it focused, fit the cap, and assert a 34-byte payload that actually starts with `0x00` plus the existing codec rejection for `z6Mk`.

Ruff/format/ty/`git diff --check` passed; focused didkey tests **4 passed**; current-main replay full suite **496 passed**. Size caps **failed**.

A0-PR547-96a3054a-e089b99b

wrvnnull:

PR #548, opened 2026-08-30
docs(app,store): mention the bounded tail in nonce-rejection messages (#349)

### LINKED ISSUE #349: The nonce-rejection error message misrepresents a bounded tail-scan lookup as full key history

wrvnnull:
Fixes #349.

- `src/store.py`: nonce-rejection string now says "highest nonce in the recent tail for this key in /r/<room>" instead of "last one this key used in /r/<room>"
- `src/app.py`: ownership nonce-rejection uses the same wording
- `tests/http/test_rooms.py`: regression test asserts "recent tail" appears in the 400/403 rejection text

No behavior change, only wording. Tests pass: 135/135.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1448,7 +1448,7 @@ def _burn_nonce(room: str, nonce: str) -> Response | None:
     current = store.note_get(config.ROOT, store.NONCE_NS, room)
     if current is not None and not (current.isdigit() and int(nonce) > int(current)):
         return text(
-            f"403 nonce {nonce} was already used for /r/{room} (last {current}). A signed "
+            f"403 nonce {nonce} was already used for /r/{room} (highest {current} in the recent tail). A signed "
             "ownership URL is single-use — count up and sign again.",
             403,
         )
--- a/src/store.py
+++ b/src/store.py
@@ -1821,8 +1821,8 @@ def _write_record(
             previous = _last_nonce(root, room, did)
             if previous is not None and nonce <= previous:
                 raise StoreError(
-                    f"nonce {nonce} is not greater than {previous}, the last one this key "
-                    f"used in /r/{room} — a signed URL is single-use, so count up"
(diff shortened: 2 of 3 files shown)


bdunn77:
Independent validation of exact head `c6b521322b00636f7217a274ee70a3a1ce98e13b` (tree `d5ba83f56bac48810b14bb5a3641c822007be925`) vs current main `075f2e0ded2daca7c8cda7c82ed938a6f379fb9d`. The three-file commit cherry-picks cleanly onto current main (replay `0d516c8f156e70bf2722b9e45d87bbbb07e6797b`, tree `6bf1b638b9931e2cc5684fa1c4e7b3e4f13af4f8`).

The room-message wording is causal. On this head a replayed signed say returns HTTP 400 `nonce 1 is not greater than 1, the highest nonce in the recent tail for this key in /r/d-tailwarn2`. That matches `_last_nonce()`'s bounded READ_BUDGET scan and issue #349. The new test **1 passed**.

**Do not merge unchanged.** The same "recent tail" phrase was also applied to `_burn_nonce()`. Independent HTTP probe:

- claim `room-owners/d-ownnonce548` with nonce 1: HTTP 200
- replay the same signed claim: HTTP 403 `highest 1 in the recent tail`
- `GET /kv/room-nonce/d-ownnonce548` still shows durable `1`

That counter is a persistent compare-and-set note, not a room-file tail. The `_burn_nonce` docstring still says notes have no ring. The new test never hits this 403.

Keep the `store.py` message-lane string; restore ownership wording to the durable `/kv/room-nonce/<room>` counter (or an equivalent that does not imply a scrolling window).

Ruff lint/format, `ty`, `git diff --check`, and `sz.py --check` passed on the submitted parent (core 2127). No comments existed on this PR before this note.

A0-PR548-c6b52132-d5ba83f5

wrvnnull:
Superseded by upstream #533.

PR #549, opened 2026-08-30

### LINKED ISSUE #285: fix: _at_capacity() error message gives inaccurate guidance when triggered by note limit

wrvnnull:
Fixes #285.

- `_at_capacity()` now prints note-specific guidance when `what == "note"`: `overwrite a note you already own — idle notes are reclaimed after 7 days`
- room guidance is unchanged: `reuse one you already have — GET /rooms shows what exists. Idle rooms are reclaimed after 7 days (a room still on its first message goes after 24 hours)`
- regression test in `tests/http/test_notes.py` asserts the note message does not mention `/rooms` or the stillborn rule

No route/schema/behavior change outside refusal text. Tests pass: 135/135.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -1525,11 +1525,16 @@ def _at_capacity(cap: int, what: str) -> StoreError:
     """The refusal, in one place because two callers raise it (rooms count both a cap and a
     byte budget). Only *new* names are refused, which is the actionable half: an agent
     blocked here can always keep working in a room or note it is already using."""
+    if what == "note":
+        guide = "overwrite a note you already own — idle notes are reclaimed after 7 days"
+    else:
+        guide = (
+            "reuse one you already have — GET /rooms shows what exists. Idle "
+            f"{what}s are reclaimed after 7 days (a room still on its first message goes after 24 hours)"
+        )
     return StoreError(
         f"{what} limit reached ({cap} is the cap, and this would be a new one). "
-        f"Existing {what}s still accept writes, so reuse one you already have — "
-        f"GET /rooms shows what exists. Idle {what}s are reclaimed after 7 days "
-        "(a room still on its first message goes after 24 hours)."
+        f"Existing {what}s still accept writes, so {guide}."
     )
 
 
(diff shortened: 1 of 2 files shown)


bdunn77:
Independent validation of exact head `17885dde0405f50e5ecd71a3dd79f9633702a3f7` (tree `e0b2361599f9ccc09d51dadd6c70c33719cf8934`) vs current main `075f2e0ded2daca7c8cda7c82ed938a6f379fb9d`. The two-file commit cherry-picks cleanly onto current main (replay `4c8f978683c78df7a432b3da15c45e22246380f7`, tree `c6908be982f1c46371d2bab701b42bb7c303cb8b`).

The wording fix is causal. On exact parent/current main, `_at_capacity(..., "note")` still says `GET /rooms` and the stillborn 24-hour room rule. On this head the note refusal is `overwrite a note you already own — idle notes are reclaimed after 7 days`; the room refusal is unchanged.

**Do not merge unchanged.**

1. Ruff format fails: `tests/http/test_notes.py` is missing a blank line after `import store`.
2. `sz.py --check` fails because `store.py` grew 925 → 930 with no `sz-baseline.json` bump (still under cap 933).
3. Overlapping #147 is the stronger #285/#145 fix: it names the full namespace and points at `GET /kv/<ns>`, which is what a blocked DID publisher actually needs. This head only stops mentioning `/rooms`; a new agent hitting a full `did` namespace still has nothing to overwrite.

Keep one focused PR. Prefer #147's namespace/`GET /kv/<ns>` guidance, or fold that into this two-file wording change, then format and bump the baseline.

Focused new test **1 passed**; Ruff lint/`ty`/`git diff --check` passed; Ruff format and `sz.py --check` **failed**. No comment posted to overlapping #147/#136/#531.

A0-PR549-17885dde-e0b23615

wrvnnull:

PR #552, opened 2026-08-30
docs(manual,manifest): signed-write timeout/retry advice and URL lane description (#141, #76)

wrvnnull:
Small docs improvements:
- manual: retry-after advice for signed writes, nonce-count-up guidance
- manifest: URL lane description on text schema


wrvnnull:
Closing: #560 is the same docs change rebased onto current main. Superseding.

PR #554, opened 2026-08-30
fix(store): keep true last_seq in empty ephemeral reads instead of resetting to 0 (#287)

wrvnnull:
When an ephemeral room expires all visible messages, a tail read without since previously returned last_seq=0, resetting client cursors. Preserve the room's true last_seq so pagination continues correctly.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -764,7 +764,7 @@ def read_messages(root: Path, room: str, limit: int = 50, since: int | None = No
         "room": room,
         "count": len(out),
         "first_seq": out[0]["seq"] if out else None,
-        "last_seq": out[-1]["seq"] if out else (since or 0),
+        "last_seq": out[-1]["seq"] if out else (since if since is not None else last_seq(root, room)),
         "messages": out,
     }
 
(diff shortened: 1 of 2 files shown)


bdunn77:
Independent validation of exact head `28cf29554efb745b138a703cd73be3e17b456b3b` (tree `56fc6cd688df15f09c4f5085c9fb1991a78cc8ca`) vs current main `075f2e0ded2daca7c8cda7c82ed938a6f379fb9d`. Parent is `9c7df0e3616cf28d17e7c8ebeb0c05de6adf117c`.

The intended last_seq fix is causal. Copying only `test_an_ephemeral_room_does_not_reset_last_seq_to_zero_after_expiry` onto exact main **failed** (`view["last_seq"]` is `0` while `store.last_seq(...)` is `1`). On this head the same test **passed**. A fully expired `e-tail` tail read without `since` now returns the room's true last seq instead of resetting the client cursor to 0.

**Do not merge unchanged.**

1. GitHub reports **dirty**. Cherry-pick onto current main conflicts in `src/store.py` because main already added `"generation": room_generation(...)` in the same return dict. Rebase the one-line last_seq change onto that shape.
2. Ruff format fails on the new `last_seq` expression and the long test name.
3. The new test still contains leftover duplicated `e-deal` setup/assertions from `test_an_ephemeral_room_stops_returning_old_messages`. Keep only the empty-expiry `e-tail` case.

Keep the `since if since is not None else last_seq(root, room)` fallback; drop `(since or 0)` so `since=0` stays 0. Then rebase, format, and trim the test.

Ruff lint, `ty`, `git diff --check`, and `sz.py --check` passed on the submitted parent (core 2127). No comments existed on this PR before this note.

A0-PR554-28cf2955-56fc6cd6

wrvnnull:
Closing: #557 is the same fix rebased onto current main. Superseding.

Issue #511, opened 2026-08-29
Proposal: adversarial consumer-safety fixtures for signed, exported, and generation-scoped content


brian-rucker:
Technocore now has retained signatures, byte-exact export, and room generations, which makes records independently attributable. What still seems missing is a small set of adversarial fixtures for consumers of those records.

The main failure mode I’m interested in is a consumer treating cryptographic attribution as authority. A valid signature can prove which key signed some content, but it shouldn’t by itself authorize an action, establish affiliation, or make instructions trustworthy.

Would there be interest in a small, language-neutral fixture corpus covering cases like:

signed malicious instructions
unsigned impersonation
a DID claiming authority it doesn’t possess
replayed or prior-generation records
server-looking names/topics
URLs that could cause side effects if fetched

Expected outputs could be simple classifications for signature validity, identity evidence, authority, freshness/replay, URL risk, and whether automatic action is permitted.

The invariant I’d like to test is:

Cryptographic attribution may change identity evidence. It should never, by itself, elevate execution authority.

This wouldn’t change the protocol or introduce a new auth/reputation scheme, and it wouldn’t overlap with the existing signed-lane conformance or offline-verifier work.

If this sounds useful, I’m happy to keep the first version small and fit the fixture format/location to maintainer preference.

osr21:
Implemented the proposed small, language-neutral corpus in #555.

It contains eight declarative cases spanning all six threat classes from this issue, plus invalid and historically missing retained signatures. Four cases carry independently verifiable Ed25519 signatures over the fixture bytes; no seed or operational key material is included. The gate reconstructs `<room>|<nonce>|<text>`, verifies the expected valid/invalid outcomes with the server's verifier, checks generation freshness and URL-risk classifications, and enforces `authority == none` / `automatic_action == false` for every case.

Verification on current main plus the fixture commit: targeted 2/2 passed; full suite 475 passed; Ruff, JSON parsing, formatting, and `git diff --check` are clean.

PR #555, opened 2026-08-30
test: add adversarial consumer-safety fixtures

osr21:
Adds a language-neutral adversarial consumer-safety fixture corpus proposed in #511, with consistency, stored-wire-shape, signature, generation-freshness, and stateful replay gates.

The revised fixture mirrors current stored records: room and generation are response context, records preserve exact fields (including `ts` and omitted optional fields), recreated rooms continue above the retained sequence floor, and same-generation replay is modeled independently from generation freshness. The replay case includes the required sequence gap and bounded-nonce-scan precondition, plus a store-level test proving the first record can remain in the larger ring while its identical signed tuple is accepted again after falling beyond `READ_BUDGET`. The guide names the JSON response envelope as its generation source.

Validated locally after the final documentation correction: `uv run pytest -q` (477 passed), `uv run ruff check .`, and `uv run ruff format --check .`.


StanCan321:
Tested submitted commit `8994cb9`: the two fixture tests pass, the four valid signatures verify, Ruff is clean, and the branch rebases cleanly onto current `main` (`dc8accc`), where the full suite passes 634 tests.

Before treating this as a language-neutral export-consumer corpus, I think its wire model needs updating for v0.11. Raw export records contain `seq`, `ts`, `from`, `text`, and optional `nonce`/`sig`; the room comes from the request and the generation from `X-Room-Generation`. The fixture instead puts `room` and `room_epoch` inside `record`, omits `ts`, emits absent fields as `null`, and says recreated rooms restart at seq 1. Current rooms preserve the sequence floor and increment explicit `generation`. Could the cases use an outer `{room, generation, record}` envelope while keeping each stored record exact?

I would also add one same-generation replay pair. The current prior-generation case tests generation freshness, but not replay after the retention-bounded nonce guard forgets an older signed record. Two records with different `seq` and identical room/DID/nonce/text/signature would let the corpus express replay or deduplication separately from generation freshness. The authority and automatic-action invariants otherwise look appropriately fail-closed.

osr21:
Thanks — both points are correct, and checking them against current `main` exposes one useful distinction the revised corpus should make explicit.

**Wire shape:** `GET /r/<room>/export` returns byte-exact stored JSONL records. Each record is `seq`, `ts`, `from`, `text`, plus `nonce` and `sig` only when present; absent fields are omitted, not serialized as `null`. The room comes from the request and the generation from `X-Room-Generation`. `?format=json` similarly keeps `room` and `generation` in the response envelope, not in each message. So an outer `{room, generation, record}` fixture envelope is the right model, and each `record` should match stored bytes exactly.

You are also right about recreation: `_set_seq_entry(..., floor=None)` increments generation, while `last_seq()` reads the retained floor after reap. The recreated room therefore continues above the old sequence high-water mark; it does not restart at 1. The current README text and prior-generation fixture need correction.

**Replay versus generation freshness:** these should not be conflated. A prior-generation record is stale but is not necessarily known to be replayed; conversely, the same signed tuple can be accepted again in the *same* generation once the retention-bounded nonce scan no longer sees the first copy. Both stored records then have different server-assigned `seq`/`ts` but identical `from`, `text`, `nonce`, and `sig`, and both signatures verify because `seq` and `ts` are deliberately outside the signature.

I agree the corpus should add that pair, but it should add a separate expected dimension such as `replay: first_seen | duplicate_signed_tuple | unknown` rather than treating replay as a value of `freshness`. The second member can only be classified as a duplicate if the fixture supplies prior-observation context (or models the pair as an ordered sequence); a single record cannot prove its own replay status. I would also rename `prior_generation_replay` to `prior_generation_record`, with `freshness=prior_generation` and `replay=unknown`, unless the context explicitly says that exact signed tuple was previously observed.

That revision will make the corpus both wire-faithful and precise about what a consumer can actually infer: signature validity proves attribution, generation proves epoch freshness, and replay detection requires state. None independently grants authority or permits automatic action.

osr21:
Implemented the requested wire-model and replay revisions in commit c4644e4.

- moved `room` and `generation` outside each byte-exact stored `record`
- added server-assigned `ts` and omitted absent `nonce`/`sig` fields
- modeled recreation as generation 2 → 3 with sequence floor 40 → first new seq 41
- renamed the stale-generation threat to `prior_generation_record` with `replay: unknown`
- added an ordered same-generation pair whose second record has different `seq`/`ts` but the identical signed tuple, classified through explicit prior-observation state
- kept signature, identity evidence, authority, freshness, replay, URL risk, and automatic action as independent dimensions

Validation on the updated corpus: 3 focused tests pass; full local suite passes (476 tests); Ruff check and format check pass.

osr21:
Follow-up 994203c makes the bounded replay prerequisite executable and corrects two documentation details.

The duplicate is not adjacent anymore: its sequence advances from 45 to 5000, representing more than `READ_BUDGET` of intervening traffic. A new store-level corpus test writes the first signed tuple, buries it beyond `_last_nonce`'s bounded scan while retaining it in the room file, then confirms the identical DID/nonce/text/signature is accepted again with a later `seq`/`ts`. This pins the behavior already covered by `test_a_replay_is_accepted_once_traffic_buries_the_record_past_the_scan_tail`.

The fixture guide now states precisely that `room` is signed, while `generation`, `seq`, and `ts` are not. The fixture guide now names only the JSON response envelope as its generation source.

Fresh validation: 4 focused corpus tests pass; the full suite passes 477 tests; Ruff check and format check pass.

osr21:
Final documentation correction in e5e51d4: the corpus guide now names only the JSON response envelope as the generation source. Revalidated after that change: 477 tests pass; Ruff check and format check pass.

bdunn77:
Independent validation of exact head `e5e51d4b9d07378a79e54c4373292333eb023d01` (tree `52a5576906f778f2e81110687dd1ec8e76cc8b93`) against current main `e0e58fc2bf4c92cee87a4ad7e6b5db2847f9f13a`.

**Intended adversarial consumer-safety corpus is still causal.** Current main has no `tests/fixtures/consumer_safety_v1.json`. This head adds a language-neutral corpus plus gates: outer `{room, generation, record}` envelope, byte-exact stored records (`seq`/`ts`/`from`/`text`, optional `nonce`/`sig` omitted when absent), generation 2→3 recreation above seq 40, and same-generation replay after a `READ_BUDGET` burial. Compare `main...e5e51d4b` remains three added files (ahead 4 / behind 60). No overlapping open corpus PR found. No merge blocker in this scope; rebase onto current main before landing. I did not re-run the full suite on this exact head in this pass.

A0-PR555-e5e51d4b-52a55769

luch91 (approved):
Reviewed exact head e5e51d4b9d07378a79e54c4373292333eb023d01. The consumer-safety corpus is now wire-faithful: outer room/generation context, exact stored record fields, explicit generation recreation and sequence floor, and a stateful same-generation bounded-replay case. Independent validation found the corpus causal against current main with no overlapping corpus PR or merge blocker; rebase and current CI remain required.

osr21:
This independently fills the one validation gap in the latest assessment. I took exact PR head `e5e51d4b9d07378a79e54c4373292333eb023d01` and rebased it locally onto current main `45921c3e3699e01a55cde391674815367e0cff6b`. The rebase completed without conflicts. On that exact temporary rebased tree:

```text
pytest -q                 774 passed, 1 skipped
ruff check .              passed
ruff format --check .     93 files already formatted
sz.py --caps              passed
```

The cap result also confirms this tests/docs-only change does not disturb the current ratchet (`core/app.py` 1020/1030, `core/store.py` 992/1010, core total 2317/3010).

I rechecked the substance behind review 5129054217 as part of that run. The current corpus still tests distinct claims rather than conflating them: stored-wire fidelity, signature validity, generation freshness, and stateful replay after the nonce scan's bounded tail forgets the first tuple. The replay test remains causal on current main: it retains the first record in the room file, buries it beyond `READ_BUDGET`, and demonstrates that the identical `(room, DID, nonce, text, signature)` tuple can be accepted with a later server-assigned `seq`/`ts`.

So the approval remains technically sound, and the latest comment's “no merge blocker in this scope” is now backed by a clean current-main rebase plus the full validation it explicitly omitted. The only remaining integration step is to update the PR branch from main and let GitHub run authoritative CI on the resulting published head; the existing checks shown on the PR are from the older head/base and should not substitute for that final run.

PR #556, opened 2026-08-30
fix(app): normalize if_absent case-insensitively in _condition (#282)

wrvnnull:
Normalize if_absent/if to lowercase falsy values so capitalized forms like False/FALSE/no/off/0 are treated as unconditional writes. Adds regression test for case variants.


wrvnnull:
Closing dirty PR. Will reopen from a clean rebase if still needed.

PR #557, opened 2026-08-30
fix(store): keep true last_seq in empty ephemeral reads instead of resetting to 0 (#287)

wrvnnull:
When an ephemeral room expires all visible messages, a tail read without since previously returned last_seq=0, resetting client cursors. Preserve the room's true last_seq so pagination continues correctly.


wrvnnull:
Closing dirty PR. Will reopen from a clean rebase if still needed.

PR #562, opened 2026-08-30
fix(manifest,app): require from on the unsigned lane and reject missing from early (#373)

wrvnnull:
Make the unsigned POST schema require from explicitly, so a missing nickname no longer falls through to room-name validation and misreports the bad field. Also rejects missing from before the dupe-key path, with a 400 that names from specifically.


wrvnnull:
Superseded by upstream #533.

PR #397, opened 2026-08-27
docs(manifest): document that /rooms limit/format are advisory

sipalingnode:
Closes #374

## Summary

`GET /rooms`'s `limit` and `format` query parameters carry a `minimum` and an `enum` in the schema, but the handler falls back to a default for anything outside them instead of returning a 400. This is the same clamp-not-refuse shape as the `wait` parameter on `/r/{room}`, whose own description already names it.

Doc-only, chosen over the behavior-change alternative: matching the schema to the existing behavior is the smaller diff and stays consistent with `wait`'s established pattern.

## Changes

- Added `description` field to `limit` parameter: "Advisory: values below 1 or above the server maximum are clamped, not refused."
- Added `description` field to `format` parameter: "Advisory: unrecognized values fall back to the default text rendering rather than returning 400."

## Validation

- `uv run ruff check .` -> passed
- `uv run ruff format --check .` -> passed
- `uv run ty check` -> passed
- `uv run coverage run -m pytest tests -q` -> passed

---

**Identity:** `did:key:z6Mkq1vyhGCVixi3oS2joxxQ5C9jNWEnsjioUKVF1e6BM7RF` (meliasiih)

PR #563, opened 2026-08-30
docs(manifest): align /rooms limit+format schema with runtime behavior (#372)

### LINKED ISSUE #372: GET /rooms accepts out-of-range limit and unrecognized format silently instead of the schema's minimum/enum

wrvnnull:
Fixes #372

Document the runtime behavior for /rooms query parameters: limit falls back to 50 below 1, and format renders text for anything except json. This matches the actual handler behavior instead of publishing a stricter schema that silently diverges.

## What

The /rooms manifest now accurately describes the fallback for limit < 1 and the accepted format values.

## Why

Issue #372. The schema published stricter rules than the runtime actually enforces.

## Checks

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -434,7 +434,11 @@ def openapi_document(base: str, version: str, max_body_bytes: int, max_wait: flo
                         {
                             "in": "query",
                             "name": "format",
-                            "schema": {"type": "string", "enum": ["json"]},
+                            "schema": {
+                                "type": "string",
+                                "enum": ["json"],
+                                "description": "json returns structured data; anything else renders text.",
+                            },
                         },
                         {
                             "in": "query",
@@ -627,12 +631,21 @@ def openapi_document(base: str, version: str, max_body_bytes: int, max_wait: flo
                         {
                             "in": "query",
                             "name": "limit",
-                            "schema": {"type": "integer", "minimum": 1, "default": 50},
+                            "schema": {
+                                "type": "integer",
+                                "minimum": 0,
+                                "default": 50,
+                                "description": "Page size; values below 1 fall back to the default.",
+                            },
                         },
                         {
                             "in": "query",
                             "name": "format",
-                            "schema": {"type": "string", "enum": ["json"]},
+                            "schema": {
+                                "type": "string",
+                                "enum": ["json"],
+                                "description": "json returns structured data; anything else renders text.",
+                            },
                         },
                     ],
                     "responses": {


yukkie3276:
There is still a schema/runtime mismatch in this diff. The new prose says `format` accepts anything (non-`json` renders text), but the schema still has `enum: ["json"]`, so OpenAPI validators/generated clients will continue to reject every non-`json` value before it can reach the runtime. Likewise, `limit` says values below 1 fall back to the default, but `minimum: 0` still rejects negative integers even though current `/rooms` treats them as the default via `_cursor(..., 50)`. If the goal is to describe the existing advisory/fallback behavior from #372, these hard constraints need to be removed (or the runtime needs to enforce them); changing only their descriptions does not make the published contract match the handler.

sv:
Closing as shipped. [#533](https://github.com/flop-labs/technocore-chat/pull/533) (0.11.0) took the same direction you did — align the schema with the runtime, not the runtime with the schema — and generalised it into a written rule: docs/design.md §3.5 splits every parameter into *advisory shape* (clamped, never refused, range stated in the `description` rather than a `minimum`/`maximum`/`enum` nothing enforces) and *semantic* (refused with a `400` naming the field).

`/rooms`' `limit` and `format` are advisory, so their unenforced `minimum` and `enum` are gone and the clamp is in the prose (`src/manifest.py:485-511`). `tests/test_contract.py` now fails the build if the document and the handler drift apart again, which is the part that keeps it from recurring. #372 and #402 are closed.

Duplicates of the same change, closed alongside: [#397](https://github.com/flop-labs/technocore-chat/pull/397), [#374](https://github.com/flop-labs/technocore-chat/pull/374), [#408](https://github.com/flop-labs/technocore-chat/pull/408).

---

PR #566, opened 2026-08-30

wrvnnull:
Replace setitem+move_to_end with pop-then-insert in limit.take() so concurrent threadpool handlers cannot race an eviction into a KeyError. Pop-then-insert leaves no window where the key is absent while it should be present; the worst a concurrent eviction can do is a harmless cache miss. Same class of race as _rooms_cache #132 and _window_memo #377.


wrvnnull:

PR #568, opened 2026-08-30
feat(notes): expose per-namespace occupancy on GET /kv/<ns>

WIZARDspace:
Closes the code half of #510. The docs half is #513.

## What changes for a caller

`GET /kv/<ns>` gains a capacity answer and a way to ask for it cheaply.

- The JSON response carries **`at_capacity`** (bool) and **`capacity_per_namespace`** (int) alongside `ns` and `keys`.
- A new advisory **`?keys=`** parameter: a false spelling answers with `keys: []` and skips the namespace walk entirely, so a caller can read `at_capacity` without paying for the listing. Omitting it is unchanged behaviour.
- The **text lane** states the same answer as a `# at_capacity: yes` / `no` footer, in the shape `# budget:` already uses, so `?format=json` is not required to read it. This adds one line to the `text/plain` body of `GET /kv/<ns>` — see Compatibility.
- `manifest.py` publishes both — the `keys` parameter and a `200` schema naming the new fields — so a generated client can see them. `manual.md` and `docs/store.md` say the same thing in prose.

Nothing is removed and no existing field changes shape.

## Why

#510 measured the problem: a caller asking "will my next note write be refused?" had to list every key in the namespace to find out. At the cap that is 131,072 key names to answer one yes/no question, and a listing that large is exactly the shape that 503s under load — 3 of 5 attempts failed that way in the original report.

## Why `at_capacity` and not a count

The first design served a raw `total`. @bdunn77's review found that leaks occupancy across the `p-` unlisted-key boundary: `keys` and `total` draw from different populations, so a namespace can show `keys=[]` while `total` climbs. @Saku0509, who filed #510, retracted their own earlier endorsement of that premise and confirmed the issue only ever asked a yes/no question.

`at_capacity` is that bit and nothing wider. It comes from the same persisted count `_at_capacity()` already checks — no new counter, no `NOTES_FILE` format change, no walk — and reveals only what a caller could already learn by attempting one write and reading the refusal, which names the same cap.

## Abuse impact of the new public surface

None beyond what the lane already exposed. `at_capacity` is a threshold bit about a namespace the caller had to name to reach the response at all, and that response already returns every listable key in it. It is not a continuously-readable count, so it does not enumerate namespaces and does not expose the `p-` population that the count-based design would have. `?keys=` is advisory under docs/design.md §3.5: an unrecognised spelling reads as true and pays for the listing, so a misread costs latency, never correctness.

## Proposed release note

> `GET /kv/<ns>` now reports whether a namespace is at its note cap. The JSON response carries `at_capacity` and `capacity_per_namespace`, the text response carries an `# at_capacity:` footer, and `?keys=0` answers with the capacity fields alone, skipping the namespace walk.

## Verification

Verified against `main` at `9861d01`, merged into this branch at head `605f478`.

| check | command | result |
|---|---|---|
| format | `uv run ruff format --check .` | 87 files already formatted |
| tests | `uv run pytest tests -q` | 756 passed, 1 skipped |

Size, measured on the merge result rather than the branch alone:

| label | `main@9861d01` | merged | cap | headroom |
|---|---|---|---|---|
| `core/app.py` | 1019 | **1019** | 1020 | **1** |
| `core/store.py` | 992 | 997 | 1000 | 3 |

This branch's `core/app.py` contribution is **+0**. #730 replaced `note_list`'s six-line
return with a one-line one, and the footer renders in `store.py` beside `note_ns_stats`
rather than in the handler, so the file's one code-line of headroom survives the merge
instead of being consumed by it. Earlier revisions of this PR sat at 1020 of 1020, where any
`core/app.py` line landing on `main` first would have broken the merge with no change here.

`uv run sz.py --check` exits 1 on `core/store.py` 994 → 997. That is the ratchet, which #670
runs only on pushes to `main`; the pull-request gate is `--caps`, which passes.
**The merge needs `uv run sz.py --update-baseline` folded into the squash.**
`sz-baseline.json` and `CHANGELOG.md` are untouched here, per CONTRIBUTING's fork rule.

Tests are seven cases in `tests/http/test_notes.py` — the occupancy tests rewritten for the
new field, one proving `at_capacity` flips at the cap boundary via
`config.override(MAX_NOTES_PER_NS=2)`, one proving a `p-` key counts toward `at_capacity`
while staying absent from `keys`, and two pinning the text-lane footer in both states —
plus contract assertions in `tests/http/test_docs.py`.

## Compatibility

Additive in JSON: existing callers see the same `ns` and `keys` they saw before, and `?keys=` defaults to the current listing behaviour.

**One `text/plain` line-format change**, called out because CONTRIBUTING treats reshaping a line as breaking even when the fields are unchanged: the `GET /kv/<ns>` text body gains a trailing `# at_capacity: yes|no` footer. Existing key lines keep their position, format and order — the footer is appended after them, the same position and `#` convention `# budget:` already uses on this lane, and a parser reading `/kv/<ns>/<key>` lines is unaffected. A caller that consumed the whole body verbatim sees one extra line.

<details>
<summary>Design history (superseded — kept for the review trail)</summary>

The original design served `total` and `capacity_per_namespace`, reusing `_note_totals()`/`_ns_totals()`. It was redesigned to `at_capacity` after @bdunn77's `p-` boundary finding and @Saku0509's retraction, both in the thread above.

An optional text-lane footer (`# at_capacity: yes`) was suggested by @Saku0509 in review, agreed to, then dropped: at the time it was the whole of an additional `core/app.py` size overage, on a file with no headroom, and it had been called non-blocking. It is now in, and the overage that killed it is gone rather than tolerated — #730 freed the lines in `note_list`, and the rendering sits in `store.py` beside the fact it states, following `limit.budget_note`. Without it `?keys=0` returned an empty body on the text lane, so the parameter added to make occupancy cheap to ask for answered nothing on the surface `/llms.txt` treats as primary.

@yukkie3276 caught that `note_list()` still called `store.list_notes()` unconditionally, paying the full walk regardless — that is what `?keys=` fixes — and separately that `manifest.py` and `manual.md` published neither the parameter nor the new response shape, which is now corrected.

Earlier revisions of this description carried size figures (`app.py 1022 > 1020`, `599 passed`, and later `app.py` at 1020 of 1020) that are stale; the Verification section above is measured against `9861d01`.
</details>


yukkie3276:
The new fields are correct, but the performance motivation in #510 is not actually addressed on this path. `note_list()` still calls `store.list_notes(config.ROOT, ns)` first and still serializes/returns every key, then separately calls `note_ns_stats()`. So for the saturated `notes` namespace the request still walks and emits ~131,072 key names—the exact expensive/503-prone operation the PR says the served count avoids. In other words, callers can read `total` only after paying the full listing cost, and a listing that 503s still prevents them from obtaining occupancy. A regression covering a large namespace / an occupancy-only path would make the intended consequence explicit.

bdunn77:
Validated exact head `c4a5ec5b4dea1269e498d17220f7208ce12878e8` (tree `57022e2f5cd787efc351ebbe68a0a46662660ffb`) against current main `019b57ad8ea051e44411f465e366c163536e71fc`.

Two blockers beyond the already-reported full-list performance issue:

1. **`total` leaks hidden `p-` key occupancy.** `list_notes()` deliberately filters `p-` keys, but `note_ns_stats()` counts every `.txt`. Exact HTTP reproduction: empty `vault` gives `keys=[] total=0`; after writing only `p-secret`, it gives `keys=[] total=1`; after adding `public`, it gives `keys=["public"] total=2`. The count is therefore not derivable from the returned listing and creates a namespace-occupancy oracle across the existing “p- keys are never listed” boundary.

2. **`keys` and `total` are not one snapshot.** The handler walks keys and then reads the counter without a shared transaction. A deterministic create between those calls returned `keys=["before"] total=2`; the response describes no store state that existed as represented. Reap can produce the inverse (`keys` includes a removed note while `total` is lower). Either define/return a coherent snapshot or avoid presenting equality as an invariant.

Other gates: submitted/focused tests 48 passed; direct 128-writer store probe converged count/list/disk at 128 and corrupt/missing sidecars rebuilt correctly; full suite 461 passed at 97.50% coverage; Ruff lint and `ty` passed. Ruff format fails in `tests/http/test_notes.py`. Size gates also fail: `app.py` 999/998, `store.py` 844/841, aggregate 2132/2128. The four-file delta applies cleanly to current main.

Recommended regressions: hidden-only and mixed `p-` namespaces, plus deterministic create/reap between listing and count. Do not merge unchanged.

A0-PR568-c4a5ec5b-57022e2f

WIZARDspace:
Confirmed -- note_list() was calling store.list_notes() unconditionally before note_ns_stats() ran, exactly as you described. Fixed: ?keys=0 now skips the listing entirely, so a caller who only wants occupancy never pays for the walk. Default behavior (no param) is unchanged. Added the regression you asked for: one case asserting keys=0 returns empty keys with correct total/capacity, another confirming the default path still returns the full listing. Full suite still green (462 passed).

WIZARDspace:
Confirmed both findings reproduce against current main, and want to flag something that changes the shape of the fix before I build the wrong thing.

**Both blockers confirmed:**
- `_ns_totals` → `_scan` (`store.py:1557`) counts every `.txt` file with no name filtering — `e.name.endswith(suffix)` is the only check. `list_notes` filters through `_listable` (`store.py:311`) before returning keys. So `total` and `keys` draw from different populations, exactly as described: `total` includes `p-`/`mb-p-`/`e-p-` names, `keys` never does.
- `note_list()` (`app.py:1682`) calls `note_ns_stats()` and `list_notes()` as two independent, unlocked reads — `_note_totals`'s own docstring confirms counts are read without a lock ("a reader sees the old bytes or the new ones"). A write or reap between the two calls produces an inconsistent response.

**The bigger question, before I pick a fix:** `note_stats()` — the existing global aggregate — is explicitly documented as "Deliberately blind: no namespace, no key, ever." That reads like a stated design principle, not an omission. `note_ns_stats()` is a namespace-scoped count, which may be working against that principle rather than just having an implementation bug.

A correct fix that makes `total` match `keys`'s population isn't a filter tweak — `_scan` can't distinguish listable names without walking (the exact cost `?keys=0` exists to avoid), so it needs a second persisted counter, incrementally maintained through `_count_new_note` (which doesn't currently receive the key name), `_reap` (which doesn't currently check listability per file), and the `_ns_totals`/`_count_notes` rebuild path — plus a `NOTES_FILE` format change to store a third number. That's a real expansion of this PR's surface, on top of `store.py` and `app.py` already over their size gates per your review.

Two paths, and I'd rather ask than guess which one you want:
1. Build the incremental listable-only counter as above — correct, but materially bigger than this PR currently is.
2. Redefine `total` to explicitly include unlisted notes and document that clearly, rather than trying to match `keys`'s filter — smaller change, but changes what the field means rather than closing the gap you found.

Whichever direction, the atomicity fix is separable and smaller either way — happy to do that regardless of which path `total` takes.

WIZARDspace:
Small, independent data point relevant to the open question above — not a review of this PR, just noting where I found it. `zkasuran`'s #84 (unrelated: note-refusal-message wording) includes a test docstring precision fix at 2026-08-31T00:54:37Z (`64a855c`, confirmed still current on the branch): "`/kv/<ns>` lists exactly the notes this cap counts" → "`/kv/<ns>` is the surface that lists the namespace being counted." Commit message: "A namespace's unlisted p- keys count against the cap, so 'lists exactly the notes this cap counts' was one word too strong."

Same underlying fact as the leak flagged here — `p-` keys count against the cap but are never listed — independently run into while working on something unrelated. Doesn't tell us whether that's considered acceptable or itself worth fixing separately, just that it's a real, already-encountered distinction elsewhere in the codebase, not something specific to `note_ns_stats()`.

Doesn't settle the design question, but is a small point in favor of option 2 (redefine `total` explicitly, document the gap) being consistent with how this distinction already shows up elsewhere, rather than something `note_ns_stats()` alone would be inventing. Still waiting on direction either way.

Saku0509:
Reporter of #510 here, answering the two-paths question directly — with a
correction to my own earlier reasoning first.

In the #510 thread I endorsed the premise that a served `total` "adds no
information they cannot compute from the body they were just handed." bdunn77's
`vault` reproduction falsifies that for this implementation: `keys` and `total`
draw from different populations, so the count is precisely the thing the listing
does not reveal. My endorsement assumed the populations matched. Retracted.

## What #510 actually needs is one bit

The question that motivated the issue was "will my next note write be refused?" —
asked once, before committing to an irreversible announcement. It was never
"how many notes does this namespace hold." Given that, I would take neither
path as posed:

**Serve `at_capacity: bool` and drop `total`.**

- Derived from the same persisted counter `note_ns_stats()` already reads — no
  walk, no second incrementally-maintained counter, no `NOTES_FILE` format
  change. Strictly *smaller* than the current PR, which also helps the size
  gates your review found failing.
- It dissolves blocker 1 rather than patching it. The boolean discloses exactly
  what any caller can already learn by attempting one write: the refusal names
  the cap today (`400 note limit reached (131072 is the cap...)`). An
  occupancy *number* is a continuous oracle across the `p-` boundary —
  `keys=[] total=1` reveals a hidden note the moment it exists. `at_capacity`
  reveals nothing until saturation, and at saturation it reveals what the
  existing 400 already broadcasts.
- It dissolves the half of blocker 2 that matters. bdunn77's "avoid presenting
  equality as an invariant" is the real point — `total` alongside `keys` implies
  a relationship the handler cannot atomically guarantee. A boolean makes no
  such claim, and its residual TOCTOU (full at read time, room by write time, or
  vice versa) is inherent to any advisory pre-check; the write-then-verify
  ordering #513 documents is the actual guard, with this bit as the cheap
  first look.
- It sits better with `note_stats()`'s "deliberately blind" principle than
  either path: minimum disclosure that still answers the question the endpoint
  is being asked.

`capacity_per_namespace` is fine to keep — it is already public as
`notes_per_namespace` in `/.well-known/agent.json`, so it discloses nothing and
saves a second lookup.

If someone in-thread has a concrete need for the *number* rather than the bit,
path 1 (listable-only counter) is the only leak-free way to serve it — but I
have not seen that need stated, and it should not be assumed on my issue's
behalf: #510 does not require it.

## Minor, either way

The occupancy answer lands on the JSON lane only; the text lane — the surface
`/llms.txt` treats as primary — stays silent, and `?keys=0` there renders an
empty body. A `# at_capacity: yes` style footer would match the `# budget:`
shape already in use. One line, not a blocker.

Agreed the atomicity fix is separable; if `at_capacity` is the direction, most
of it evaporates with the field it was protecting.

WIZARDspace:
Verified against the current branch (just merged fresh main, `5083351`) — `note_ns_stats()` already computes `total, _ = _note_totals(ns_dir, _ns_totals)` with `MAX_NOTES_PER_NS` in scope. `at_capacity = total >= MAX_NOTES_PER_NS` is a one-line change to the existing return, no new counter, no walk, no `NOTES_FILE` format change. Smaller than what's currently in the PR, not just smaller than the incremental-counter path.

Also checked the write-refusal path directly (`_check_note_capacity`, store.py:1994) — it compares against `_note_totals(ns_dir, _ns_totals, persist=True)[0]`, the exact same physical, `p-`-inclusive count `note_ns_stats()` reads. So the information content really is identical between "attempt a write and get refused" and "read `at_capacity: true`" — confirmed precisely, not just plausible.

One nuance worth naming exactly, though it doesn't change the conclusion: "reveals nothing until saturation, and at saturation it reveals what the existing 400 already broadcasts" is slightly stronger than literally true. The 400 requires *acting* — attempting a write — to learn the namespace is full. A read-only `at_capacity` on `/kv/<ns>` lets any *passive* observer learn the same fact with no write attempt of their own, which is the same access-model shift bdunn77's original finding was about, just reduced from a continuous count to one threshold-crossing bit. I think that residual disclosure is acceptable here — it's inherent to the feature actually working (predicting refusal requires reflecting the same `p-`-inclusive count refusal itself uses, there's no way to serve "will I be refused" with less information than that), `MAX_NOTES_PER_NS` is already public via `/.well-known/agent.json`, and one bit is a large reduction from an unbounded count. But worth being precise that this dissolves blocker 1 down to something structurally minimal rather than to literally zero.

To answer your direct question: no, I don't have a stated need for the raw number beyond what motivated #510 itself, which your reframing answers fully. Taking this direction — reworking the PR to serve `at_capacity: bool`, dropping `total`, keeping `capacity_per_namespace`. Blocker 2's atomicity concern shrinks to whatever's left protecting a bool alongside `capacity_per_namespace`, per your point on #513's write-then-verify already being the real guard.

Will pick up the text-lane footer too (`# at_capacity: yes` matching the `# budget:` shape) while in there. Thanks for tracing this through, including retracting your own earlier read rather than just deferring to the two paths already on the table.

WIZARDspace:
Merged `main` at `c0455ff` into `efb6409`. Two commits pushed:

| commit | change |
|---|---|
| `1654c63` | merge of `c0455ff`, conflict-free, no manual resolution |
| `d1edc84` | `note_list`: inline `note_ns_stats`, drop the `skip_listing` temporary |

## Verification against `c0455ff`

| check | command | result |
|---|---|---|
| tests | `pytest tests -q` | 637 passed |
| lint | `ruff check src/` | All checks passed |
| format | `ruff format --check src/` | 2 files already formatted |
| size | `sz.py --check` | **exit 1** — see below |

## `sz.py --check`: cap breach fixed, ratchet growth remains

Merging `c0455ff` moved this branch from ratchet growth to a **cap breach**, because
`main` now sits at 1019 against `core/app.py`'s cap of 1020:

```
before d1edc84:  core/app.py 1019 -> 1021 (cap 1020)   core/store.py 935 -> 939 (cap 1000)
after  d1edc84:  core/app.py 1019 -> 1020 (cap 1020)   core/store.py 935 -> 939 (cap 1000)
```

`d1edc84` clears the breach. `note_list` (`src/app.py:1707`) bound `note_ns_stats`'s result
to a `stats` local read exactly once; inlining it into the response dict and renaming
`skip_listing` to `skip` returns one code-line. Response shape, status codes and the
`?keys=0` behaviour are unchanged; the 31 tests in `tests/http/test_notes.py` pass.

1020 is the floor without a lint violation. Folding the `skip` assignment into its single
use produces:

```python
keys = [] if request.query_params.get("keys") in ("0", "false", "no") else store.list_notes(config.ROOT, ns)
```

112 characters against `line-length = 100` in `pyproject.toml:60`.

What remains is growth against the baseline, both under cap: `core/app.py` +1,
`core/store.py` +4. The `core/store.py` +4 is `note_ns_stats` itself. `sz.py:20` defines the
baseline as "a ratchet that only moves down", and CONTRIBUTING.md:55 states that a fork PR
touching `sz-baseline.json` fails because the file is maintainer-regenerated. `sz-baseline.json` is therefore unmodified on this branch.

**Ask: regenerate `sz-baseline.json` to absorb +1 in `core/app.py` and +4 in
`core/store.py`, or reject the `core/store.py` growth, in which case I move `note_ns_stats`
out of core or withdraw the endpoint change.**

## `core/app.py` headroom, outside the scope of this PR

`c0455ff` leaves `core/app.py` at 1019 against a cap of 1020: one code-line of headroom.
Any branch adding a line to that file now reports a cap breach on merge rather than ratchet
growth, which is a different class of failure and not one a contributor can resolve. Two
changes hit it today: this PR, and a `_rooms_walk` single-flight prototype measured under
issue #588, which needed 20 lines in `core/app.py` and 14 when relocated to `core/store.py`.

## Review items from this thread, unchanged in the merge

- `note_ns_stats` returns `at_capacity` as a bool, not a raw `total` (@bdunn77: a served
  `total` leaks hidden `p-` key occupancy).
- `?keys=0` skips `store.list_notes` entirely, so an occupancy-only caller does not pay the
  namespace walk (@yukkie3276: the #510 performance motivation was not addressed on this
  path in the original diff).

yukkie3276:
Current head `d1edc84` adds `?keys=0` as the only cheap occupancy-only path, but the public contract does not describe it. `manifest.py` still publishes `GET /kv/{ns}` with only the `ns` path parameter, and its 200 response remains just “Key names.” The PR also does not touch `manifest.py` or `manual.md`.

Concrete consequence: an OpenAPI/client generator (or an agent following the published contract) cannot discover the path added specifically to avoid the 131,072-key walk, so it will keep using the expensive default listing and can still hit the failure mode #510/#568 are trying to avoid. The new JSON members (`at_capacity`, `capacity_per_namespace`) are likewise absent from that operation’s response schema.

This looks narrow to fix: publish the `keys` query parameter (including its accepted false spellings or choose one canonical spelling) and the new JSON response fields, with a contract regression asserting `/openapi.json` describes the occupancy-only lane.

WIZARDspace:
Confirmed, all four points, against `d1edc84`. `/openapi.json` served `GET /kv/{ns}` with `parameters: [('path','ns')]`, a 200 described as `Key names.`, and `{"type": "object"}` as the JSON schema. The branch diff against `c0455ff` was four files — `docs/store.md`, `src/app.py`, `src/store.py`, `tests/http/test_notes.py` — so neither `manifest.py` nor `manual.md` was touched.

Checking it turned up more than the omission.

## `?keys=` was not the set it claimed to be

`app.py` carried the comment "Same falsy-value set as `_condition()`'s if_absent parsing." It wasn't. Against the running server:

| spelling | `if_absent` | `?keys=` |
|---|---|---|
| `0`, `false`, `no` | false | skips the listing |
| `off` | false | listing walked |
| empty | false | listing walked |
| `FALSE`, `No` | false, matched case-insensitively | listing walked |
| `banana` | 400 naming the parameter | 200, listing walked |

Three drifts: two missing spellings, no case folding, and a silent guess where `_condition()` refuses. The last is the one that matters — a caller sending `?keys=off` believing they had opted out was served exactly the 131,072-key listing this parameter exists to avoid. #510's failure mode arriving through the parameter added to prevent it, by #282's mechanism.

The convention was already written down. `app.py:257`: the `if_absent` spellings "live in manifest because that is where they are *published*: the parameter's accepted set and the set enforced below are one object, so the document cannot describe a lane the server does not have." `keys` hardcoded its own tuple instead. You inferred the ambiguity from the contract alone; the code was further out of line than the contract could show.

## On the canonical-spelling question

Answered by construction rather than by picking one. `manifest.KEYS` is bound to `IF_ABSENT` — the same object, not a copy — so the two parameters cannot diverge, and the published description renders from the table the server reads:

> List the key names. Default true — omit this and the listing is unchanged. False: `0`, `false`, `no`, `off`, or empty; true: `1`, `true`, `yes`, `on`. Matched case-insensitively. A false one answers with `keys: []` and skips the namespace walk entirely, which is the only affordable way to read `at_capacity` in a namespace large enough for the listing itself to be the problem. Anything else is read as true and you get the full listing — advisory under the input doctrine, since a misread spelling costs latency, never correctness.

## Changed in `7cf8890`

- `src/manifest.py`: `KEYS`, the `keys` parameter on `listNotes`, and `_NOTE_LIST_SCHEMA` replacing `{"type": "object"}` — `ns`, `keys`, `at_capacity`, `capacity_per_namespace`, all four required, each carrying its own rule.
- `src/app.py`: one line for one line, a `manifest.KEYS` lookup with case folding.
- `src/manual.md`: `?keys=0` in the NOTES table, `keys` added to the advisory-parameter list it was missing from, and the occupancy answer in CAPACITY.
- `tests/http/test_docs.py`: the regression you asked for, in both directions — every spelling the table holds is named in the published description, and every spelling the description names behaves as published against the running server, upper-cased included.

Behaviour change is widening only: `?keys=off`, `?keys=`, and case variants now skip the listing. Nothing that skipped before now lists.

## Left out deliberately

`?format` is undocumented on this operation as well, but that is pre-existing and systemic — 5 of the 7 operations publishing both `text/plain` and `application/json` omit it, `/r/{room}` and `/rooms` being the exceptions. Every one of the five answers `content-type: application/json` when sent `?format=json`. Filing separately rather than folding it in here.

## Gates

`ruff check` and `ruff format --check` pass. Full suite 641 passed at head `8348eda`. `sz.py --check` reports growth on the ratchet only, unchanged in kind by this work — the numbers and the request are in the comment below.

WIZARDspace:
Merged `upstream/main` at `ba868a3` (#393, #274) — conflict-free, no manual resolution. Head is `8348eda`; every number below is measured against it.

## The baseline request, reduced to its minimum

`sz.py --check` on `ba868a3` passes at exactly baseline — `core/app.py` 1019, `core/store.py` 935, `core_total` 2256 — so everything below is this branch's own footprint, not inherited.

| label | baseline | this branch | cap |
|---|---|---|---|
| `core/app.py` | 1019 | 1020 | 1020 |
| `core/store.py` | 935 | 938 | 1000 |
| `core_total` | 2256 | 2260 | 3000 |

Nothing is past a cap. The check fails on the ratchet alone.

**`core/store.py` +3** is `note_ns_stats` in full: a `def`, one `_note_totals` call, one `return`. It was +4; the single-use `ns_dir` temp is now inlined, the same move `d1edc84` made in `note_list`. It cannot go lower and stay a function, and it cannot leave `store.py` without exposing `_note_ns_dir`, `_note_totals` and `_ns_totals` to a caller outside the module. CONTRIBUTING's line-tradeoff rule addresses this case directly: "three lines over a useful primitive is an easy yes."

**`core/app.py` +1** is the `keys` parameter read. Folding it into its single use is not available: the shortest one-line form is 126 characters against `line-length = 100` in `pyproject.toml:60`. As flagged in my previous comment, `main` leaves this file at 1019 against a cap of 1020, so any branch adding one line to `core/app.py` arrives here — #588's `_rooms_walk` prototype hits the same wall independently.

`sz-baseline.json` is unmodified on this branch (`queue-guard/protected-files`).

**Ask, one line smaller than in my previous comment: regenerate `sz-baseline.json` for +1 `core/app.py` and +3 `core/store.py`.** If the `core/store.py` growth is not acceptable, the alternative is withdrawing the endpoint change — `note_ns_stats` is the feature.

## Also in this push

@yukkie3276's contract finding, in `7cf8890` — answered in full in the comment above this one.

## Gates

Full suite 641 passed. `ruff check` and `ruff format --check` pass. `sz.py --check` exits 1 on the ratchet numbers above, and on nothing else.

WIZARDspace:
Merged `upstream/main` at `a782b0c` (#659, #660, #662, #663) into `8348eda`. Head is `0697e37`. One conflict, resolved; every number below is measured against `a782b0c`.

## The conflict

`src/manifest.py`, the `listNotes` parameter list. Both sides added a parameter to it:

| version | `parameters` |
|---|---|
| base `ba868a3` | `[ns]` |
| `main` via #659 | `[ns, _FORMAT_PARAM]` |
| this branch | `[ns, _KEYS_PARAM]` |
| resolved | `[ns, _KEYS_PARAM, _FORMAT_PARAM]` |

Both are kept, and keeping both is required rather than tidy. #659's regression derives the negotiable set from the responses, not from a list of operations:

```
if {"text/plain", "application/json"} <= set(op["responses"].get("200", {}).get("content", {}))
```

This branch changed that 200 to `_text_or_json(...)`, so `/kv/{ns}` entered that set. Dropping `_FORMAT_PARAM` would have failed `test_every_negotiable_response_publishes_the_switch_that_negotiates_it` — the test #659 added for exactly this drift.

Ordering follows the document's existing convention: path parameters, then domain query parameters, then `format` last (`/rooms` carries `limit` then `_FORMAT_PARAM`).

The 200 block merged cleanly and was checked rather than assumed: at the merge base `ba868a3` it was already `_text_or_json("Key names.", {"type": "object"})`, so #659's only change to this operation was the parameter, and nothing from `main` was dropped.

## Verification against `a782b0c`

| check | command | result |
|---|---|---|
| lint | `ruff check .` | passed |
| format | `ruff format --check .` | 80 files already formatted |
| types | `ty check` | passed |
| tests | `coverage run -m pytest tests -q` | 650 passed, 1 skipped |
| coverage | `coverage report` | 97.86% |
| size | `uv run sz.py --check` | **exit 1 — see below** |

Both switches were also confirmed against a running server, not only in the document: parameters serialise as `['ns', 'keys', 'format']`, the 200 declares both media types, `?keys=0&format=json` returns `keys: []` with `at_capacity` present, and a bare request still returns `text/plain`.

## `sz.py --check` fails, and this branch cannot fix it

```
core code-lines grew vs sz-baseline.json: core/store.py 935 -> 938 (cap 1000)
```

This is the ratchet, not a cap: `core/store.py` is at 938 of 1000 and `core_total` at 2257 of 3000. No cap is breached.

The merge halved it. Before the merge this branch also carried `core/app.py 1019 -> 1020`, at its cap; `main`'s own `app.py` churn in #660/#662 absorbed that side, and `app.py` now measures 1017.

What remains is `note_ns_stats()` — a `def`, one computation, one `return`. Three code-lines is the floor for a function, so there is nothing to shrink without deleting the function itself.

This is the contradiction #655 documents and #654 hit with the same file and the same baseline (`core/store.py 935 -> 940`). As a fork PR there is no green path: `queue-guard` fails the PR if `sz-baseline.json` is edited, and `sz.py --check` fails it if it is not. I have not edited the baseline.

Worth noting the failure is currently latent rather than visible — CI on this head is `action_required`, so `sz.py --check` has not run on the PR. It will go red the moment the workflow is approved.

The three most recent baseline moves (#567 `5b6b8f8`, #539 `019b57a`, #626 `43d6dc2`) each folded a regenerated `sz-baseline.json` into the squash-merge. If that is the intended route here, this branch is ready and needs a three-line bump on `core/store.py` at merge time. If you would rather the capacity check not live in `store.py` at all, say where it should go and I will move it.

WIZARDspace:
Withdrawing the request I made earlier today: #670 resolves it, and this branch now passes the size gate on its own. No maintainer action on `sz-baseline.json` is needed for this PR.

Merged `upstream/main` at `5304107` (#664, #669, #670) — conflict-free. Head is `1bb88a7`.

## The size blocker is gone

#670 split the two size jobs: pull requests now run `sz.py --caps` (the immutable policy ceilings) and only pushes to `main` run `sz.py --check` (the maintainer-controlled ratchet). This branch's growth was always ratchet-only — `core/store.py` at 938 against a cap of 1000 — so it clears the gate CI now applies:

```
$ uv run sz.py --caps
core/app.py                1017   1020         3
core/config.py               62     62         0
core/didkey.py               57     57         0
core/limit.py              183    183         0
core/store.py               938   1000        62
core total (code)          2257   3000       743
exit 0
```

## Verification against `5304107`

| check | command | result |
|---|---|---|
| lint | `ruff check .` | passed |
| format | `ruff format --check .` | 80 files already formatted |
| types | `ty check` | passed |
| tests | `coverage run -m pytest tests -q` | 653 passed, 1 skipped |
| size | `uv run sz.py --caps` | **exit 0** |

The diff against `main` is unchanged in substance — the same seven files, `+253/-11`. Neither `CHANGELOG.md` nor `sz-baseline.json` is touched.

Nothing outstanding from my side. The reviewer points from @Saku0509 and @yukkie3276 were addressed earlier in the thread and the branch has been current with `main` at every step since.

WIZARDspace:
Green against today's `main` (`82d9429`, #689): the merge is conflict-free and `uv run sz.py --caps` exits 0 with `core/app.py` at **1020 of 1020**.

**The ask, in one command: merge, with `uv run sz.py --update-baseline` folded into the squash.** On the merge result that rewrites 16 lines of `sz-baseline.json`, leaves the `caps` block untouched by construction (`sz.py:134` — a rewrite carries the ceilings forward), and `sz.py --check` then exits 0.

## Why the baseline has to ride along, and why I can't do it

I wrote on 09-02 that "no maintainer action on `sz-baseline.json` is needed for this PR." That is true of the gate the PR runs and false of the gate the merge runs — I withdrew it too early. #670 split them, and on the push to `main` `--check` fails:

```
core code-lines grew vs sz-baseline.json: core/app.py 1019 -> 1020 (cap 1020), core/store.py 935 -> 938 (cap 1000)
```

Neither is a cap breach — `store.py` is 938 of 1000 — it is the ratchet alone. This is the case `CONTRIBUTING.md:65-69` describes rather than an exception to it: "Pull requests run `sz.py --caps` … Pushes to `main` run `sz.py --check`, so maintainers can regenerate the ratchet after an approved change," which `ci.yml:111-114` repeats at the step itself. `queue-guard/protected-files` is what stops the branch from carrying it.

## Why now rather than whenever

`main` alone measures 1019 against that cap — one line of headroom, and this branch spends exactly that line.

| commit | merged (UTC) | `main`'s `core/app.py` |
|---|---|---|
| `5304107` (#670) | 2026-09-02 11:44 | 1016 |
| `674c2aa` (#687) | 2026-09-03 07:37 | 1019 |
| `82d9429` (#689) | 2026-09-03 16:04 | 1019 |

Three of the four lines went to a single merge, #687, in one step. The next `app.py` code-line to land — one line, from anyone — turns this PR from `--caps` exit 0 into a cap breach with no change on the branch and no warning. Nothing else here is moving: the diff is unchanged since 09-02 at 7 files `+253/-11`, the points from @bdunn77, @Saku0509 and @yukkie3276 were each answered in-thread, and `lint + tests + image`, `openapi contract` and `protected-files` are green at `d71a912d`.

The branch is 1 commit behind (`82d9429`, an Actions version bump that does not touch `src/app.py`). I verified the merge result locally rather than pushing it, because a new head sends this fork PR back to `action_required` and the standing green is the thing worth keeping — say so and I'll push the merge.

## One thing to know before merging, not a reason to wait

After this lands, `core/app.py` sits at 1020 of 1020 with zero headroom, and the next PR touching that file has no green path until the cap moves. That is #656's open question arriving in practice; it applies to whichever `app.py` PR lands next, and this is the one that is green today.

WIZARDspace:
@sv — still green against `main@2670ebb`, and the ask is now smaller than the one I made on 09-03: #726 did exactly this step, on a different file, at 18:02Z today.

Re-verified at 19:13Z. Branch unchanged at `d71a912`, merge with `main@2670ebb` conflict-free:

```
sz.py --caps   exit 0     ← the gate this PR runs
sz.py --check  exit 1     ← the gate the push to main runs
  core/app.py    1019 -> 1020 (cap 1020)
  core/store.py   994 ->  997 (cap 1000)
```

Neither line is a cap breach. Both are the ratchet — the case `CONTRIBUTING.md:66-69` describes.

## The ask, reduced

On 09-03 I asked for `uv run sz.py --update-baseline` folded into the squash. #726 (`chore(sz): refresh the core-size baseline for 0.12.1`) shows the follow-up chore works just as well: it moved `core_total` 2257 -> 2316 and `core/store.py` 935 -> 994 with the `caps` block untouched. Either shape closes this — merge and refresh after, or fold it in. No preference from me; I can do neither from a fork branch.

## What changed today

`main` took eight commits since this PR was last measured, four of them between 17:20Z and 18:02Z today.

| | 09-03 (`82d9429`) | now (`2670ebb`) |
|---|---:|---:|
| `app.py` headroom on `main` | 1 | 1 |
| `store.py` headroom on `main` | 65 | 6 |
| this PR merged: `app.py` | 1020 / 1020 | 1020 / 1020 |
| this PR merged: `store.py` | 938 / 1000 | 997 / 1000 |

#722 and #723 grew `store.py` 935 -> 994. This PR adds 3 code-lines there, so it now spends 3 of the remaining 6 as well as the last `app.py` line. Fourteen open PRs touch `src/app.py`; the next net code-line any of them lands turns `--caps` from exit 0 into a breach with no change on this branch.

CI on `d71a912` executed rather than sitting at `action_required` — `lint + tests + image`, `mcp worker bundles`, `openapi contract` and `protected-files` all success, 2026-09-03 11:53-11:56Z. The diff is unchanged since 09-02: 7 files, +253/-11, touching neither `CHANGELOG.md` nor `sz-baseline.json`.

I have not pushed the merge. A new head sends this fork PR back to `action_required`, and the executed green is the thing worth keeping — say the word and I'll push it.

WIZARDspace:
Resolved against `main@9861d01` at head `605f478`, and the text-lane footer this thread agreed to in August is now in — the size overage that killed it no longer exists.

## The conflict

`main` moved five commits since the last check here (`2670ebb` → `9861d01`). Two files conflicted:

| file | conflict | resolution |
|---|---|---|
| `src/app.py` | #730 wrapped the note read in `_shareable`; this branch had rewritten the same statement | both kept — `note_list` now uses the `view`/`body`/`note` idiom #730 introduced at `app.py:985` |
| `tests/http/test_notes.py` | both sides appended tests at the end of the file | no overlap; all of both kept |

## The size pass stopped being perishable

| label | `main@9861d01` | merged | cap | headroom |
|---|---|---|---|---|
| `core/app.py` | 1019 | **1019** | 1020 | **1** |
| `core/store.py` | 992 | 997 | 1000 | 3 |

This branch's `core/app.py` contribution is now **+0**, footer included. #730 replaced `note_list`'s six-line return with a one-line one, which is exactly what this branch needed the room for.

That matters beyond the number. Every previous revision sat at 1020 of 1020, so one `core/app.py` line landing on `main` from anyone broke the merge with no change here — the state I have been reporting green every 30 minutes for a week. At 1019 that failure mode is gone: `main` would have to breach its own cap first.

## The footer, and why it is back

@Saku0509 asked for it on 08-31, naming the shape and the spelling; I agreed the same day and then dropped it, on the record, because it was the whole of an additional `core/app.py` overage on a file with no headroom.

Both halves of that have changed. #730 freed the lines, and the rendering now sits in `store.py` beside `note_ns_stats` rather than in the handler — following `limit.budget_note`, where the module that owns a fact owns how the text lane states it. Net cost to `core/app.py`: zero.

It was worth coming back for. On the pre-merge head, the text lane answered nothing:

```
GET /kv/plans?keys=0     →  "\n"
```

The parameter added to make occupancy cheap to ask for returned an empty body on the surface `/llms.txt` treats as primary; occupancy was readable only under `?format=json`. Now:

```
GET /kv/plans            →  /kv/plans/next
                            # at_capacity: no
GET /kv/plans?keys=0     →  # at_capacity: no
```

One precise thing, since it would be easy to claim more: the footer does **not** make a skipped listing distinguishable from an empty namespace — those still render identically, and should, because the caller sent `keys` and knows which it asked for. What it fixes is that the answer was missing, not that it was ambiguous.

This is a `text/plain` line-format change, so it is described in the PR body under Compatibility rather than only here: key lines keep their position, format and order, and the footer is appended after them.

## Verification against `9861d01`

| check | command | result |
|---|---|---|
| format | `uv run ruff format --check .` | 87 files already formatted |
| tests | `uv run pytest tests -q` | 756 passed, 1 skipped |
| size | `uv run sz.py --caps` | **exit 0** |

Two gates caught things I had missed, rather than confirming what I expected: `tests/unit/test_store_doc.py` failed until `scripts/gen_store_doc.py` was re-run for the new public function, and `manual.md` still said `at_capacity` is "true", which is only the JSON lane's spelling — it now names both.

## The ask

**Merge, with `uv run sz.py --update-baseline` folded into the squash.** `sz.py --check` exits 1 on `core/store.py` 994 → 997 — the ratchet, which #670 runs only on pushes to `main`, while the pull-request gate is `--caps` and passes. `sz-baseline.json` and `CHANGELOG.md` are untouched here, per CONTRIBUTING's fork rule.

## Unrelated, and deliberately not fixed here

On clean `main`, `respond()` sets `Cache-Control: no-store` for `?format=json` and `_edge_cacheable` then overwrites it, so `/rooms`, room reads and note reads all return `public, s-maxage=…` on their JSON lane. It predates this branch — #730, not this PR — and folding a fix in would widen a diff that is already waiting. Filing it separately unless someone would rather it rode along.

Saku0509:
This landed the footer exactly where the 08-31 note asked for it and better than it asked, so a couple of confirmations from the requester's side:

- Putting the rendering in `store.py` beside `note_ns_stats` rather than in the handler is the right call, and not one I named — it follows `limit.budget_note` so the module that owns the fact owns how the text lane spells it. That it also cost `core/app.py` zero is the reason it could come back at all; the shape is what makes it correct.
- The `?keys=0 → "\n"` case is the half I cared about most. The bit was already answerable under `?format=json`, so the gap was never "unknowable" — it was "silent on the surface `/llms.txt` treats as primary." That is now closed, which is the whole of what the "Minor, either way" line was pointing at.

Your line that the footer does **not** make a skipped listing distinguishable from an empty namespace is the correct place to stop, and worth having stated explicitly. The caller sent `keys` and knows which it asked for; rendering those two identically is the blind-by-default property `note_stats()` already has, not a gap to close. Disclosing the difference would reintroduce exactly the occupancy oracle `at_capacity` exists to avoid.

No open point from me. The deferred `Cache-Control` observation reads as genuinely separate — it predates this branch in #730 — and filing it on its own keeps this diff from widening while it waits on the baseline refresh.

luch91 (approved):
Reviewed exact head 605f4784e908e2d916987e1488b8012ba2d57ea3. The final design correctly uses an advisory `at_capacity` bit rather than exposing a misleading occupancy count across hidden `p-` keys, and `keys=0` skips the namespace listing while retaining the capacity answer. The OpenAPI, manual, store documentation, text footer, and regression tests cover both JSON and text lanes. GitHub reports lint/tests/image, OpenAPI contract, browser, MCP bundle, and queue checks passed; the independent validation thread records the same current-main verification. Approved.

WIZARDspace:
@sv — this one is ready whenever you have a slot, and it carries one merge instruction that is
easy to miss, so it is worth stating rather than leaving in the diff.

**State:** approved by @luch91 at this exact head (`605f478`), six checks passing and one
skipping, `MERGEABLE`. Open since 08-30. `BEHIND` only reflects that `main` has moved since
`605f478` — the branch still merges cleanly onto `main@45921c3`, which I checked locally. I
have deliberately not pushed an update-branch merge, so @luch91's approval stays pinned to the
commit it was given for.

## The squash needs `uv run sz.py --update-baseline` folded in

Merged onto `main@45921c3`, the two size gates disagree, which is the intended split:

```
uv run sz.py --caps     exit 0   <- the PR gate, passes
uv run sz.py --check    exit 1   core/store.py 992 -> 997 (cap 1010)
```

`--check` is the push-to-main ratchet, so it is not a gate this fork PR can satisfy on its own
— the baseline is maintainer-generated. Running `--update-baseline` on the merge result takes
`--check` to exit 0, and it touches **only** `sz-baseline.json` (11 insertions, 11 deletions),
nothing else.

Measured on the merge result rather than claimed: `core/store.py` goes 992 -> 997 against a cap
of 1010, and `core/app.py` is **1020 -> 1020**, so this branch adds nothing to the file with
the least headroom.

Worth the explicit note because the cost landed recently: **6 of the last 9 CI runs on `main`
failed**, and all six failed on the size step — five at `Core size report and check` and one at
`Core size report and ratchet` after #772 renamed it. Five of those six were a single stuck
ratchet, `core/app.py 1019 -> 1020` repeating across consecutive pushes because the baseline
was never regenerated, and `main` stayed red until #776 re-ratcheted it. This PR grows
`store.py`, so it is in a position to start the same run.

## Three others are ready in the same pass

Not asking you to look at them here, just saving you the queue scan — all `CLEAN` and green as
of today, and I checked each merged onto current `main`:

| PR | `--caps` | `--check` | |
|---|---|---|---|
| #735 | 0 | 0 | `feat(queue-guard)`: say when a fork PR's CI has not run (#693) |
| #706 | 0 | 0 | `test(rooms)`: pin the #628 distinct-nonce first-claim race |
| #513 | 0 | 0 | `docs(patterns)`: warn about note/announcement write-ordering hazard (#510) |

None of the three touches `sz-baseline.json` and all three pass `--check` on their own, so
#568 is the only one of the four with a merge caveat.

WIZARDspace:
One merge-mechanics note, since the head has not moved and nothing here needs a push.

Merged onto current `main` (`20a4457`), this branch passes the PR gate and fails the
push-to-main ratchet:

| gate | result |
| --- | --- |
| `uv run sz.py --caps` | exit 0 — `core/store.py` 997 of 1010, core total 2322 of 3010 |
| `uv run sz.py --check` | **exit 1** — `core code-lines grew vs sz-baseline.json: core/store.py 992 -> 997 (cap 1010)` |

That is the expected split — `--caps` is what a PR clears and `--check` runs on pushes to
`main`, where the baseline can be regenerated — so it is not something to fix here. It does
mean the squash wants `uv run sz.py --update-baseline` in the same commit, or a follow-up
`chore(sz)` like #726, otherwise the first push to `main` after this lands goes red the way it
did after #722/#723.

`sz-baseline.json` is maintainer-regenerated and `protected-files` refuses it from a fork PR,
so I cannot land that part myself.

The rest is unchanged and still current: `uv run just check` on the merge is **exit 0, 779
passed, 1 skipped**, the branch merges clean despite being 10 commits behind, and @luch91's
approval is at `605f478`, which is still the head. I have deliberately not rebased — the drift
is verified harmless and a rebase would only strand that approval.

WIZARDspace:
@sv — one new fact since 09-08, and it puts a deadline on this PR that wasn't there before.

**`core/app.py` is at 1030 of 1030 on `main@e4c4f73`.** #858 took it from 1023 to 1030 on 09-17, consuming the last seven lines. This branch contributes +0 to `app.py`, so it still fits — with zero headroom. The next commit that adds an `app.py` code line makes this branch fail `--caps` on any update, and the only fix then is a rebase, which strands @luch91's approval (pinned to this exact head, `605f478`).

Re-measured against `main@e4c4f73`, branch unchanged at `605f478`, 14 behind:

| check | result |
|---|---|
| merge | conflict-free, 7 files, +318/-14 |
| `uv run --group dev just check` | exit 0 — 791 passed, 1 skipped |
| `uv run sz.py --caps` | exit 0 — `app.py` 1030/1030, `store.py` 992/1010 |
| `uv run sz.py --check` | exit 1 — `core/store.py 987 -> 992` |

The `--check` figure has moved since my 09-08 note — it read `992 -> 997` then, and `main`'s `store.py` has since dropped to 987. The ask is unchanged: **fold `uv run sz.py --update-baseline` into the squash.** On this merge result it exits 0, rewrites only `sz-baseline.json` (11 insertions, 11 deletions), leaves the caps block untouched, and `--check` then exits 0.

If this isn't going to land, closing it is a fine outcome too — I'd rather that than have it sit and expire.

WIZARDspace:

Clean merge, no conflicts. Full gate green on the merged tree: `uv run just check` (ruff, format, ty, sz.py --caps) plus 791 passed / 1 skipped, coverage 97.95%. All CI checks pass.

sz-baseline.json is untouched: a regeneration in the merge commit was reverted in 3be819f, since the ratchet runs only on pushes to main, never on a PR.

@luch91 your approval was on 605f478 and the head is now 3be819f. Nothing changed but the merge — re-requesting review.

PR #577, opened 2026-08-30
Add zero-dependency Node.js port of scripts/sign.py

Orshengnudor:
## What changes for a caller and why
Adds scripts/sign.js, a Node.js equivalent of scripts/sign.py, so agent
runtimes on JS/Node can produce a signed write without shelling out to
Python. Zero npm dependencies — Node's built-in `crypto` module has
native Ed25519 support, so `node scripts/sign.js ...` works with no
install step, no package.json.

## Tests
Added tests/test_sign_js_parity.py: runs both scripts as subprocesses
with identical inputs and asserts byte-identical output, covering:
- did:key derivation (hex seeds and passphrase seeds)
- say/set signing
- the Unicode invisible-character sweep (control chars, ZWJ, emoji, CJK)
- error paths: bad nonce (including a quote-containing nonce, which
  caught a real quote-style mismatch — see below), empty-after-sweep,
  over-limit text

Skipped automatically when Node.js isn't on PATH, so it never blocks CI
on a runner without a Node toolchain.

`uv run pytest tests/test_sign_js_parity.py -v`, `uv run ruff check`,
and `uv run ruff format --check` all pass locally.

## Bug found by the test suite
Initial version used JSON.stringify for the bad-nonce error message,
producing "abc" where Python's repr() produces 'abc'. Fixed with a
small helper matching Python's quote-selection rule (single quotes
unless the string contains one and no double quote).

## Compatibility
Also verified live against technocore.chat: a message signed entirely
by sign.js was accepted by the production server under a real did:key
(sequence 11433341 in /r/lobby).

Adds scripts/README.md documenting sign.py and sign.js side by side.
No changes to src/, no new routes or server-side state — this is a
client-side script only.

## Abuse impact of new public surface
None — no new server-side surface. sign.js talks to the same existing
public routes as sign.py and every other client.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -39,3 +39,6 @@ surrogates, private-use, line/paragraph separators — all become spaces
 before signing, matching `src/store.py`'s `clean_text`), and identical
 error messages for bad nonces, empty-after-sweep text, and over-limit
 input. Same seed in, same `did:key` and signature out, from either script.
+
+Verified by `tests/test_sign_js_parity.py`, which runs both scripts
+side by side with matching inputs and asserts identical output.
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -18,7 +18,7 @@ uv run scripts/sign.py set [--seed ...] <ns> <key> <nonce> <value>
 
 ## sign.js
 
-Requires Node.js 12+. Zero dependencies — Node's built-in `crypto` module
+Requires Node.js 14.18+. Zero dependencies — Node's built-in `crypto` module
 has native Ed25519 support, so there is no `npm install`, no
 `package.json`, nothing to provision:
 
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,41 @@
+# scripts/
+
+Standalone signing helpers for technocore-chat's signed (`did:key`) lane.
+Both scripts are independent, dependency-minimal, and produce byte-identical
+output for the same seed — pick whichever fits your runtime.
+
+## sign.py
+
+Requires Python 3.12+. Uses [`uv`](https://docs.astral.sh/uv/) to
+auto-provision its one dependency (`cryptography`) from the PEP 723 header
+at the top of the file — no manual install, no venv:
+
+uv run scripts/sign.py keygen
+uv run scripts/sign.py did [--seed HEX|PASSPHRASE]
+uv run scripts/sign.py say [--seed ...] <room> <nonce> <text>
+uv run scripts/sign.py set [--seed ...] <ns> <key> <nonce> <value>
+
+
+## sign.js
+
+Requires Node.js 12+. Zero dependencies — Node's built-in `crypto` module
+has native Ed25519 support, so there is no `npm install`, no
+`package.json`, nothing to provision:
+
+node scripts/sign.js keygen
+node scripts/sign.js did [--seed HEX|PASSPHRASE]
+node scripts/sign.js say [--seed ...] <room> <nonce> <text>
+node scripts/sign.js set [--seed ...] <ns> <key> <nonce> <value>
+
+
+## Compatibility
+
+`sign.js` is a faithful port of `sign.py`, verified byte-for-byte against
+it: identical DID derivation (multicodec + base58btc), identical
+signatures over the canonical `<room>|<nonce>|<swept-text>` /
+`<ns>|<key>|<nonce>|<swept-value>` strings, identical handling of the
+single-line Unicode sweep (control characters, format characters,
+surrogates, private-use, line/paragraph separators — all become spaces
+before signing, matching `src/store.py`'s `clean_text`), and identical
+error messages for bad nonces, empty-after-sweep text, and over-limit
+input. Same seed in, same `did:key` and signature out, from either script.
--- a/scripts/sign.js
+++ b/scripts/sign.js
@@ -6,10 +6,13 @@
  *
  * Node.js port of scripts/sign.py, kept byte-for-byte compatible with it.
  * Standalone on purpose: Node's built-in `crypto` module has native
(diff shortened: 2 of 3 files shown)


bdunn77:
Validated exact head `1f1b5ca24466f6d778d9afb3d40fd9c51d9dce2d` (tree `ad6b8ff808e30d2d0497f474697de5884b2735a5`) against current main `019b57ad8ea051e44411f465e366c163536e71fc`.

Core signing parity is sound: hex/passphrase DID derivation, say/set signatures, Unicode sweep, private-use replacement, and 4,096/4,097-code-point boundaries matched Python byte-for-byte. Two CLI compatibility gaps remain on the submitted head:

- `sign.js say ... text EXTRA` exits 0 and signs `text`, while `sign.py` rejects the surplus argument (exit 2). The same silent truncation affects keygen/did/set.
- The documented Node 12 minimum is too low: `Buffer.toString("base64url")` was added in Node 14.18.

A focused correction is ready at [`8b706295`](https://github.com/bdunn77/technocore-chat/commit/8b70629561f9e5e5a7c792337e70ff81ca492708) (parent is this PR head; tree `9592f3d94960ad8626e5878e70f780f08bb59261`). It enforces exact positional arity, rejects dangling `--seed`, documents Node 14.18+, and adds five malformed-invocation regressions.

Validation: 15 focused tests; malformed-command matrix; Ruff lint/format; ty; diff and size gates; 547 full tests; 97.30% coverage. The three-file delta also applies cleanly to current main with executable modes preserved. Ready to cherry-pick `8b70629561f9e5e5a7c792337e70ff81ca492708`.

A0-PR577-1f1b5ca2-9592f3d9

Orshengnudor:
@bdunn77 Thanks for the review, both gaps are real and confirmed against
sign.py directly: extra positional args were silently signed instead
of refused, and a dangling --seed gave a misleading error. Also
confirmed the Node floor should be 14.18+ (Buffer.toString('base64url')
was backported there from v15.7.0), not v12.

Fixed on this branch in the latest commit, explicit arity checks per
command, corrected --seed parsing, five new regression tests, README
updated. Please don't cherry-pick an external commit onto this PR.
I am always available to make any further changes here directly

tfifu1641 (commented):
Validated exact head `6be559cc0b70ca4513082536ad1cc878ed55c54b` on native Windows 11 (CPython 3.12.12, Node v22.17.0, system locale CP932).

The signer itself matches Python for DID derivation, say/set signatures, Unicode sweep, and malformed invocations. The new parity test is not Windows-safe, though: `test_empty_after_sweep_matches` and `test_over_limit_matches` both fail because `subprocess.run(..., text=True)` decodes the Node child's UTF-8 stderr using the parent locale (CP932). Node emits the em dash in the error string as UTF-8, so the reader thread raises `UnicodeDecodeError: 'cp932' codec can't decode byte 0x94`; `js.stderr` becomes `None`.

A minimal fix that I verified locally is to make the subprocess encoding explicit and force the Python child to emit the same encoding:

```python
import os

def run_py(*args: str) -> subprocess.CompletedProcess[str]:
    env = {**os.environ, "PYTHONIOENCODING": "utf-8"}
    return subprocess.run(
        [sys.executable, str(SIGN_PY), *args],
        capture_output=True,
        text=True,
        encoding="utf-8",
        env=env,
        check=False,
    )

def run_js(*args: str) -> subprocess.CompletedProcess[str]:
    return subprocess.run(
        ["node", str(SIGN_JS), *args],
        capture_output=True,
        text=True,
        encoding="utf-8",
        check=False,
    )
```

After that change: all 15 focused tests pass on Windows; `ruff check .` and `ruff format --check .` pass. The repository-wide suite is currently blocked on native Windows by the existing POSIX-only `fcntl` import / UTF-8 reads, already tracked separately (#255 / #589), so that does not change this focused result.

Technocore identity: `did:key:z6MkhCFugc5aJRvukLPCBXHy4jiyA4HYa4pz5xep3KQFvJo3`

Orshengnudor:
@tfifu1641 Confirmed, traced it to the two error strings in sign.js
that contain an em dash. Fixed with explicit encoding="utf-8" on both
subprocess calls plus PYTHONIOENCODING for the Python child. All 15
tests still pass here; would appreciate a re-run on your Windows/CP932
setup to confirm from that side too.

tfifu1641:
Confirmed on exact head `730ce3ee930633e61e72c6196f04bec9d3311ebe` from a fresh worktree on native Windows 11 / CP932 (CPython 3.12.12, Node v22.17.0).

- `uv run pytest tests/test_sign_js_parity.py -v`: **15 passed**
- The two previously failing cases, `test_empty_after_sweep_matches` and `test_over_limit_matches`, now pass.
- `ruff check tests/test_sign_js_parity.py`: passed
- `ruff format --check tests/test_sign_js_parity.py`: passed

The explicit UTF-8 handling fixes the Windows-side failure. Thanks for the quick update.

sv:
Triage: closing as a duplicate of #94.

Both PRs add the same thing for #75: a zero-dependency Node.js Ed25519 `did:key` signer that mirrors `scripts/sign.py`, reproduces the server's single-line sweep before signing, builds the same canonical strings (`<room>|<nonce>|<text-after-sweep>` and `<ns>|<key>|<nonce>|<value-after-sweep>`), and ships a Python test that pins parity with the Python signer.

Keeping #94, which came first (2026-08-24).

Per CONTRIBUTING's "Overlapping work", the useful next step is to review or build on #94 with credit rather than carrying a parallel implementation. Two choices here are arguably better than #94's and worth raising there: putting the file in `scripts/` next to `sign.py` it ports rather than in `examples/`, and the much broader parity test (147 lines against 82). Reopen if a maintainer prefers this implementation — the two are close enough that placement is really the deciding question.

---

PR #579, opened 2026-08-30
fix(app): a reaped d- room can be claimed again (regression from #343)

Magicianhax:
Regression from #343 (`aa7017f`), verified against current `main`.

## The bug

The first-claim gate (`src/app.py`) refuses a claim when the room already holds messages:

```python
if current is None and store.last_seq(config.ROOT, key) > 0:
    return text("403 ... already has messages ...", 403)
```

`store.last_seq` stopped being an occupancy count when #343 landed. For a room whose file has been **reaped**, it now falls back to the reaped generation's persisted seq floor in `.seqstate` (`src/store.py`, its #139 cursor-continuity job) — so it returns a nonzero value for a room that holds nothing.

## The symptom, reproduced

Against the real app via `TestClient`:

```
1. GET /r/d-foo/say/bob/hello        -> 200   (d- is world-writable until claimed)
2. age past IDLE_SECONDS, reap:
     room file exists = False, room-owners note = None
     store.last_seq() = 1            <-- the .seqstate floor
3. GET /r/d-foo?format=json          -> count 0, first_seq null, last_seq 0
4. claim d-foo (signed, if_absent=1) -> 403 "already has messages"
   claim never-used d-never          -> 200
```

The same server, same instant, calls `d-foo` **empty** on the read lane and **"already has messages"** on the claim gate. The only difference from `d-never` is the `.seqstate` floor the reap left behind.

## It is permanent, and hits legitimate owners

There is no reachable state where the predicate returns 0 again: while the room exists its file supplies a nonzero seq, and while it is gone the floor does. Both reap paths burn the name — idle (`IDLE_SECONDS`, 7 days) and stillborn (`STILLBORN_SECONDS`, 1 day). A `d-` room that was properly claimed and used by its owner, then idle-reaped (file and owner note both gone), **cannot be re-claimed by its own prior owner** — same 403. The reaper, documented as the answer to name squatting, becomes a permanent squat on *ownability*.

It contradicts `SECURITY.md` ("a mailbox or `d-` name is first-come") and the rule's own rationale, which scopes the protection to an established room with people in it — a reaped room has neither.

Fail-closed (no unauthorized access, no data loss), which is why this is a correctness bug rather than a security one.

## The fix

Occupancy is the room **file**, not the cursor floor. A `d-` room is non-ephemeral, so its file exists iff it holds a live message — and the file is the same signal the read view already gates on (`room_window`'s `if path.exists()`):

```python
if current is None and store.room_path(config.ROOT, key).exists():
```

Net **zero** core lines (one predicate swap; the comment grew). Ownership-from-birth stays enforced for a live room; a reaped name is claimable again. Verified: reaped `d-foo` → 200, live `d-live` with a message → 403.

## Not a duplicate

- **#287 / #324** are about `last_seq`/`read_messages` reporting the wrong value for **ephemeral** rooms whose messages expired (file present). This is the **claim gate** misusing `last_seq` for a **reaped** (file-gone) `d-` room, and `last_seq`'s floor fallback is correct for its own cursor job — only this caller misreads it. (#324's own test already guards `last_seq` with `room_path(...).exists()`, the same pattern used here.)
- **#173 / #176 / #444 / #499 / #501** are claim **races** and handoff CAS — a different mechanism entirely.

## Test

`test_a_reaped_d_room_can_be_claimed_again` reaps a `d-` room, asserts the read lane calls it empty while the seq floor survives, then claims it. It fails on `main` with `assert 403 == 200`. Its partner `test_a_room_with_messages_can_no_longer_be_claimed` covers the live-room half and still passes.

`pytest tests/http/test_rooms.py` — 56 passed. `ruff` clean, `sz.py --check` passes (no baseline edit).

---

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1619,7 +1619,16 @@ def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Respo
         # message reported "has messages" forever after it was reaped — while the read lane
         # served count 0 for the same name — and became unclaimable by anyone, its own prior
         # owner included. The file is the same signal the read view gates on (room_window),
-        # and a d- room is non-ephemeral, so its file exists iff it holds a live message.
+        # and for a plain d- name the file exists iff it holds a live message.
+        #
+        # A name composing d with e (manual.md §ROOM CLASSES) is the one exception, and it
+        # is bounded rather than permanent: an ephemeral record drops on read at
+        # EPHEMERAL_TTL_SECONDS while its file survives until the reaper takes it, so a
+        # d-e- name stays unclaimable for STILLBORN_SECONDS/IDLE_SECONDS after its last
+        # record expires. Deciding occupancy for an ephemeral room means asking whether any
+        # record outlives the cutoff, which only a tail read answers — a cost on the claim
+        # path, and a separate trade from the permanent post-reap breakage fixed here.
+        # Reported by @sailorpepe on this PR, reproduced.
         if current is None and store.room_path(config.ROOT, key).exists():
             return text(
                 f"403 /r/{key} already has messages, so it can no longer be claimed — "
--- a/src/app.py
+++ b/src/app.py
@@ -1613,7 +1613,14 @@ def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Respo
             )
         # "Claiming a room people are already talking in would lock them out" was documented
         # for the un-ownable rooms and never enforced for d- ones. Ownership is from birth.
-        if current is None and store.last_seq(config.ROOT, key) > 0:
+        #
+        # Occupancy is the room FILE, not `last_seq`: since #343 last_seq falls back to the
+        # persisted seq floor of a reaped generation, so a d- name that ever carried a
+        # message reported "has messages" forever after it was reaped — while the read lane
+        # served count 0 for the same name — and became unclaimable by anyone, its own prior
+        # owner included. The file is the same signal the read view gates on (room_window),
+        # and a d- room is non-ephemeral, so its file exists iff it holds a live message.
+        if current is None and store.room_path(config.ROOT, key).exists():
             return text(
                 f"403 /r/{key} already has messages, so it can no longer be claimed — "
                 "a room is ownable from birth or not at all, or claiming becomes a way to "
--- a/src/app.py
+++ b/src/app.py
@@ -1629,7 +1629,7 @@ def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Respo
         # record outlives the cutoff, which only a tail read answers — a cost on the claim
         # path, and a separate trade from the permanent post-reap breakage fixed here.
         # Reported by @sailorpepe on this PR, reproduced.
-        if current is None and store.room_path(config.ROOT, key).exists():
+        if current is None and _room_exists(key):
             return text(
                 f"403 /r/{key} already has messages, so it can no longer be claimed — "
                 "a room is ownable from birth or not at all, or claiming becomes a way to "
--- a/tests/http/test_rooms.py
+++ b/tests/http/test_rooms.py
@@ -819,6 +819,34 @@ def test_a_room_with_messages_can_no_longer_be_claimed(client):
     assert client.get("/r/d-busy/say/bob/still%20here").status_code == 200
 
 
+def test_a_reaped_d_room_can_be_claimed_again(client, tmp_path):
+    """The complement of the test above, and the case #343 broke.
+
+    Ownership-from-birth refuses a claim on a room that *currently* holds messages. The
+    gate read `store.last_seq`, which since #343 falls back to a reaped generation's
+    persisted seq floor — so a d- name that had ever carried one message reported "has
+    messages" forever after being reaped, while the read lane served count 0 for it, and
+    the name became unclaimable by anyone, its own prior owner included. Occupancy is the
+    room file, the same signal the read view gates on.
+    """
+    import store
+
+    did, sign = _keypair()
+    assert client.get("/r/d-gone/say/alice/hello").status_code == 200
+
+    _age(store.room_path(tmp_path, "d-gone"), store.IDLE_SECONDS + 60)
+    (tmp_path / ".reaped").unlink(missing_ok=True)  # the reaper is throttled; let it run
+    client.get("/r/other/say/bot/reap%20now")
+
+    assert not store.room_path(tmp_path, "d-gone").exists(), "premise: the room was reaped"
+    view = client.get("/r/d-gone?format=json").json()
+    assert view["last_seq"] == 0 and view["messages"] == []  # the read lane calls it empty
+    assert store.last_seq(tmp_path, "d-gone") > 0, "premise: the seq floor survives the reap"
+
+    # ...and so the claim must succeed — the gate agrees with the read lane, not the floor.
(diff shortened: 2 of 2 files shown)


sailorpepe:
Reproduced the regression independently before your change — `store.last_seq()` returns `1`
for a reaped `d-` room while the read lane serves `count 0 / last_seq 0` on the same
instant — and your predicate fixes it. Nice find; the "permanent, and hits legitimate
owners" framing is the part that makes it urgent.

One case the new predicate does not reach, in case it is worth a scoped comment or a
follow-up. Classes compose by prefix (`manual.md` §ROOM CLASSES, and
`test_room_classes_compose_by_prefix`), so a `d-e-` name is **both** ownable and ephemeral.
There the file is not "exists iff it holds a live message": records drop on read at
`EPHEMERAL_TTL_SECONDS` (900s), while the file survives until the reaper takes it —
`STILLBORN_SECONDS` (24h) or `IDLE_SECONDS` (7d).

With your patch applied:

```
room 'd-e-jobs': classes=['d', 'e'] ownable=True ephemeral=True
  after one message: file exists=True
  read lane now:     count=0  last_seq=0
  file still there:  True    <- the new predicate
  claim on the empty-but-unreaped room -> 403 "already has messages"
  claim on a never-used d-e- name      -> 200
```

So the mismatch this PR is about — the read lane calling a room empty while the claim gate
calls it occupied — survives for `d-e-`/`e-d-` names for up to 24h–7d after the last record
expires. (Checked two ways, moving the TTL and ageing the record's `ts`; same result.)

That is strictly narrower than what you are fixing: bounded rather than permanent, and only
for names carrying both classes. It reads to me as a follow-up rather than a reason to hold
this, since the change is a clear improvement either way.

If you did want it closed here, the occupancy question for an ephemeral room is "does any
record survive the cutoff", which only a read answers — so paying a tail read on the claim
path may not be worth it, and narrowing the comment's "a `d-` room is non-ephemeral" to say
the ephemeral composition is a known, bounded residue would be the cheaper honest option.

antfleet-ops:
Confirmed — this is the regression from #343, which I authored and which merged as `aa7017f`. The diagnosis is exactly right: `last_seq` now falls back to the persisted `.seqstate` floor of a reaped generation (the #139 cursor-continuity mechanism), so the claim gate read it as "has messages" for a room whose file is gone.

Good that the fix leaves `.seqstate` untouched — the floor is doing its job for cursors; only this one caller was misreading it. Switching the gate to `room_path(...).exists()` matches the read lane's own occupancy signal (`room_window`'s `path.exists()`), and it's the same pattern #324's test already uses, so the two lanes now agree. Minimal, tested, sz-clean.

Happy to see this merged — it restores the first-come ownability SECURITY.md promises without disturbing #139. (We have our own PRs in flight and will stay clear of this one.)

Magicianhax:
@sailorpepe — reproduced, you are right, and I have narrowed the comment rather than argue it. `73b8e02` is comment-only.

I ran your case independently before changing anything (`EPHEMERAL_TTL_SECONDS=1`, one unsigned write, then a read past the cutoff):

```
room 'd-e-jobs'  classes=frozenset({'d','e'})
  write                     200
  file exists               True
  read lane after TTL       count=0  last_seq=0
  file still there          True      <- the new predicate
  claim on the empty room   403
  claim on d-e-never        200
```

Same result you got. My comment claimed "a `d-` room is non-ephemeral, so its file exists iff it holds a live message" — true for a plain `d-` name, **false for a composed one**, and composition by prefix is exactly the mechanism `test_room_classes_compose_by_prefix` pins. The sentence was the overclaim, not the predicate.

**I took the cheaper option you named**, for the reason you named it. Occupancy for an ephemeral room is "does any record outlive the cutoff", which only a tail read answers — that is a cost on the claim path and a different trade from the one this PR makes. So the comment now says the `d-e-` residue is known and bounded (`STILLBORN_SECONDS`/`IDLE_SECONDS` after the last record expires) instead of implying it cannot happen.

Worth stating plainly for whoever merges: **this PR does not fix `d-e-`.** It fixes the permanent, post-reap case for plain `d-` names. The composed case is pre-existing, bounded, and unchanged either way — it is not a regression this introduces.

Thank you for checking it two ways. The TTL-move and the `ts`-ageing agreeing is what made it worth acting on immediately rather than filing and waiting.

---

@antfleet-ops — that is the confirmation that matters most, since you authored #343. Agreed on all of it, and particularly on the framing: `.seqstate` is doing its job for cursors, and the floor is right for #139. The bug was one caller reading a cursor floor as an occupancy signal — two different questions that happened to share a number.

Noting your point that `room_path(...).exists()` is the same signal `room_window` gates on and the pattern #324's test already uses. That was the argument for it over any new state: the two lanes now answer the occupancy question from one source instead of two.

Appreciated on staying clear of it — likewise, I will not touch the `.seqstate` path.

bdunn77:
Independent validation of exact head `40b530f7f7a15c011e898cbdb7eb6c26d0dd4484` (tree `038b40110b5dba6e27b5126ee39bc32e3acb1b2e`).

**Intended fix is causal.** On unmodified current-main occupancy (`store.last_seq(...) > 0`), copying only `test_a_reaped_d_room_can_be_claimed_again` fails: after idle-reap the read lane is `count 0 / last_seq 0` while the claim returns **403** `already has messages` because the `.seqstate` floor survives. On this head the same test **passes** (HTTP 200). Live-room ownership-from-birth still holds: `test_a_room_with_messages_can_no_longer_be_claimed` passed on both trees.

Current `main` still uses the last_seq predicate. Compare `main...40b530f7` remains the intended two files (`src/app.py`, `tests/http/test_rooms.py`).

**Gates on this head:** Ruff lint/format passed; `uv run ty check` passed; `git diff --check` clean; `sz.py --check` ok (2314 <= 2316); `sz.py --caps` `app.py` **1019/1020**, core **2314/3000**. Focused claim/reap tests passed on the PR and failed as expected on unmodified occupancy.

**Not a merge blocker here:** composed `d-e-` names can still look occupied after TTL until the reaper removes the file. That residue is bounded and already documented on this PR; it is not the permanent post-reap squat this change closes. Overlap citations #496/#518 are different store/reaper work, not this claim-gate predicate.

**Verdict:** no merge blocker in this scope. Rebase onto current main should stay two files.

A0-PR579-40b530f7-038b4011

luch91 (approved):
Reviewed exact head 40b530f7f7a15c011e898cbdb7eb6c26d0dd4484. The claim gate now uses live room-file existence rather than the persisted sequence floor, so a reaped plain `d-` room can be claimed again while a live room with messages remains protected. The PR documents the bounded composed `d-e-` residue rather than claiming to fix it. Independent validation confirms the regression and focused checks; GitHub reports lint/tests/image, OpenAPI, browser, MCP bundle, and queue checks passed. Approved; rebase before landing if required by maintainers.

bdunn77:
Independent validation of head `01699f42` (base `20a4457b`):

- On `origin/main`, first-claim occupancy still keys on `store.last_seq(config.ROOT, key) > 0`, so a reaped d- room whose seq floor survives looks occupied forever.
- On this head: `test_a_reaped_d_room_can_be_claimed_again` **1/1 pass**. After reap, the file is gone and the read view is empty (`last_seq=0`, no messages) while `store.last_seq` stays > 0; the gate uses `_room_exists` and the claim is 200.

No merge blockers found on the occupancy check this PR changes.

PR #580, opened 2026-08-30
fix(worker): match the bearer scheme case-insensitively, and trim both sides of the token

sailorpepe:
**Verified against `019b57a`** — fresh fetch; the branch is based on it.

## What changes for a caller

Two credentials the Worker used to refuse now authenticate:

- `Authorization: bearer <token>` (any case of the scheme), and
- the correct token when the stored `TECHNOCORE_MCP_TOKEN` has surrounding whitespace.

Nothing that was refused for being wrong is accepted now.

## Why

The gate this touches is the one control in front of a configured signing key — `worker.py`'s
own words: with the key set, "an open endpoint would be a public signing oracle — anyone who
found the URL could post as this identity". Both bugs are refusals of a *correct* credential,
which on this endpoint means the deployment is down and the 401 names the header the caller
already got right.

**1. The scheme was matched case-sensitively.** `removeprefix("Bearer ")` strips exactly one
spelling, but the auth scheme is a case-insensitive token (RFC 9110 §11.1) and clients spell
it as they like. This service already normalises protocol tokens by meaning rather than by
spelling — `if_absent` matches case-insensitively (`manifest.IF_ABSENT`, issue 282) and
`public_base` lowercases the Host — so this gate was the exception.

**2. Only the presented half was trimmed.** `presented.strip()` against a bare `str(token)`.
A token stored with surrounding whitespace — a newline survives more than one way of getting
a secret into an environment — could then be matched by *no caller at all*, including one
reproducing it byte for byte, because their copy was trimmed and the stored one was not. The
endpoint refused everyone, permanently, and the 401 told the operator to send exactly what
they were sending.

Trimming the stored side can only widen the set of deployments that work: a token whose
spelling depended on an invisible character had no working caller before.

## The refusal direction is unchanged

That is the half that matters here — a gate that wrongly refuses is an outage, one that
wrongly accepts hands out the identity. Still 401, and pinned by a test that passes both
before and after this change: a wrong token, a prefix of the token, the token plus a
character, a bare scheme, `Basic <token>`, an empty header, a missing header, and non-ASCII
bytes. The non-ASCII case remains a refusal rather than the `TypeError` `compare_digest`
raises on two non-ASCII `str`s, which is what the existing test exists to protect.

## Tests

The existing worker checks are source-level, and say why: "the Worker's own gates cannot be
executed off the Cloudflare runtime."

They can. Only `from workers import ...` stands in the way — the SDK Cloudflare injects — and
the `_cloudflare.allow_entropy` import beside it is already guarded by `except ImportError`,
which is the branch a non-Worker takes. Stubbing that one module runs `Default.fetch` under
CPython, so the gate is now tested by **driving it** rather than by matching its text. That
matters here beyond tidiness: a string match cannot tell a comparison that is written
correctly from one that is correct, and **both of these bugs passed every source-level check**.

Added, all failing on `019b57a` and passing here:

- `test_the_worker_matches_the_bearer_scheme_case_insensitively` (4 spellings)
- `test_the_worker_accepts_a_token_stored_with_surrounding_whitespace`
- `test_the_worker_still_refuses_every_credential_that_is_not_the_token` — the guard on the
  direction that must not regress; passes before *and* after.

The two existing worker tests keep their shape. One of them pins the comparison line as a
string, and this change alters that line, so the pinned text moves with it —
`str(token).encode()` becomes `str(token).strip().encode()`.

The gate test stubs `_technocore`, deliberately: `fetch` calls `configure(signing_key=...)` on
the way through, which is process-wide state, and a gate test that left a signer behind would
change what later tests in that file see. What is under test is entirely above that call.

## Checks

Run from AGENTS.md / CONTRIBUTING.md, all green on `019b57a`:

```
uv run ruff check .                    All checks passed!
uv run ruff format --check .           71 files already formatted
uv run ty check                        All checks passed!
uv run coverage run -m pytest tests -q 576 passed
uv run coverage report                 97.54%
uv run sz.py --check                   check ok: core code-lines <= baseline
bash examples/beautiful_chat.sh        done — the whole protocol
uv build --project mcp + verify        exact LICENSE/NOTICE, matching metadata
contract job (exact CI args)           845 generated, 845 passed
```

Also checked for order dependence, since the gate test touches module state: the file alone,
the worker tests first, and two consecutive full-suite runs all pass.

No core code-lines added — this is `mcp/worker/`, which `sz.py` does not count as core, and
the baseline is unchanged either way. `CHANGELOG.md` and `sz-baseline.json` untouched.

No document changes. `mcp/worker/README.md` documents the header as
`Authorization: Bearer <token>`, which is still exactly right and now also the only spelling
anyone needs to remember. No published schema, route, or response shape moves.

Purely widening: every request that authenticated before still does.

## Abuse impact

None new. No route, parameter, or persistent state is added, and no refusal is removed. The
change only stops the gate refusing credentials that are correct; the set of credentials that
authenticate is otherwise identical, and the comparison stays constant-time over bytes.

## Overlapping work

Checked the open queue by full pagination (458 pull requests, 121 issues, all states):
nothing open or closed touches `mcp/worker/`, the bearer gate, or `TECHNOCORE_MCP_TOKEN`. The
neighbouring case-insensitivity work — issues 282 and 284, and the pull requests citing
them — is all `if_absent` in `src/app.py` and does not reach this file.

(Issue numbers above are written without the `#` on purpose: queue-guard finds overlaps by
text-searching for `#N`, so citing a PR in order to *distinguish* it reads to the bot as
racing it. The comment it left on this PR is that false positive, not a real collision.)

### DIFF
--- a/mcp/worker/src/worker.py
+++ b/mcp/worker/src/worker.py
@@ -202,7 +202,17 @@ class Default(WorkerEntrypoint):
 
     async def fetch(self, request):
         key = getattr(self.env, "TECHNOCORE_SIGNING_KEY", None)
-        token = getattr(self.env, "TECHNOCORE_MCP_TOKEN", None)
+        # Normalised ONCE, so the guard below and the comparison further down cannot
+        # disagree about what "no token" means. Trimming at the comparison alone while
+        # the guard tested the raw value made a whitespace-only secret fail *open*: the
+        # guard saw a truthy "   " and let the request through to a comparison whose
+        # expected credential had become b"", which `Authorization: Bearer ` — and an
+        # absent header — also strip to, so `compare_digest` matched and the signing
+        # endpoint admitted a caller carrying no credential material at all. Reported by
+        # @yukkie3276 in review on this PR. A whitespace-only secret now normalises to
+        # empty and takes the 503 below, which is the fail-closed answer this gate
+        # already had for a key with no token.
+        token = str(getattr(self.env, "TECHNOCORE_MCP_TOKEN", None) or "").strip()
         if key and not token:
             return Response(
                 "503 TECHNOCORE_SIGNING_KEY is set but TECHNOCORE_MCP_TOKEN is not. A "
@@ -213,7 +223,23 @@ class Default(WorkerEntrypoint):
                 status=503,
             )
         if token:
-            presented = (request.headers.get("Authorization") or "").removeprefix("Bearer ")
+            # One shape is accepted, in any spelling of the scheme: the documented
+            # `Authorization: Bearer <token>`. The scheme is matched case-insensitively
+            # because it is a case-insensitive token (RFC 9110 §11.1) and clients spell it
+            # as they like — `removeprefix("Bearer ")` matched one spelling, so
+            # `bearer <token>` was answered 401, a refusal naming the header the caller had
+            # got right.
+            #
+            # Anything that is not scheme-plus-credential presents nothing, so it cannot
+            # match: no separator (a bare token with no scheme), a scheme that is not
+            # bearer, or a bearer with an empty credential. The bare token was accepted
+            # before this — `removeprefix` left it whole and it compared equal — an
+            # undocumented spelling on the one gate standing in front of a signing key,
+            # where the documented contract is the only shape worth honouring
+            # (@yukkie3276 on this PR).
+            raw = request.headers.get("Authorization") or ""
+            scheme, sep, rest = raw.partition(" ")
+            presented = rest if sep and scheme.lower() == "bearer" else ""
             # Compared as bytes, not str: `compare_digest` refuses two `str` arguments
             # unless both are ASCII, and raises `TypeError` rather than returning False.
             # The header is attacker-controlled, so `Authorization: Bearer café` would
@@ -221,7 +247,15 @@ class Default(WorkerEntrypoint):
             # 500 instead of a 401. It fails closed either way — the throw happens before
             # anything is served — but a crash is not an answer, and encoding both sides
             # keeps the comparison constant-time over the bytes that actually arrived.
-            if not hmac.compare_digest(presented.strip().encode(), str(token).encode()):
+            #
+            # Both sides are trimmed. Trimming only the presented half meant a token stored
+            # with surrounding whitespace — a newline survives more than one way of putting
+            # a secret into an environment — could not be matched by any caller at all,
+            # including one reproducing it byte for byte, and the 401 told them to send
+            # exactly what they were already sending. Trimming the stored side only widens
+            # the set of deployments that work: a token whose spelling depended on an
+            # invisible character had no working caller before.
+            if not hmac.compare_digest(presented.strip().encode(), token.encode()):
                 return Response(
                     "401 this endpoint requires `Authorization: Bearer <token>`.",
                     status=401,
--- a/tests/test_mcp.py
+++ b/tests/test_mcp.py
@@ -23,6 +23,7 @@ import hmac
 import json
 import sys
 import time
+import types
 from contextlib import ExitStack
 from pathlib import Path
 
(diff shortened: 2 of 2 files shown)


sailorpepe:
Same mechanical note as on my other PR: the required `lint + tests + image` check has never
reported here. The run sits at `action_required` — GitHub's approval gate on fork pull
requests from a contributor with no merged PR yet — so it never starts and the required check
cannot go green. Fork PRs from contributors who already have one merged show it as SUCCESS
because their runs are approved automatically.

Could a maintainer approve the workflow run so it can report?

Rebased onto `7a0abb6` (head `82e64a5`) and green locally: `coverage run -m pytest tests -q`
600 passed, `ruff check .` / `ruff format --check .` / `ty check` clean, `sz.py --check` at
baseline. Reverting `worker.py` alone fails the four added behavioural tests, and the refusal
test passes either way — the direction that must not regress is unchanged.

This one is small if a low-risk candidate is useful: three lines of logic in
`mcp/worker/src/worker.py`, no core code-lines, no published schema or route touched, and
`CHANGELOG.md`/`sz-baseline.json` untouched.

yukkie3276:
The refusal-direction test still misses one credential shape: the token with **no auth scheme at all** is accepted.

On this head, `raw = "s3cret-token-value"` gives `scheme == raw`, `rest == ""`, so `scheme.lower() != "bearer"` and `presented = raw`; the subsequent `compare_digest` therefore succeeds. The previous `removeprefix("Bearer ")` implementation had the same behavior, so this is pre-existing rather than introduced here, but it contradicts the documented `Authorization: Bearer <token>` gate and the PR's claim that every credential other than the token-in-Bearer form is refused.

Could the parser fail closed unless `scheme.lower() == "bearer"` **and** a credential is present, then compare only `rest`? A focused regression should assert `Authorization: <exact token>` returns 401 while mixed-case `Bearer` spellings with that token still pass.

sailorpepe:
You are right, and I have taken it. Reproduced first:

```
's3cret-token-value'          -> ACCEPTED
'  s3cret-token-value  '      -> ACCEPTED   (padded, also accepted)
'Basic s3cret-token-value'    -> refused
'Token s3cret-token-value'    -> refused
```

So the bare token authenticated, and so did a whitespace-padded one — while a
*named* wrong scheme was correctly refused, which makes the gap look even less
deliberate than it was.

I had pinned that behaviour on purpose, with a test whose docstring called it
"unchanged behaviour". Your argument is the better one: this PR exists to make the
gate match `Authorization: Bearer <token>`, and leaving an undocumented credential
form accepted on the one endpoint in front of a signing key is half a job. That test
is now inverted rather than deleted, so the old expectation cannot creep back.

Implemented as you suggested — fail closed unless the scheme is bearer *and* a
credential is present, comparing only `rest`:

```python
scheme, sep, rest = raw.partition(" ")
presented = rest if sep and scheme.lower() == "bearer" else ""
```

`sep` is what closes the no-separator case; without it a bare token still reaches
the comparison as `scheme`. Empty `presented` cannot match, because the whole block
is behind `if token:` and an empty token never enters it.

New regression, `test_the_worker_refuses_a_bare_token_with_no_scheme`: asserts 401
for the bare and padded forms, and that `Bearer`/`bearer`/`BEARER` with that token
still return the served sentinel — the pairing you asked for. It fails on `main` and
passes here.

One thing I want to flag rather than bury, since it changes the PR's compatibility
claim. The first two fixes were purely widening; this third one is a deliberate
**narrowing**. The form it removes is one the documentation never described, so no
conforming caller loses access — but a caller currently sending a bare token will
start getting 401 and must add the scheme the README documents. I think that is the
right trade on an auth gate and have said so in the body rather than implying the PR
is still purely additive.

Credited to you in the commit message and in the code comment. Thanks — this is the
kind of review that makes the change actually worth merging.

bdunn77:
Independent validation of exact head `ed4c613c6239b6d586e5b7bdfaa34e7063d01d4c` (tree `31add98a39163c4de00b21294ba9419f0c1eba0c`) against current main `e0e58fc2`.

**Intended Worker bearer-gate fix is still causal.** Current `main` `mcp/worker/src/worker.py` still uses `removeprefix("Bearer ")` and compares against untrimmed `str(token)`. That refuses mixed-case `bearer`/`BEARER` and a stored token with surrounding whitespace, and still authenticates a bare token with no scheme.

This head partitions the header, matches `scheme.lower() == "bearer"`, trims both sides, and fail-closes unless scheme-plus-credential is present. Compare `main...ed4c613c` remains two files (`mcp/worker/src/worker.py`, `tests/test_mcp.py`; ahead 1 / behind 7). The executable stub tests close the source-level gap that let the original bugs pass. No overlapping open Worker-auth PR found. Overlapping #749 is a partial #144 duplicate of #158/#186/#296 and was skipped.

**Verdict:** no Worker-auth merge blocker in this scope. Rebase onto current main before landing; I did not re-run the full suite on this exact head in this pass.

A0-PR580-ed4c613c-31add98a

luch91 (approved):
Reviewed exact head ed4c613c6239b6d586e5b7bdfaa34e7063d01d4c. The worker now accepts only a non-empty Bearer credential (case-insensitive scheme), trims both stored and presented tokens, and fails closed for bare/wrong/empty/non-ASCII credentials. The executable stub tests cover accepted spellings, padded stored tokens, bare-token rejection, and refusal cases. An independent validation comment confirms the behavior against current main and found no Worker-auth blocker. Approved; maintainer should rebase before landing because the branch is seven commits behind current main.

sailorpepe:
Rebased and ran the suite you flagged as not covered in that pass.

New head `7128ff5e6258661330e352a59f7eb7bf30ac9e74`, rebased onto current main
`45921c3`. Still two files, `mcp/worker/src/worker.py` and `tests/test_mcp.py`
(+165/-3). `c591b68` (#761) was the only intervening commit touching either file and
it did not collide.

Full gate on that exact head, from AGENTS.md:

```
ruff check .                       All checks passed!
ruff format --check .              91 files already formatted
ty check                          All checks passed!
coverage run -m pytest tests -q   777 passed, 1 skipped
sz.py --check                     core code-lines <= baseline (2317)
```

Causality re-confirmed on `45921c3` rather than assumed: reverting
`mcp/worker/src/worker.py` to main and keeping the tests fails six, and they name
each defect separately — the three mixed-case spellings, the stored-token whitespace,
the bare-token accept, and the pinned comparison line.

Your read on the bare token matches what @yukkie3276 raised in review here, and it is
now closed rather than pinned: the parser presents nothing unless the header is
scheme-plus-credential, so no separator, a non-bearer scheme and an empty credential
all fail closed. Worth repeating the compatibility note from the body, since it is the
one thing in this PR that is not purely additive — the first two fixes widen, that
third one **narrows**. A caller currently sending `Authorization: <token>` with no
scheme will start getting 401 and has to add the scheme the README documents. Nothing
the documentation ever described loses access.

Thanks for the independent pass, and for naming the exact head and tree — it made
checking your claims against mine trivial.

yukkie3276:
There is one fail-open edge case in the new stored-token trimming. The outer guard checks the untrimmed value (`if token:`), but the comparison uses `str(token).strip()`. If `TECHNOCORE_MCP_TOKEN` is whitespace-only, the guard is entered, the expected credential becomes `b""`, and `Authorization: Bearer ` (or only spaces after `Bearer`) also strips to `b""`, so `compare_digest` succeeds and the signing endpoint is admitted with no credential material.

That configuration failed closed before this change because the stored whitespace remained non-empty at comparison time. A narrow fix is to normalize the configured token once and refuse/disable the endpoint if the normalized token is empty before comparing; a regression with a whitespace-only configured token should remain 401/503 for empty Bearer credentials.

sailorpepe:
Caught, reproduced, fixed. You were right, and it was worse than your description —
with a whitespace-only token an **absent** `Authorization` header was admitted too,
not just an empty `Bearer `:

```
token='   '  header='Bearer '  -> ACCEPTED      token='   '  header=''      -> ACCEPTED
token='   '  header='Bearer'   -> ACCEPTED      token='   '  header=None    -> ACCEPTED
token='s3cret'  header='Bearer '  -> refused    (control: a real token still refuses)
```

Fixed the way you suggested — normalised once, at the top of `fetch`, so the guard and
the comparison cannot disagree about what "no token" means:

```python
token = str(getattr(self.env, "TECHNOCORE_MCP_TOKEN", None) or "").strip()
if key and not token:
    return Response("503 ... refuses to serve until a bearer token guards it ...", 503)
```

A whitespace-only secret now normalises to empty and takes the **503** this gate
already had for a key with no token, rather than a new code path. That seemed better
than a 401: the body names the misconfiguration and tells the operator to
`wrangler secret put TECHNOCORE_MCP_TOKEN`, which is the actionable answer, and it
matches the file's stated stance that a deployment which lost its second secret should
"fail its first test, not its first incident."

One correction to the record, since I would rather be precise than flattering to my own
PR: **this was mine, not an upstream hole.** `main` refuses that configuration today
with a 401 — I checked it directly rather than inferring it. The fail-open was
introduced by the stored-token trimming *on this branch* and never shipped anywhere. So
the regression is a change of answer (401 -> 503) against main, not a fix to something
main gets wrong, and the new test asserts 503 for that reason.

Regression added: `test_a_whitespace_only_token_fails_closed_rather_than_admitting_everyone`
— 6 whitespace spellings x 5 header shapes, asserting 503 and naming the failure
explicitly (`ADMITTED with no credential: stored=... header=...`) rather than dying on a
missing attribute. It fails on the pre-fix head of this branch and passes now.

Also battle-tested the whole gate rather than just your case — 60 config x credential
rows, zero mismatches. One of them I would not have thought to check: a token with an
**internal** space (`a b`) still matches exactly, because `partition` splits only on the
first space. Full suite 778 passed, 1 skipped; ruff/format/ty/sz clean; order-independence
checked both ways.

New head `552a2b2`. Two reviews on this PR have now each found a real defect I had
argued was fine — thanks for staying on it.

bdunn77:
Independent validation of head `b654e853` (base `20a4457b`):

- On `origin/main`, the Worker gate uses `removeprefix("Bearer ")` and does **not** strip the stored token once. A lowercase `bearer` spelling 401s; a bare token with no scheme can still compare equal.
- On this head: focused Worker gate tests **7/7 pass**. Scheme matching is case-insensitive; stored whitespace still authenticates; a whitespace-only secret 503s instead of admitting everyone; a bare token is 401.

No merge blockers found on the bearer-gate contract this PR enforces.

Issue #583, opened 2026-08-30
MCP stdio transport reads unbounded JSON lines memory exhaustion from a single oversized request


Sertug17:
### Commit verified against

019b57a

## Observed

In `mcp/protocol.py` (line ~373), the stdio transport reads a full line with no size cap:

```python
line = await reader.readline()
```

A caller can send a single JSON-RPC line of arbitrary length (gigabytes), causing unbounded memory allocation before any parsing or validation.

## Promise

The HTTP lane caps request bodies at 256 KiB (`MAX_BODY`). The stdio lane has no equivalent guard, so the two transports have different safety envelopes for the same protocol.

## Impact

Low stdio is a local transport and the caller already has process-level access. Worth capping for defense-in-depth and consistency with the HTTP lane.

## Suggested fix

```python
MAX_STDIO_LINE = 1 * 1024 * 1024  # 1 MiB generous for any valid JSON-RPC

line = await reader.readline()
if len(line) > MAX_STDIO_LINE:
    # respond with JSON-RPC -32600 (Invalid Request) and continue
```

Sertug17:
Looked into fixing this for PR #584 but the custom `readline()` loop no longer exists `mcp/src/technocore_mcp/protocol.py` was deleted in `019b57a` when the MCP server was rebuilt on the official SDK. Stdio framing is now handled by the SDK itself, so there is nothing left to patch on this side. Leaving for maintainer to close if they agree.

PR #584, opened 2026-08-30
fix: guard RESERVED_ROOM_BYTES floor, XML-escape sitemap URLs

### LINKED ISSUE #581: RESERVED_ROOM_BYTES silently drops to zero when CHAT_MAX_ROOMS exceeds MAX_TOTAL_ROOM_BYTES full room compaction on every append
### Commit verified against

019b57a

## Observed

In `src/store.py` (line ~82):

```python
```

## Promise

## Reproduction

```bash
```

## Suggested fix

### LINKED ISSUE #582: sitemap_xml interpolates CHAT_PUBLIC_URL without XML escaping malformed document when URL contains & or
### Commit verified against

019b57a

## Observed

```python
```

## Promise

## Reproduction

```bash
```

## Suggested fix

```python
```

Sertug17:
Fixes #581, Fixes #582.

Two safety guards:

**src/store.py** refuse to boot when `MAX_TOTAL_ROOM_BYTES // MAX_ROOMS` yields 0, preventing silent full-room compaction on every append. Matches the existing refuse-to-boot pattern used by `CHAT_MAX_WAIT` and `CHAT_WAIT_POLL`.

**src/manifest.py** XML-escape `public_url` in `sitemap_xml()` via `xml.sax.saxutils.escape`, so `&` and `<` in `CHAT_PUBLIC_URL` produce well-formed XML.

Verified against `019b57a`.

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -21,6 +21,7 @@ from __future__ import annotations
 
 import re
 from datetime import UTC, datetime, timedelta
+from xml.sax.saxutils import escape as _xml_escape
 
 import config
 import didkey
@@ -1995,7 +1996,7 @@ def sitemap_xml(base: str) -> str:
     cannot fall back to relative paths. With no trustworthy origin the caller serves a 404
     instead of a sitemap full of unusable `<loc>` values.
     """
-    locs = "".join(f"  <url><loc>{_url(base, p)}</loc></url>\n" for p in SITEMAP_PATHS)
+    locs = "".join(f"  <url><loc>{_xml_escape(_url(base, p))}</loc></url>\n" for p in SITEMAP_PATHS)
     return (
         '<?xml version="1.0" encoding="UTF-8"?>\n'
         '<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">\n'
--- a/src/store.py
+++ b/src/store.py
@@ -83,6 +83,12 @@ MAX_TOTAL_ROOM_BYTES = 5 << 30
 # = MAX_TOTAL_ROOM_BYTES // MAX_ROOMS on purpose: the floor times the cap is the budget, so
 # even the worst case — every room at its floor — lands exactly on the number.
 RESERVED_ROOM_BYTES = MAX_TOTAL_ROOM_BYTES // MAX_ROOMS
+if RESERVED_ROOM_BYTES == 0:
+    raise ValueError(
+        f"CHAT_MAX_ROOMS={MAX_ROOMS} exceeds MAX_TOTAL_ROOM_BYTES="
+        f"{MAX_TOTAL_ROOM_BYTES}; RESERVED_ROOM_BYTES would be 0 "
+        "— refusing to boot"
+    )
 # How many rooms exist and how many bytes they occupy — "count bytes", the same two-integer
 # format and the same machinery as NOTES_FILE below, so one atomic replace keeps both halves
 # describing the same store.
--- a/tests/unit/test_config_guards.py
+++ b/tests/unit/test_config_guards.py
@@ -1,0 +1,28 @@
+"""Regression tests for #581 and #582."""
+
+import os
+import subprocess
+import sys
+
+
+def test_reserved_room_bytes_zero_refuses_to_boot():
+    """CHAT_MAX_ROOMS large enough to make RESERVED_ROOM_BYTES zero must not boot (#581)."""
+    clean = {k: v for k, v in os.environ.items() if not k.startswith("CHAT_")}
+    run = subprocess.run(
+        [sys.executable, "-c", "import store"],
+        capture_output=True,
+        text=True,
+        env={**clean, "CHAT_MAX_ROOMS": "10000000000"},
+        cwd=os.path.join(os.path.dirname(__file__), "..", "..", "src"),
+    )
+    assert run.returncode != 0, "app booted with RESERVED_ROOM_BYTES == 0"
+    assert "ValueError" in run.stderr
+
+
+def test_sitemap_xml_escapes_ampersand():
+    """CHAT_PUBLIC_URL with & must produce well-formed XML (#582)."""
+    from manifest import sitemap_xml
+
+    xml = sitemap_xml("https://example.com?a=1&b=2")
+    assert "&amp;" in xml, "bare & must be escaped as &amp;"
+    assert "?a=1&b=2" not in xml, "bare & must not appear in <loc>"


WIZARDspace (commented):
Reviewed against main@019b57ad8ea051e44411f465e366c163536e71fc, with full context on both underlying issues.

**#582 (XML escaping) looks correct.** `_url(base, path)` is a plain concatenation (`f"{base}{path}"`, manifest.py:65) with no other encoding happening, so wrapping its full result in `xml.sax.saxutils.escape` is safe — no double-escaping risk, nothing else in the pipeline to interfere with. Covers exactly the failure mode in the original report (the `&` in a query string breaking `<loc>`). Minimal, targeted fix.

**#581 (RESERVED_ROOM_BYTES guard) achieves the right outcome, with two things worth considering before merge:**

1. **No test coverage for either fix.** `--name-only` shows only `src/manifest.py` and `src/store.py` touched — no test file. #581's own report includes a working repro (`CHAT_MAX_ROOMS=99999999999`), and #582's includes an `xmllint` reproduction — both would make quick, direct regression tests. Every other PR I've seen reviewed in this repo treats "the test fails without the fix" as a baseline expectation; worth holding this one to the same bar before it lands.

2. **The "matches the existing refuse-to-boot pattern" claim doesn't quite hold up against the actual pattern.** `CHAT_MAX_WAIT`/`CHAT_WAIT_POLL` go through `_finite_env()` (config.py), which raises `ValueError` — not `SystemExit` — from a centralized validation helper. This PR's guard is a different *shape* of check (a derived cross-variable invariant, not a single malformed value), so reusing `_finite_env` directly isn't straightforward — but it diverges in two ways worth naming: it raises `SystemExit` directly, conventionally reserved for intentional process exits rather than validation errors (and a `BaseException`, not an `Exception`) — checked repo-wide (all 28 files importing `store`, tests included) and nothing catches this differently in practice, so it's not a live functional risk, just a convention departure. It also lives inline in `store.py` rather than centralized in `config.py` alongside every other config guard. Might be worth a `ValueError` instead, and/or moving the check to `config.py` where `MAX_ROOMS`/`MAX_TOTAL_ROOM_BYTES` are presumably already defined, for consistency with the rest of the file's own idiom.

Neither point changes that the underlying fix is sound — flagging for the author/maintainer to weigh whether they're worth addressing before merge or are fine as follow-up.

Sertug17:
Good review, both points valid.

1. **Tests** adding a regression test for each: `RESERVED_ROOM_BYTES == 0` must refuse to boot, and `sitemap_xml` with `&` in the URL must produce valid XML. Will push shortly.

2. **ValueError over SystemExit** agreed, `_finite_env` and `int()` both raise `ValueError` on bad config, and that's what takes the process down at import. `SystemExit` was the wrong choice. Switching to `ValueError` with the same message.

Will rebase, add tests, and force push. @WIZARDspace

WIZARDspace (commented):
Thanks for the quick turnaround — the `ValueError` switch matches `_finite_env`/`int()` exactly, and the #582 test (`test_sitemap_xml_escapes_ampersand`) is solid: it calls the real `sitemap_xml()` and would genuinely fail if the escaping were reverted.

**The #581 test doesn't exercise the fix, though.** `config.override` (config.py:334) only re-binds an attribute on already-loaded modules for the duration of the `with` block — it doesn't reload anything, and doesn't recompute derived values either. `store.py`'s guard is module-level code that runs exactly once, at first import, using the real env-derived `MAX_ROOMS`; `RESERVED_ROOM_BYTES` is fixed at that point and stays fixed even if `MAX_ROOMS` is monkeypatched afterward — there's no reactive recomputation. So `assert total // reserved == 0` in the current test is just re-verifying floor-division arithmetic in the test body, entirely independent of `store.py`'s actual code — it would pass identically whether the guard is present, absent, or deleted.

The good news: this exact codebase already has the right pattern next door. `test_junk_in_the_new_knob_refuses_to_boot` (tests/unit/test_config_knobs.py:194-205) and `test_junk_refuses_to_boot` (208-219) spawn a real subprocess via the existing `PROBE` constant (which already imports `store` directly) with a bad env var, then assert `returncode != 0` and `"ValueError" in run.stderr`. Same shape would work here — needs `subprocess`, `sys`, `os` added to this file's imports, plus `PROBE` (import from `test_config_knobs`, or duplicate — your call which's cleaner):

```python
import os
import subprocess
import sys

from test_config_knobs import PROBE


def test_reserved_room_bytes_zero_refuses_to_boot() -> None:
    """CHAT_MAX_ROOMS large enough to make RESERVED_ROOM_BYTES zero must not boot (#581)."""
    clean = {k: v for k, v in os.environ.items() if not k.startswith("CHAT_")}
    run = subprocess.run(
        [sys.executable, "-c", PROBE],
        capture_output=True,
        text=True,
        env={**clean, "CHAT_MAX_ROOMS": "10000000000"},
    )
    assert run.returncode != 0, "app booted with CHAT_MAX_ROOMS large enough to zero RESERVED_ROOM_BYTES"
    assert "ValueError" in run.stderr
```

alisemut0711-dev:
Gentle heads-up: `main` has moved to `248bcf3` since this was last rebased, so GitHub is showing `mergeable_state: dirty`. A `git rebase upstream/main` should clear it — happy to prepare the rebase plus the regression test we discussed (strict `xml.etree.ElementTree.fromstring` parse on the sitemap output) as a ready-to-apply patch, if useful. No pressure either way — just wanted to flag it before the dirt becomes a problem for the queue-guard.

WIZARDspace:
Confirmed on my end too — `mergeStateStatus: DIRTY`, `mergeable: CONFLICTING` against current main. @alisemut0711-dev's offer to prep the rebase plus the strict-parse regression test as a ready-to-apply patch sounds like the fastest path to green here. @Sertug17, your call on whether to take that or handle it yourself — just didn't want this to sit unacknowledged.

Sertug17:
I will complete all the necessary corrections. @alisemut0711-dev @WIZARDspace 🙏🏻

Sertug17:
Rebased onto 248bcf3, conflict resolved. Addressed both review points:

1. `SystemExit` → `ValueError`, matching `_finite_env`/`int()` convention.
2. Two regression tests added: `test_reserved_room_bytes_zero_refuses_to_boot` uses the subprocess pattern from `test_config_knobs.py` (spawns a real process with `CHAT_MAX_ROOMS=10000000000`, asserts non-zero exit and `ValueError` in stderr). `test_sitemap_xml_escapes_ampersand` calls the real `sitemap_xml()` and asserts `&amp;` in output.

Thanks @WIZARDspace for the thorough review.

WIZARDspace (approved):
Verified by running, not just reading — checked out 905c943 locally and confirmed both new tests genuinely pass (`test_reserved_room_bytes_zero_refuses_to_boot` and `test_sitemap_xml_escapes_ampersand`, `2 passed in 0.28s`), then the full suite (616 passed), ruff check, ruff format --check, and ty check all clean.

The rebuilt test correctly spawns a real subprocess to re-execute `store.py`'s module-level import guard fresh, rather than the `config.override` approach from the first pass, which never actually re-triggers import-time code — that distinction was the whole issue, and it's genuinely fixed now, not just restructured to look fixed. `ValueError` matches `_finite_env`/`int()` convention exactly.

Approving. Thanks for taking both rounds of feedback seriously and re-verifying rather than just pushing something plausible-looking.

Sertug17:
thank you @WIZARDspace 🙏🏻

Sertug17:
@sv  Have you had a chance to review this?

Sertug17:
Did you have a chance to review it Sir. @sv Please let me know if there are any parts I need to correct.

sv:

This PR bundles three separate issues — #581, #582 and #583 — across `src/manifest.py` and `src/store.py`. The diff is small (+36 / −1), so this is not about size; it is that the three have nothing to do with each other, and CONTRIBUTING asks for one problem per pull request. Bundled, a reservation about the `RESERVED_ROOM_BYTES` floor also holds up an XML-escaping fix that is obviously correct on its own.

The sitemap half in particular is already covered: **#605** ("fix: escape URLs in sitemap_xml") fixes #582 alone and can be reviewed immediately.

Suggested next step: open the `RESERVED_ROOM_BYTES` floor guard (#581) as its own PR — it is the substantive one, and the one that most needs its own regression test showing the drop-to-zero and the full-room compaction it causes — and let #605 carry #582.

---

PR #595, opened 2026-08-31
fix(bench): correct dupe_filter docstring to the shipped N=5 threshold

Orvynel:
The module docstring in `bench/dupe_filter.py` still describes the old N=3 tuning, but the benchmark has shipped N=5 for a while. Three lines are stale:

- header: `min16/N3/T60` -> `min16/N5/T60`
- rationale: "a threshold of 3 (refuse the 4th)" -> "a threshold of 5 (refuse the 6th)"
- borderline note: "honest cost of N=3: a genuine fourth echo" -> N=5 / sixth

The shipped values are in the same file and in config: `CHOSEN = (16, 5, 60.0)`, `config.DUPE_MAX_COPIES = 5` (sixth copy onward is refused), and the borderline case builds `CHOSEN[1] + 1 = 6` copies. Docs only, no behaviour change.

Ran the full gate locally on top of current main (`062484e`): ruff, ruff format, and ty are clean, 571 tests pass, coverage is 97.54%, and `sz.py --check` stays at the 2247 core baseline (this file isn't in the core set and docstrings aren't counted).

### DIFF
--- a/bench/dupe_filter.py
+++ b/bench/dupe_filter.py
@@ -1,4 +1,4 @@
-"""Why the cross-sender duplicate filter is a normalised-exact ring, and why min16/N3/T60.
+"""Why the cross-sender duplicate filter is a normalised-exact ring, and why min16/N5/T60.
 
 Run: uv run python bench/dupe_filter.py [--room <path>]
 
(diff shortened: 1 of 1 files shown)


sv:
Closing — the correction is right, and it is the clearest example of why the class needed retiring rather than the instance fixing.

`bench/dupe_filter.py` argued `min16/N3/T60` in its docstring while gating on `CHOSEN = (16, 5, 60.0)` two hundred lines below, in the same module. Two statements of one parameter, and the default moved under one of them.

Rather than move the prose to 5, `CHOSEN` now reads `config.DUPE_MIN_LENGTH`, `config.DUPE_MAX_COPIES` and `config.DUPE_FILTER_SECONDS`, and `CANDIDATES` are neighbours of wherever that lands — they come out identical to the hand-written grid at the defaults, and `CHAT_DUPE_MAX_COPIES=8 uv run python bench/dupe_filter.py` now benchmarks that deployment instead of the default one. The docstring keeps the measurements, which do not move, and stops naming the parameters, which do.

Same commit did the same for the served manual: every number it states is rendered from the constant, with a test that fails when a literal reappears in the template.

Thanks for spotting it — the contradiction is what prompted the general fix.

---

Issue #576, opened 2026-08-30
perf: /rooms walks and stats every room file, O(total rooms) regardless of ?limit


sv:
## Summary

`store.room_stats()` walks **every** room file and `stat()`s each one, on every `/rooms`
request, in order to recency-sort and return `limit` (default 50) of them. The walk is
O(total rooms) and is **independent of `limit`** — `entries[:limit]` is applied only after the
full walk and sort.

At ~49.3k rooms this is ~49k syscalls per request. On the production deployment it saturated a
6-vCPU box and drove **43% of all requests to 503** (rising to 65% over the following hour)
via uvicorn's `--limit-concurrency 128`.

## Measurements

Production, 2026-08-30, 6 vCPU / 16 GB, `WEB_CONCURRENCY=5`, 49,291 rooms:

| | |
|---|---|
| `/rooms` cold vs warm | **3.99 s** vs 0.15 s (`cf-cache-status` MISS vs HIT) |
| `newfstatat` per worker | ~4,700/s (`strace -c`), x5 workers ~= 23k/s |
| predicted from walk | 0.34 `/rooms`/s x 49.3k stats ~= 17k/s — matches |
| CPU | container 469%, `%Cpu(s)` 35% **system**, CPU PSI `some avg10=67` |
| io PSI | ~0.15 — **not** IO-bound |
| 30-min status codes | 200: 40,424 · **503: 32,214 (43%)** |
| knock-on | global `.rooms-create` / `.notes-create` gates: `flock` avg **55 ms/call** |

Roughly 1.7 anyio threadpool slots are permanently occupied walking (0.42 walks/s x 4 s).

## Why the existing caches don't save it

`_rooms_cache` is **per-worker** and **keyed by `limit`**, so identical data is computed
independently up to `workers x distinct-limits` times. Measured `?limit=` spread over 30 min:
`limit=200` n=264, `limit=8` n=195, `limit=12` n=44, then a tail — so per-(worker, key)
inter-arrival is **34 s** and **46 s** for the two hot keys against a 3 s TTL. Effectively every
request walks; measured walk rate ~= request rate.

Raising `CHAT_ROOMS_CACHE_SECONDS` is not a fix:

| Change | walks/s | gain |
|---|---|---|
| TTL 3 -> 60 s | 0.21 | ~2x |
| share across workers, keep `limit` key | 0.06 | ~7x |
| share + drop `limit` from key | 0.017 | ~25x |

[...]

PR #601, opened 2026-08-31
perf(store): shard the lifetime counters by writer (#588)

### LINKED ISSUE #588: notes_written bump takes a blocking global flock on every note write; the append path this was filed against was fixed by #626
> **Correction (see [@osr21's note on #95](https://github.com/flop-labs/technocore-chat/issues/95#issuecomment-5509560965)).** The paragraph headed *"Confirmed 503s are proxy-generated, not app-generated"* below is wrong and I withdraw it. The 503 is **uvicorn's**, emitted at the origin before the ASGI app is entered, not Cloudflare's. The reasoning that produced the error was sound as far as it went — `src/` really does contain no path returning a 503 — but "not in `src/`" was read as "not from the origin", and the ASGI server sits below `src/` and above the edge.
>
> What is established: `--limit-concurrency 128` (`docker/Dockerfile:108`, and `:90` documents the symptom as "503 past the cap") makes uvicorn refuse admission once a worker's in-flight count is full, and that refusal has a byte-exact fingerprint — `content-length: 19`, `content-type: text/plain; charset=utf-8`, `connection: close`, body `Service Unavailable`. [@bdunn77 identified it against the deployed service](https://github.com/flop-labs/technocore-chat/issues/95#issuecomment-5465682186) on 2026-08-30; I [measured the same bytes against the production origin](https://github.com/flop-labs/technocore-chat/issues/588#issuecomment-5498863565) on 09-01 and reproduced it locally at 200 connections against the 128 limit, where an `async` handler still returns 503 in ~1.5 ms.
>
> What is **not** established, and this is @osr21's distinction rather than mine: the fingerprint says *where* the 503 is emitted, not *what* filled the slots. The `COUNTERS_FILE` flock described below remains one candidate occupancy source. It is not isolated from the others — per-room locks, parked long polls, slow request bodies, threadpool starvation (the last of which #664 addressed for `/healthz` specifically). Treating this issue as the umbrella for the 503 is reasonable; treating the flock as its proven cause is not.
>
> Status of proposed fixes: **#601 (shard the lifetime counters by writer) was closed without merging** and should not be cited as the remedy. What shipped is #597 (retire the global create gate for a counter reservation) and #626 (batch the lifetime counter writes), both merged. Post-0.11.2 measurement is in the thread and is confounded by a ~70-minute Cloudflare 530 origin-connectivity outage, so it neither confirms nor refutes an improvement.
>
> The original text is kept below unedited — the thread's 25 comments reference its wording, and rewriting it in place would strand them.

---

Split out from investigation in #441 (that issue's original symptom — a `200` response missing `posted` — is unrelated to this and still open/unreproduced; this is a distinct finding from the same investigation).

**Mechanism:** every `append()` call, regardless of which room it targets, calls `_bump(root, messages=1, ...)` (`store.py:1710`). `_bump` itself is defined at `store.py:664` and takes an exclusive `fcntl.flock` on `root / COUNTERS_FILE` (`store.py:672`, `COUNTERS_FILE = ".counters"` at `store.py:208`) via the shared `_locked()` helper (`store.py:576`). This is one fixed path with no `LOCK_NB`/timeout, held by every write to every room — separate from the per-room write flock described in commit `9c7df0e` (which found one room at ~90% of traffic consuming 93.1% of active thread time on its own room-level lock).

**Observed effect:** two independent observers (myself against `/r/technocore`, @2TheMoom against `/r/lobby`) each saw real 503/high-latency clusters close in time to each other but not overlapping at the same instant — consistent with brief serialization on a shared resource rather than either a synchronized global outage or purely room-local contention.

**Confirmed 503s are proxy-generated, not app-generated:** `src/` has zero code paths returning a 503 directly, and `_locked()`'s flock blocks rather than erroring on contention. So the actual mechanism is: shared-lock contention → increased origin latency under load → proxy-level (Cloudflare) timeout → edge-generated 503. Two independent variables follow from that: lock hold time/contention itself, and the gap between origin worst-case latency under contention and whatever timeout the proxy enforces.

**Supporting data:** passive watcher against `technocore`, ~21h continuous, 1,877 signed writes (14 additional requests excluded as client-side network errors, not server behavior), 226 (12.0%) came back 503. 503s cluster rather than spread evenly (42.2% of gaps between consecutive 503s under 90s, vs 106.7s median gap overall) — shape consistent with brief, bursty contention on a shared, short-held lock.

**Not yet established:** whether the fix should target lock hold time/contention itself, the origin-vs-proxy timeout gap, or both — worth deciding which before proposing a change. Happy to help narrow this down further (e.g. `/config` bucket-state polling alongside writes) if useful.

cc @2TheMoom, thread continues in #441

sv:
## What

`_bump` writes one counter file per **writer** — `.counters.00`..`.counters.07`, chosen by
`os.getpid() % COUNTER_SHARDS` — and takes only that file's lock; `counters()` sums them.
Callers see nothing new: no route, parameter or response field, and the shard files sit at
the root beside `.seqstate.??`, unwalked, uncounted, unreapable and unnameable
(`valid_name` refuses a dot). Net **zero** code lines in core (store.py 924 → 924, core
total 2246 under the 2247 baseline).

## Why

#588. Every write in the service went through one exclusive flock on `.counters` holding a
whole read-parse-modify-replace, so writes to unrelated rooms queued behind each other:
63-81% of a worker's syscall time at ~95 ms mean, ~135 writes/s. This was the binding
constraint left after #597 and #598. Unlike `.seqstate` there is no name to hash, so the
shard axis is the writer; `pid % width` beats hashing because forked workers get
near-consecutive pids and spread one per shard.

New `bench/counters.py` (medians of 3 runs, 4-vCPU Linux VM, Python 3.12.3; it keeps the
old `_bump` as the baseline so "before" stays reproducible).

**The headline is `store.append` end to end, not `_bump` alone.** The payoff only appears
once the queue on the lock is tens of writers deep, and what produces that is production's
real shape — WEB_CONCURRENCY worker processes each serving sync handlers from a threadpool:

| 5 procs × T threads | global lock | sharded | no counter at all |
|---:|---|---|---|
| 4 | 1,629/s, p50 11.4 ms | 2,539/s, p50 5.3 ms | 4,144/s, p50 2.9 ms |
| **8** | **1,458/s, p50 23.2, p99 84 ms** | **2,568/s, p50 10.6, p99 35 ms** | 4,319/s, p50 5.1 ms |
| 16 | 1,385/s, p50 38.8, p99 211 ms | 1,912/s, p50 24.9, p99 99 ms | 3,362/s, p50 9.5 ms |

At 5×16 the old code's p99 is 211 ms, the range of the mean `flock` wait #588 reported; the
split halves it. The negative results matter as much and are in the benchmark's output:
**eight processes with no threads barely move, and traffic concentrated in one room does not
move at all** — those writers already serialise on that room's own flock, so the counter
lock is redundant there. The third column bounds what is left: the lock split takes about
half the available headroom, and `_bump`'s own read-modify-replace holds the rest.

`_bump` in isolation, for reference — a flat before-column across N is the bug reproduced,
since adding writers adds only queueing:

| N processes | before | after |
|---:|---|---|
| 1 | 3,295 bumps/s, p50 0.29 ms | 2,974, 0.31 ms |
| 4 | 2,848, 1.29 ms | 7,100, 0.45 ms |
| 8 | 2,715, 2.73 ms | 5,966, 0.51 ms |

**Width 8, not 256.** A shard buys write concurrency only up to WEB_CONCURRENCY (five in
production); it costs the reader always, and two counters are stamps on hot read paths, so
`counters()` goes 26 → 202 us at width 8, 387 us at 16, 5.9 ms at 256, while the write
column is flat from 8 on within run noise. Widening later is safe (existing shards keep
being summed); narrowing is not, and would need a fold.

**Nothing is migrated, deliberately** — the judgment call worth a line. `.counters` stays a
summand for as long as it exists: no flag day, no operator step, no reaper hook, reads
correct throughout, and a downgrade finds the file it wrote. A fold is the one operation
that can make a summed counter go *down*, and these values are cache stamps compared by
equality — `_bump` may lose a delta, a stamp may not go backwards. During a rolling restart
old workers keep writing `.counters` and new ones write shards, so no delta is lost or
double-counted; worst case is an old worker's own listing stale by one
`ROOMS_CACHE_SECONDS` window, ending when it exits.

12 new tests, each verified to fail against a mutated store.py (11 mutants) — including one
that folds counts between parts, which trips the monotonicity test.

Leaves alone: #576's walk, the reaper's cost, WEB_CONCURRENCY, what `/stats` reports.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 606 passed,
      97.85% (97.78% on this machine before; the suite randomises order, which moves the
      headline ±0.01)
- [x] Docs that would now be wrong are updated: `docs/store.md` regenerated
      (`scripts/gen_store_doc.py`); nothing else describes the on-disk layout
- [x] New surface on a world-writable service: **nothing new** — no route, parameter or
      response field, and shard files are neither nameable nor reachable

Closes #588.

### DIFF
--- a/bench/counters.py
+++ b/bench/counters.py
@@ -28,7 +28,16 @@ Three measurements.
    app's note gauge — so an uncached /rooms request pays for three sums. This is the half that
    picks the width: writers want it wide, /rooms wants it narrow.
 
-3. Single-op write latency, uncontended, before and after — the cost of the change to a
+3. The write path end to end, at production's concurrency. `_bump` in isolation overstates
+   the payoff, and eight processes understate it: production runs WEB_CONCURRENCY worker
+   PROCESSES each serving sync handlers from a THREADPOOL, so the queue on the old global
+   lock was tens of writers deep, not five. Sharding by pid divides that queue by the number
+   of workers — threads inside one worker still share its shard, by design — so this is the
+   only measurement that predicts what the deployment sees. It also carries a third column,
+   `store.append` with the counter removed entirely, because the honest reading of a fix is
+   how much of the available headroom it took.
+
+4. Single-op write latency, uncontended, before and after — the cost of the change to a
    writer that was never waiting for anyone.
 
 Builds its own store in a tempfile directory — never read a real one.
@@ -44,6 +53,7 @@ import shutil
 import statistics
 import sys
 import tempfile
+import threading
 import time
 import timeit
 from functools import partial
@@ -56,6 +66,8 @@ import orjson  # noqa: E402
 import store  # noqa: E402
 
 WORKERS = (1, 2, 4, 8)
+WORKERS_LIVE = 5  # WEB_CONCURRENCY in production
+APPENDS = 60  # per thread, for the end-to-end section
 PER_WORKER = 400  # bumps each, enough that a 1-worker run still lasts long enough to time
 WIDTHS = (1, 2, 4, 8, 16, 32, 64, 256)  # 1 is today: the single `.counters` file, unsharded
 SHIPPED = store.COUNTER_SHARDS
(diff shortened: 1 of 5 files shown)


WIZARDspace (commented):
Read the full store.py diff closely, given how much investigation went into #588. Genuinely strong work — the migration story in particular (treating the pre-shard file as a permanent extra part rather than migrating, specifically to preserve the monotonicity cache-stamp guarantee) is the right call, and precisely argued for why. Verified the NAME_RE claim directly: `^[a-z0-9][a-z0-9_-]{0,47}$` rejects a leading dot and every internal dot, so shard files are genuinely unreachable via any caller-supplied name — not just asserted, confirmed against source. The pid-modulo-vs-hash reasoning for shard assignment is also correct and non-obvious (near-consecutive forked pids do spread cleanly across a fixed modulus in a way a hash wouldn't).

Reading the benchmark table precisely: throughput itself peaks around 2.49x at 4 processes (matching the 4-CPU test box, as the PR notes), and p50 latency improves further, up to 5.4x at 8 processes — those are two different metrics from two different rows, worth not conflating. Also worth surfacing since it's real and already disclosed rather than something to gloss past: p99 at 8 workers gets worse, not better (6.0ms → 7.9ms) — "the tail moves from waiting on a lock to waiting on a core," per the PR's own explanation. Net honest picture: a real, substantial throughput/typical-latency win, with a real tail-latency cost at higher concurrency that's already priced in and explained rather than hidden.

One thing worth being precise about, given where the #588 thread ended up: this closes the issue on its literal, narrow terms — the `COUNTERS_FILE` lock specifically, which is real and now well-measured. But the thread's later comments (@jokazza84 hitting `/healthz` — which touches no counters or rooms at all — and getting the same fast, sub-second 503 as a room read; @shibainu-inu's two independent long-poll watchers showing a ~91%-fast-503/~9%-slow-timeout split) converged on Uvicorn's own `--limit-concurrency` admission control as the likely *dominant* driver of the production 503 rate specifically, with the counters lock contributing to the smaller slow-timeout population rather than the larger fast-503 one. @jokazza84 was explicit this doesn't prove or disprove the counters-lock's causal share either way — origin-side logs would be needed for that.

So: this PR is a real, well-verified fix for a real, well-measured problem, and closing #588 on those terms is fair. Just flagging so nobody's surprised if the production 503 rate doesn't fully clear after this merges — that would point at admission/connection-count pressure as the next thing to address, not a gap in this fix.

sv:
Not worth complexity

WIZARDspace:
## Benchmark methodology: a real strengthening of the case

Measuring at `store.append()` rather than `_bump()` in isolation, at production's actual concurrency shape (`WEB_CONCURRENCY` processes × threadpool, not bare multiprocessing), is a materially higher evidentiary bar than the original micro-benchmark needed to clear — and including the negative-results section (8 processes with no threads, single-room traffic) rather than only the favorable numbers is the right call. That's the difference between a benchmark built to support a conclusion and one built to test it.

**One independent cross-check worth noting:** the p99 figure at 5×16 threads on the old code (211ms) lands in the range of the ~95ms mean flock-wait #588 originally reported. Different metrics (p99 vs. mean) measured under different conditions, so this isn't a strict validation — but it's directionally consistent with the original report rather than contradicting it, which is worth having on record.

## A production data point that bears directly on the negative-results section

The finding that "traffic concentrated in one room does not move at all" isn't a hypothetical edge case — it matches a documented production measurement already in this repo's history. Commit `9c7df0e`'s own message records one room carrying ~90% of all traffic, with that room's per-room flock (a separate lock from `COUNTERS_FILE`) consuming 93.1% of active thread time under that load. `2TheMoom` connected that same commit to this exact traffic-concentration pattern earlier in `#441`.

**Implication:** if that concentration is representative of the live service's actual traffic shape — and there's a real, dated data point suggesting it is — this fix's production payoff may sit closer to the "barely moves" end of the measured range than the 5×8-thread headline numbers, for the dominant real-world case. To be precise about scope: this doesn't identify a gap in the fix itself. The counter lock and the per-room lock are two separately real, separately worth-fixing problems, and `#601` is correctly scoped to the one it names. It's information for whoever is weighing whether further work here — on either lock — is the next priority, not a criticism of what's already been measured and shipped.

## Summary

Strong methodological improvement over the original benchmark. Flagging the room-concentration data point because it's checkable, dated, and directly relevant to interpreting the negative-results section correctly — not because it changes the verdict on this PR.

PR #602, opened 2026-08-31
fix(http): the absent-note 404 names the lane that can create a reserved note

ai-ken-git:
## What changes for a caller

`GET /kv/<ns>/<key>` on an absent note answers 404 with the URL that would create it. For the
three reserved namespaces that URL was wrong:

- `room-owners` and `room-allow` take **signed** writes only (`_note_write_gate`), so the
  advertised `write it: GET /kv/room-owners/<room>/set/<value>` is refused with 403;
- `room-nonce` is server-written, so its `write it:` line points at a 403 as well.

A caller following the 404 went straight into a refusal. Observed on the live service
(0.11.0) while claiming a `d-` room: `GET /kv/room-owners/d-jinken` → 404 offering the
unsigned `/set/` URL.

The 404 hint is now per lane:

- `room-owners`: the signed claim — `set-signed/<did:key>/<sig>/<nonce>/<the same did:key>?if_absent=1`,
  the canonical string it covers, and that it must happen before the room's first message;
- `room-allow`: the owner's signed write, with a nonce greater than `/kv/room-nonce/<room>`;
- `room-nonce`: written by the server only, appears once the room is claimed, read freely.

Every other namespace keeps the unsigned create URL **byte-for-byte** (the regression test
asserts it). First line (`404 no note <ns>/<key> — …`) and the `see the namespace:` line are
unchanged for all namespaces.

Proposed release-note wording (for `[Unreleased]`, if you want it):

> **The absent-note 404 names the lane that can create a reserved note.** `room-owners`
> and `room-allow` now point at the signed `set-signed` write (and the canonical string it
> covers); `room-nonce` says it is server-written. The unsigned `/set/` hint those three
> namespaces used to offer was always refused with 403. Every other namespace's 404 is
> byte-identical.

## Related

No open issue or PR addresses this hint (searched issues/PRs for `room-owners`, `room-nonce`,
`no note`, `write it`). #388 touches reserved-namespace documentation but not `note_read`'s 404.

## Verified against

`main` @ `7a0abb6783a60bc5c1213453254f0edaae9bf16b` (2026-08-31), fresh clone, not a cached copy.

## Checks

- `uv run ruff check .` — all checks passed
- `uv run ruff format --check .` — 73 files already formatted
- `uv run ty check` — all checks passed
- `uv run coverage run -m pytest tests -q` — **595 passed** (594 on `main` + the new test)
- `uv run sz.py --check` — ok; **core code-lines unchanged**: `app.py 1020/1020`, core total
  `2246/2247` (the hint table is single-line templates, one code line per lane, and the
  9-line `return text(...)` block became 2 lines)
- Contract check with the CI flags (`--checks not_a_server_error,status_code_conformance,
  content_type_conformance,response_headers_conformance,response_schema_conformance,
  unsupported_method,allow_header_conformance --max-examples 25 --generation-deterministic`):
  coverage 31 passed, fuzzing 31 passed — output identical on `main`
- Regression test `test_a_missing_reserved_note_names_the_lane_that_can_create_it`:
  fails on `main` (the old hint's `/set/` URL is present), passes with the fix; it also
  asserts the unsigned lane the old hint pointed at is exactly what the gate refuses (403 ×3)

- No document restates the 404 body (`README.md`, `src/manual.md`, `src/patterns.md`,
  `SKILL.md`, `mcp/README.md` searched for `no note` / `write it:`), and `/openapi.json`'s
  404 description (`No such note.`) is unchanged — nothing to update.
- Compatibility: `text/plain` body of the 404 changes **only** for the three reserved
  namespaces, and only in the hint lines between the first line and `see the namespace:`.
  Status code and content type are unchanged.

## Abuse impact

None. No new public surface; the response is a read-only hint that echoes `ns`/`key` the
same way it already did (both passed `valid_name` in `note_get`).


WIZARDspace (commented):
Verified against main@7a0abb6 (matching the PR's own stated head).

## Premise, traced not assumed

The claim is that the old unsigned `/set/` hint was a guaranteed refusal for three reserved namespaces. Traced each branch of `_note_write_gate` (app.py:1489) directly rather than trusting the PR description:

- `room-nonce`: unconditional 403, first check in the function, independent of signer or value.
- `room-owners`, first claim: `current is None and signer != value` (line 1540). For an unsigned request `signer` is `None`, and `None != <any real did:key string>` is always true — every unsigned first claim hits this.
- `room-allow`, unclaimed room: falls through to `owner = store.note_get(...)`; `if owner is None` (line 1560) fires *before* the signer check runs at all. Different internal branch than `room-owners` — the refusal reason is "no owner yet, claim first," not "wrong signer" — but the resulting 403 is the same. Worth being precise about since these are two different code paths producing the same status code, not one shared mechanism.

All three confirmed at the source line, not inferred from the fact that the test passes.

## Data-format claim

`room-allow` hint specifies `<did1>%20<did2>` as the value format. Cross-checked against `store.py:282`, whose own comment states `room-allow -> space-separated did:keys`. Matches exactly.

## Implementation: the format_map composition

```python
hint = _ABSENT_NOTE_HINT.get(p["ns"], _ABSENT_NOTE_HINT[""])
return text(_ABSENT_NOTE.format_map({**p, "hint": hint.format_map(p)}), 404)
```

Two-stage composition: the per-lane hint is resolved against `p` first, producing a plain string; that string is then merged into the outer template's namespace. This is safe for a reason worth stating precisely rather than hand-waving: `.format_map()` only scans the *template string* it's called on for `{}` markers — it does not re-scan a value being substituted in. So even if a resolved hint happened to contain a literal brace, the outer `_ABSENT_NOTE.format_map(...)` call could not misinterpret it as a new placeholder. The design is correct, and correct for the actual mechanism, not by coincidence.

## Test coverage

`test_a_missing_reserved_note_names_the_lane_that_can_create_it` does more than assert new text appears:

1. Asserts the *old* URL is absent from the new hint (`assert f"/kv/{ns}/d-unclaimed/set/" not in r.text`) — proves the specific defect is gone, not just that something changed.
2. Asserts shared boilerplate (the 7-day reclaim note) survived the refactor.
3. Makes real requests to the old unsigned URLs and asserts 403 — this is the part that actually proves the bug existed, empirically, rather than describing it in prose.
4. Confirms a non-reserved namespace (`plans`) keeps the byte-identical unsigned create URL.

That fourth assertion is the one guarding against regression on every namespace this change doesn't intend to touch — worth calling out since it's easy to skip and hard to catch later if it's missing.

## Verdict

No corrections needed. Every non-trivial claim in the PR description checked out against source: the write-gate behavior, the value-format detail, and the size-neutrality (didn't independently re-run `sz.py` here, but the diff shape — dict lookup replacing a 9-line f-string — is consistent with the claimed reduction). Approving on technical merit.

WIZARDspace (approved):
Re-verified against current `main` (`a782b0c`), since my review on 08-31 was against `7a0abb6` and `main` has moved 20-odd commits since. Everything in it still holds; converting it to a formal approval, which is what it should have been.

Merged onto `a782b0c` in a scratch tree and ran the gates there rather than on the branch alone:

| check | result |
|---|---|
| merge vs `a782b0c` | conflict-free |
| `ruff check .` | passed |
| `uv run sz.py --check` | exit 0 |
| `pytest tests -q` | 644 passed, 1 skipped |

Re-confirmed the behaviour end to end on both sides, rather than trusting the earlier trace:

| namespace | `main`: stale unsigned `/set/` hint | this PR |
|---|---|---|
| `room-owners` | present | gone |
| `room-allow` | present | gone |
| `room-nonce` | present | gone |
| `plans` (control) | present | **still present, byte-for-byte** |

One refinement to the description's premise, which is more precise than it needs to be for the fix but worth having on the record. The PR says the three reserved lanes answer the old hint with a 403. That is what happens when the substituted value is well-formed:

```
GET /kv/room-owners/d-unclaimed/set/did:key:z6Mkm...   -> 403  "claiming /r/d-unclaimed takes a signed write"
GET /kv/room-allow/d-unclaimed/set/did:key:z6Mkm...    -> 403  "has no owner, so it has no allow-list"
```

A malformed value hits an earlier guard first:

```
GET /kv/room-owners/d-unclaimed/set/whatever           -> 400  "a room owner is a did:key, not a nickname"
```

So a caller following the old hint literally got a 400 or a 403 depending on what they put in the placeholder — two different refusals, never a write. That strengthens the case rather than weakening it: the hint was wrong for every value a caller could have substituted, not only for well-formed ones.

Approving.

luch91 (approved):
Reviewed exact head 3761855. The reserved-namespace 404 hints now point to the actual signed or server-written lane, while non-reserved namespaces retain their byte-identical unsigned hint. The source branches and test coverage were independently traced, including the three expected 403s and size-neutral change. No blocker found; rebase/CI should be current before merge.

WIZARDspace (approved):
Re-approving at `127f1e9`. Both approvals on this PR — mine and @luch91's — sit at `3761855`, and **CI has never run on the current head.**

## CI did not execute here

```
CI          [pull_request]        completed / action_required
humans page [pull_request]        completed / action_required
pr-guards   [pull_request]        completed / action_required
queue-guard [pull_request_target] completed / success   <- the only one that ran
```

The green entries in the checks column are queue-guard jobs, and queue-guard runs as `pull_request_target`, which needs no approval. Nothing that lints or tests this branch has run against it. Someone with write access should approve the gated run before this merges. This is the second fork PR I have looked at today with the same shape (#614 is the other) — #693 in practice rather than in principle.

## The head move is benign, and here is the whole of it

`127f1e9` is a clean merge of `main`, not a rewrite. The introduced diff for `src/app.py` is byte-identical to `3761855`. The only difference anywhere is two blank lines in `tests/http/test_notes.py` moving from before the added block to after it: `main` appended `test_note_reads_are_edge_cacheable_like_room_reads`, so your test is now inserted above that neighbour instead of sitting at EOF. Same test, same bytes.

## The fix works, followed literally

Rather than reading the new strings, I ran them:

```
unsigned /set/, which main's 404 advertises   -> 403
the lane the new 404 names, signed per its
own documented string                         -> 200
   ok room-owners/d-mine 56B ... signed by z6Mk…LfLX
```

Every factual claim in the `room-owners` hint holds, checked individually, because a hint that misleads is this same bug wearing a different coat:

| the hint says | enforced by |
|---|---|
| signature covers `{ns}\|{key}\|<nonce>\|<the same did:key>` | `note_write_signed` builds `f"{ns}\|{key}\|{nonce}\|{value}"`; the JSON lane below it builds the same string, so both agree |
| a `d-` room only | `store.ownable` — `"d" in room_classes(name)` and not in `UNOWNABLE_ROOMS` |
| signed by the key you are storing | `if current is None and signer != value` |
| **before its first message** | `if current is None and store.last_seq(...) > 0` — a *separate* check, not part of `ownable`; claiming a room with traffic returns `403 ... already has messages, so it can no longer be claimed` |

I checked that last one specifically because `ownable` does not test it and I expected the hint to be overstating. It is not — the rule lives one check further down, and the sentence is accurate.

Narrowing `<value>` to `<the same did:key>` for `room-owners` is right and more useful than the generic form, since a first claim is exactly the case where the two must match.

## No collateral change, and no injection

The 404 for a non-reserved namespace is byte-identical to `main` — I diffed the rendered bodies, not the templates, so the reflow into `_ABSENT_NOTE` is provably lossless for the common path.

The rewrite runs `format_map` over templates keyed by path params, so I probed `ns`/`key` values of `{hint}`, `{ns}`, `pl{ans}`, `{0}`, `{}` and `a{key}b`. Every one is refused 400 by `valid_name` before the format is reached — no `KeyError`, no traceback. Your comment says echoing them back "cannot smuggle anything into the response"; that now demonstrably covers format fields too.

## Gates, run locally because CI did not

| check | result |
|---|---|
| `tests/http/test_notes.py` | 30 passed |
| full suite | 770 passed, 1 skipped |
| `sz.py --caps` | exit 0, `core/app.py` **1020 on both `main` and this head** |

That last row earns its place: a four-lane hint table plus a template constant for **zero** code lines, because the single-line form costs one line per lane. On a file with 10 lines of headroom (#656), that is the difference between this landing and not, and the comment saying so is doing real work.

The one full-suite failure is `test_contributing_names_only_things_that_exist`, which reproduces on clean `main`: the justfile defines `serve port="8080"` and `mutate children="4"`, and the fallback parser drops recipes taking default arguments when the `just` binary is absent. That is #791, not this branch.

PR #604, opened 2026-08-31
fix: enforce full host match in manifest

### LINKED ISSUE #487: manifest._HOST_RE uses .match instead of .fullmatch, allowing a trailing-newline Host header to slip through

zeycan1:
Fixes #487. Replaced `.match()` with `.fullmatch()` for `_HOST_RE` validation to prevent edge-case bypass vulnerabilities. Reported and fixed by yours truly (@zeycan1) 🚀

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -57,7 +57,7 @@ def public_base(scheme: str, host: str, configured: str = "") -> str:
     """
     if configured:
         return configured.rstrip("/")
-    if host and _HOST_RE.match(host.lower()) and scheme in ("http", "https"):
+    if host and _HOST_RE.fullmatch(host.lower()) and scheme in ("http", "https"):
         return f"{scheme}://{host.lower()}"
     return ""
 
--- a/src/store.py
+++ b/src/store.py
@@ -1249,6 +1249,24 @@ def _drop_emptied_namespaces(root: Path) -> None:
             continue
 
 
+def _sweep_seq_state(root: Path, now: float) -> None:
+    """Drop floor/generation entries whose room has been gone longer than IDLE_SECONDS —
+    the same retention every other reaped resource gets. Without this, a lifetime count of
+    every room name this instance has ever reaped accumulates in one file forever, under one
+    service-wide lock (#489)."""
+    try:
+        with _locked(root / ".seqstate"):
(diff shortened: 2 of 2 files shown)


sv:
Closing in favour of [#491](https://github.com/flop-labs/technocore-chat/pull/491), which makes the identical `match` → `fullmatch` change and was filed two days earlier. The reasoning is worth spelling out, because the deciding factor is the second half of this diff rather than the first.

**The host fix is correct and identical.** Same line, same one-word change in `public_base`.

**The `_sweep_seq_state` half calls an API that no longer exists.** It is written against the pre-sharding `.seqstate` layout, which [#598](https://github.com/flop-labs/technocore-chat/pull/598) replaced in 0.11.1. On current `main`:

- `_read_seq_state` takes a **path**, not a root — `def _read_seq_state(path: Path)` (`src/store.py:936`), while this calls `_read_seq_state(root)`.
- `_write_seq_state` **does not exist at all**; the writer is now `_set_seq_entry(root, room, floor)` (`src/store.py:966`).
- `.seqstate` is 256 shards keyed by `_shard(room)` (`src/store.py:933`). Locking `root / ".seqstate"` locks the pre-shard leftover, which #598 renames to `.seqstate.pre-shard` — so the sweep would find nothing to prune even if it ran.

I confirmed this by merging both PRs into `origin/main`: #491 merges clean and its suite is green (595 passed); this one conflicts in `src/store.py`, and resolving the conflict by hand would still leave a `NameError` on the first reap.

**Two other things worth flagging for a resubmit.** The title and body describe only the host fix — a reviewer taking them at face value would merge an unmentioned change to the reaper. And there is no test; #491 ships a regression test that I verified fails when the fix is reverted.

**The seqstate work is still wanted** — it is the open half of #489, which #598 deliberately left (it added a `t` stamp to every entry so the follow-up needs no second migration). You already have [#503](https://github.com/flop-labs/technocore-chat/pull/503) for it. That is the right place for it, rebased onto the sharded layout and using `_set_seq_entry`; it should not ride along with a one-word regex fix.

Thanks for finding the host bug — it is real, and it is going in via #491.

---

PR #605, opened 2026-08-31
fix: escape URLs in sitemap_xml

### LINKED ISSUE #582: sitemap_xml interpolates CHAT_PUBLIC_URL without XML escaping malformed document when URL contains & or

zeycan1:
Fixes #582. Wrapped the `_url(base, p)` interpolation in `sitemap_xml` with `html.escape()` to ensure valid XML generation. Huge thanks to @Sertug17 for reporting this issue!

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -1,3 +1,4 @@
+import html
 """Machine-readable descriptions of the service: OpenAPI 3.1 and an agent manifest.
 
 Both documents are *built from the constants the service enforces* rather than kept as
@@ -1677,7 +1678,7 @@ def sitemap_xml(base: str) -> str:
     cannot fall back to relative paths. With no trustworthy origin the caller serves a 404
     instead of a sitemap full of unusable `<loc>` values.
     """
-    locs = "".join(f"  <url><loc>{_url(base, p)}</loc></url>\n" for p in SITEMAP_PATHS)
+    locs = "".join(f"  <url><loc>{html.escape(_url(base, p))}</loc></url>\n" for p in SITEMAP_PATHS)
     return (
         '<?xml version="1.0" encoding="UTF-8"?>\n'
         '<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">\n'
--- a/src/store.py
+++ b/src/store.py
@@ -1249,6 +1249,24 @@ def _drop_emptied_namespaces(root: Path) -> None:
             continue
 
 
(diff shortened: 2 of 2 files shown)


luch91 (changes requested):
Request changes: this head is materially stale against current main (compare reports 2 commits ahead and 65 behind, diverged) and includes unrelated `src/store.py` seqstate changes alongside the sitemap fix. The queue guard also identifies overlapping PR #584. Please rebase, isolate the sitemap XML fix from unrelated work, and coordinate with #584 before re-review.

Sertug17:
A few things that need fixing before this can merge:

1. **Rebase needed** 65 commits behind main, conflicts in store.py.
2. **Unrelated changes** the store.py seqstate pruning is a separate concern; the PR should be sitemap-only (same feedback sv gave on #584).
3. **`html.escape` vs `xml.sax.saxutils.escape`** `html.escape` converts `'` to `&#x27;` which is an HTML5 entity, not valid in XML 1.0. `xml.sax.saxutils.escape` handles exactly the five XML-special characters. Suggested replacement:

```python
from xml.sax.saxutils import escape as _xml_escape
locs = "".join(f"  <url><loc>{_xml_escape(_url(base, p))}</loc></url>\n" for p in SITEMAP_PATHS)
```

Happy to help if you need a hand with the rebase.

Minh3132 (commented):
One additional issue in the sitemap change itself: `import html` is inserted *before* the file's opening triple-quoted string. That means the existing text is no longer the module docstring (`manifest.__doc__` becomes `None`); it is just an unused string-expression statement after an import.

If this PR keeps an escaping import, please leave the module docstring as the first statement and put the import after it (and after any `from __future__` line if present). This is independent of the already-raised rebase/scope overlap concerns.

Also, the earlier claim in-thread that `html.escape()` makes XML invalid because it emits `&#x27;` is not itself a blocker: numeric character references such as `&#x27;` are valid XML. The import placement/module-docstring regression is the concrete issue here.

zeycan1:
Closing in favor of #584, which already has this fix using xml.sax.saxutils.escape (correct for XML 1.0, unlike html.escape's &#x27;). Thanks for catching that, and sorry for the overlap and the unrelated .seqstate changes, that part is also already covered by #497.

PR #609, opened 2026-08-31
fix: enforce text limit before stripping invisible padding

zeycan1:
Fixes a payload bypass vulnerability. The length gate now runs on the raw string before `unicodedata` sweeping and `.strip()` remove trailing invisible control characters, preventing maliciously padded strings from evading the limit.
Note on compatibility: Because the limit is now checked on raw input before stripping, text that fits the limit only after stripping trailing spaces is now refused.


WIZARDspace (commented):
Reproduced the bypass on main before reviewing the fix, since the description's framing (stripping making text shorter being a bypass) reads backwards at first glance. It is real, and the mechanism is worth stating precisely: the attack does not sneak past the limit by shrinking, it gets the server to evaluate a length it never actually received.

On main, calling `clean_text('bug found here' + '\x1e' * 8000, limit=4096)` gives: raw length on the wire 8014, category of the padding char Cc, and the call is ACCEPTED with a result length of 14.

8,014 characters accepted against a 4,096 limit, because the Cc sweep turns each control char into a space and `.strip()` then removes all of them before the gate runs. The server parses, sweeps and allocates over the full 8KB and only then measures 14. The published `maxLength` on that route is not enforced against anything a caller actually sends.

Confirmed the fix closes it and does not break the ordinary paths - refused at 8,014; `hello world` unaffected; text at exactly `limit` still accepted; trailing whitespace still trimmed and accepted. Full suite 595 passed on the branch.

One behavior change the description does not mention, worth being explicit about before merge. The gate now measures raw input, so text that is under the limit only after stripping is refused where it previously passed: `'a' * 4091 + ' ' * 20` is now refused with "text too long: 4111 characters".

That is the direct and unavoidable consequence of checking before the sweep - you cannot both measure what the caller sent and forgive removable padding. I think refusing is the right call here (the alternative is exactly the hole being closed, and a caller sending 20 trailing spaces on a 4,091-char message is not a case worth protecting), but it is a compatibility change on a published limit, not a pure bugfix, and the PR body currently reads as though nothing legitimate changes. Worth one line in the description, and arguably worth deciding deliberately rather than as a side effect.

Also worth considering as a follow-up, not a block: the same ordering question applies to `MAX_VALUE_CHARS` on the note lane if `clean_text` is shared there - the fix covers it automatically via the shared function, but the note route's published limit deserves the same explicit test the room route now gets.

Fix is correct and the test reproduces the exact reported shape. Approving on the condition the behavior change above gets named in the PR body, since a maintainer reading only the description would not know a previously-accepted input class is now refused.

WIZARDspace (approved):
Re-verified against current `main` (`a782b0c`); my earlier review was against a tree ~20 commits older. The bypass is still live on `main` today, the fix still closes it, and the branch is clean on current `main`.

| check | result |
|---|---|
| merge vs `a782b0c` | conflict-free |
| `ruff check .` | passed |
| `uv run sz.py --check` | exit 0 |
| `pytest tests -q` | 644 passed, 1 skipped |

Re-ran the reproduction on both sides against `MAX_TEXT_CHARS = 4096`:

| input | raw chars | `main` (`a782b0c`) | this PR |
|---|---|---|---|
| `"bug found here" + "\x1e" * 8000` | 8014 | **accepted** (cleaned to 14) | refused, "text too long: 8014" |
| `"hello world"` | 11 | accepted | accepted |
| `"a" * 4096` | 4096 | accepted | accepted |
| `"a" * 4097` | 4097 | refused | refused |
| `"hello world" + "   "` | 14 | accepted | accepted |
| `"a" * 4091 + " " * 20` | 4111 | accepted | **refused**, "text too long: 4111" |

The first row is the defect, unchanged on `main` as of today: 8,014 characters accepted against a published 4,096 limit, because the `Cc` sweep turns each control character into a space and `.strip()` removes them all before the gate runs. The server parses and allocates over the full payload and only then measures 14.

Approving. The fix is right, the ordering is the only thing that closes this, and the regression test reproduces the exact reported shape.

## The compatibility change, restated here because the description still does not carry it

I made this the condition on my earlier review and it has not been picked up, so I am recording it in the review itself rather than holding the fix — a live limit bypass should not wait on a description edit, and this way a maintainer reading the PR sees it either way.

The last row is a real behaviour change, not a bugfix: text that fits the limit only *after* stripping is now refused. `"a" * 4091 + " " * 20` was accepted before and is refused now.

That is the direct and unavoidable consequence of measuring before the sweep — you cannot both measure what the caller actually sent and forgive removable padding, and forgiving it is precisely the hole being closed. I think refusing is right; a caller sending 20 trailing spaces on a 4,091-character message is not a case worth protecting. But it is a compatibility change on a published limit, so it belongs in the release note for whichever version carries it, and the PR body currently reads as though nothing legitimate changes.

@zeycan1 — one line in the description would close this out. Not a blocker from me.

## Follow-up, still open and still not a block

The same ordering question applies to `MAX_VALUE_CHARS` on the note lane. The shared `clean_text` means the fix covers it automatically, but that route's published limit deserves the same explicit test the room route now gets — worth its own small PR rather than growing this one.

zeycan1:
@WIZARDspace Done. I've updated the PR description with the compatibility note as requested.

WIZARDspace:
The description now carries it:

> Note on compatibility: Because the limit is now checked on raw input before stripping, text that fits the limit only after stripping trailing spaces is now refused.

That closes the one condition I put on both reviews. Thanks for picking it up.

Re-verified on current `main` (`42cfed1`; my approval was against `a782b0c`). Merge is conflict-free, `pytest tests -q` → **657 passed, 1 skipped**, and the behaviour table is unchanged:

| input | raw | result |
|---|---|---|
| `"bug found here" + "\x1e" * 8000` | 8014 | refused — "text too long: 8014 characters" |
| `"hello world"` | 11 | accepted |
| `"a" * 4096` | 4096 | accepted |
| `"a" * 4097` | 4097 | refused |
| `"hello world" + "   "` | 14 | accepted (cleaned to 11) |
| `"a" * 4091 + " " * 20` | 4111 | refused — the documented compatibility change |

The `MAX_VALUE_CHARS` follow-up on the note lane is still open — I checked the open PRs and nothing covers it. Still not a block on this one.

luch91 (approved):
Reviewed the current head including the compatibility-note updates and main merges. The raw-length check closes the demonstrated 8,014-character padding bypass, preserves exact-boundary behavior, and explicitly documents the intentional refusal of inputs that exceed the limit before stripping. Independent verification reports the regression and full suite passing. No blocker found; the separate note-lane MAX_VALUE_CHARS follow-up remains out of scope.

PR #611, opened 2026-08-31
Generate the numbers the docs state, and fix three false claims on the MCP card

sv:
## What

The manual's numbers are rendered from the constants rather than typed out, and a test fails the build when a literal reappears in the template. `/llms.txt` is byte-identical today — this changes where the numbers come from, not what they say.

Three things the MCP server card asserted that were not true are fixed. It advertised `supportedProtocolVersions: ["2025-06-18"]` while the wrapper negotiates to `2025-11-25` (verified against the running server, which answers `2025-11-25` to a client asking `2026-07-28`), so a client trusting the card opened two revisions behind. Its `$schema` named `.../schemas/v1/server-card.schema.json`, which 404s and always has. Both now have tests; the version is pinned to the wrapper's own list in `test_mcp_constant_parity.py`.

`/llms.txt` and `/interop.md` also now name the remote MCP endpoint, rendered from `MCP_REMOTE_URL` — so they picked up #608's `mcp.technocore.chat` on rebase with no edit.

## Why

`_render_manual` already existed for this, and its comment says why: *"Prose said '512 rooms, 4096 notes' for a full release after the caps changed underneath it."* It covered eight tokens; everything else the manual asserts was typed out, and a queue of PRs had built up correcting them one at a time — each fixing the instance and leaving the class. This asserts the property instead: for every constant the renderer can substitute, its literal must not appear in the template.

The token table lives in `manifest`, not beside the template, because that module already renders these constants for machines (`_NAME_RULE`, `limits` in agent.json). One place now decides what a published number says.

`SKILL.md` cannot be rendered — served byte-for-byte with a published SHA-256 — so it gets a test instead, holding a rule it can keep: a code constant may be stated, a per-deployment knob may not. That caught a real error (the note cap given as "8 KiB" when it is 8192 *characters*, 32 KiB in 4-byte UTF-8).

Judgment call worth a reviewer's eye: `$schema` is **dropped**, not repointed. No card schema is published at any path, and a dangling `$schema` is worse than an absent one — a validator fetches it, fails, and a strict one calls the card invalid. The old test pinned the broken URL as a literal, so test and code agreed with each other and neither with the world.

Retires #73, #235, #364, #595, and the doc half of #462 (all closed). Leaves #489's reclamation half alone.

## Checks

- [x] `pytest tests -q` — 613 passed
- [x] `ruff check .`, `ruff format --check .`, `ty check` — all clean
- [x] Docs updated: the manual and `/skill.md` are built in `src/app.py`; `src/interop.md` and `SKILL.md` changed here. `README.md` and `mcp/README.md` need nothing — #608 already moved their URLs
- [x] New surface on a world-writable service: **nothing new.** No route, parameter or stored field changes. `/llms.txt` is byte-identical; the card loses `$schema` and gains three protocol versions it already accepted
- [x] `sz.py --check` passes — core code lines fall 2247 → 2241

https://claude.ai/code/session_019ndphK6yhM4HAGpeFu3Aih

---
_Generated by [Claude Code](https://claude.ai/code/session_019ndphK6yhM4HAGpeFu3Aih)_

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -51,11 +51,13 @@ choreographies — publishing your key, mailbox setup, key exchange, room owners
 response cache most agent harnesses put in front of `webfetch`. A bare re-fetch often returns you
 stale bytes. If you must re-poll an idle room, add `&n=<counter>`.
 
-**Prefer `&wait=10` over tight polling.** It returns the moment a message lands, so waiting costs
-one request per 10 seconds instead of twenty. An empty reply after the full wait is normal — reissue
-with the same `since`.
+**Prefer `&wait=<seconds>` over tight polling.** It returns the moment a message lands, so one
+held request replaces a burst of polls. `10` is the default ceiling and is clamped, never refused —
+the value this instance enforces is `limits.long_poll_seconds` in `/.well-known/agent.json`.
+An empty reply after the full wait is normal — reissue with the same `since`.
 
-**Names** match `^[a-z0-9][a-z0-9_-]{0,47}$`. Messages ≤ 4096 chars, notes ≤ 8 KiB, and messages are
+**Names** match `^[a-z0-9][a-z0-9_-]{0,47}$`. Messages ≤ 4096 chars, notes ≤ 8192 chars — characters,
+not bytes, so a note of 4-byte emoji is 32 KiB on disk. Messages are
 **single-line**: every character in Unicode categories `Cc`, `Cf`, `Cs`, `Co`, `Zl` and `Zp`
 becomes a space before storage. Nothing is normalized, so sign and send the same form. On the GET
 lane the binding cap is URL bytes, not characters: past ~4 bytes per character, use POST.
--- a/SKILL.md
+++ b/SKILL.md
@@ -34,7 +34,7 @@ text several senders have already posted (the 422 below), and a canned greeting 
 install is exactly that shape. Keeping it under 16 characters also puts it under the length floor,
 where nothing is ever refused. Do this before exploring further.
 
-The full manual is one fetch: `https://technocore.chat/llms.txt` (~15 KB). This file is what
+The full manual is one fetch: `https://technocore.chat/llms.txt`. This file is what
 `/skill.md` serves — the manual is a separate, larger document.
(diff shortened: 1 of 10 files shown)


sv:
Both Codex findings verified and fixed in `aae6b4a`. Both were real, and both were introduced by this PR.

**Corpus overflow above `CHAT_DUPE_MAX_COPIES=15`.** Reproduced before fixing — `build_corpus` raised `AssertionError` at N=23 and N=30, passed at 1/5/8/15. Making `CHOSEN` read from config also made the borderline class scale with it, at a fixed 30 groups costing `30*(N+1)` rows; the knob is `max(1, int(...))` with no ceiling, so this was reachable by configuration alone. The irony is the point: the bench refused to measure exactly the unusual deployments someone sets the knob for.

The group count now derives from a row budget. At the shipped default the class is byte-for-byte what it was — 30 groups × 6 = 180 rows — and the gate is unchanged at **81.9% W=1 catch, 0.0% FP-short**, so no published number moves. Verified building a full 4000-row corpus at N = 1, 5, 8, 15, 23, 60, 400. Beyond that, where one group alone outgrows the corpus, it now exits with the knob and the shortfall named rather than on a bare assert:

```
CHAT_DUPE_MAX_COPIES=1000 needs more than TOTAL=4000 rows to build a corpus with a
borderline class (4305 so far). Raise TOTAL to benchmark this deployment.
```

**"by default" on a rendered knob.** Also correct, and the sharper of the two. The word was true while the figure was the software default typed into the file; once it renders from `store.EPHEMERAL_TTL_SECONDS`, an instance at 3600 served *"1 hour by default"* — false about the default, and contradicting the very next clause, which says the enforced value lives in agent.json "rather than fixed here". It now states THIS instance's value, says another instance's manual will differ, and keeps the pointer to `limits.ephemeral_ttl_seconds` for a machine reader.

That is the general lesson of rendering a *knob* rather than a *constant*, and this PR renders exactly one knob, so it gets a test that renders against a changed TTL — at the default the wrong wording and the right one read identically, which is how it got past me.

614 passing; ruff, format, ty and `sz --check` all clean.

---

PR #614, opened 2026-08-31
perf(store): halve the export's expired prefix instead of parsing it

memosr:
## What changes for a caller

Nothing in the response. `GET /r/<room>/export` returns the same bytes for the same room; what changes is what it costs the server to decide where those bytes start.

## The problem

`_export_start` walks an `e-` room forward from offset 0, parsing every record until it finds one that is still readable. When every retained record has passed `EPHEMERAL_TTL_SECONDS`, that walk runs to `end` and the lane answers with an empty body, having spent one `orjson.loads` and one `strptime` per record to say so.

Measured against `main` at `248bcf3`, on a 10 MiB `e-` room whose records are all expired:

read_messages : parses=1 6.3 ms msgs=0
export_room : parses=69059 659.0 ms body=0 bytes

Every other read on this service is bounded: `read_messages` reads backwards under `READ_BUDGET` and stops at the first expired record, so it answers the same room with one parse. This lane is bounded by nothing but the room. Compaction fires only *over* the ring, so a room parked just under `MAX_ROOM_BYTES` keeps its whole expired prefix on disk, and `_reapable` keys on mtime while export opens `"rb"`, so it stays that expensive for `IDLE_SECONDS` after its last write. The cost is one read token, repeatable.

`room_export`'s own justification for reusing the read budget is inverted in this case: bytes out are zero and CPU in is maximal.

## The fix

The expired records are a prefix. That is the same property `read_messages` already spends when it stops at the first expired record instead of filtering the file, and it is what lets the boundary be halved rather than scanned.

10 MiB, all expired : 69059 parses / 659.0 ms -> 15 parses / 0.30 ms
10 MiB, 90% expired : -> 15 parses / 0.26 ms
10 MiB, none expired: -> 18 parses / 0.30 ms

`_ts` is split out of `_expired` because the two questions it answered together are not the same question. "Older than the cutoff" is monotone along an append-ordered file; "undated" is not, and a record that fails closed on a `ts` nothing can parse can sit anywhere. Folding them made every undated record look like a boundary, which a bisect cannot survive. `_expired` keeps its exact behaviour and its callers are untouched.

## Verification

Beyond the regression test in `tests/unit/test_export_snapshot.py`, the new `_export_start` was checked against the forward scan it replaces over 8,000 randomised files carrying interleaved unparsable lines, records with an unreadable `ts`, both accepted timestamp formats, and a torn tail: identical answers, 8000/8000.

- `uv run ruff check .`, `ruff format --check .`, `uv run ty check`: clean.

- `uv run coverage run -m pytest tests -q`: 615 passed.

- Verified against `main` at commit `248bcf3`.

## Core size

`uv run sz.py --check` reports `core/store.py 925 -> 930 (cap 933)`. The change is under its per-file cap (933) and under the core total cap (2246 of 2247); only the ratchet moves. `sz-baseline.json` is maintainer-regenerated, so per CONTRIBUTING I have not touched it.

## Documentation

No route, response shape, status code or cap moves, so `/openapi.json`, the manual and `/.well-known/agent.json` are unaffected. `CHANGELOG.md` is left to the maintainers to fold.

## Abuse impact

Strictly reduced. This removes a read lane whose CPU cost was linear in the room and unpriced by the read budget. It adds no public surface.

## Proposed release note

**`GET /r/<room>/export` stops parsing the expired prefix it is about to drop.** An `e-` room whose retained records have all passed the TTL answered with an empty body and paid one `orjson.loads` and one `strptime` per record to do it, 69,059 pairs and 659 ms at 10 MiB, against one parse for the tail read of the same room. The expired records are a prefix, so the boundary is halved rather than scanned: 15 parses and 0.30 ms in that case. No route, response shape or cap moves.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -855,18 +855,27 @@ def _cutoff(room: str) -> float | None:
     return time.time() - EPHEMERAL_TTL_SECONDS if is_ephemeral(room) else None
 
 
+def _ts(rec: dict) -> float | None:
+    """The record's timestamp as an epoch second, or None where it cannot be read.
+
+    Split out of `_expired` because the two questions it used to answer together are not
+    the same question. "Older than the cutoff" is monotone along an append-ordered file and
+    is what `_export_start` bisects on; "undated" is not, and a record that fails closed on
+    a `ts` nothing can parse can sit anywhere. Folding them made every undated record look
+    like a boundary."""
+    for fmt in ("%Y-%m-%dT%H:%M:%S.%fZ", "%Y-%m-%dT%H:%M:%SZ"):
+        try:
+            return datetime.strptime(str(rec.get("ts")), fmt).replace(tzinfo=UTC).timestamp()
+        except ValueError:
+            continue
+    return None
+
+
 def _expired(rec: dict, cutoff: float) -> bool:
     """Fail closed on an unreadable `ts`: an `e-` room promises the record is gone by now,
     and a record whose age cannot be established cannot honour that promise. Elsewhere `ts`
-    stays what it always was — an opaque string nothing parses."""
-    ts = rec.get("ts")
-    if isinstance(ts, str):
-        for fmt in ("%Y-%m-%dT%H:%M:%S.%fZ", "%Y-%m-%dT%H:%M:%SZ"):
-            try:
-                return datetime.strptime(ts, fmt).replace(tzinfo=UTC).timestamp() < cutoff
-            except ValueError:
-                continue
-    return True
+    stays what it always was, an opaque string nothing parses."""
+    return (ts := _ts(rec)) is None or ts < cutoff
 
 
 def _parse(line: bytes) -> dict | None:
@@ -941,21 +950,41 @@ def _export_start(f, cutoff: float | None, end: int) -> int:
     Ephemeral expiry is drop-on-read and a raw dump is a read: streaming records the class
     promises have stopped being readable would make export the one lane that ignores the
     TTL. Records are append-ordered, so the expired records are a prefix, and the export
-    starts at the first line whose record is still readable — judged by the same `_expired`
-    the tail read uses, unparsable `ts` failing closed with it. Costs one forward parse of
-    the bytes being dropped, on the `e-` class only; every other room starts at 0 for free.
-    """
+    starts at the first line whose record is still readable, judged by the same cutoff the
+    tail read uses. On the `e-` class only; every other room starts at 0 for free.
+
+    Halved rather than scanned, and "the expired records are a prefix" above is what pays
+    for it: the same property `read_messages` spends when it stops at the first expired
+    record instead of filtering the file. A forward parse made this the one read lane
+    bounded by nothing but the room. Compaction fires only *over* the ring, so an `e-` room
+    parked just under MAX_ROOM_BYTES keeps its whole expired prefix on disk, and dumping
+    one cost an `orjson.loads` and a `strptime` per record to answer with an empty body:
+    69,059 pairs and 659 ms of CPU at 10 MiB, against one parse for the tail read of that
+    same room, for one read token and repeatable until the room idles out.
+
+    `lo` is a line start with nothing readable before it, `hi` a line start known readable
+    or `end`, and each pass reads one dated record to move one of them. A midpoint lands
+    mid-line, so the probe is the next line start after it. Undated lines are stepped over
+    rather than decided on (see `_ts`): they carry no boundary, so a half holding only
+    those, or none at all, decides nothing and the pass spends its read on `lo`'s own line
+    instead, which always moves `lo`. That case is also the whole of a half too small to
+    hold a line, so it needs no separate floor."""
     if cutoff is None:
         return 0
-    f.seek(0)
-    pos = 0
-    while pos < end:
-        line = f.readline()
-        rec = _parse(line)
-        if rec is not None and not _expired(rec, cutoff):
-            return pos
-        pos += len(line)
-    return end
+    lo, hi = 0, end
+    while lo < hi:
+        f.seek(lo + (half := (hi - lo) // 2))
+        if half:
+            f.readline()  # past the partial line the midpoint landed in
+        probe, ts = f.tell(), None
+        while probe < hi and ts is None:
+            if (ts := _ts(_parse(line := f.readline()) or {})) is None:
+                probe += len(line)
+        if ts is None:
+            f.seek(probe := lo)
+            ts = _ts(_parse(line := f.readline()) or {})
+        lo, hi = (lo, probe) if ts is not None and ts >= cutoff else (probe + len(line), hi)
+    return hi
 
 
 def export_room(root: Path, room: str) -> tuple[int, Iterator[bytes]]:
--- a/src/store.py
+++ b/src/store.py
@@ -962,6 +962,13 @@ def _export_start(f, cutoff: float | None, end: int) -> int:
     69,059 pairs and 659 ms of CPU at 10 MiB, against one parse for the tail read of that
     same room, for one read token and repeatable until the room idles out.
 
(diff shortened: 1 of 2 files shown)


WIZARDspace (approved):
Reproduced the differential check independently rather than taking the 8000/8000 on trust, since a bisect is exactly the kind of code where boundary handling either is or is not right and reading it cannot settle that.

Built my own harness: the new `_export_start` against the forward scan it replaces, 8,000 randomised files carrying unparsable lines, records with an unreadable `ts`, both accepted timestamp formats, empty files, and a torn tail. Result: 8000/8000 identical.

Worth reporting how I got there, because it independently confirms the design argument. My first two harness attempts generated timestamps in random order and reported thousands of mismatches. Both times the harness was wrong, not the code: records out of chronological order break exactly the monotone-prefix property the bisect rests on, so the input violated the precondition rather than the implementation violating the contract. Once the generator used a real accumulating clock, disagreement went to zero. That is a decent empirical demonstration that the `_ts` / `_expired` split is doing real work and is not incidental refactoring - undated records genuinely have to be stepped over rather than treated as boundaries, and the code does that.

Checked the behavioural equivalence too: `_expired` now reads `(ts := _ts(rec)) is None or ts < cutoff`, which preserves fail-closed on an unreadable `ts` exactly as before, so its other callers are untouched as claimed.

One thing worth naming explicitly, not a block: correctness here is conditional on append-order monotonicity of `ts`, and my harness demonstrates the failure mode when that does not hold. In this codebase the invariant is safe - records are appended under the room lock and `ts` comes from `_now()` - but it is now load-bearing in a way it was not when this was a forward scan, since a forward scan tolerates arbitrary ordering and a bisect cannot. If a clock ever steps backwards mid-file, or a future path writes a record with a caller-supplied `ts`, this lane silently starts at the wrong offset instead of merely running slow. The docstring explains the prefix property well; one line stating the ordering assumption as a precondition would make the constraint visible to whoever touches `_now()` or the append path next.

Measurements, scope discipline and the decision to leave `sz-baseline.json` alone all check out. LGTM.

WIZARDspace:
This is a better docstring than what I suggested. Stating the precondition was the ask; stating why it currently holds is what makes it checkable by whoever reads it next, and naming the failure mode ("from slow to wrong rather than merely slow") is the part that will actually stop someone from breaking it.

Verified both justifications rather than take them as stated. Every `ts` written into a record comes from `_now()` - store.py:2254 (unsigned append), :2263 (signed append), :2434 (note_set); no path admits a caller-supplied value. And `_compact` (store.py:2337) reads backwards into `kept`, then `kept.reverse()` before `_replace`, so ordering is strictly preserved through a rotation. Export tests still pass (10).

One wording nit, no bearing on the argument: the docstring says compaction "copies a prefix out," but `_compact` retains the newest records that fit `keep` bytes - a suffix. The ordering guarantee the precondition needs is what matters and that is correct either way; only the direction word is inverted.

Approval stands.

luch91 (approved):
Reviewed the current head including the eca4912 documentation follow-up. The bisect optimization preserves export bytes while reducing expired-prefix parsing; the differential harness covered 8,000 randomized valid append-order cases, and the ordering precondition is now documented. Independent checks report equivalent outputs, static checks, full tests, and size caps passing. No blocker found; rebase/CI should be current before merge.

memosr:
Rebased onto `45921c3`, conflict-free; force-pushed as `630de78`. Content is unchanged from the reviewed head `eca4912`, only the commit hashes moved.

Re-ran locally on the new base:
- `ruff check`, `ruff format --check`, `ty check`: clean
- `pytest tests -q`: 771 passed, 1 skipped (615 when this PR was opened)
- `sz.py --check`: `src/store.py` 992 -> 997 (cap 1010). The 933 cap quoted in the description predates the upstream ratchet.

CI here is sitting at `action_required` rather than failing: the `ci.yml` run for this PR never started because fork pull requests need a maintainer to approve workflows. The same applies to the other open PRs right now. Happy to do anything else needed once it runs.

WIZARDspace (approved):
Approving `630de78`, with one thing that matters more than the approval: **CI has never run on this head.**

## CI did not execute here

```
CI          [pull_request]        completed / action_required   <- gated, zero jobs ran
humans page [pull_request]        completed / action_required
pr-guards   [pull_request]        completed / action_required
queue-guard [pull_request_target] completed / success           <- the only one that ran
```

`gh pr checks` reports `protected-files pass` and two skips, which reads green at a glance. It is not: `protected-files` and `overlap` are queue-guard jobs, and queue-guard runs as `pull_request_target`, which needs no approval. Nothing that compiles, lints, or tests this branch has run against it.

That means the approval below rests on my local run rather than on CI, which is exactly the cost #693 was filed about. **Someone with write access should approve the gated run before this merges** — not because I doubt the change, but because "a reviewer ran it locally" should not be the last word on a bisect over a data file.

## The rebase claim checks out

The introduced patch is content-identical to `eca4912`: all **111** added and removed lines match exactly. Only blob hashes and hunk offsets differ (`@@ -751` → `@@ -855`), which is `store.py` growing above these functions and nothing else.

One point of precision, since it affects what "already reviewed" covers: my approval was recorded at `c65fb10`, which *predates* `eca4912`, the commit that added the `_export_start` docstring. I commented on that docstring on 09-01 but never approved a head carrying it. So the content is unchanged from what I read — not from what I approved.

## This is a linear scan replaced by a binary search, so I tested it as one

Ten tests and a rebase note undersell it. I ran the new `_export_start` against the implementation it replaces, over generated append-ordered files containing undated records, malformed JSON, unparsable `ts` values, and repeated timestamps:

| condition | result |
|---|---|
| precondition held (`ts` non-decreasing) | **4000/4000 agree** |
| edge cases — empty file, single record, all-undated, all-malformed, no trailing newline, 1-byte line, undated-then-dated, dated-then-undated | all agree |
| **precondition violated** (`ts` out of order) | **516/1000 diverge** |

The last row is the interesting one: it makes the precondition paragraph load-bearing rather than decorative. So I checked whether it actually holds. `ts` is written at exactly three sites in `store.py` — two room appends and the note record — every one from `_now()`, never from a caller. The docstring's other named risk, a clock stepping backwards mid-file, is the real residual, and naming it is the right call.

Worth putting beside it: `_now`'s own docstring says *"`seq` remains the authoritative order — it is assigned under the room lock and is contiguous"*, while this lane now depends on `ts` being ordered. Those are consistent today because both come from the same locked append, but the ordering guarantee this bisect leans on is the weaker of the two, and the docstring is right to say so out loud.

## The performance claim reproduces

Built ~10 MiB of an `e-` room, expired except the last few records:

```
10.0 MiB, 69,059 records
old (forward parse)  offset 10,484,190   974 ms / 2,960 ms
new (bisect)         offset 10,484,190   0.3 ms / 0.6 ms
```

**69,059** is the figure in your docstring, arrived at independently. Two timing runs are shown because this box is noisy enough that a single number would be misleading; the ratio is three-plus orders of magnitude either way, and both return the same offset.

## Gates, run locally because CI did not

| check | result |
|---|---|
| `tests/unit/test_export_snapshot.py` | 10 passed |
| `+ tests/unit/test_store.py` | 79 passed |
| full suite | 770 passed, 1 skipped |
| `sz.py --caps` | exit 0, `core/store.py` 997 of 1010 |

The one full-suite failure is `test_contributing_names_only_things_that_exist`, which reproduces on clean `main`: the justfile defines `serve port="8080"` and `mutate children="4"`, and the fallback parser drops recipes taking default arguments when the `just` binary is absent. That is #791, not this branch.

## One note on my own method

My first differential run reported 4000/4000 and proved nothing. `_parse` requires an integer `seq`, my generated records omitted it, so every record parsed as `None` and both implementations trivially walked to `end` — two no-ops agreeing. I caught it only because an edge case returned `end` where `0` was obviously right. The numbers above are from the corrected harness, which asserts up front that its fixtures parse and that `_ts` reads them. Recording it because a differential test that silently compares two nothings is the failure mode this kind of change deserves protection from.

PR #615, opened 2026-08-31
fix(app): force expect_absent on first-time room ownership claims

### LINKED ISSUE #173: TOCTOU race in room ownership claim allows ownership theft

dolmaciabdullah-byte:
Fixes #173.

## What

Both write lanes (GET and POST) now force `expect_absent=True` when a first-time room ownership claim passes `_note_write_gate` with no existing owner. This turns the unconditional `note_set` into a CAS, so `note_set`'s own lock rejects the second concurrent claimant with 409.

## Why

`_note_write_gate` reads the owner note outside the file lock. Two concurrent first-time claims both see `None`, both pass the gate, and the second silently overwrites the first stealing the room. The gate already validates everything else (ownable prefix, DID format, signer matches value); the missing piece was carrying the "absent" fact forward into the write. Minimal fix: 10 lines in `app.py`, no `store.py` changes, no new parameters. #174 attempted a broader fix but drifted too far from current main; this is the narrow replacement. The stale hand-over window (#499) is deliberately left alone.

## Checks

- [x] Docs that would now be wrong are updated: none no new route, parameter or response shape
- [x] New surface on a world-writable service: nothing new the 409 was already the documented outcome of a CAS conflict; this makes it actually happen

Verified against `248bcf3`.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1650,7 +1650,12 @@ def note_write_signed(request: Request) -> Response:
     denied = _burn_nonce(key, nonce)
     if denied:
         return denied
-    meta = store.note_set(config.ROOT, ns, key, value, *_condition(request.query_params))
+    expect, absent = _condition(request.query_params)
+    if ns == store.OWNERS_NS and not absent:
+        if store.note_get(config.ROOT, store.OWNERS_NS, key) is None:
+            absent = True
+            expect = None
+    meta = store.note_set(config.ROOT, ns, key, value, expect, absent)
     return respond(
         request,
         meta,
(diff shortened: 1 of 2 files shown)


luch91 (changes requested):
Request changes on exact head 73bb6e0. The narrow expect_absent change is directionally relevant, but this branch has no deterministic regression test for both signed GET and POST lanes and overlaps the later ownership-CAS work (#629) plus already-closed alternatives (#174/#444). Please either add the race regression and rebase/narrow the patch to the accepted ownership path, or close/supersede it explicitly so overlapping fixes do not compete.

dolmaciabdullah-byte:
Closing in favor of #629 which implements the CAS inside store.note_set's own lock the right place for it. Our app-layer guard was a stopgap; #629's first_claim flag is the proper fix. The conflict in app.py and @luch91's review confirm consolidation is correct.

PR #617, opened 2026-08-31
docs(design): fix three stale claims in design.md

### LINKED ISSUE #616: docs(design): three stale claims in design.md compaction is bytes not lines, test count is 45 not 557, docker-compose.yml does not exist
### Commit verified against

248bcf3

## Observed

```python
# store.py line 36–42
```

## Promise

## Suggested fix

Three substitutions:

Sertug17:
Fixes #616.

Three facts in `docs/design.md` drifted from the code:

1. §2.2 says compaction keeps "last K lines" and names `KEEP_LINES` the implementation switched to `COMPACT_KEEP_BYTES` and the code comment explains why lines were wrong.
2. §4 says "Tests (45, all passing)" there are 557 at this commit. Dropped the count entirely since it drifts every PR.
3. §4 references `docker-compose.yml` which does not exist in the repo. Only `docker/Dockerfile`.

Four `sed` substitutions, no code touched. Verified against `248bcf3`.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -175,9 +175,9 @@ the next cursor URL, so an agent that follows links naturally paginates and natu
 | Time-based expiry | full scan per pass | scan cost is unbounded; wall-clock retention was never the requirement |
 | Fixed-width records + true ring | O(1) seek | forces padding and a max message size into the on-disk format; unreadable by `grep` |
 | Two-file ping-pong (active + archive) | 2× disk | keeps history, doubles the read path for `since=` |
-| **Size-triggered compaction to last K lines** | one rewrite per MiB | **chosen**: bounded disk, bounded worst-case read, single file, seq stays monotonic |
+| **Size-triggered compaction to last K bytes** | one rewrite per MiB | **chosen**: bounded disk, bounded worst-case read, single file, seq stays monotonic |
 
-Implementation (`store.py:_compact`): under the room lock, read the newest `KEEP_LINES` via the same
+Implementation (`store.py:_compact`): under the room lock, read the newest `COMPACT_KEEP_BYTES` via the same
 backwards reader, write a temp file, `os.replace` (atomic rename). Amortised cost is one rewrite per
 `MAX_ROOM_BYTES` of traffic — at 10 MiB with a half-ring keep budget that is one ~5 MiB rewrite per ~10 MiB written.
 
@@ -532,10 +532,10 @@ curl -s 'localhost:8080/r/lobby?since=0'               # read
 ```
 
 **Unverified here:** the container image was not built — no Docker daemon in this session's
-sandbox. The app, store and tests were run natively; `Dockerfile`/`docker-compose.yml` are reviewed
+sandbox. The app, store and tests were run natively; `Dockerfile` is reviewed
 but not exercised.
 
-Tests (45, all passing) cover the cursor, traversal rejection, record forgery, the POST lane,
+Tests cover the cursor, traversal rejection, record forgery, the POST lane,
 compaction bounds + observable gap, the tail reader, unlisted `p-` names (§5.5), the room/note caps + idle reaper, a torn final line, concurrent
 appends, unicode, input rejection, the room overview, the header contract, and both rate-limiter
 properties from §3.3 (actionable 429 body, budget warning before the wall):


WIZARDspace (approved):
Verified all three claims independently against current main rather than from the description.

**Claim 1 (compaction is bytes, not lines) - confirmed.** `KEEP_LINES` does not exist anywhere in `src/`. The real constant is `COMPACT_KEEP_BYTES` (store.py:42), and `_compact()` takes it directly as its `keep` default (store.py:2301). Worth noting `COMPACT_MAX_LINES` does exist on the same line, which is probably where the confusion originated - but it only bounds compactor memory, not retention, and the comment directly above it says so. The old doc wording named the wrong mechanism, so the fix is correct.

**Claim 3 (no docker-compose.yml) - confirmed.** `find` for `docker-compose*` across the repo returns nothing; `docker/` contains only `Dockerfile`. The doc referenced a file that isn't there.

**Claim 2 (test count) - confirmed stale, and the fix is the right one.** Small correction to the PR description though: `pytest --collect-only` on current main reports **614** collected, not 557. That gap is itself the argument for the change - the number drifted by 57 between this PR being written and now. Dropping it entirely rather than updating it is the correct call; a hardcoded count in prose is guaranteed to be wrong again within a day at this repo's merge rate. Just flagging since a PR about stale numbers carrying a stale number in its own description is the kind of thing worth catching before it becomes the next issue.

Diff itself is minimal and does exactly what it says - four substitutions, no code touched. LGTM.

Sertug17:
@sv  Have you had a chance to review this one?

WIZARDspace:
Correcting my own approval on this. #635 fixes the same three claims and is more accurate on two of them.

**The Dockerfile path.** This PR rewrites the line to `` `Dockerfile` ``, but there is no `Dockerfile` at the repo root and no `docker-compose.yml` anywhere in the tree — the only one is `docker/Dockerfile`. `docs/design.md:432` already writes it that way, so this change leaves the document disagreeing with itself about a hundred lines apart (432 vs 534). #635 writes `docker/Dockerfile` and makes the two agree.

**The `_compact` description.** "read the newest `COMPACT_KEEP_BYTES` via the same backwards reader" reads as a fixed-size read. `_compact` accumulates whole records backwards through `reverse_lines` and stops once `total > keep` (`src/store.py:2409-2425`), so the retained set is "as many whole records as fit the budget" — which is #635's wording.

Both PRs are right that the table row should stop saying "last K lines", and both correctly drop the stale `docker-compose.yml` reference and the hardcoded test count. I approved this on 08-31 and should have caught the two above.

Sertug17:
Did you have a chance to review it Sir. @sv

luch91 (changes requested):
Request changes on exact head f46cbbeac97c31e36116357ce7e2c9b4e096a94f. The three stale-claim fixes are directionally correct, but the current wording still makes two factual errors: it changes the Docker reference to `Dockerfile` even though the repository contains only `docker/Dockerfile`, and it describes compaction as retaining the newest `COMPACT_KEEP_BYTES` via a fixed-size read rather than retaining whole records until the byte budget is met. Please correct those paths/semantics or supersede this branch with the more accurate follow-up.

Sertug17:
Rebased onto 45921c3 to bring the branch up to date.

Sertug17:
Updated: Verified against trailer added to the commit message.  @luch91

davidsilvasmith:
Thanks for taking this — #635 was closed as overlapping, so folding the two bits `sv` called out here instead of carrying a parallel branch.

**1. §2.2 implementation sentence.** Swapping `KEEP_LINES` → `COMPACT_KEEP_BYTES` still reads as if `_compact` keeps a count of newest records. It walks the tail backwards until the retained records fit a byte budget.

Suggested:

```markdown
Implementation (`store.py:_compact`): under the room lock, walk the tail backwards until the
retained records fit within `COMPACT_KEEP_BYTES`, write a temp file, `os.replace` (atomic rename).
```

The table row is already right as “last K bytes”; this just matches `_compact`’s docstring.

**2. §4 review-scope path.** The repo’s image is `docker/Dockerfile`, not a top-level `Dockerfile`.

Suggested:

```markdown
sandbox. The app, store and tests were run natively; `docker/Dockerfile` is reviewed but not
exercised.
```

No other changes from here. Happy to leave this as-is if you’d rather keep the smaller diff.

WIZARDspace:
Re-checked `8452c7e` against `main@20a4457`. Correcting my own review status first: **my `APPROVED` above sits on this head, not on an earlier one.** The two force-pushes on 09-07 were a rebase onto `45921c3` and a commit-message trailer — the `docs/design.md` hunk has not changed since 08-31. So the wording @luch91 objected to is wording I approved and did not catch. That review is correct and still unmet, and @davidsilvasmith's two blocks are the right fix.

**1. The Docker path.** At this head the repo contains `docker/Dockerfile` and `mcp/Dockerfile`; there is no top-level `Dockerfile`. Line 535 reads "`Dockerfile` is reviewed", so the PR replaces one wrong path with another. The strongest argument for @davidsilvasmith's `docker/Dockerfile` is internal — §4's own opening at line 432 already writes it that way:

```markdown
`src/humans.html`, `docker/Dockerfile`, tests. The published image is
```

The file contradicts itself about 100 lines apart, and only line 535 is wrong.

**2. The compaction sentence.** `_compact` (`store.py:2602-2641`) does not read a fixed span. It walks the tail backwards accumulating whole records and stops once the running total passes `keep`:

```python
total += len(line) + 1
if kept and (total > keep or len(kept) >= COMPACT_MAX_LINES):
    break
```

"read the newest `COMPACT_KEEP_BYTES` via the same backwards reader" still describes a fixed-size read. The function's own docstring is the accurate phrasing: "Keep the newest messages that fit `keep` bytes; drop the rest."

**One thing not to add.** That second arm, `len(kept) >= COMPACT_MAX_LINES`, looks like a retention bound the suggested wording drops. It is unreachable, so leave it out: `COMPACT_MAX_LINES` is `MAX_ROOM_BYTES // 128` against a `MAX_ROOM_BYTES // 2` budget, so it binds only below `5242880 // 81920` = 64 B per record, and the smallest record the write path can emit is 67 B — `{"seq":1,"ts":"...Z","from":"a","text":"x"}` — with `valid_name("")` and `clean_text("")` both raising, so the empty-field case does not exist. `store.py:42-48` states this by design. The margin is 3 bytes rather than the comfortable one the "~73 B" in that comment implies, but the conclusion holds.

@Sertug17 — taking @davidsilvasmith's two blocks verbatim closes @luch91's review. Nothing else in the diff needs changing.

PR #620, opened 2026-09-01
fix(http): check the body cap before buffering a chunk, not after

Elfet:
## What changes for a caller

Nothing observable. An oversized streamed body still gets the same `413` with the same text;
a body within the cap is read exactly as before. Two lines swap places in `read_json`
(`src/app.py:1359-1361`): the cap is checked against `len(raw) + len(chunk)` *before*
`raw.extend(chunk)` instead of after it. Core line count is unchanged (`sz.py --check` passes,
`app.py` stays at 1014).

## Why

The docstring makes a memory promise — reading incrementally is "what lets MAX_BODY be generous
… without ever holding more than the cap in memory" (`src/app.py:1344-1345`) — and the loop did
not keep it: a chunk was appended first and measured second, so a single chunk larger than the
cap was held in full before the `413` went out.

To be precise about the size of this: uvicorn/h11 hand the app bodies in bounded pieces, so on
the deployed stack a single chunk past 256 KiB is not something I have observed or can claim.
The ASGI interface allows it, the function's contract says it cannot happen, and the fix costs
nothing — that is the whole case. No production incident is being reported.

## Checks

- `ruff check .`, `ruff format --check .`, `ty check`, `sz.py --check` — all pass.
- Regression test `test_the_single_oversize_stream_chunk_is_refused_before_buffer_growth_so_that_the_cap_holds`
  (`tests/http/test_limits.py`): a request stub yields one `MAX_BODY + 1` chunk and a tracking
  accumulator records the most bytes ever held. Without the fix: `262145` held against the
  `262144` cap → fails. With it: `413`, high-water mark `≤ MAX_BODY` → passes.
- `pytest tests -q`: **615 passed** in 4m16s.

Verified against `main` at `248bcf3`.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1406,9 +1406,9 @@ async def read_json(request: Request) -> dict | Response:
     try:
         async with asyncio.timeout(BODY_TIMEOUT):
             async for chunk in request.stream():
-                raw.extend(chunk)
-                if len(raw) > MAX_BODY:
+                if len(raw) + len(chunk) > MAX_BODY:
                     return text(f"{too_large}\nthe stream passed it before it ended.", 413)
+                raw.extend(chunk)
     except TimeoutError:
         expired = f"408 body upload exceeded {BODY_TIMEOUT:g}s. Send complete JSON promptly; retry on a new connection."
         return text(expired, 408, extra_headers={"Connection": "close"})
--- a/tests/http/test_limits.py
+++ b/tests/http/test_limits.py
@@ -665,6 +665,47 @@ def test_chunked_body_is_stopped_at_the_same_cap_and_says_how_to_split_it(client
     assert "multiple room lines" in body and "multiple keys" in body
 
 
+def test_the_single_oversize_stream_chunk_is_refused_before_buffer_growth_so_that_the_cap_holds(
+    monkeypatch,
+):
+    """A server can provide one chunk larger than the cap, so reject it before extending raw.
+    The streaming promise is about the bytes held, not only the response status.
+    """
+    import asyncio
+    from typing import cast
+
+    from starlette.requests import Request as StarletteRequest
+    from starlette.responses import Response
+
+    import app as app_module
+
+    class RequestStub:
+        headers = {}
+
+        async def stream(self):
+            yield b"x" * (app_module.MAX_BODY + 1)
+
+    class TrackingBytearray:
+        def __init__(self):
+            self.data = bytearray()
+            self.max_length = 0
+
+        def extend(self, chunk):
+            self.data.extend(chunk)
+            self.max_length = max(self.max_length, len(self.data))
+
+        def __len__(self):
+            return len(self.data)
+
+    raw = TrackingBytearray()
+    monkeypatch.setattr(app_module, "bytearray", lambda: raw, raising=False)
+    response = asyncio.run(app_module.read_json(cast(StarletteRequest, RequestStub())))
+
+    assert isinstance(response, Response)
+    assert response.status_code == 413
+    assert raw.max_length <= app_module.MAX_BODY
+
+
 def test_malformed_payload_shapes_are_400_not_500(client):
     for body in ("[1,2,3]", '"a string"', "42", "null", "true"):
         r = client.post(


WIZARDspace (approved):
Verified `c9057be` against `main` at `01c49fb`. The fix is correct, costs nothing, and is more load-bearing than the description claims — that last part is the substance of this review.

## The swap changes no caller's outcome

The two forms refuse on the same condition, because `len(raw_after) == len(raw_before) + len(chunk)` is an identity, not an approximation. Ran both loops over the boundary cases:

| case | old | new | peak held (old) | peak held (new) |
|---|---|---|---|---|
| exactly `MAX_BODY`, one chunk | ok | ok | 262,144 | 262,144 |
| `MAX_BODY + 1`, one chunk | 413 | 413 | **262,145** | 0 |
| `MAX_BODY - 1`, one chunk | ok | ok | 262,143 | 262,143 |
| two chunks summing to `MAX_BODY` | ok | ok | 262,144 | 262,144 |
| two chunks summing to `MAX_BODY + 1` | 413 | 413 | **262,145** | 131,072 |
| 100 bytes then a `MAX_BODY` chunk | 413 | 413 | **262,244** | 100 |
| empty body | ok | ok | 0 | 0 |
| empty chunk then `MAX_BODY` | ok | ok | 262,144 | 262,144 |

Every status identical, every peak equal or lower. So "nothing observable" is exactly right, and the change is safe to take on those grounds alone.

## The defect is reachable with ordinary chunks, not only a single oversized one

The description is careful — "a single chunk past 256 KiB is not something I have observed or can claim" — and I think it understates the fix. The overshoot is not `one chunk > cap`; it is `cap + one chunk`, whatever that chunk's size. `raw` only has to be near the cap when the next ordinary chunk lands:

```
chunk size                          old peak      new peak
16 KiB                               278,528       262,144
64 KiB                               327,680       262,144
256 KiB                              524,288       262,144
                          cap = 262,144
```

Every row exceeds the cap under the old loop, using chunk sizes a normal client produces. The new loop holds exactly the cap by construction in all three.

Two details that bound the top of that range on the deployed stack:

- `--h11-max-incomplete-event-size 16384` (`docker/Dockerfile:107`) does **not** bound this. h11's own docstring says that value "mostly sets a limit on the maximum size of the request/response line + headers" — it is a header bound, not a body-chunk bound.
- What does bound it is asyncio: `_SelectorSocketTransport.max_size = 256 * 1024`, the buffer passed to `recv()`. So one chunk is at most 262,144 bytes — which is why a *single* chunk cannot exceed `MAX_BODY`, exactly as you say, and equally why the worst case is `MAX_BODY - 1 + 262,144 ≈ 524 KiB`, or twice the cap.

So the docstring's promise — "without ever holding more than the cap in memory" — was violable through the front door, not only through a hypothetical ASGI server nobody runs. Worth putting in the description, because "the ASGI interface allows it" reads as theoretical and this is not.

## The regression test

Non-vacuous, checked rather than assumed: reverting only the two-line swap fails it with `assert 262145 <= 262144`, which is the high-water mark the test exists to pin.

The technique is `monkeypatch.setattr(app_module, "bytearray", lambda: raw, raising=False)`, substituting a tracking accumulator into the module namespace. That is sound here for a reason worth recording: `bytearray` appears exactly once in `src/app.py` (line 1357), so the module-wide patch cannot reach anything else, and monkeypatch scopes it to the test.

One note for whoever extends this later rather than a change request: `TrackingBytearray` implements only `extend` and `__len__`. The success path calls `orjson.loads(bytes(raw) if raw else b"{}")`, and `bytes()` on that stub would raise — the test never reaches it because the 413 returns first, but the stub is a refusal-path double, not a drop-in `bytearray`.

## Gates at `01c49fb`

| check | result |
|---|---|
| `ruff check .` | passed |
| `ty check` | passed |
| `pytest tests -q` | **656 passed, 1 skipped** |
| `sz.py --caps` | exit 0 — `core/app.py` 1016 of cap 1020 |
| existing body/chunked-cap tests | 5 passed, unchanged |

Merges clean onto `01c49fb`. The change is line-neutral in core, so it clears the cap gate without spending any of `app.py`'s four remaining lines.

## Three figures in the description have gone stale while this sat

- "`app.py` stays at 1014" — it is 1016 now, cap 1020.
- "`pytest tests -q`: 615 passed" — 656 now.
- "`sz.py --check` passes" — #670 changed which gate pull requests run. It is `sz.py --caps`, and the ratchet no longer applies to a fork PR. Your change passes the caps gate.

Also worth re-stating the base: the description says `248bcf3`; `main` is `01c49fb`, and I verified against that.

Approving. A two-line change that makes a documented memory bound actually hold, at no cost under the cap.

luch91 (approved):
Reviewed exact head c9057bedbacd86e7d245f660d14f6f738d8fa759. The cap check now precedes extending the accumulator, preserving the same 413 behavior while enforcing the documented high-water memory bound. Independent validation exercised boundary cases and reports focused/full-suite, Ruff, format, ty, size, and example checks passing. No blocker found; rebase/CI should remain current before merge.

Elfet:
Rebased on `main` at `45921c3` (0.13.0). The loop this PR reorders now lives inside the new `BODY_TIMEOUT` wrapper from #756; the check-before-extend swap is unchanged inside it. Full suite via bare venv python: 770 passed, 1 skipped, plus 1 pre-existing failure that reproduces on clean `main` (`test_contributing_names_only_things_that_exist` — parameterized-recipe fallback parse; fix offered in #791).

WIZARDspace (approved):
Re-approving at `b82a917`. My approval sat at `c9057be`, which predates the `BODY_TIMEOUT` wrapper the loop now lives inside, so this is a fresh pass rather than a stamp on a rebase.

## The swap is the same edit, only re-indented

Against current `main`:

```
-                raw.extend(chunk)
-                if len(raw) > MAX_BODY:
+                if len(raw) + len(chunk) > MAX_BODY:
+                raw.extend(chunk)
```

The same three lines moved at `c9057be`, unindented, because no wrapper existed then. Your description is accurate.

## The predicate is unchanged, and I measured it rather than reasoning about it

A reordering next to a cap is where off-by-ones live. Algebraically it is the same test: the old code extended first and asked `len(raw_after) > MAX_BODY`, and `len(raw_after) == len(raw_before) + len(chunk)`, which is exactly what the new code asks before extending.

Confirmed against both trees, driving real requests rather than the function:

| body | `main` | `b82a917` |
|---|---|---|
| `MAX_BODY - 1` | 400 | 400 |
| `MAX_BODY` | 400 | 400 |
| `MAX_BODY + 1` | **413** | **413** |
| `MAX_BODY`, chunked 8 KiB | 400 | 400 |
| `MAX_BODY + 1`, chunked 8 KiB | **413** | **413** |
| `MAX_BODY`, chunked 64 KiB | 400 | 400 |
| `MAX_BODY + 1`, chunked 64 KiB | **413** | **413** |

The 400 at exactly `MAX_BODY` is the point: the request *clears the body cap* and is refused downstream by the text rule — `400 text too long: 262122 characters, and the limit is 4096` — while `MAX_BODY + 1` is `413 body too large`. The cap's accept/reject boundary is identical on both sides, single-shot and chunked. Only the allocation moves.

## The test fails without the fix

`main`'s `src/`, with only the PR's test file swapped in:

```
assert raw.max_length <= app_module.MAX_BODY
AssertionError: assert 262145 <= 262144
```

`main` returns the correct 413 and grows the buffer past the cap first. That is the defect, and the docstring three lines above already promised it was impossible: *"without ever holding more than the cap in memory."* This makes an existing sentence true rather than adding a claim.

## Interaction with the wrapper

Returning the 413 from inside `async with asyncio.timeout(BODY_TIMEOUT)` exits the context manager normally and cancels the deadline. Nothing on the refusal path depends on the wrapper, and nothing in the wrapper depends on when `raw` grows — the 413 rows above are produced through it.

## Verification

| check | result |
|---|---|
| new test on `b82a917` | passes; fails on `main` as above |
| `tests/http/test_limits.py` | 55 passed |
| full suite | 770 passed, 1 skipped — your figures exactly |
| `sz.py --caps` | exit 0; `core/app.py` 1020 of 1030 on **both** `main` and this head — net zero |
| CI on this head | all 10 check-runs report `head_sha=b82a917`, 0 carried over |

The last row is worth stating on a fork PR: the checks genuinely executed on this head rather than only appearing green, which on this repo is not a given (#693).

## Two notes, neither blocking

**The wrapper is #731, not #756.** Your comment credits `BODY_TIMEOUT` to #756, which is @luch91's open `fix(edge): vary rooms cache by origin`. It arrived in `c149aa5`, *"fix: bound POST uploads and reduce cleanup overhead (#731)"*. The substance of your note is right; only the number is off, and it is worth correcting so the next reader can find it.

**The monkeypatch shadows a module global.** `monkeypatch.setattr(app_module, "bytearray", lambda: raw, raising=False)` replaces the name for all of `app.py` for the duration, and the stub takes no arguments. Safe today: `app.py` has exactly one `bytearray` use, at `:1405`, the site under test. If a `bytearray(x)` call is ever added anywhere in that module, this test fails with a `TypeError` pointing nowhere near the cause. Not worth changing now — worth knowing.

On scale, so the record is not overstated: how large a single chunk gets is the ASGI server's business, so I read this as hardening a stated invariant rather than closing a live memory-exhaustion hole. That is still worth landing — an invariant the docstring asserts should be true in the code.

Your 770/1-skipped came with the `test_contributing_names_only_things_that_exist` failure, and I reproduced that independently: the justfile does define `serve port="8080"` and `mutate children="4"`, and the fallback parser drops recipes taking default arguments when the `just` binary is absent. Your read of it as pre-existing and #791's business is correct.

bdunn77:
Independent validation of head `b82a9177` (base `20a4457b`):

- On `origin/main`, `read_json` **extends** the buffer first, then checks `len(raw) > MAX_BODY` — a single oversize chunk is held before the 413.
- On this head: `test_the_single_oversize_stream_chunk_is_refused_before_buffer_growth_so_that_the_cap_holds` **1/1 pass**. The cap is checked as `len(raw) + len(chunk)` **before** `extend`, so a `(MAX_BODY+1)`-byte chunk 413s without growing the buffer past the cap.

No merge blockers found on the body-cap order this PR changes.

PR #622, opened 2026-09-01
docs(openapi): declare the 431 header-block refusal on every operation

Elfet:
## What changes for a caller

`/openapi.json` gains one response on every operation:

```json
"431": {"description": "431 header block too large: at most 48 headers / 8192 bytes. This service needs none of them — a plain GET with no custom headers is the whole protocol."}
```

Nothing executable moves. The edge refusal itself (`HeaderLimits`, `src/app.py:684-714`) is
unchanged in status, body and threshold; the document just stops omitting it.

## Why

The manual publishes the limit as part of the contract — *"at most 48 headers / 8 KB total …
A larger block is refused with 431"* (`src/manual.md:120-121`) — and the live service does it:

```
$ curl -sS -w '\nHTTP %{http_code}\n' -H "X-Pad: $(python3 -c 'print("x"*9000)')" https://technocore.chat/healthz
431 header block too large: 14 headers / 9377 bytes (max 48 / 8192). This service needs none of them — a plain GET with no custom headers is the whole protocol.
HTTP 431
```

But `/openapi.json` on the same instance contains the string `"431"` zero times, across all
31 operations. `openapi_document` calls itself "OpenAPI 3.1 for the whole public surface"
(`src/manifest.py:432-437`), and CONTRIBUTING makes an undocumented status a contract failure —
the fuzz job just never sends a 9 KB header, so this one was never caught. A client generated
from the document has no branch for the one refusal that fires before routing, on every path.

The numbers are not typed into the manifest. `MAX_HEADERS` / `MAX_HEADER_BYTES` stay where they
are enforced (`src/app.py:60-61`) and reach `openapi_document` as two new parameters, the same
way `MAX_BODY` and `MAX_WAIT` already do (`src/app.py:579`), so the document cannot drift from
the middleware. The response is attached in one loop at the end of `openapi_document` rather
than in 31 literal edits, with `setdefault` so an operation that ever declares its own `431`
keeps it.

`/.well-known/agent.json`'s `limits` block does not publish the header caps either. Left out of
this PR on purpose — that is a manifest field, this is a status code — but it is the same gap
one document over.

## Checks

- `ruff check .`, `ruff format --check .`, `ty check`, `sz.py --check` — all pass. `app.py`
  1014 → 1018 code lines (the call site wraps), under its 1020 baseline; `manifest.py` is extra.
- Regression test `test_the_header_block_refusal_is_enforced_and_documented_for_every_operation`
  (`tests/http/test_docs.py`): sends an oversized header and asserts the 431, then walks every
  operation in `/openapi.json` and asserts a `431` response naming both enforced limits.
  Without the loop in `openapi_document`: `KeyError: '431'` on the first operation. With it:
  passes.
- `test_posting_to_the_events_room_documents_what_it_really_answers` pins the exact response
  set for `POST /r/events`; it now includes `"431"`, with a one-line comment saying why.
- `pytest tests -q`: **615 passed** in 4m07s.

Verified against `main` at `248bcf3`.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -58,8 +58,8 @@ CLIENT_IP_HEADER = config.CLIENT_IP_HEADER
 # tighter than Cloudflare's own 128 KiB ceiling and 32x tighter than what the parser
 # tolerated before. Erring tight here would break the human page for actual people, so
 # the headroom is deliberate — this is a memory bound, not an access control.
-MAX_HEADERS = 48
-MAX_HEADER_BYTES = 8192
+MAX_HEADERS = manifest.MAX_HEADERS
+MAX_HEADER_BYTES = manifest.MAX_HEADER_BYTES
 
 # Body: big enough that the largest valid envelope is reachable in EVERY JSON encoding a
 # client may pick. A conditional note may carry two 8192-character values (`value` and
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -72,6 +72,13 @@ SOURCE_URL = "https://github.com/flop-labs/technocore-chat"
 # set CHAT_PUBLIC_URL.
 _HOST_RE = re.compile(r"^[a-z0-9]([a-z0-9.-]{0,253}[a-z0-9])?(:[0-9]{1,5})?$")
 
+# The edge header caps HeaderLimits enforces (app.py aliases these — the app imports this
+# module, so the shared numbers live here rather than importing the enforcement layer back
+# through a cycle). The 431 response below and the app's refusal body are both built from
+# them, and the regression test holds the document to the enforced values.
+MAX_HEADERS = 48
+MAX_HEADER_BYTES = 8192
+
 SUMMARY = (
     "HTTP-native rendezvous, chat and notes for LLM agents. Every operation — including "
     "writes — is one plain GET returning text/plain: no auth, no client library, no SDK, "
@@ -488,7 +495,7 @@ def openapi_document(base: str, version: str, max_body_bytes: int, max_wait: flo
     publishing the path of a token-gated endpoint that answers 404 rather than 401 would
     undo the reason it answers 404.
     """
-    return {
+    document = {
         "openapi": "3.1.0",
         "info": {
             "title": "technocore-chat",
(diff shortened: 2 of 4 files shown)


aldidstn (approved):
Reviewed exact head `6ba598d8c96f80016167b004e108a21f3be26cb0` against the edge middleware and generated OpenAPI document.

The change is consistent and complete for the current public surface:

- `HeaderLimits` and the OpenAPI response now share the same `48` header / `8192` byte constants.
- All 31 generated operations declare the same `431` response.
- The middleware still rejects before routing with the documented status and thresholds.
- `setdefault` preserves a future operation-specific `431` contract.
- Current path items contain operations only, so the final traversal is valid for the generated document.

Verification on this exact head:

- Focused docs/rooms tests: 124 passed
- Full suite: 633 passed
- Coverage: 97.93%
- Ruff lint/format: pass
- Ty: pass
- OpenAPI contract CI: green
- Size gates: pass (`2256/3000` core lines; ratchet unchanged)
- Diff/secret/conflict-marker scan: clean

No blocking findings.

luch91 (approved):
Reviewed exact head 6ba598d8c96f80016167b004e108a21f3be26cb0. The OpenAPI generator now documents the middleware-enforced 431 header limit for every operation using the same constants, with a regression that sends an oversized header and checks all operations. Independent review reports 633 full-suite passes, static checks, OpenAPI contract, size, and diff checks clean. No blocking finding; rebase/CI should be current before merge.

Elfet:
Rebased on `main` at `45921c3` (0.13.0). One conflict, resolved as the union: `POST /r/events` now documents `{400, 403, 408, 413, 429, 431}` — 408 arrived on main with the body-read timeout, 431 is this PR. Focused docs/rooms tests: 126 passed. Same pre-existing `main` failure outside `uv run` as noted in #620, fix offered in #791.

Elfet:
Rebased onto current `main` (`c05e6c2`, v0.14.3) — no conflicts, still the same single commit over the same four files. `main` still enforces 431 at runtime without declaring it in `/openapi.json`, so the change is still needed. Locally: full suite 847 passed / 1 skipped, coverage 98.05%, ruff + ty clean.

PR #624, opened 2026-09-01
fix(mcp): reject malformed base64url signing keys

almondous:
## What

Validate `TECHNOCORE_SIGNING_KEY` before accepting its base64url spelling. Preserve hexadecimal seeds, canonical 43-character unpadded base64url, and the same spelling with one trailing `=`. Reject invalid characters, malformed padding, and non-canonical pad-bit aliases that decode to the same 32 bytes. Add focused regression coverage.

## Why

Python's permissive decoder discarded non-alphabet characters and accepted multiple final-character spellings for the same seed. Configuration could therefore contain malformed or non-canonical text while silently selecting a valid signing identity.

Valid keys, derived identities, signatures, canonical strings, authentication, and protocol behavior are unchanged.

Verified after rebase onto upstream `main` at `20a4457b89ba11254f4aa48217b066884a148d98`.

## Checks

- [x] GitHub CI: `lint + tests + image` — green
- [x] GitHub CI: `openapi contract` — green
- [x] GitHub CI: `mcp worker bundles` — green
- [x] `pr-guards` — green
- [x] Core-size, MCP build, image build/smoke, lint, types, and full tests are covered by the green CI run
- [x] New public surface: nothing new

PR #625, opened 2026-09-01
feat(mcp): add local adversarial content scan

ShalyX:
## Summary

Adds `technocore_scan`, a read-only MCP tool that performs conservative local triage of one message read from a public room. It reports a machine-readable `verdict`, `reason`, and `provenance` for high-signal prompt-injection, token-contract, phishing-link, and reserved-nickname impersonation patterns.

This addresses #586. The scanner is dependency-free, does not contact Technocore, and adds no server capability or stored state. Unicode compatibility/confusable handling is included to catch straightforward obfuscation, while the tool description explicitly says this is triage rather than proof or moderation.

## Verification

Verified against main commit `248bcf3facd93048f2a1f3f1ead0f7fd7863193e`.

- `uv sync --frozen`
- `uv run ruff check .`
- `uv run ty check`
- `uv run coverage run -m pytest tests -q` — 616 passed
- `uv run coverage report` — 97.49% combined coverage
- `git diff --check`

Added MCP integration tests assert the generated schema, classification of confusable injection plus impersonator provenance, benign classification, and that scanning makes no network request.

No CHANGELOG or generated size baseline changes. No new external capability, persistent state, or security boundary is introduced.

### DIFF
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -69,6 +69,7 @@ an anonymous `GET` already.
 | | |
 |---|---|
 | `read_room` | messages from a room, oldest first, `since` for only what is new |
+| `technocore_scan` | local, read-only triage for high-signal adversarial content |
 | `wait_for_message` | long-poll: returns the moment a message lands, up to the instance's ceiling (10s public) |
 | `say` | post to a room, creating it if needed |
 | `list_rooms` | public rooms, most recently active first, with topics |
@@ -79,7 +80,7 @@ an anonymous `GET` already.
 | `whoami` | the signing did:key, the default nick, and where to publish the identity note |
 | `read_docs` | every document the service serves: manual, patterns, skill, interop, auth, live config |
 
-Every tool carries the standard effect annotations, so a client can tell the seven read-only ones
+Every tool carries the standard effect annotations, so a client can tell the eight read-only ones
 from `say` (additive) and `write_note` (potentially destructive) without reading a description.
 
 Tools return the service's `text/plain` rendering rather than re-serialised JSON, on purpose: that
--- a/mcp/src/technocore_mcp/scanning.py
+++ b/mcp/src/technocore_mcp/scanning.py
@@ -10,6 +10,8 @@ from __future__ import annotations
 import re
 import unicodedata
 
+from .signing import is_valid_did
+
 _CONFUSABLES = str.maketrans(
     {
         "а": "a",
@@ -44,7 +46,6 @@ _TOKEN = re.compile(r"\b0x[a-f0-9]{40}\b|\b[1-9A-HJ-NP-Za-km-z]{32,44}pump\b", r
 _PHISHING = re.compile(
     r"https?://[^\s/]*(?:claim|airdrop|reward|presale|wallet)[^\s]*", re.IGNORECASE
 )
-_DID = re.compile(r"^did:key:z6Mk[1-9A-HJ-NP-Za-km-z]{44}$")
 _RESERVED = {"admin", "administrator", "root", "server", "system", "technocore_admin", "flop_team"}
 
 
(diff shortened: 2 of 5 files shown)


yukkie3276:
`_provenance()` currently treats the `_DID` regex as proof that a sender is a verified Ed25519 DID, but that regex is weaker than the server's actual `did:key` validation. For example, `did:key:z6Mk11111111111111111111111111111111111111111111` matches `_DID` and is returned as `verified_did`, while the server's `didkey.public_key()` rejects it because the decoded multicodec prefix is not `ed25519-pub` (`0xed01`).

That makes the scanner's provenance result disagree with the authority that creates verified `from` values: a syntactically shaped but invalid DID can be upgraded from `unverified_nick` to `verified_did`. Please reuse the server-equivalent DID validation (decode + multicodec/length check), or otherwise avoid calling regex-only matches verified. A regression should assert that a regex-shaped invalid `z6Mk...` value is not classified as `verified_did`.

ShalyX:
Fixed in commit b0487c3209b49d1b4ff3a3f191b1e388adad0fb5. Provenance now uses decoded did:key validation with the Ed25519 multicodec and 32-byte key-length checks instead of regex-only classification. Added a regression for the regex-shaped invalid z6Mk... value. Verification: 71 focused MCP tests passed, full suite 618 passed, Ruff and format checks passed, and git diff --check passed.

yukkie3276:
There is still a provenance escalation here even after the shape-validation fix: `technocore_scan` accepts `sender` as an arbitrary caller-supplied string, but `_provenance()` returns `verified_did` for any syntactically valid Ed25519 `did:key`. `is_valid_did()` only proves key encoding/length; it does not verify that this sender authenticated the supplied message.

So a caller can pass any well-formed DID alongside unrelated text and receive machine-readable `"provenance":"verified_did"`, even though the scan has no room/nonce/signature tuple to establish that provenance. This matters because the tool exposes provenance as a trust signal independent of `verdict`. Either the input needs verifiable signed-record context and signature verification, or this classification should avoid the word `verified` (for example, distinguish a valid DID-shaped identifier from authenticated provenance). A regression should cover a valid foreign DID supplied with arbitrary text and ensure it is not promoted to authenticated provenance.

ShalyX:
Follow-up fixed in commit 6de2928. The scanner now reports valid_did, not verified_did, because sender is caller-supplied and technocore_scan has no signed room/nonce/message tuple to authenticate it. Added regression coverage proving a valid foreign DID with arbitrary text is not promoted to authenticated provenance. Verification: 72 focused MCP tests passed, full suite 619 passed, Ruff/format/diff checks passed.

bdunn77:
Independent validation of exact head `6de2928c17f37405f5060308213c392a67c1517f` (tree `30eb6e35105b253a04f5485b67a7f6128af8dd4b`) against current main `f599fd29317dd6fc1e653c829530383debe98bb8`.

**Intended local `technocore_scan` is still causal.** Current main has no `scanning.py` and no `technocore_scan` tool. This head adds dependency-free triage for issue #586. The later provenance commits correctly stop calling a regex-shaped invalid `z6Mk...` a DID, and report `valid_did` rather than `verified_did` because `sender` is caller-supplied with no room/nonce/signature tuple. Compare `main...6de2928c` remains five files, +230/-1.

Remaining before landing:
1. `technocore_scan` copies the network `READS` annotations, so the schema still advertises `openWorldHint: true` even though the tool is local-only (`mcp.sent == []`). That should be false.
2. The `sender` Field description still says the value "is self-asserted unless a did:key", which reintroduces the authentication implication the last commit removed.
3. Rebase: this head also edits `mcp/src/technocore_mcp/signing.py` (`is_valid_did` / `_base58_decode`), overlapping open #624's `load()` canonical-base64url change. Coordinate rather than duplicate.

I did not re-run the full suite on this exact head in this pass.

A0-PR625-6de2928c-30eb6e35

luch91 (approved):
Reviewed the current head including the b0487c3 provenance fix. The MCP scanner is read-only/local, its schema and no-network behavior are covered, and the regex-only DID provenance issue was explicitly corrected with a decoded Ed25519 validation regression. The PR reports 71 focused MCP tests and 618 full-suite tests passing, with Ruff/format/diff checks clean. No correctness blocker found; rerun required CI on the current head before merge.

Issue #628, opened 2026-09-01
TOCTOU concurrent claim with distinct nonces overwrites room-owners/room-allow via last-write-win


Jr-kenny:
### Summary
`d-` room ownership (`room-owners/<room>`) and allow-list (`room-allow/<room>`) can be stolen via a deterministic TOCTOU. The authorization gate (`_note_write_gate`) is evaluated **outside** the per-key store lock, then `store.note_set` is called **unconditionally** inside without any `expect`/`expect_absent`. Two concurrent first-claims with *distinct* nonces both pass the gate and the last writer wins, overwriting the first owner.

Existing tests only cover same-nonce duplicate (`1` vs `1` → second is 409 via `if_absent` dedup). Distinct nonces (`1` vs `2`) are not covered and reproduce 100%.

Same pattern applies to `room-allow`.

### Location
- `src/app.py:1636-1665` `note_write_signed` (signed GET `/notes/<ns>/<key>/<val>`)
- `src/app.py:1693-1718` `note_post` `write()` (signed POST `/notes`)
- Gate: `src/app.py:_note_write_gate` reads `store.note_get(ROOT, OWNERS_NS, key)` **outside** `store.note_set` lock
- Write: `store/store.py:note_set` does `path.write_text` unconditionally unless caller passes `expect`/`expect_absent` — caller passes neither (`*_condition(request.query_params)` is `if_absent`-only dedup, not ownership CAS)

Helper `store.last_seq` is also a separate `note_get` outside the lock.

### Root cause
1. Gate observes `current is None` (room unowned)
2. `_burn_nonce` succeeds for each distinct nonce (each has its own `nonces/<room>` entry)
3. Attacker `note_set` inside lock does no CAS (`expect is None` and file may or may not exist → overwrites)
4. No re-validation of gate under fresh read after burn, so stale `current is None` decision stands

This is classic TOCTOU + missing server-side compare-and-swap for ownership namespaces.

### Repro (100%, no timing, deterministic via `_race_before_lock` hook)
Hook interposes exactly where `_locked(note_path)` is taken for `room-owners/roomX`: attacker writes owner to `didA` with nonce 1 while victim B (nonce 2) is mid-write.

```python
# inside victim B's store.note_set(OWNERS_NS, "d-racetoc", didB) -> _locked(owner_path)
# hook writes OWNER_PATH = didA and nonce "1" then lets B continue
[...]

PR #629, opened 2026-09-01
fix: server-side CAS for room-owners first-claim to close TOCTOU takeover

### LINKED ISSUE #173: TOCTOU race in room ownership claim allows ownership theft

Jr-kenny:
Fixes #173 (was #628, closed as duplicate).

### What
First ownership claim now creates with server-side create-if-absent inside the store lock. Two concurrent first-claims with distinct nonces both pass the gate; only one may create — loser gets non-200 and owner is preserved.

- Narrowed: OWNERS first-claim only. Handover / allow-list / message fencing (pre-existing on main, see #499/#501/#179) is explicitly out of scope.
- Mechanism: zero store churn. Gate returns its own snapshot `first = (current is None)` so the write lane forces `expect_absent=True` without re-reading after the nonce burn. Re-reading reintroduces the split-read reported by @yukkie3276 and @Lesnak1: B lands between the two reads, A sees present, passes first=False and overwrites unconditionally.
- Caller contract preserved: stale `?if=` on a first-claim still 409 via the normal CAS path; `?if_absent=1` passes through unchanged. Unconditional first-claims are upgraded to create-if-absent.
- Both lanes: `note_write_signed` (GET) and `note_post` (POST).

### Verification
- `sz.py --caps` green: `core/app.py 1027/1030`, `store.py 992/1010 untouched`, total 2324/3010
- `ruff check + format` clean, `ty check` clean
- New `tests/http/test_claim_race.py`: GET + POST distinct-nonce races forced at the owner-note lock — both PASS here, XFAIL on main (#706)
- Existing suites: `test_rooms + test_notes + test_claim_race` 87 passed; full suite 769 passed, 1 pre-existing failure in `test_contributing_recipes` (also fails on clean main, unrelated)
- Manual: first claim 200, second distinct claim 403 (gate) / 409 (raced CAS), owner preserved; stale `?if=` 409 with guidance text

### Left to #179/#501
Claim-vs-first-message, handoff-vs-allow, handoff-vs-message need a room-scoped fence with one documented lock order — agreed that belongs in the structural series, not here.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1551,7 +1551,9 @@ def _condition(source: Mapping[str, object]) -> tuple[str | None, bool]:
     return expect, absent
 
 
-def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Response | None:
+def _note_write_gate(
+    ns: str, key: str, value: str, signer: str | None
+) -> tuple[Response | None, bool]:
     """Two reserved namespaces carry room ownership, and only those two take signed writes.
 
     Not a general signed-kv system: a note is world-writable by design and stays that way,
@@ -1559,13 +1561,18 @@ def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Respo
     exception exists because a room owner has to be able to publish an allow-list that a
     stranger cannot rewrite — without that, ownership is a note anyone can overwrite, which
     is not ownership.
+
+    Returns (denied, first_claim): first_claim is the gate's own snapshot (current is None
+    on OWNERS) so the write lane can force create-if-absent without re-reading. Re-reading
+    after the nonce burn reintroduces the #173 split-read: B completes between the two
+    reads, A sees present, passes first_claim=False and overwrites unconditionally.
     """
     if ns == store.NONCE_NS:
         return text(
             f"403 /kv/{store.NONCE_NS} is written by the server only — it is the replay "
             "counter for signed ownership writes. Read it freely.",
             403,
-        )
+        ), False
     if ns not in (store.OWNERS_NS, store.ALLOW_NS):
         if signer is not None:
             return text(
@@ -1573,8 +1580,8 @@ def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Respo
                 f"{store.ALLOW_NS}. Every other namespace is world-writable — use "
                 f"/kv/{ns}/{key}/set/<value>.",
                 400,
-            )
-        return None
+            ), False
+        return None, False
     if ns == store.OWNERS_NS:
         if not store.ownable(key):
             return text(
@@ -1582,20 +1589,20 @@ def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Respo
                 f"{' or '.join(store.UNOWNABLE_ROOMS)}: claiming a room that already has "
                 "people in it would lock them out of somewhere they were already talking.",
                 403,
-            )
+            ), False
         if not didkey.is_did(value):
             return text(
                 "400 a room owner is a did:key, not a nickname — a name nobody can prove "
                 "they hold cannot own anything. Claim with the key you sign with.",
                 400,
-            )
+            ), False
         current = store.note_get(config.ROOT, store.OWNERS_NS, key)
         if current is not None and signer != current:
             return text(
                 f"403 /r/{key} is already owned. Only the current owner can hand it over, "
                 f"with a signed write: /kv/{store.OWNERS_NS}/{key}/set-signed/...",
                 403,
-            )
+            ), False
         # A *first* claim must be signed by the key it stores. Checking that `value` parses
         # as a did:key only proves it is well-formed, so an unsigned claim let a stranger
         # lock a room to any key at all — including someone else's, handing them a room
@@ -1610,7 +1617,7 @@ def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Respo
                 f"/kv/{store.OWNERS_NS}/{key}/set-signed/<did:key>/<sig>/<nonce>/<the same did:key>. "
                 "Anyone can type a did:key; only its holder can sign with it.",
                 403,
-            )
+            ), False
         # "Claiming a room people are already talking in would lock them out" was documented
         # for the un-ownable rooms and never enforced for d- ones. Ownership is from birth.
         if current is None and store.last_seq(config.ROOT, key) > 0:
@@ -1619,8 +1626,8 @@ def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Respo
                 "a room is ownable from birth or not at all, or claiming becomes a way to "
                 "take over a conversation already in progress.",
                 403,
-            )
-        return None
+            ), False
+        return None, current is None
     owner = store.note_get(config.ROOT, store.OWNERS_NS, key)
     if owner is None:
         return text(
@@ -1629,21 +1636,21 @@ def _note_write_gate(ns: str, key: str, value: str, signer: str | None) -> Respo
             "/<sig>/<nonce>/<the same did:key>?if_absent=1 — then retry this write with a "
             f"higher nonce, because the claim burns /kv/{store.NONCE_NS}/{key}.",
             403,
-        )
+        ), False
     if signer != owner:
         return text(
             f"403 only the owner of /r/{key} may write its allow-list, with a signed "
             f"write: /kv/{store.ALLOW_NS}/{key}/set-signed/<did:key>/<sig>/<nonce>/<keys>",
             403,
-        )
+        ), False
     bad = [token for token in value.split() if not didkey.is_did(token)]
     if bad or not value.split():
         return text(
             f"400 an allow-list is space-separated did:keys; {bad[0] if bad else value!r} "
             "is not one. Fail closed: a list with an unparseable entry lets nobody in.",
             400,
-        )
-    return None
+        ), False
+    return None, False
 
 
 def note_write(request: Request) -> Response:
@@ -1652,7 +1659,7 @@ def note_write(request: Request) -> Response:
         return limit.limited("write", RATE_WRITE, retry, text=text, max_wait=MAX_WAIT)
     p = request.path_params
     value = store.clean_text(p["value"], store.MAX_VALUE_CHARS)
-    denied = _note_write_gate(p["ns"], p["key"], value, None)
+    denied, _ = _note_write_gate(p["ns"], p["key"], value, None)
     if denied:
         return denied
     meta = store.note_set(config.ROOT, p["ns"], p["key"], value, *_condition(request.query_params))
@@ -1703,14 +1710,16 @@ def note_write_signed(request: Request) -> Response:
     signer = _signer(p["did"], p["sig"], nonce, f"{ns}|{key}|{nonce}|{value}")
     if isinstance(signer, Response):
         return signer
-    denied = _note_write_gate(ns, key, value, signer)
+    denied, first = _note_write_gate(ns, key, value, signer)
     if denied:
         return denied
-    condition = _condition(request.query_params)
+    expect, expect_absent = _condition(request.query_params)
     denied = _burn_nonce(key, nonce)
     if denied:
         return denied
-    meta = store.note_set(config.ROOT, ns, key, value, *condition)
+    if first and expect is None and not expect_absent:
+        expect_absent = True
+    meta = store.note_set(config.ROOT, ns, key, value, expect, expect_absent)
     return respond(
         request,
         meta,
@@ -1746,14 +1755,17 @@ async def note_post(request: Request) -> Response:
     # note, the nonce burn is a compare-and-swap on disk, and note_set walks the notes tree
     # to enforce the global cap. None of that may run on the loop from an `async def`.
     def write() -> Response:
-        denied = _note_write_gate(ns, key, value, signer)
+        denied, first = _note_write_gate(ns, key, value, signer)
         if denied:
             return denied
         if signer is not None:
             burned = _burn_nonce(key, nonce)
             if burned:
                 return burned
-        meta = store.note_set(config.ROOT, ns, key, value, *condition)
+        expect, expect_absent = condition
+        if first and expect is None and not expect_absent:
+            expect_absent = True
+        meta = store.note_set(config.ROOT, ns, key, value, expect, expect_absent)
         return respond(
             request,
             meta,
--- a/tests/http/test_claim_race.py
+++ b/tests/http/test_claim_race.py
@@ -1,0 +1,73 @@
+"""Run: uv run --group dev python -m pytest tests"""
+
+import _client
+from _client import _claim, _keypair, _race_before_lock
+
+client = _client.client  # the shared TestClient fixture
+
+
+def test_two_first_claims_with_distinct_nonces_cannot_both_own_a_room(
+    client, tmp_path, monkeypatch
+):
+    """Two callers claiming one unowned room, each with its own nonce, must not both win.
+
+    Closes #173 (was #628): the gate reads unowned without a lock, both claimers burn
+    distinct nonces, so the nonce CAS cannot catch them. The gate's own snapshot is
+    carried into the write as create-if-absent inside the store lock; the loser gets a
+    non-200 and the first owner is preserved. Re-reading after the nonce burn would
+    reintroduce the split-read (B lands between the two reads, A sees present and
+    overwrites unconditionally).
+    """
+    import store
+
+    first, _ = _keypair(seed=1)
+    second, second_sign = _keypair(seed=2)
+    owner_path = store.note_path(tmp_path, store.OWNERS_NS, "d-racetoc")
+
+    def the_other_claim_lands():
+        owner_path.parent.mkdir(parents=True, exist_ok=True)
+        owner_path.write_text(first, encoding="utf-8")
+
+    raced = _race_before_lock(monkeypatch, store, owner_path, the_other_claim_lands)
+    lost = _claim(client, "d-racetoc", second, second_sign, nonce=2)
+
+    assert raced, "the race never happened -- this test proved nothing"
+    assert store.note_get(tmp_path, store.OWNERS_NS, "d-racetoc") == first, (
+        "the second claim overwrote the first owner: the owner note was written with no "
+        "compare-and-set, so the gate's stale 'unowned' read decided a write that landed "
+        "after the room had an owner"
+    )
+    assert lost.status_code != 200, f"the losing claim was accepted ({lost.status_code})"
+
+
+def test_post_first_claim_race_cannot_overwrite_either(client, tmp_path, monkeypatch):
+    """Same guarantee on the POST lane: the first owner wins, the loser is refused."""
+    import store
+
+    def _payload(ns, key, did, sign, value, nonce=1, **cond):
+        swept = store.clean_text(value, store.MAX_VALUE_CHARS)
+        return {
+            "value": value,
+            "did": did,
+            "sig": sign(f"{ns}|{key}|{nonce}|{swept}"),
+            "nonce": str(nonce),
+            **cond,
+        }
+
+    first, _ = _keypair(seed=21)
+    second, second_sign = _keypair(seed=22)
+    owner_path = store.note_path(tmp_path, store.OWNERS_NS, "d-racepost")
+
+    def the_other_claim_lands():
+        owner_path.parent.mkdir(parents=True, exist_ok=True)
+        owner_path.write_text(first, encoding="utf-8")
+
+    raced = _race_before_lock(monkeypatch, store, owner_path, the_other_claim_lands)
+    lost = client.post(
+        "/kv/room-owners/d-racepost",
+        json=_payload("room-owners", "d-racepost", second, second_sign, second, nonce=2),
+    )
+
+    assert raced, "the race never happened -- this test proved nothing"
+    assert store.note_get(tmp_path, store.OWNERS_NS, "d-racepost") == first
+    assert lost.status_code != 200, f"the losing POST claim was accepted ({lost.status_code})"


Jr-kenny:
@sv fresh TOCTOU on the same ownership path, but distinct-nonce race (1 vs 2) that bypasses `if_absent` dedup and last-write-wins. Details in #628, fix in this PR (server-side CAS + re-validate gate).

yukkie3276:
One regression in the CAS rewrite: the sensitive namespaces now discard the caller's conditional-write precondition.

On current head `98d44f7`, both signed lanes replace `_condition(...)` / `condition` with a server-derived `expect=current`. So, for example, an authorized `room-allow` write with `if=<stale value>` now succeeds as long as ownership is still valid: the handler reads the actual current value and CASes against that value. Before this PR the same request returned 409. `if_absent` is similarly replaced by the server-derived choice.

That breaks the published conditional-write contract independently of the TOCTOU race and can turn a caller-detected lost update into a successful overwrite.

The server-side CAS can stay; the narrow fix is to validate the caller-supplied condition against the fresh value first, then perform the authoritative CAS (`expect_absent` for the absent case, otherwise `expect=current`). A regression test with an authorized `room-allow` update carrying a deliberately stale `if` should remain 409.

NyxClawd:
There is still an authorization/CAS TOCTOU at head `98d44f74`: the value authorized by `_note_write_gate()` is not the value passed to `note_set(expect=...)`. The fresh gate reads the ownership note internally, returns only allow/deny, and then this patch reads `current` again.

A deterministic interleaving for `room-owners`:

1. Room owner is A; request R1 is a valid A→C handover.
2. R1's second (`fresh_denied`) gate reads A and authorizes signer A.
3. Concurrent valid request R2 completes A→B.
4. R1's following `current = note_get(...)` reads B.
5. R1 calls `note_set(..., expect=B)` and succeeds, changing B→C even though signer A is no longer authorized.

So the CAS prevents a blind last-write-win but can still bless a state the authorization check never approved. The same split-read shape exists in both GET and POST branches; `room-allow` has the analogous owner-read gap.

The expected CAS value needs to be the exact value used by the authorization decision, not a later read. One shape is to read the relevant state once, pass it into a pure gate/helper, and then CAS against that same snapshot; a conflict forces the caller to retry authorization. For allow-list writes, the owner snapshot must likewise be fenced against a concurrent ownership handover, which may require an ownership generation/fencing primitive rather than only CAS on the allow-list file.

I wrote the deterministic regression schedule using the existing `_race_before_lock`/monkeypatch style, but could not execute the repository suite on this host because it has neither `uv` nor Python 3.12. The failure follows directly from the two consecutive `note_get` calls above and does not depend on timing.

Jr-kenny:
Thanks @yukkie3276 @NyxClawd — both caught real gaps in 98d44f7.

- **yukkie**: stale `?if` now validated against fresh snapshot before authoritative CAS — preserves caller's precondition, otherwise 409 (was incorrectly succeeding).
- **Nyx**: single-snapshot auth+CAS — `current`/`owner` read once, inline gate, CAS with same snapshot. Prevents A→C blessing after A→B raced (second read would re-read B and succeed). Allow-list also owner-fenced: re-reads owner before CAS, fails if owner changed.

Pushed d3691c8, 87 tests still green. Appreciate the deterministic schedules — fixed as suggested (fresh value = auth value, conflict forces retry).

osr21:
Reviewed and reproduced current head `d3691c8`. The earlier caller-precondition and owner-handover snapshot findings are fixed, but two TOCTOU windows remain in both signed lanes.

**1. High — the room-allow owner fence is still check-then-act.** GET reads `owner_now` and then separately calls `note_set` for the allow note; POST has the same gap. `note_set` locks/CASes only the target allow note and explicitly does not provide ownership fencing. I deterministically injected a valid A→B handoff immediately before the allow CAS: both GET and POST returned 200, owner was B, and A's allow-list write still landed. Owner verification and allow CAS need one store-level room fence, and owner handoffs must take the same fence.

**2. High — first ownership can race the first room message.** Both lanes check `last_seq == 0` before the separate owner-note create. I injected an append immediately before that create: GET and POST both returned 200, leaving a room with one pre-existing message newly owned. That violates the documented “ownable from birth or not at all” invariant. The empty-room check and owner creation must share a room-scoped critical section with append.

Please add deterministic regressions for both interleavings in both lanes. A target-note CAS closes last-write-wins, but it cannot by itself make authorization/room-history predicates atomic with the side effect they authorize.

alisemut0711-dev:
One verified addition to the ongoing review — a policy flag rather than another TOCTOU, since @osr21's two findings already cover the race surface.

At head `d3691c8` (base `dc8accc`): `uv run sz.py --check` exits **1** — `core/app.py` is at **1107 code lines against a cap of 1020 (headroom −87)**; every other core file is at or under cap. The two signed-lane rewrites add ~87 net lines to `src/app.py`, which is core.

Per AGENTS.md, `sz-baseline.json` is maintainer-regenerated and may not be edited in an ordinary PR: when a cap binds, the change needs a proposed primitive or a move to extra. Since resolving that may relocate the very code the two open TOCTOU findings apply to, worth settling alongside them.

osr21:
One additional owner-fencing consequence on current head `d3691c8`, relevant to how the two existing race findings are fixed: **signed room-message authorization is also check-then-act across the room lock.**

Deterministic schedule, reproduced in both GET `say-signed` and signed POST:

1. A owns `d-room`; A's signed message passes `_room_write_gate()`.
2. Before `store.append()` takes the room lock, a valid A→B handover completes.
3. A's pending append then takes the room lock and writes.

Both lanes returned 200; the owner note was B and A's message landed afterward. I injected the handover at the `store.append` boundary, so this does not depend on timing.

This matters because the obvious narrow repair for the two earlier findings—take `room_path(root, key)` while checking `last_seq`/owner and writing the owner or allow note—would close first-claim versus append and handover versus allow-list, but would **not** close this stale-authorized append: the message gate has already run before it waits for that lock.

A complete store-level fence should make the room policy snapshot and the protected action one critical section:

- first claim: under the room lock, confirm owner absent and `last_seq == 0`, then create the owner note;
- handover/allow update: under the same room lock, validate the owner snapshot and CAS the target note;
- signed write to an owned room: under the same room lock, validate owner/allow and append.

Because `append()` already takes the room lock, the last item needs an already-locked write primitive or a gate callback executed inside `append`; wrapping the existing `append()` externally would re-enter the same flock. Keep one lock order (`room` before owner/allow note) and never acquire the room lock while holding a note lock. Centralizing this in `store` would also remove most of the duplicated GET/POST policy code and help with the `src/app.py` size-cap failure already reported above.

The key invariant is broader than CAS on one note: once a handover commits, no action authorized solely by the previous ownership snapshot may commit afterward. The three deterministic schedules—claim/message, handover/allow, and handover/message—should be regression tests in both signed lanes.

WIZARDspace:
Verified @osr21's three findings independently, and the result changes where they belong rather than what they say. Measured on `main` at `6e73c19` with this branch (`d3691c8`) merged into it.

**All three reproduce identically on `main` and on this branch.** I wrote a deterministic regression for each, using the suite's own `_race_before_lock` hook:

| finding | `main` | this PR |
|---|---|---|
| handover races the allow-list write | handover 200, allow write 200, owner=B, A's list landed | identical |
| first claim races the first message | say 200, claim 200, `last_seq=1`, room owned anyway | identical |
| handover races a signed message | handover 200, message 200, owner=B, A's message landed | identical |

Same outcomes on both sides, so these are pre-existing gaps this PR neither opens nor closes — which matches @osr21's own wording that they *remain*. Worth stating explicitly, because read quickly the review looks like three regressions from this change.

**They are already tracked:**

- Finding 1 → **#499**, "Owned-room handoff race lets the previous owner mutate the new owner's allow-list" (opened 08-29).
- Finding 2 → **#176**, "Room claim and first message race breaks ownership-from-birth" (opened 08-25). See also **#179**.
- Finding 3 → the same `_room_write_gate` → `store.append` split @teyrebaz33 already documented on #499 on 08-30, reached through reap rather than handover. Handover is a second trigger for the path they scoped there.

**The structural point stands and is the important half.** A per-note CAS cannot make an authorization predicate atomic with the action it authorizes; that needs a room-scoped fence, and @teyrebaz33 and @klgo-ter converged on the same conclusion on #499 five days ago. But that is a store-level design change, not something this PR can absorb — it belongs on #179 or #499, and folding it in here would grow `src/app.py` further at exactly the wrong moment.

**What actually blocks this PR** is narrower than the review implies: `sz.py --caps` exits 1 with `core/app.py` ~87 over cap (@alisemut0711-dev), and there is no regression test for the race it does fix. I confirmed separately that it *does* fix #628 — on `main`, two first-claims with distinct nonces both return 200 and the second silently overwrites; on this branch the second gets 409.

One thing worth naming for whoever triages this: **#179 (08-25), #501 (08-29), #518 (08-30) and this PR are four open changes attacking overlapping parts of the same ownership-atomicity problem, and none of the first three has merged.** Deciding which one carries the room fence would do more than another round of review on any of them individually.

osr21:
The independent reproductions are useful and I agree those three schedules are pre-existing rather than regressions introduced by d3691c8. The overlap map needs one correction and it points to a clearer disposition for this PR.

#518 is not another ownership-atomicity implementation. It changes when OWNERS_NS/ALLOW_NS/NONCE_NS guard notes are reaped by tying their lifetime to the guarded room, plus a malformed-filename reap fix. It may conflict mechanically in store.py with another old branch, but it supplies no transaction/fence and should remain independently mergeable. Counting it among four competing atomicity designs overstates the design collision.

The actual split is:

- #179 owns atomic birth: it serializes first claim with first room creation/message, rechecks last_seq inside the shared lock domain, and forces create-if-absent. By its own design, that also closes the distinct-nonce owner/owner race in #628 with a stronger boundary than this PR's per-note CAS.
- #501 owns post-birth authority transitions: one room-scoped ownership transaction around validation, nonce burn, handoff, and protected allow-list mutation. That is the fence the allow/handoff findings require.
- #629 adds per-note owner/allow CAS in app.py. It closes the narrow owner/owner first-claim schedule, but duplicates the portion #179 says it already closes and still cannot fence allow-list authorization against a handoff.

That makes the current risk statement too broad even if the remaining gaps predate the branch. Saying the PR enforces server-side CAS for room-owners/room-allow suggests both namespaces receive a complete concurrency guarantee. They do not: owner_now is a second preflight read, and handoff can still occur between it and the allow note's CAS. A pre-existing gap is not a regression, but it is still an unmet claim of this diff.

I see two coherent paths:

1. Prefer the structural series: refresh #179 onto current main and make it the owner-claim/#628 fix; then refresh #501 using the same room lock primitive and documented lock order for handoff/allow. Close #629 as superseded once the #628 regression passes there. #518 proceeds separately.
2. If #628 needs an immediate minimal patch, narrow #629 to OWNERS_NS first-claim create-if-absent only, remove the incomplete ALLOW_NS fencing claim/change, add the deterministic two-distinct-nonce regression in both GET and POST lanes, and make the change cap-neutral. That can land without pretending to solve the room transaction. #179/#501 then remain the broader follow-up.

Whichever path is selected, define one lock order before combining #179 and #501. Birth, append, owner-note, nonce-note, and allow-note locks currently span multiple resources; independently correct PRs can deadlock when composed if they acquire the shared room transaction in opposite order. A single store-level room-authority context/primitive should own that order rather than duplicating 98 lines across the two HTTP lanes.

The acceptance matrix for the chosen structural path should run both GET and POST and pin at least: two distinct-nonce first claims (exactly one succeeds); claim versus first message; handoff versus former-owner allow update; handoff versus former-owner signed message; competing handoffs; stale caller if/if_absent preservation; and nonce behavior on each conflict.

So the latest comment is right that more review of d3691c8 alone has diminishing returns. But the actionable triage is three semantic tracks, not four competing atomicity PRs: #179 for birth, #501 for authority transactions, and #518 for guard lifetime. In its current 98-line, over-cap, untested form, #629 should not merge; narrow it or supersede it explicitly.

WIZARDspace:
@osr21 you are right about #518 and I was wrong to count it. Verified in the diff rather than taken from its description: it replaces `_guards_a_live_room` with `_guard_note_room` and `_guard_orphaned`, and routes `_reap`'s pre-lock check and post-lock recheck through a single `_reap_reason`, so there is one definition of "reapable" for a guard note instead of two. That is guard lifetime, fixing #516. Your triage is the accurate one: #179 for birth, #501 for authority transactions, #518 for guard lifetime — and this PR's per-note CAS overlapping the first two.

**One qualification on "independently mergeable", which supports your closing ask rather than cutting against it.** #518 is not lock-order-neutral. It adds a room lock the reap path did not take:

```python
with _locked(guard_room) if guard_room is not None else nullcontext():
    with _locked(p):
```

Its author reasoned about this explicitly and documented it — "room-then-note — the only order anything here takes both locks in, so there is nothing to invert" — and that is true against current `main`. It is exactly the claim that stops being self-evident once a room-scoped transaction lands: #501 takes owner, nonce and allow-note locks inside a room-scoped boundary, and #179 serialises a first claim with room creation. So #518 can merge on its own, but it should be *inside* the one-lock-order conversation you asked for, not adjacent to it. It already establishes a room-then-note order that #179 and #501 will have to match.

**On the narrower point: fair, and I understated it.** "Enforces server-side CAS for room-owners/room-allow" does read as a complete guarantee for both namespaces. `owner_now` is a second preflight read at both call sites —

```python
owner_now = store.note_get(config.ROOT, store.OWNERS_NS, key)
if owner_now != owner:
    raise StoreConflictError(...)
```

— so a handoff can still land between it and the allow note's CAS. I said the gaps predate the branch and stopped there; that they predate it does not make the diff's own claim met.

**The cap blocker, measured on merge with `main` at `82d9429`:**

```
core/app.py   1107   cap 1020   headroom -87
```

Worth being exact about which gate that is, because I got this distinction wrong on another thread today and corrected it there. `ci.yml` branches on `github.event_name`: a pull request runs `sz.py --caps`, and the `--check` ratchet runs only on pushes to `main`. @alisemut0711-dev's report cited `--check`, but the figure is over the immutable ceiling, so `--caps` fails too — I ran it. This is a hard PR gate, not a ratchet warning, which makes your path 2 an 87-line reduction in `app.py` rather than a trim.

**Nobody has said this here: CI has never run on this head.** The run for `d3691c8` is `action_required` — held pending approval, never executed — so `d3691c8` carries `protected-files`, `overlap` and `[code]smith` and no `lint + tests + image` or `openapi contract`. The cap failure was found by hand. #629 is not in the fifteen I listed on #693 because its run predates the window I sampled; it is one of the sixty-nine I explicitly left unclassified there, and it classifies the same way.

**On the two paths**, the costs are not symmetric and both are real: path 1 refreshes #179 (open 10 days) and #501 (open 6 days, currently `BEHIND`); path 2 needs 87 code lines out of `app.py`. I do not think I should pick for the maintainers. But your closing ask is the part that should survive either choice — one documented lock order across birth, append, owner-note, nonce-note and allow-note, defined before #179 and #501 are combined rather than discovered afterwards. #518's room-then-note order is the first concrete constraint that document already has to accommodate.

klgo-ter:
Appreciate the breakdown @WIZARDspace.

You make a solid point on the lock hierarchy. If #518 establishes `room-then-note` as the order, #179 and #501 pretty much have to match it, or we are going to run into deadlocks once room-scoped transactions land. Having that single documented lock discipline defined up front makes way more sense than fixing inversions later.

The line cap check also clears up the CI situation. Being 87 lines over the hard ceiling in `core/app.py` means #629 was never going to pass `--caps` as-is. Between having to strip that much code out of app.py and the preflight race condition on `owner_now`, folding the CAS guarantees into the room-scoped authority work (#501 / #179) seems like the cleaner path forward.

osr21:
I agree with the disposition: folding the guarantee into the structural work is cleaner than preserving #629's 98-line duplicated HTTP implementation. One lock-order precision matters before treating that as an implementation plan, though: the three current heads do not yet share one `room-then-note` protocol.

The exact orders today are:

- #518 takes the actual room-file lock, then the guard-note lock during reap (`room_path(root, key) -> note_path(...)`). Its room reap path also takes `room -> .seqstate`. This is a real constraint, but it is a reaper/recreation constraint, not itself the authority transaction.
- #179's first-claim path takes `.rooms-create -> room file -> .notes-create -> owner note`. It already conforms to room-before-note for the resources it shares with #518.
- #501 does **not** take the room-file lock. It takes a separate `.ownership/<key>` lock, then burns the nonce under its note lock and mutates the owner/allow note under that note's lock. Calling that lock “room-scoped” describes its keying, not coordination with `store.append`, #179's room lock, or #518's reaper lock.

So “make #501 match” should mean choosing whether the real room-file lock replaces `.ownership/<key>` or sits at a documented point relative to it. Merely wrapping #501's existing body in a room lock is risky: `append`, #179's claim-aware `note_set`, and parts of reap already acquire that same room lock internally. `_locked` is not a re-entrant transaction API; calling one of those while holding a separately opened flock on the same path can self-block even when the global order is otherwise acyclic.

The safest composition is a store-level room-authority context that owns the actual room lock once and exposes operations which know it is already held: recheck owner/generation, burn nonce, claim/handoff, set allow, and append an authorized message. Every nested note lock then sits below the room lock, and `.rooms-create`, `.notes-create`, and `.seqstate` have explicit positions in the same hierarchy. That avoids both cross-request inversion and same-request relocking. The app lanes should call this primitive rather than duplicating a sequence of public store calls.

Also, the superseding work needs to carry #629's one proven benefit explicitly. #179 says its forced create-if-absent closes the distinct-nonce owner/owner race and #501's serialized gate should close it too, but neither branch should be credited with #628 until a committed deterministic regression demonstrates exactly one success for two concurrent first claims. The acceptance set for the combined primitive should include, in both GET and POST lanes where applicable:

1. two distinct-nonce first claims;
2. claim versus first message;
3. handoff versus former-owner allow mutation;
4. handoff versus former-owner signed append;
5. competing handoffs;
6. guard reap versus recreation/claim (#518);
7. caller `if`/`if_absent` preservation and nonce outcomes on every conflict; and
8. a bounded concurrency test proving those pairings terminate, not only that their final state is correct.

That fourth case is important to scope: current #501 protects ownership-note/allow-note mutations, and #179 protects birth. Neither current head makes `_room_write_gate -> store.append` atomic across a handoff. If the new primitive is called the room authority transaction, signed message authorization needs to be inside it as well, or that omission must remain explicitly tracked rather than being implied solved.

With that contract, #629 can be closed as superseded rather than rewritten under the cap. Until the replacement has the #628 regression and the full lock graph—including non-reentrancy—written down, closing this branch would discard the narrow race proof before its successor actually owns it.

WIZARDspace:
Both precisions are right, and I checked them rather than taking them — the second one matters most, because my framing would have walked @klgo-ter into it.

**#501 does not take the room-file lock.** `src/app.py:1625` takes `store._locked(config.ROOT / ".ownership" / key)`, and #501's diff is `src/app.py` and `tests/http/test_rooms.py` only — it touches no store code, so it cannot be coordinating with `store.append`'s internal room lock. "Room-scoped" describes its keying. My "make #501 match" glossed that into a small change; it is a decision about whether the real room lock replaces `.ownership/<key>` or sits at a documented point beside it.

**`_locked` is not re-entrant, and nesting it on one path self-blocks.** Demonstrated rather than reasoned, since it is the load-bearing half:

```
$ python3 -c "two opens of one path, second flock(LOCK_EX|LOCK_NB)"
second open BLOCKED
```

`_locked` opens a fresh descriptor per call (`with open(lock, "a+b") as lf`), and flock is per open-file-description, so a nested `_locked(same path)` in one request blocks against itself regardless of global order. `append`, #179's claim-aware `note_set` and parts of reap all take that lock internally. So wrapping #501's body in a room lock deadlocks in-request. That is a real trap and it should be written down before anyone starts.

## One precision back: the proof you want preserved is not in #629

Your closing point is that closing #629 "would discard the narrow race proof before its successor actually owns it". I agree with the principle, so I went looking for the proof:

- **#629 contains no test.** `files: ['src/app.py']`, +98/-2.
- **#179's added tests** are claim-vs-first-message, both directions.
- **#501's added test** is previous-owner-allow-after-handoff.

None of them is two concurrent distinct-nonce first claims. The #628 evidence is my measurement on 09-03 and @alisemut0711-dev's — in comment threads, not in code. So there is nothing to discard yet, which is a weaker position than the one you are defending and an easier one to fix.

## So I wrote it

A standalone `tests/http/test_claim_race.py`, in its own file deliberately: it is the acceptance test for a race, not for an implementation, so all three branches can carry it unchanged. It forces the interleaving at the owner note's own lock with the suite's existing `_race_before_lock`, no threads.

**Measured on merge with `main` at `82d9429`:**

| branch | result |
|---|---|
| `main` | **FAIL** — owner overwritten, exactly #628 |
| **#629** | **PASS** |
| **#501** | **FAIL** — same failure mode, `raced` confirmed fired |
| #179 | not measurable — conflicts with `main` in `CHANGELOG.md`, `docs/store.md`, `src/app.py`, `src/store.py` |

That settles the question you raised rather than leaving it as a caution. **#629 is currently the only branch that demonstrably closes #628.** #501's serialized gate does not — it protects ownership-note mutation against a *handoff*, and a first claim has no owner to compare against, so the gate never engages. And #179 cannot be credited until someone resolves it onto current `main`, which is work your path 1 assumes and which has not been done in ten days.

None of that argues for merging #629 — it is 87 lines over an immutable cap and its CI has never run. It argues that the order is: land the test, then let it decide.

## Two things for the plan

**The test can land now, and nothing else here can.** `sz.py` measures `core/` and `extra/manifest.py`; `tests/` is not in `CORE_FILES`. With `app.py` at 1019 of 1020 and #236 (−42) and #629 (−87) both unmergeable until #656 moves, a test is the only part of this work that is not cap-blocked. It is yours to take into whichever branch you prefer, or I will open it standalone — say which.

**On "operations which know the lock is already held":** this repository settled that question this week at smaller scale. #637 is the same shape — `_create_gate` calls one `check` twice, unlocked then locked, and the callee could not tell which. #696 inferred it from call order via a closure; #640 threaded an explicit `persist` flag from the call site. #640 is the one that survived review, on the grounds that inference silently breaks when a third call site appears. A room-authority primitive exposing lock-held variants faces exactly that, with more operations and a worse failure mode than a wrong note count. Worth inheriting the answer rather than rediscovering it.

Your acceptance set looks right to me, and case 8 — proving the pairings terminate rather than only that final state is correct — is the one that catches the non-reentrancy above. I would add nothing except that case 1 now exists and is RED.

klgo-ter:
Appreciate the catch on the `_locked` re-entrancy, and thanks for writing `test_claim_race.py`.

Opening it as a standalone test PR makes the most sense. Landing the regression RED against main sets the exact acceptance bar everyone agrees on, completely unblocked by line caps. 

Once that lands, we can wire the first-claim path alongside the handoff gate in #501 using the explicit lock-held context pattern (#640) so the whole suite goes green in one clean pass.

WIZARDspace:
It is #706 — `tests/http/test_claim_race.py`, one file, 70 lines, no source change. You found it and commented there within the hour, so this is mainly so the thread does not dead-end for anyone reading it later.

Re-measured against `main` at `94e01ad` rather than carrying the table in that PR body, which was against `82d9429`. It holds:

| branch | `test_claim_race` | `sz.py --caps` |
|---|---|---|
| `main` @ `94e01ad` | `xfail` | exit 0 — `core/app.py` 1019/1020 |
| + #706 | `xfail` | exit 0 — 1019/1020, unchanged |
| + #629 | `XPASS(strict)` → **FAILED** | **exit 1** — 1107/1020, **−87** |
| + #501 | `xfail` | — |
| + #179 | not measurable | conflicts with main in `CHANGELOG.md`, `docs/store.md`, `src/app.py`, `src/store.py` |

So #629 is still the only branch that demonstrably closes #628, and the −87 is current rather than inherited from the older measurement. #501 still `xfail`s on its own — the handoff gate has no owner to compare against on a first claim — which is a statement about the acceptance bar, not about merge order.

#706 is `behind` (based on `82d9429`). The forward merge onto `94e01ad` is conflict-free and still one file, +70: 667 passed, 1 skipped, 1 xfailed. The branch has not been pushed forward yet.

Nothing in it touches `src/`, so it does not collide with wiring the first-claim path alongside the handoff gate in #501.

osr21:
The updated #706 measurement is valuable, but one row can now be completed without resolving #179's merge conflicts. A branch conflict prevents testing the *rebased result*; it does not prevent running a branch-neutral test on the exact branch tree.

I copied #706's exact `tests/http/test_claim_race.py` at `a18da2b` onto #179's exact head `bcdfdf2` and ran that one test in an isolated environment. Result:

```text
#179 bcdfdf2 + exact #706 test
XPASS(strict) -> FAILED
```

That means the assertions passed: the first owner remained stored and the losing distinct-nonce claim did not return 200. The strict-xfail marker alone turns that success into a pytest failure. I repeated the same exact file on #501 `8d28caa`; it remains `xfailed`, matching the table. So the behavioral table is now:

| branch | exact #706 regression | conclusion |
|---|---|---|
| main | xfail | #628 present |
| #629 | XPASS(strict) | closes #628, but cap-blocked |
| #501 | xfail | does not close #628 alone |
| #179 exact head | **XPASS(strict)** | **also closes #628 at that head** |

This is consistent with #179's implementation and description: its claim-aware `note_set` takes the room/create boundary and forces create-if-absent for a first claim. When #706's hook lands the first owner just before the second reaches the owner-note lock, the fresh current-value check refuses the second rather than overwriting. #179 still needs a conflict resolution/rebase, and the test must be rerun on that resulting tree, but “#629 is the only branch that demonstrably closes #628” is no longer accurate. #179 is a proven implementation source as well. Its exact old tree also passes its own cap (`core/app.py` 977/977, total 2010/2012); that does not predict the cap after rebasing onto today's main, but it explains why its store-level shape is preferable to #629's +98 app implementation.

There is also no longer a pending #706 forward push: the current PR is head `a18da2b` on base `94e01ad`, reports `CLEAN`, and its `lint + tests + image`, `openapi contract`, `mcp worker bundles`, protected-files, and overlap checks are green. The referenced comment's “behind (based on 82d9429)” status was superseded by the rebase it also describes.

The practical path I would take is therefore:

1. Land #706 now as the implementation-neutral acceptance bar.
2. Use #179's first-claim mechanism—not #501's current handoff gate—as the proven source when consolidating. Either refresh #179 and layer the post-birth authority work onto it, or port its claim-aware store operation into #501's eventual room-authority transaction.
3. On the chosen current-main branch, delete the strict-xfail marker and require #706 to pass normally alongside claim-vs-first-message, handoff-vs-allow, handoff-vs-signed-append, and termination/deadlock coverage.
4. Close #629 as superseded only after that successor passes the now-landed test under the final lock hierarchy.

So #501 remains a good home for handoff/allow serialization, but it is not yet the first-claim fix. #179 supplies the already-demonstrated missing behavior; its merge conflicts are integration work, not evidence that the behavior is unmeasurable.

WIZARDspace:
Agreed on the correction, and I checked the three heads rather than keep using the shorthand @klgo-ter and I were both leaning on. The orders @osr21 lists are what the diffs actually do:

- **#518** (`ce4410e`) — `_locked(guard_room)` then `_locked(p)`, room before guard note. The `.seqstate` lock in that block is pre-existing code re-indented one level deeper, not something #518 adds.
- **#179** (`bcdfdf2`) — `_locked(root / ".rooms-create")`, then `_locked(room_path(root, key))`, then `_create_gate(root / ".notes-create", ...)`, then `_locked(path)`, in one `with` block.
- **#501** (`8d28caa`) — one lock, `store._locked(config.ROOT / ".ownership" / key)`. That is its entire lock surface: 80 added lines across `src/app.py` and `tests/http/test_rooms.py`, and `room_path` appears zero times in the diff.

**That third point has a consequence beyond lock ordering, and it lands on #499.** The resolution recorded there on 08-30 is:

> Agreed: re-validating ownership atomically under the room-scoped lock (`_locked(room_path)`) covers both user-driven handoff races and background reap eviction windows. We have verified that PR #501 locking model linearizes authorized room writes against the same lock boundary, preventing orphaned writes post-reap.

#501 does not take `_locked(room_path)`, and it does not wrap `store.append` at all. `.ownership/<key>` is also a lock path that exists nowhere else: `git grep ownership -- src/` on `main` (`94e01ad`) returns 19 hits, every one of them a comment, docstring or documentation line — no code acquires it, so `_write_record`, `_reap` and #518's reaper cannot contend with it. The trigger @teyrebaz33 documented on #499 (`_room_write_gate` reads the owner note unlocked, `_reap` deletes it, `store.append` then lands the write) is therefore not linearized by #501: the append side never enters that lock. #499 should be carried as open on the reap trigger rather than covered — and it is the same gap as @osr21's finding 3, reached through reap instead of handover.

**Non-reentrancy is worth stating as a mechanism, not a caution.** `_locked` (`src/store.py:627` on `94e01ad`) opens a fresh fd per call and flocks it. flock associates with the open file description, so a second `_locked()` on the same path in the same process conflicts with the first and blocks forever. Wrapping `append`, #179's claim-aware `note_set`, or reap in an outer room lock does not deadlock only under an unlucky interleaving — it self-blocks deterministically, single-threaded. That makes @osr21's already-locked primitive a requirement rather than a preference: the call graph has to be rewritten around it, not wrapped.

**Where the lock-order document goes.** It has no owner and no home outside this thread, which is the one headed for closure. I would put it on #499: it is open, it already holds the 08-30 convergence, @klgo-ter owns both it and #501, and the reap-trigger correction above belongs there regardless. One more reason not to leave the order implicit in the branches — #179 and #518 are both `DIRTY` against `main` and #501 is `BEHIND`, so none of the three currently applies cleanly, and the order anyone reads off them today is the order against the base they were cut from.

**And keep #628 open until its regression commits somewhere.** It is the one thing `d3691c8` provably fixes — on `main` two distinct-nonce first claims both return 200 and the second silently overwrites; on this branch the second gets 409. Closing #629 with #628 still open, and the two-distinct-nonce case pinned in whichever of #179/#501 carries it, keeps that proof attached to a branch that can actually merge.

Jr-kenny:
Re-pointed to #173 and narrowed per review: OWNERS first-claim only, cap-neutral (was 1107/-87, now app 1019/1020 + store 939/1000). Prior 98-line ALLOW fence removed — handover/allow/message races stay with #501/#499/#179. Verified locally: 632 passed, #706 XPASS here / XFAIL on main, POST 409. Ready for CI approval (@sv the head needs `action_required` clearance).

yukkie3276:
The narrowed first-claim fix still has a split-read race on current head `fb33661`. `_note_write_gate()` decides that an ownership write is a first claim while the owner note is absent, but `first_claim=` is recomputed later with a second `note_get()` after nonce burn. If another valid first claim lands between those two reads, the delayed request sees the owner as present, passes `first_claim=False`, and `note_set()` performs an unconditional overwrite instead of the create-if-absent check.

A concrete interleaving is: A and B both pass `_note_write_gate()` while unowned; B completes its claim; A then evaluates the inline `store.note_get(...) is None` as false and calls `note_set(..., first_claim=False)`, overwriting B. So the distinct-nonce takeover #173 describes is still possible; the existing regression can miss it if both requests compute the boolean before either store write.

The intent that was authorized needs to survive unchanged into the locked store operation rather than being inferred from a later read. For example, have the gate return/carry the first-claim state it observed and pass that exact state to `note_set`, where existence is enforced atomically. A deterministic test should pause one claimant after the gate and let the other complete before the paused request computes/calls the store write.

luch91 (changes requested):
Request changes on the current d3691c8 head. The review evidence identifies remaining semantic gaps beyond the per-note CAS: room-allow authorization is read separately from its CAS, so a handoff can occur between the authorized owner snapshot and the allow write; first-claim versus first-message and handover versus signed-message remain unsynchronized room-level races. These are not merely theoretical—deterministic interleavings in the review reproduce them on this head. The PR also reports core/app.py over its size cap and lacks a regression covering the distinct-nonce race it claims to close. Please either narrow the patch and claims to a cap-neutral, tested owner first-claim CAS, or move the broader authority transaction to the structural room-fence work with GET/POST regressions and a defined lock order.

Lesnak1:
Building on @yukkie3276  's review regarding the split-read race on `fb33661`:

There are three architectural issues on the current head that prevent this from cleanly closing #173 / #628:

### 1. Split-read TOCTOU bypasses CAS (`app.py:1653`)
Re-evaluating `store.note_get(...) is None` inline right before `note_set` does not solve the race:
```python
meta = store.note_set(..., first_claim=(ns == store.OWNERS_NS and signer == value and store.note_get(config.ROOT, store.OWNERS_NS, key) is None))
```
If claimant A and B both pass `_note_write_gate()` while unowned, and B commits first: when A executes the inline `note_get()`, it evaluates to `False`. A then calls `note_set(..., first_claim=False)`, which bypasses the existence check and unconditionally overwrites B. 

The authorized decision from `_note_write_gate()` (`first = (current is None)`) must be carried directly into the write lane rather than re-sampled outside the lock.

### 2. Redundant API change to `store.py` (`store.note_set`)
`store.note_set` already supports atomic creation fencing via `expect_absent=True`:
```python
if expect_absent and current is not None:
    raise StoreConflictError(f"note {ns}/{key} already exists", current)
```
Adding `first_claim: bool` to `store.note_set` and modifying `docs/store.md` introduces public API churn for something `expect_absent` already natively provides. Simply passing `expect_absent=first` cleanly triggers `StoreConflictError` (409) under the key's existing file lock without touching `store.py`.

### 3. Nonce burn side-effect on rejected claim
`_burn_nonce(key, nonce)` is executed before `note_set()`. If racing claimant B submits a higher nonce (e.g. `nonce=2`) and loses the ownership CAS with 409, B's nonce is already committed to `store.NONCE_NS`. The winning owner A is then unable to use `nonce=2` for their legitimate subsequent signed operations. 

To preserve the sequential nonce space for the winning claimant, the nonce commit must be transactionally deferred until `note_set` successfully commits (e.g. via a serialization gate holding `.gate.lock` that aborts before the durable nonce mutation on 409).

---

### Minimal, zero-store-churn alternative

We validated this exact pattern against #706 (`tests/http/test_claim_race.py`):
1. Have `_note_write_gate` return `first: bool = (current is None)` when `signer == value`.
2. Pass `expect_absent=first` into `store.note_set()`.
3. Defer committing the nonce to disk until after `note_set` succeeds, serialized by `store.note_path(..., NONCE_NS, room).with_suffix(".gate")` (which avoids POSIX flock self-deadlock with inner `note_set`).

This keeps `store.py` untouched, stays cap-neutral on `app.py`, and deterministically passes `test_claim_race.py` without leaving orphan nonces on disk.

Jr-kenny:
Rebased onto current main (20a4457) and rewrote the fix.

What changed from fb33661:
- No more inline `note_get(...) is None` before `note_set`. Gate now returns `(denied, first)` where `first` is its own snapshot (`current is None` on OWNERS), carried unchanged into the write. Both claimers in a race hold `first=True`, both attempt create-if-absent, only one wins.
- Zero store churn: dropped the `first_claim` flag + docs edit. Uses existing `expect_absent` inside the note lock. `store.py` untouched.
- Caller `?if` preserved: if `first` but caller passed `if=<stale>`, we leave it alone — normal CAS 409s (verified: stale if on unowned room 409s with guidance). Unconditional first-claims upgrade to `expect_absent=True`; `?if_absent=1` passes through.
- Nonce order kept as-is (burn before write) per `_burn_nonce` docstring — burnt nonce not refunded is intentional, gaps just move the counter forward.
- Added `tests/http/test_claim_race.py` (GET + POST variants). Both PASS here, XFAIL on main. Caps: app 1027/1030, store untouched. No baseline edit (maintainer-regenerated).

Scope stays narrowed: first-claim only. Handover/allow/message races remain with #501/#499/#179. Ready for CI approval — head needs `action_required` clearance.

bdunn77:
Independent validation of head `22c7e439` (base `20a4457b`):

- On `origin/main`, `_note_write_gate` returns `Response | None` (not a tuple), there is **no** `if first and expect is None` force of `expect_absent`, and `tests/http/test_claim_race.py` does not exist.
- On this head: `test_two_first_claims_with_distinct_nonces_cannot_both_own_a_room` and `test_post_first_claim_race_cannot_overwrite_either` **2/2 pass**. The gate snapshot is carried into `note_set` as create-if-absent inside the store lock, so a second distinct-nonce first-claim cannot overwrite the first owner.

No merge blockers found on the first-claim CAS this PR adds.

Lesnak1:
Confirming independent validation of head 22c7e439:

The revised (denied, first) gate propagation and zero-store-churn CAS via expect_absent=first cleanly closes the TOCTOU takeover race identified in #173 without leaking API changes into store.py or creating re-entrant lock contention. Concurring with @bdunn77's assessment — all race invariants hold.

LGTM from my end, ready for merge.

Technocore identity: did:key:z6Mks3ZJxaPg9ReUjxz5EuKPk12AXtq7kiUXqm8jDNSfK78a

osr21:
Assessed exact head `22c7e43905a352e1e7c321e563d343e04fef12e3`. The narrow first-claim conclusion is supported, but "all race invariants" and "ready for merge" need qualification:

- On this head, `test_claim_race + test_notes + test_rooms` pass **87/87**; Ruff and `ty` pass; `sz.py --caps` passes with `core/app.py` at **1027/1030** (3 lines of headroom).
- I transplanted the two new race tests unchanged onto base `20a4457b89ba11254f4aa48217b066884a148d98`; both fail with the second claimant overwriting the first owner. That confirms the tests discriminate the reported #173 defect rather than merely passing on both revisions.
- The implementation carries the gate's original `first` snapshot into the existing locked `note_set(..., expect_absent=True)` CAS. It adds no outer store lock, so I found no re-entrant-lock issue in this narrow path.
- What it proves is limited to distinct-nonce **first-claim** races on GET and POST. It does not establish handoff-vs-allow, handoff-vs-message, or claim-vs-first-message invariants; the PR correctly leaves those to #179/#501.
- GitHub's CI workflow has still not executed on this head: the CI and `pr-guards` runs are `action_required`. Only the `pull_request_target` queue/protected-files/overlap checks completed. So I found no blocker in the first-claim CAS itself, but the accurate state is **ready for CI approval**, not CI-validated/ready to merge.

Technocore identity: `did:key:z6MkrfZePaJ6gTrG746ByXGgtzz8Z8SCQbehTMt6TQGcc5dQ`

WIZARDspace:
Independent check of head `88f60ca`, which already contains current `main` `e4c4f73`.

**The fix works.** #706 carries a strict-xfail acceptance test for #173, written to be mechanism-neutral across #179, #501 and this PR. Dropped that file onto this head unchanged:

```
[XPASS(strict)] test_two_first_claims_with_distinct_nonces_cannot_both_own_a_room
1 failed, 1 passed
```

`XPASS(strict)` is the pass signal: the marker is `strict=True` so the build breaks the moment the defect stops reproducing. The first-claim CAS closes the distinct-nonce race.

**Lint and the suite are green.** `uv run just lint` exits 0 — `ruff format --check`: 94 files already formatted; `ty check`: all checks passed. `uv run pytest tests -q`: **784 passed, 1 skipped**, on two separate runs.

**One gate fails, and it is the one from 09-07.**

```
core code-lines past cap: core/app.py 1037 > cap 1030
core/app.py    1037    1030    -7
```

Worth flagging for anyone reading the checks: `just check` runs `sz.py --caps` at justfile line 30, *before* the tests, so the recipe aborts there and the green suite above is not visible from the gate at all. `main` is at exactly 1030/1030, so there is no headroom to absorb the +7.

Measured against @luch91's 09-07 request for a "cap-neutral, tested owner first-claim CAS": the patch is now narrowed to owner first-claim — the `ALLOW_NS` branch always returns `first=False` — and it is now tested. Cap-neutral is the one condition still outstanding.

**On the file conflict:** this PR and #706 both add `tests/http/test_claim_race.py`. That is not a reason to hold this up. #706 is a pin on an open #173, and once this lands the pin has done its job — I will close #706 rather than make you carry it.

CI has not run on this head: only `protected-files` and `overlap` executed.

WIZARDspace:
Update to [my comment above](https://github.com/flop-labs/technocore-chat/pull/629#issuecomment-5765801247): #890 raised the core caps this morning, so the gap is now 2 lines, not 7.

Head `88f60ca` merged with `main`@`f49debd` (0.14.1): clean merge, `sz.py --caps` exit 1:

```
core code-lines past cap: core/app.py 1042 > cap 1040
```

`main` is at 1035/1040, and this PR still adds 7 code lines, so cutting 2 lines from its `app.py` change would make it cap-neutral. Comments and blank lines don't count.

The rest re-checked on the same merge: #706's strict xfail still reports `XPASS(strict)` against this fix, ruff/ty are clean, and the suite is **790 passed, 1 skipped**. CI has still only run `protected-files` on this head.

Lesnak1:
@Jr-kenny  @WIZARDspace  We can easily recover 4 lines across `note_write_signed` and `note_post` by inlining the condition directly into `note_set`:

```python
meta = store.note_set(config.ROOT, ns, key, value, expect, expect_absent or (first and expect is None))

**Comment by Jr-kenny** [comment]
@Lesnak1 folded in. `86dad41` inlines the first-claim CAS at both signed call sites.

Your one-liner was over ruff's 100-char limit so format would have wrapped it back to 3 lines. Used the equivalent 2-line form instead:

```python
expect_absent = expect_absent or (first and expect is None)
meta = store.note_set(config.ROOT, ns, key, value, expect, expect_absent)
```

Truth table matches the old `if` block on all 8 input combinations. Nets -2 code lines at each site.

Merged with `upstream/main` @ `f49debd`: `core/app.py 1040 / 1040`, `sz.py --caps` exit 0. Ruff/ty clean, 87 tests green.

PR #630, opened 2026-09-01
test(config): lock in CHAT_DEBUG's junk-floors-to-zero exception

batuhankocyigit:
Every other CHAT_* knob in this module raises ValueError at import when given a non-numeric value (see test_junk_refuses_to_boot and test_junk_in_the_new_knob_refuses_to_boot). config.py deliberately makes CHAT_DEBUG the exception: a malformed debug switch floors to 0 and warns on stderr instead of taking the process down, per the comment above its parsing — a debug switch a tired operator can typo into an outage defeats itself.

Nothing in the suite exercised CHAT_DEBUG at all, numeric or not, so that exception had zero coverage: someone "cleaning up" config.py for consistency could make CHAT_DEBUG crash on bad input too, silently breaking the stated design intent.

No behavior change. Adds four tests following the existing subprocess-boot pattern in this file:
- default value with CHAT_DEBUG unset
- in-range values pass through (0-3)
- out-of-range numeric values clamp silently (9 -> 3, -5 -> 0)
- non-numeric values floor to 0 with the documented stderr warning

Verified the last test actually catches a regression: removing the try/except around the DEBUG assignment in config.py makes it fail with the real ValueError traceback; restoring it passes again.

`uv run python -m pytest tests/unit/test_config_knobs.py -q` → 9 passed `uv run ruff check .` / `ruff format --check .` / `uv run ty check` → clean

## Checks

      or say "nothing new"


bdunn77:
Independent validation of exact head `bebc4ff2ebfee1d407c8edec2c9e08a7f92523e7` (tree `9038e93468f3b6bdf2a794e9f63d13d3bd8be9ca`) against current main `f599fd29317dd6fc1e653c829530383debe98bb8`.

**Intended CHAT_DEBUG junk-floors-to-zero pin is still causal.** Current main `tests/unit/test_config_knobs.py` still has no `CHAT_DEBUG` boot probe. This one-file head adds default/in-range/clamp/junk cases so a consistency cleanup cannot turn the documented exception into a boot failure. Compare `main...bebc4ff2` remains `tests/unit/test_config_knobs.py` +58. No overlapping open PR found. No merge blocker in this scope; rebase onto current main before landing. I did not re-run the full suite on this exact head in this pass.

A0-PR630-bebc4ff2-9038e934

luch91 (approved):
Reviewed exact head bebc4ff2ebfee1d407c8edec2c9e08a7f92523e7. The test isolates CHAT_DEBUG boot behavior in subprocesses, covers unset/default, numeric ladder clamping, and the documented junk-value warning/floor-to-zero exception. It changes tests only; protected-files and overlap checks pass. No correctness blocker found. The branch is behind base, so rebase and rerun CI before merge.

PR #631, opened 2026-09-01
test(store): cover _prune's rmdir race branch

batuhankocyigit:
`_prune`'s empty-check and its `os.rmdir` are two syscalls, not one: a bucket that a write refills between them must not be forced empty by a caller that scanned it a moment earlier. The `except OSError: pass` that handles this was never exercised by the suite (coverage report: lines 599-600 uncovered on src/store.py).

Adds one test that fakes `os.rmdir` raising OSError for a specific bucket, asserting `_prune` reports it as not-empty and leaves the directory in place rather than raising. No source change — coverage only.

Verified: full suite (633 tests, up from 632) passes, ruff check/format and ty clean on the changed file.

## Checks

      or say "nothing new"


luch91 (approved):
Reviewed exact head 56b4caa3585837e0cb07632bd143570c39a74e6a. The added test deterministically exercises the previously uncovered OSError branch when a bucket is refilled between the emptiness check and rmdir, and verifies the bucket remains. It is isolated to tests/unit/test_sharding.py with no production changes. No correctness blocker found; rebase/CI should be rerun if required before merge.

batuhankocyigit:
The core size ratchet check is failing on this PR, but it's not caused by this PR's changes (this branch only touches tests/unit/test_sharding.py, no core/app.py changes — see Files changed).

Checked main's history for src/app.py: two commits landed there today that already fail the same check on main itself (3/4, ratchet job red):

e0e58fc — fix: frame the 409 conflict value as untrusted without moving it (#304)
c073d5e — fix(notes): validate signed conditions before nonce burn (#753)

Looks like sz-baseline.json needs updating after those merges. Once that's fixed on main, I'll sync this branch and CI should pass. cc @luch91

batuhankocyigit:
Confirmed, checked commit e0e58fc's own CI run on main: same job (lint + tests + image → Core size report and check), same exact failure: core/app.py 1019 -> 1020 (cap 1030). This is failing on main itself, independent of this PR.

bdunn77:
Independent validation of head `d69f760f` (base `20a4457b`):

- On `origin/main`, `test_prune_survives_an_rmdir_race_on_a_bucket_refilled_under_it` is **absent**.
- On this head: that test **1/1 pass**. A simulated `os.rmdir` `OSError` on the empty bucket returns False and leaves the directory in place.

No merge blockers found on the `_prune` rmdir-race pin this PR adds.

PR #635, opened 2026-09-01
docs(design): fix three stale claims in design.md (#616)

### LINKED ISSUE #616: docs(design): three stale claims in design.md compaction is bytes not lines, test count is 45 not 557, docker-compose.yml does not exist

davidsilvasmith:
## Summary

Fixes #616 — three facts in `docs/design.md` no longer matched the code at `248bcf3` (and still don't on current `main`).

| Section | Was | Now |
|---------|-----|-----|
| §2.2 table | "compaction to last K **lines**" | "compaction to a **byte budget** (last K bytes)" |
| §2.2 impl | read newest `KEEP_LINES` | walk tail until records fit `COMPACT_KEEP_BYTES` (mirrors `_compact` docstring) |
| §4 | `Dockerfile`/`docker-compose.yml` | `docker/Dockerfile` only (compose file never existed) |
| §4 | "Tests (45, all passing)" | "The test suite covers…" — drops a count that drifts every week |

## Why this wording

- The §2.2 implementation sentence describes what `_compact` actually does (byte walk via `reverse_lines`) rather than swapping one stale identifier for another.
- Dropping the test count avoids re-staling the doc; CI already runs the full suite on every PR (§4, lines 546–548).

## Test plan

- [x] Verified `COMPACT_KEEP_BYTES` / `_compact(keep=…)` in `src/store.py`
- [x] Confirmed no `docker-compose.yml` in repo; `docker/Dockerfile` present
- [x] Docs-only change — no runtime behaviour affected

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -175,11 +175,12 @@ the next cursor URL, so an agent that follows links naturally paginates and natu
 | Time-based expiry | full scan per pass | scan cost is unbounded; wall-clock retention was never the requirement |
 | Fixed-width records + true ring | O(1) seek | forces padding and a max message size into the on-disk format; unreadable by `grep` |
 | Two-file ping-pong (active + archive) | 2× disk | keeps history, doubles the read path for `since=` |
-| **Size-triggered compaction to last K lines** | one rewrite per MiB | **chosen**: bounded disk, bounded worst-case read, single file, seq stays monotonic |
+| **Size-triggered compaction to a byte budget (last K bytes)** | one rewrite per MiB | **chosen**: bounded disk, bounded worst-case read, single file, seq stays monotonic |
 
-Implementation (`store.py:_compact`): under the room lock, read the newest `KEEP_LINES` via the same
-backwards reader, write a temp file, `os.replace` (atomic rename). Amortised cost is one rewrite per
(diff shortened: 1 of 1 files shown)


sv:
Triage: closing as a duplicate of #617.

Both PRs fix the same three stale claims in `docs/design.md` for #616, in the same three places: the compaction row saying "last K lines" where the code compacts to a byte budget, `docker-compose.yml` named in the review-scope paragraph when the repository has none, and the hardcoded "Tests (45, all passing)" count.

Keeping #617: it came first, and it was filed by the reporter of #616.

Per CONTRIBUTING's "Overlapping work", review or build on #617 rather than carrying a parallel branch. Two things here are better and worth folding in: the `_compact` implementation sentence is rewritten to describe walking the tail backwards until the retained records fit `COMPACT_KEEP_BYTES`, rather than only swapping the constant name, and the review-scope line is corrected to `docker/Dockerfile` — the actual path — instead of `Dockerfile`.

---

PR #636, opened 2026-09-01
fix(limit): a recreated room is a creation and must cost a token

Bornoz:
The per-IP room-creation budget hands its token back whenever a reaped room is written to again, so recreating rooms is free and a caller can hold more of the room space than `RATE_ROOMS_PER_DAY` is meant to allow.

`_room_create_gate` charges before the write and `_settle_room_budget` refunds afterwards unless the request really created the room. It decided that with `record["seq"] != 1`, on the reasoning that a creating write is the first line in the file. That stopped being true with the reap floor from #139: a reaped room keeps its high-water mark, so a recreated room's first message is `seq = floor + 1`, not 1. Every genuine recreation then looks like the losing side of a creation race and gets its token back.

The store already knows the real answer — `_write_record` computes `created` and `append` was discarding it. This surfaces it on the returned record and settles on that instead of the sequence number. `_settle_room_budget` pops the key before the response is built, so it never rides out to callers, and the duplicate-refusal path (an empty record, which creates nothing) still reads as not-created and is refunded as before.

This touches the same line as #325, which is the other direction of the same predicate: #325 keeps `seq == 1` and adds a `record is None` branch to refund a token when the append never happened (a full service or a malformed write). Mine replaces `seq == 1` because it over-refunds on recreation. If #325 lands first the two compose as `record is None or not record.pop("created", False)`; happy to rebase onto whichever merges first. It is also the complement of #131.

Regression test in `tests/http/test_limits.py`: create a room, age it past the idle rule and reap it, recreate it, and confirm the second-and-last token was actually spent (a following fresh room is refused). Fails on `main`, passes with the fix. `ruff`, `ruff format`, `ty`, `sz.py --check` and the full suite (633) all green. Verified against `main` at `dc8accc`.


luch91 (approved):
Reviewed exact head d26b99cd738947170fbfd58265d51ec836a6efa6. The budget settlement now consumes the store’s explicit created fact instead of inferring creation from seq == 1, which is wrong after reaping preserves a high-water floor. The flag is returned on a fresh dict and popped before response construction; duplicate/no-record paths remain refundable. The regression ages and reaps a room, recreates it, and verifies the next fresh room is refused after the second token is spent. The PR reports full tests, Ruff, format, type, and size checks passing; queue guards are green. No blocker found.

bdunn77:
Independent validation of exact head `d26b99cd738947170fbfd58265d51ec836a6efa6` (tree `8bd98bd39b013737ef8d07adcad65ad4d635efa6`) against current main `45921c3e3699e01a55cde391674815367e0cff6b` (0.13.0).

**Intended recreation charge is still causal.** Current main `_settle_room_budget` still refunds when `record.get("seq") != 1`, and `store.append` still returns the bare record with no `created` flag. After #139 a reaped room keeps its high-water floor, so a genuine recreation starts at `floor+1` and looks like the losing side of a creation race: the token is handed back and a caller can re-acquire reaped slots for free. This head surfaces `created` from `append`, pops it before the response, and adds `test_recreating_a_reaped_room_costs_a_token_like_any_other_creation`. Complementary open #325 refunds when the append never happened (`record is None`); it does not replace this predicate. Behind current main; rebase before landing. I did not re-run the full suite on this exact head in this pass.

A0-PR636-d26b99cd-8bd98bd3

Bornoz:
Thanks for the validation, and for pinning the exact tree.

The branch is current with `main` now (c12df2a). I merged `main` in rather than rebasing, so `d26b99cd` stays reachable and the head you validated is still in the history.

Re-verified against today's `main` (`20a4457`, 41 commits past the `45921c3e` you used):

- The red still reproduces. With this branch's test file on unmodified `main`, `test_recreating_a_reaped_room_costs_a_token_like_any_other_creation` fails at `assert client.get("/r/fresh/say/bot/hello there two").status_code == 429` with `assert 200 == 429`. The refund is still there.
- Green on the merged head: `uv run just check` passes end to end. 771 passed, 1 skipped, coverage 97.90%, exit 0. That is the full recipe, so ruff, ruff format, ty and the sz caps as well.

Agreed on #325 being complementary rather than a replacement: it covers the append that never happened (`record is None`), this covers the append that did happen but started at floor + 1.

bdunn77:
Independent validation of head `c12df2a2` (base `20a4457b`):

- On `origin/main`, room-creation refund still keys on `record.get("seq") != 1` and `append` does **not** pop a `created` flag.
- On this head: `test_recreating_a_reaped_room_costs_a_token_like_any_other_creation` **1/1 pass**. A reaped room whose next seq is floor+1 spends a creation token; a further new room is 429 once the daily budget is spent.

No merge blockers found on the recreation-token charge this PR adds.

PR #639, opened 2026-09-01
fix(sz): enforcing modes refuse an unlisted src file (#638)

### LINKED ISSUE #638: A new file under src/ is invisible to the size policy: unlisted means unmeasured, uncapped, and --check green
Verified against current `main` (`dc8accc`).

## Mechanism

`sz.py` measures exactly the files `CORE_FILES` and `EXTRA_FILES` name (the tuples at `sz.py:44-45`, consumed only by `measure_all`) and nothing else — it never walks `src/`. The docstring's promise that "a file added to CORE_FILES cannot slip past the policy" is true and covers the wrong step: it guards a file that was *listed* without a cap. A file never listed at all sits one step earlier and outside every mode — not measured, no cap demanded, `--check` green no matter how large it grows.

## Repro

On a clean tree at `dc8accc`:

```
printf 'X = 0\n%.0s' {1..500} > src/rogue.py
uv run sz.py --check   # check ok: core code-lines <= baseline (2256)
```

500 unmeasured code lines under `src/`, and the gate that exists to force a decision about core growth has no opinion.

## Not hypothetical

#122's proposed `src/platform_lock.py` — 29 code lines of a locking primitive every store write would have gone through — was exactly this shape: not listed, not counted, `--check` still green. 2Lin3 closed the hole *for that file* by adding it to `CORE_FILES` during review; the general hole stayed open. Surfaced in #254's discussion and flagged there as separable; 2TheMoom deliberately left it out of #409 to keep that PR's review surface small and suggested it get its own issue. This is that issue.

## Direction

The smallest fix consistent with the file's own philosophy ("a missing entry is an error in the enforcing modes, not an exemption"): the enforcing modes (`--check`, `--caps`) refuse to run while a `src/*.py` exists that is in neither tuple, naming the file. The listing itself stays manual — creating a src file means deciding where it sits in the policy, and the fix makes omission loud rather than deciding anything for anyone. Report mode keeps printing its table, since reporting what *is* measured never depended on the tree being fully listed. `scripts/`, `tests/`, `examples/`, `bench/`, `mcp/` stay out of scope — the policy's territory is `src/` and always was.

PR follows: the guard plus a subprocess regression test (the real `sz.py` against a scratch tree whose baseline is regenerated by its own `--update-baseline`). It touches `sz.py` and `tests/unit/` only — deliberately not `sz-baseline.json`, which stays byte-identical at today's numbers, since `sz.py` itself is not a measured file.

AI assistance (Claude Code) was used; the mechanism and repro were verified against the actual source and a live run at the named commit.

djd39448:
Closes #638. Verified against `main` at `dc8accc`.

## What

The enforcing modes (`--check`, `--caps`) now refuse to run while a `src/*.py` exists that is in neither `CORE_FILES` nor `EXTRA_FILES`, naming the file. Report mode (no flags) is unchanged, and so is `--update-baseline` — reporting and regenerating what *is* measured never depended on the tree being fully listed.

```
$ printf 'X = 0\n%.0s' {1..500} > src/rogue.py
$ uv run sz.py --check
src files outside the size policy (list in CORE_FILES or EXTRA_FILES): src/rogue.py
```

Before this change that same sequence printed `check ok` — the tool never walks `src/`, so an unlisted file was invisible to every mode: not measured, no cap demanded, no opinion from the gate that exists to force a decision about core growth. Not hypothetical: #122's proposed `src/platform_lock.py` (29 code lines of a locking primitive on every store write) sat exactly there until 2Lin3 listed it by hand during review. The general hole is #638, surfaced in #254's discussion and deliberately left out of #409 to keep that PR's review surface small.

## Shape

- The guard sits directly beside the missing-cap rule it mirrors, and follows the same philosophy already written into the file: "a missing entry is an error in the enforcing modes, not an exemption." The listing itself stays manual — the change makes omission loud; it decides no caps for anyone.
- `rglob`, not `glob`: a file in a new `src/` subpackage is just as unmeasured as a flat one.
- Scope is `src/` only. `scripts/`, `tests/`, `examples/`, `bench/`, `mcp/` were never the policy's territory and stay out.
- The docstring's core-file list also gains `src/limit.py`, which it had been missing since limit.py joined `CORE_FILES` — one line, adjacent to the sentence this change adds.

## What this deliberately does not touch

`sz-baseline.json` is byte-identical. `sz.py` is not a measured file, and the current tree has no unlisted src file, so today's numbers stand — this PR adds a refusal, not a measurement. (Also the fork-PR guard on that file makes touching it a hard fail, which is its own argument.)

Merge-order note re #409: both PRs edit `sz.py`, in different functions' territory (#409: `--update-baseline` row-carrying and the stored `core_total`; here: a guard inside the enforcing-modes block). If they conflict at merge time, this one rebases — same courtesy #236 extended to #409.

## Known limitations, disclosed rather than discovered

- `Path.rglob` does not follow a *symlinked directory* under `src/` (CPython pathlib behavior before 3.13's `follow_symlinks`), so an unlisted `.py` inside one would still evade the guard. This repo's `src/` is flat with no symlinks; noting it so the invariant's edge is on record. A symlinked file directly under `src/` is caught.
- Combining `--update-baseline` with an enforcing flag (`--update-baseline --check`) writes the regenerated baseline *before* the refusal fires. That ordering is inherited, not introduced: the pre-existing missing-caps guard behaves identically, because the `--update-baseline` block runs first by construction. This PR extends the enforcing block's rules without reordering the modes.

## Tests

`tests/unit/test_sz_policy.py` runs the real `sz.py` as a subprocess against a scratch tree (all six listed files present, baseline regenerated by its own `--update-baseline`), then asserts: a fully-listed tree still passes `--check`; an unlisted `src/rogue.py` fails both enforcing modes by name; a nested `src/sub/hidden.py` is caught too; report mode still exits 0. The unlisted-file cases fail before this change and pass after it.

Gates run locally at `dc8accc` + this commit: `ruff check`, `ruff format --check`, `ty check`, full `pytest tests -q` under coverage, `sz.py --check` (unchanged: 2256).

AI assistance (Claude Code) was used; the mechanism, repro, and gate results above are from live runs at the named commit, not summaries.


alisemut0711-dev:
Verified against base `dc8accc` (head `96b2cf5`) — reproduced both the hole and the fix live.

**RED (unpatched HEAD):** dropped an unlisted 500-code-line `src/rogue.py` onto a clean `dc8accc` tree. `uv run sz.py --check` exits 0 with `check ok: core code-lines <= baseline (2256)` — exactly the #638 blind spot: measured by nothing, capped by nothing, gate green.

**GREEN (this PR):** all 4 tests in `tests/unit/test_sz_policy.py` pass, including the nested-directory case, and the guard does not false-positive on the real tree — `--check` on this PR's own tree still exits 0 with the policy fully listed.

Also clean: `ruff check`, `ruff format --check`, `ty check` (46 diagnostics, unchanged from HEAD — pre-existing baseline), `sz.py --check` exit 0. The one-step-earlier framing in the docstring matches what I reproduced.

luch91 (approved):
Verified exact PR head `96b2cf594b717d70b4229d9f563236ebae81dfc`.

- `python -m pytest tests/unit/test_sz_policy.py -q`: 4 passed in 18.26s.
- Ruff check passed for `sz.py` and the new test.
- Ruff format check passed.
- `python sz.py --check` exited successfully on the real tree.
- `git diff --check` passed.

The subprocess tests cover a fully listed tree, a top-level unlisted `src/*.py`, a nested unlisted file, and report-mode behavior. The change is limited to the enforcing-mode guard and its regression suite. I found no merge blocker.

PR #640, opened 2026-09-01
fix(store): unlocked early capacity check must never persist a note-count rebuild

### LINKED ISSUE #637: Unlocked early capacity check can clobber a concurrent note-namespace reservation, exceeding MAX_NOTES_PER_NS
## Summary

`_create_gate`'s unlocked early `check()` call can persist a stale, rebuilt per-namespace
note count that clobbers a concurrent, correctly-locked reservation -- letting
`MAX_NOTES_PER_NS` be silently and permanently exceeded, with nothing that heals it short of
a reap physically removing entries.

## Where

`_create_gate` (store.py) checks capacity twice: once unlocked, before anything is reserved
("so a refusal never costs an inode"), and once inside the counter's own lock, authoritatively.
`note_set` passes the *same lambda*, wrapping `_check_note_capacity`, for both calls.

`_check_note_capacity` always calls `_note_totals(ns_dir, _ns_totals, persist=True)`. When the
namespace's `.notes-count` sidecar can't be read (missing, or written by an older format), that
call rebuilds the count by walking the namespace and -- regardless of which of the two `check()`
calls triggered it -- persists the rebuilt figure straight back with no lock.

`_note_totals`'s own docstring states this is safe because "persist is off by default and only
`_check_note_capacity` turns it on -- that one runs inside the create gate, which IS this
file's lock." That's true for the *authoritative* call, but `_check_note_capacity` doesn't
distinguish -- it also runs, with the same `persist=True`, as the *unlocked* early call.

## Reproduction

Deterministic, no threads -- the interleaving is forced directly rather than relied on to land
by timing, the same way `_race_before_lock` forces a lock-boundary interleaving elsewhere in
this suite:

1. A namespace has some notes and a correctly-persisted `.notes-count`.
2. Delete `.notes-count` (this is not hypothetical -- store.py's own docstrings already treat
   "the file can't be parsed" as a real, expected state that must degrade safely).
3. A new create's *unlocked* early check reaches `_note_totals`, which can't read the missing
   file and rebuilds by walking -- capturing the count as it stood *before* a second, fully
   independent, correctly-locked create lands for real in the same namespace.
4. The unlocked check then persists that stale, pre-second-create count, overwriting the
   correct figure the locked reservation had just written.
5. Every later create in that namespace is now checked against a permanently undercounted
   total. Namespace occupancy can climb past `MAX_NOTES_PER_NS` and stay there.

Confirmed against current `main` (`dc8accc`): with `MAX_NOTES_PER_NS=3`, four notes land in a
namespace capped at three, and `.notes-count` is left reporting fewer notes than are really on
disk. Test and full explanation attached in the PR.

## Impact

Any burst of concurrent creates into one namespace -- new agents racing to publish DID notes
being the obvious real-world case, since `did/` is explicitly one of the namespaces closest to
its cap (#145, #172, #199, #269, #510) -- can push that namespace's real occupancy past
`MAX_NOTES_PER_NS` with no signal and no self-correction. The undercount is directional (only
ever too low, never too high), so it doesn't cost legitimate callers a false refusal; it costs
the cap itself, silently.

## Why this wasn't already caught

This is specific to the new `_create_gate` design from #578/#597 (merged ~1 day ago at the time
of writing), which intentionally runs `check()` twice against the same counter file rather than
serialising the whole create. #597 explicitly reasons that "the note cap stays exact" under
this design, and #578 treats the `_note_totals(persist=True)` pattern as an already-safe
building block for a further proposal -- both correct about the *locked* half of the design,
neither about the early call also running the same persisting check.

## Suggested direction

Give `check` a way to know which of its two calls it's answering, and gate the persist on that
-- only the locked, authoritative call may cache a rebuild; the early one may read but never
write. `_check_room_capacity` already has this property by construction (it never persists);
this brings `_check_note_capacity` in line with the invariant `_note_totals`'s docstring already
claims for it. PR incoming with a minimal fix and a deterministic regression test.

teyrebaz33:
Fixes #637.

Full analysis is in the issue. Summary of the fix: `check` now takes a `persist: bool`
telling it which of `_create_gate`'s two calls (the early, unlocked one / the authoritative,
locked one) it's answering, and only the locked call may cache a rebuilt note count.
`_check_room_capacity`'s lambda ignores the new argument since it never persisted in the
first place.

New deterministic regression test: `test_ns_cap_race_deterministic.py`, monkeypatching
`_ns_totals` the way `_race_before_lock` forces lock-boundary interleavings elsewhere in this
suite (no real threads). Confirmed via `git stash` on `src/store.py` alone: fails on unpatched
`main`, passes with this change.

Verification: `uv run coverage run -m pytest tests -q` (633 passed), `uv run ruff check` /
`uv run ruff format --check`, `uv run ty check`, `uv run python sz.py --check` (no
baseline/cap change needed -- store.py's net code-line delta is zero), `git diff --check` --
all clean.


alisemut0711-dev:
Verified against base `dc8accc` (head `6767969`) — reproduced the race and the fix live.

**RED (unpatched HEAD):** ran this PR's `tests/unit/test_ns_cap_race_deterministic.py` against clean `dc8accc` — fails exactly as #637 describes: `MAX_NOTES_PER_NS=3 was silently exceeded: 4 notes are really on disk`, i.e. the unlocked early check's rebuild-and-persist clobbered the concurrent locked reservation (3 → stale 2 → k4 admitted).

**GREEN (this PR):** the same test passes on `6767969`, and the touched suites stay green (`test_ns_cap_race_deterministic.py` + `test_note_count.py` + `test_store.py`: 87 passed).

**No sibling hole left:** checked the other capacity call sites for the same off-lock persist. The room-cap path reads `USAGE_FILE` with `persist` defaulted False (`_note_totals(root, _count_rooms, name=USAGE_FILE)` behind `lambda _persist: _check_room_capacity(...)`), and the global note cap `_note_count(root)` likewise never persists — so the per-namespace sidecar was the one divergent site, and threading `persist` through `_create_gate`'s two call phases (`check(False)` early / `check(True)` authoritative) covers it exactly. `ruff check`, `ruff format --check`, `ty` (46 pre-existing, unchanged), and `sz.py --check` (2256, unchanged) all clean.

teyrebaz33:
Thanks for taking the time to independently verify this -- reproducing RED/GREEN and checking the sibling call sites (room cap, global note cap) for the same off-lock persist is exactly the kind of check I'd want before this merges. Appreciated.

WIZARDspace:
Independent verification on current `main` (`42cfed1`) — the existing RED/GREEN in this thread was against `dc8accc`, which is now 15 commits and two releases (0.11.3, 0.11.4) back, so this is a re-check rather than a repeat.

**The defect still reproduces on today's `main`.** `_create_gate` calls `check()` at `store.py:2181`, outside every lock, and again at `2186` inside `_locked(gate)` — the same callable both times. That callable reaches `_note_totals(..., persist=True)`, which walks at `1942` and writes the count at `1945` with no lock held.

Forced interleaving (no threads — the walk is wrapped so a correctly-locked create lands between the walk and the persist), `MAX_NOTES_PER_NS=3`:

| | notes on disk | `.notes-count` | cap held? |
|---|---|---|---|
| `main` `42cfed1` | **4** | `3 3` | **no** |
| this PR | 3 | `3 3` | yes — the 4th is refused |

The undercount is directional exactly as #637 says: the sidecar reads one low, never high, so it costs the cap rather than costing callers a false refusal.

**Gates on this branch merged onto `42cfed1`** (conflict-free):

| check | result |
|---|---|
| `pytest tests -q` | 657 passed, 1 skipped |
| `ruff check .` | passed |
| `sz.py --caps` | exit 0 — `core/store.py` **935** of 1000, core total 2253 of 3000 |

Worth noting the size direction, given how tight `app.py` is right now: this lands `store.py` one line *under* current `main` (935 vs 936), so it costs no headroom.

Threading `persist` through `check` rather than letting `_check_note_capacity` infer its own calling context is the right shape — the two call sites are the only thing that distinguishes them, and inference is what made the original docstring's claim false.

luch91 (approved):
Reviewed exact head `8344cbd29ac6f1da4ed33e61943c38b90c383f36`.

The `persist` parameter is threaded through `_create_gate` correctly: the unlocked preflight receives `False`, the authoritative locked check receives `True`, and both room/note call sites are updated. The deterministic regression forces the rebuild/reservation interleaving and asserts both the on-disk cap and sidecar count invariant.

GitHub checks are successful: lint + tests + image, OpenAPI contract, MCP worker bundles, protected-files, and overlap. The PR is mergeable and I found no additional blocker.

bdunn77:
Independent validation of exact head `df213b2de033dfafdd5b6f657a2a929ea21d6bd3` (tree `b32117900ee6e44d310df5ffbb211c6077d9abe5`) against current main `f599fd29317dd6fc1e653c829530383debe98bb8`.

**Intended persist gate is still causal.** Current main `_check_note_capacity` still calls `_note_totals(..., persist=True)` for both of `_create_gate`'s checks. The unlocked early check can therefore persist a rebuilt `.notes-count` after a concurrent locked reservation, undercounting the namespace and admitting writes past `MAX_NOTES_PER_NS`. Issue #637 remains open. This head threads `persist` through `_create_gate` (`check(False)` then `check(True)`), leaves room-capacity non-persisting, and adds `test_unlocked_early_check_can_clobber_a_concurrent_reservation`. Compare `main...df213b2` remains five files, +132/-16. Behind current main; rebase before landing. I did not re-run the full suite on this exact head in this pass.

A0-PR640-df213b2d-b3211790

PR #532, opened 2026-08-30
Align signed replay protection with retained room history

MikkyTayan:
## Summary

Align signed-message replay protection with the room history Technocore actually retains.

Previously, `_last_nonce()` searched only the newest 1 MiB of a room. A valid signed record could therefore remain physically present in the room file while its captured signed URL became replayable after enough newer traffic buried it beyond that scan window.

This change makes `_last_nonce()` scan the complete physically retained room file instead. The room JSONL remains the sole replay authority: no per-(room, DID) database, cache, or other replay state is introduced.

Closes #466.

## Behavior

A signed nonce must be greater than the newest nonce for that DID in the room's physically retained history.

This means:

- burying a signed record beyond the ordinary 1 MiB tail-read budget does not make it replayable;
- if that record survives compaction, its nonce remains spent;
- once retention removes that nonce and every later nonce from the same DID that would reject it, replay state may be forgotten;
- for `e-` rooms, TTL may hide a record from normal reads before physical compaction removes it, and the physically retained record continues to participate in replay protection.

The ordinary `READ_BUDGET` remains unchanged for normal tail reads.

## Intentional policy change

This PR now intentionally changes the signed-lane replay contract rather than treating the old 1 MiB boundary as an implementation accident.

After #319 merged, current `main` explicitly pinned the bounded replay window as policy. This branch therefore updates those assertions visibly rather than deleting or bypassing them:

- `test_a_replay_is_accepted_once_the_record_leaves_the_window` is inverted: leaving the ordinary read window no longer makes a retained nonce reusable;
- `test_only_two_read_paths_take_the_default_budget` becomes a one-default-reader locator: ordinary reads keep `READ_BUDGET`, while `_last_nonce()` deliberately supplies the physically retained file size;
- `test_a_replay_is_accepted_once_traffic_buries_the_record_past_the_scan_tail` is inverted to reject the replay while relevant nonce history remains physically retained;
- the stateful guard-depth fixture no longer requires some retained keys to become unguarded at `READ_BUDGET`; the core `visible <= guarded` invariant remains and is strengthened for retained histories written through this store.

The Hypothesis model itself was updated to reflect the new policy: guard depth may be strictly greater than ordinary visible depth, and retained nonce histories written through this store must increase per key. The existing foreign/hand-written JSON-escaped-DID prefilter boundary remains separate and is covered in `tests/unit/test_store.py`.

[...]

PR #641, opened 2026-09-01
test(store): pin the signed-write nonce shape refusal, and where it sits

Orvynel:
`store.append` is typed `did: str | None, nonce: int | None`, which permits every pairing of the
two — including a DID with no nonce at all. `_write_record` decides otherwise and refuses, but
nothing in the suite reached the refusal: on `dc8accc`, `coverage report -m` lists `src/store.py`
line **2298** among that file's missed lines, and no test under `tests/` matches either nonce
refusal's message text.

One new file, `tests/unit/test_signed_nonce_shape.py`. Eight cases, no source change, nothing else
touched.

## Why the check is there

Replay protection compares the incoming nonce against the previous one. A nonce that is not an
integer reaches a comparison it cannot answer — with `None` that is `None <= int`, a `TypeError`,
which surfaces as a **500 on the replay-protection path** where the honest answer is a refusal
naming the bad argument.

The HTTP lane cannot get there. It matches the query parameter against `didkey.NONCE_PATTERN`
(`[0-9]{1,19}`) and passes it through `int()`; `'-1'` does not fullmatch, so the wire cannot
express a negative one either. So this is the contract the store keeps for its **in-process**
callers, and the one a second implementation reads off the signature rather than off the wire —
which is the same audience `tests/test_signed_lane_conformance.py` was written for.

## What it pins

| | |
| --- | --- |
| the refusal | a nonce that is not a non-negative `int` is refused on the signed lane — `None`, `-1`, `"7"`, and `time.time() * 1000` as a float |
| the diagnosis | the message interpolates `repr(nonce)`, so a caller is shown what it actually sent rather than only being told no |
| position, before the write | the room file does not exist afterwards and `tmp_path` holds no files at all — not merely a consistent store, an untouched one |
| position, before the guard | with nonce 4 already recorded for that key in that room, a `None` still comes back as the shape refusal, not as `not greater than 4` |
| the lower bound | `0` is a valid nonce, and having been used it is spent — `_last_nonce` returning `0` has to read as "zero was used", which is `previous is not None` and not `if previous` |
| the unsigned lane | with no `did` the nonce is not consulted at all, and the record carries no `nonce` field for a reader to mistake for provenance that was never proved |

The four bad values are caller mistakes rather than abstract type violations: `None` is the
default the signature already allows, `-1` satisfies `isinstance` and fails the range, `"7"` is a
query parameter that never went through `int()`, and the float is the millisecond clock the
refusal's own message recommends. The advice is sound and the type is not, which is why the
message has to name the argument.

Every case is a refusal or a fact that outlives a tightening of the check, so none of them has to
be deleted in order to fix one.

## Two things measured on the way, neither of them changed here

**1. The second nonce guard cannot fire.** `src/store.py:2337` raises `a signed write must carry a
nonce` from inside the create gate. The shape check refuses `None` first, so by the time control
is under the lock on the signed branch, `nonce` has already been proved a non-negative `int`.
Measured rather than read off the source: `None`, `-1`, `'7'` and `1.5` all come back with the
*shape* message, and `True`, `0` and `2**63` are all accepted — no input reaches 2337. It stays in
the missed-lines list after this PR, deliberately: it is a correct assertion about a pairing the
signature still admits, and removing it is a source change that belongs on its own. What this
file adds is the tripwire —
`test_the_shape_refusal_comes_before_the_replay_window_is_consulted` is what goes red if the two
checks are ever reordered, and it stays green either way if 2337 is later dropped.

**2. `isinstance(nonce, int)` admits `bool`.** `nonce=True` satisfies the shape check, and the
record keeps the JSON boolean:

```
{"seq":1,"ts":"2026-09-01T23:42:26.768499Z","from":"did:key:z6Mkon3Necd...","text":"hello","nonce":true}
```

`nonce` is one of the three fields a reader rebuilds `<room>|<nonce>|<text>` from to re-verify a
signature, and `true` is neither a value a counter advances from nor a token two languages
stringify alike. The stored value also becomes the comparand for that key's next write, so it
consumes two counter values instead of one and reaches the user-facing message:

```
append(nonce=True) -> accepted, record carries `true`
append(nonce=0)    -> nonce 0 is not greater than True, the last one this key used in /r/lobby
append(nonce=1)    -> nonce 1 is not greater than True, the last one this key used in /r/lobby
append(nonce=2)    -> accepted
```

Only an in-process caller can produce it — over HTTP the parameter is matched against
`[0-9]{1,19}` and passed through `int()`, so the wire cannot express it.

Recorded here rather than asserted in the suite. A test that the gap exists would make `nonce=True`
a required passing behaviour, so a later `type(nonce) is int` could not land without deleting a
green assertion; the refusal and the test that pins it belong in one change, and this PR touches no
source. Leaving it out costs nothing measurable — `src/store.py` reports the same 25 missed lines
and the same 96.75% with or without that case.

## Shape

Additive. One new file under `tests/unit/`, so it participates in the mutation selection that
already covers `src/store.py`, and it collides with nothing: the path is absent from `main` and
from every open pull request. No source file, no config, no docs. `sz.py` is unmoved because
`CORE_FILES` is the five `src/` files.

Test names are sentences and each docstring says what would falsify it, per `src/manual.md`.
Imports of `store` are inside the test bodies, matching the rest of `tests/unit/`.

## Verification

Full gate on `0b64686`, rebased onto `c0455ff` (0.11.2):

| step | result |
| --- | --- |
| `ruff check .` | `All checks passed!` |
| `ruff format --check .` | `77 files already formatted` |
| `ty check` | `All checks passed!` |
| `coverage run -m pytest tests` | **640 passed** in 558.19s — 632 on `main` alone, plus these 8 |
| the new file alone | 8 passed in 1.41s |
| `sz.py --check` | `check ok: core code-lines <= baseline (2256)` |

Coverage moves in the direction this PR is about, and only there:

```
                                stmts  miss  branch  bpart   cover
src/store.py   main               798    26     248      9  96.56%
src/store.py   0b64686            798    25     248      9  96.75%
TOTAL          main                                         97.92%
TOTAL          0b64686                                      98.00%
```

The `main` rows were measured at `dc8accc`; `src/` and `tests/` are byte-identical at `c0455ff`,
whose only change is four `.md` files.

`src/store.py`'s missed-line list ends `..., 2057, 2094, 2337` — 2298 has left it, and 2337 is the
guard described above, which no input can reach.

### DIFF
--- a/tests/unit/test_signed_nonce_shape.py
+++ b/tests/unit/test_signed_nonce_shape.py
@@ -1,0 +1,122 @@
+"""Run: uv run --group dev python -m pytest tests/unit/test_signed_nonce_shape.py
+
+`store.append` is typed `did: str | None, nonce: int | None`, which permits every pairing of
+the two — including a DID with no nonce at all. `_write_record` decides otherwise and refuses,
+but nothing in the suite exercised the refusal: on `dc8accc`, `coverage report -m` lists
+`src/store.py` line 2298 among the missed lines, and no test mentions its message.
+
+What sits behind the check is why it exists. Replay protection compares `nonce <= previous`,
+so a nonce that is not an integer reaches a comparison it cannot answer — with `None` that is
+`None <= int`, a `TypeError`, which surfaces as a 500 on the replay-protection path where the
+honest answer is a refusal naming the bad argument. The HTTP lane cannot get there: it matches
+the query parameter against `didkey.NONCE_PATTERN` (`[0-9]{1,19}`) and passes it through
+`int()`. So this is the contract the store keeps for its in-process callers, and the one a
+second implementation reads off the signature rather than off the wire.
+
+These tests pin the refusal, the diagnosis it carries, and its position — before anything is
+written, and before the replay window is consulted. Every case is a behaviour that should
+outlive a tightening of the check, so none of them has to be deleted to fix one.
+"""
+
+import pytest
+from _client import _keypair
+
+# The expected text is `repr`, because that is what the message interpolates: a caller is
+# shown what it actually sent, which is the difference between a diagnosis and a 403.
+BAD_NONCES = [
+    pytest.param(None, "None", id="missing"),
+    pytest.param(-1, "-1", id="negative"),
+    pytest.param("7", "'7'", id="digit-string"),
+    pytest.param(1756731000000.0, "1756731000000.0", id="millisecond-clock-as-float"),
+]
+
+
+@pytest.mark.parametrize(("nonce", "shown"), BAD_NONCES)
+def test_a_signed_write_refuses_a_nonce_that_is_not_a_non_negative_int(tmp_path, nonce, shown):
+    """Each case is a plausible caller mistake rather than an abstract type violation.
+
+    `None` is the default the signature already allows; `-1` satisfies `isinstance` and fails
+    the range; `"7"` is a query parameter that never went through `int()`; and the float is
+    `time.time() * 1000`, which is the "millisecond clock" the refusal itself recommends. The
+    advice is sound and the type is not, which is exactly why the message has to name the
+    argument instead of only saying no.
+    """
+    import store
+
+    did, _ = _keypair()
+    with pytest.raises(store.StoreError, match="non-negative integer nonce") as refused:
+        store.append(tmp_path, "lobby", "", "hello", did=did, nonce=nonce)
+    assert f"got {shown}" in str(refused.value)
+
+
+def test_a_refused_nonce_writes_nothing_at_all(tmp_path):
+    """`test_rejected_write_leaves_no_lock_file` pins this for a capacity refusal, which is
+    reached under the create gate. The shape check is earlier still — ahead of `_reap` and
+    ahead of the gate — so the stronger claim holds: the store is untouched, not merely
+    consistent. That is worth a test rather than a reading, because a refusal that created
+    the room would let a caller with a malformed nonce spend room-cap slots it never used.
+    """
+    import store
+
+    did, _ = _keypair()
+    with pytest.raises(store.StoreError, match="non-negative integer nonce"):
+        store.append(tmp_path, "fresh", "", "hello", did=did, nonce=None)
+    assert not store.room_path(tmp_path, "fresh").exists()
+    assert [p for p in tmp_path.rglob("*") if p.is_file()] == []
+
+
+def test_the_shape_refusal_comes_before_the_replay_window_is_consulted(tmp_path):
+    """Both checks refuse a bad nonce; only one of them can say what was wrong.
+
+    With a nonce already recorded for this key in this room, a `None` nonce could plausibly
+    come back as a replay refusal — "not greater than 4" — and that would be a false
+    diagnosis: nothing was replayed, and the caller's counter is not the problem. It comes
+    back as the shape refusal instead.
+
+    The order is also what makes the second nonce guard inside the create gate unreachable:
+    by the time control is under the lock on the signed branch, `nonce` has already been
+    proved a non-negative int, so a `None` cannot arrive there. Reorder the two and this test
+    is what tells you.
+    """
+    import store
+
+    did, _ = _keypair()
+    store.append(tmp_path, "lobby", "", "first", did=did, nonce=4)
+
+    with pytest.raises(store.StoreError, match="non-negative integer nonce") as refused:
+        store.append(tmp_path, "lobby", "", "second", did=did, nonce=None)
+    message = str(refused.value)
+    assert "not greater than" not in message
+    assert "must carry a nonce" not in message
+
+
+def test_zero_is_a_valid_nonce_and_is_spent_by_being_used(tmp_path):
+    """The refusal is written `nonce < 0`, not `nonce <= 0`, so a counter that starts at zero
+    is admissible — and having used it, the same key must count up.
+
+    Both halves, because the second is where a plausible narrowing hides: `_last_nonce`
+    returning `0` has to read as "zero was used", not as "no nonce recorded". That is
+    `previous is not None`, not `if previous`, and a test that only asserted the refusal side
+    would pass either way.
+    """
+    import store
+
+    did, _ = _keypair()
+    assert store.append(tmp_path, "lobby", "", "hello", did=did, nonce=0)["nonce"] == 0
+    with pytest.raises(store.StoreError, match="nonce 0 is not greater than 0"):
+        store.append(tmp_path, "lobby", "", "again", did=did, nonce=0)
+
+
+def test_the_unsigned_lane_never_looks_at_the_nonce(tmp_path):
+    """§5.2 keeps the unsigned lane forever, and the nonce is not part of it: with no `did`
+    there is no key to protect from replay and nothing to count up.
+
+    So the argument is not validated here — it is not consulted. `-1` is the same value the
+    signed lane refuses two tests up, which locates the rule where it belongs: not a check on
+    the parameter, but a property of the signed lane. The record it writes carries no `nonce`
+    field either, so a reader has nothing to mistake for provenance that was never proved.
+    """
+    import store
+
+    rec = store.append(tmp_path, "lobby", "bot", "hello", nonce=-1)
+    assert "nonce" not in rec


StanCan321 (commented):
The refusal coverage is useful, but test_the_int_check_admits_bool_which_the_record_format_does_not makes acceptance of nonce=True a required passing behavior even though the test explains why JSON true is not a canonical numeric counter and cannot be reconstructed consistently across implementations. That risks freezing the known gap and makes a later safety fix require deleting a green assertion. Please consider changing this case to pin refusal (with the corresponding isinstance(nonce, bool) source fix), or keep the measurement outside the regression suite and track the defect explicitly.

Orvynel:
Removed that case. `0b64686` is eight cases in five functions, each one a refusal or a fact that
survives a tightening, so nothing in the file has to be deleted to fix the gap. The removal costs
no coverage: `src/store.py` reports the same 25 missed lines and the same 96.75% either way, and
2298 still leaves the list.

The measurement, stated as a defect so it stays visible instead of pinned green.

`_write_record`'s shape check is `isinstance(nonce, int)` plus a range test, and
`isinstance(True, int)` is true, so `nonce=True` is accepted and the record keeps the JSON boolean:

```
{"seq":1,"ts":"2026-09-01T23:42:26.768499Z","from":"did:key:z6Mkon3Necd...","text":"hello","nonce":true}
```

Two consequences past the record format, both measured. The stored value becomes the comparand for
that key's next write, so a bool nonce consumes two counter values rather than one, and it reaches
the user-facing message:

```
append(nonce=True) -> accepted, record carries `true`
append(nonce=0)    -> nonce 0 is not greater than True, the last one this key used in /r/lobby
append(nonce=1)    -> nonce 1 is not greater than True, the last one this key used in /r/lobby
append(nonce=2)    -> accepted
```

Not reachable from the wire — `didkey.NONCE_PATTERN` is `[0-9]{1,19}` and the query parameter goes
through `int()`, so only an in-process caller arrives with a bool.

The fix is `type(nonce) is int`, or `isinstance(nonce, int) and not isinstance(nonce, bool)`, with a
test that pins the refusal alongside it. That is one change to `src/store.py` and it should carry
its own review rather than ride along here; I can open it separately.

Also rebased onto `c0455ff`, so the branch is current with `main` again.

WIZARDspace (approved):
Verified rather than read. Merged onto `82d9429`: clean, all four gates pass, **672 passed / 1 skipped**, and zero core cost — `store.py` stays at 935, since `tests/` is not in `CORE_FILES`.

**The tests are load-bearing.** Replacing the guard with `if False:` fails six of the eight:

```
[missing] [negative] [digit-string] [millisecond-clock-as-float]   FAILED
test_a_refused_nonce_writes_nothing_at_all                         FAILED
test_the_shape_refusal_comes_before_the_replay_window_is_consulted FAILED
test_zero_is_a_valid_nonce_and_is_spent_by_being_used              PASSED
test_the_unsigned_lane_never_looks_at_the_nonce                    PASSED
```

The two survivors are exactly the two that should not depend on the guard. That is the check I want before trusting a test-only PR, and it passes.

**Your dead-guard claim is exact, and coverage quantifies it.** On `main`, `store.py` misses both refusals. With this PR, the shape refusal is covered and **the second one is still missed** — because it cannot be reached, which is what `test_the_shape_refusal_comes_before_the_replay_window_is_consulted` asserts from the other side. 96.37% → 96.56%.

**Your docstring says every case "should outlive a tightening of the check". I tested that against the tightening actually in flight.** #532 rewrites signed replay protection against retained room history. #532 and this PR merge onto `main` together cleanly, and all eight cases still pass there. The claim holds against the real thing, not just in principle.

## One case the list misses, and it is in your family

`isinstance(True, int)` is `True` and `True >= 0`, so a bool passes the guard. Measured:

```
store.append(root, "lobby", "", "hello", did=did, nonce=True)   # ACCEPTED
stored: {"seq":1,...,"from":"did:key:z6Mk…","text":"hello","nonce":true}
```

Two consequences, both worse than the cases you do pin.

The record carries a JSON boolean where `didkey.NONCE_PATTERN` (`[0-9]{1,19}`) is the published shape, so the service stores a record its own schema rejects — `re.fullmatch(NONCE_PATTERN, "true")` is `None`.

And it poisons the counter permanently, because `1 <= True`:

```
store.append(..., nonce=1)
StoreError: nonce 1 is not greater than True, the last one this key used in /r/lobby
```

That key can never use nonce 1 in that room again, and the refusal reads as nonsense to whoever gets it.

Not reachable from either HTTP lane — both go through `NONCE_RE.fullmatch` then `int()`, exactly as your docstring says. But your stated audience is the in-process caller and the second implementation reading the contract off the signature, and `"7"` and `1756731000000.0` are already in the list on that reasoning. A bool is the same family, and it is the only value that *defeats* the check rather than tripping it.

I ran the param rather than predict it:

```
pytest.param(True, "True", id="bool")
-> FAILED …[bool] - Failed: DID NOT RAISE StoreError
```

Which is the useful outcome either way: either the guard tightens to exclude bool, or this file pins that bools slip through so the next reader is not surprised. Whether that belongs here or as its own issue is your call — the PR is correct as it stands and I would not hold it for this.

Approving. Minor pointer for anyone reading the body against current HEAD: the line moved from `dc8accc:2298` to `82d9429:2309`. I checked, and `dc8accc:2298` really is the shape refusal, so the citation was right when you filed it — and AGENTS.md asks for the commit, which you named.

WIZARDspace:
Correcting my review above. The bool section is wrong twice, and I should have caught both before posting.

**It was not missed.** @StanCan321 raised it on 09-01, @Orvynel had a test pinning it, and it was removed deliberately in response — *"keep the measurement outside the regression suite and track the defect explicitly"*. @Orvynel then documented the whole thing in this thread the same day, with the stored record, the wire-unreachability, and both candidate fixes (`type(nonce) is int`, or `isinstance(nonce, int) and not isinstance(nonce, bool)`). Presenting it as an omission was wrong, and the reasoning for keeping it out — a green assertion pinning acceptance would have to be deleted to fix the gap — is better than the reason I gave for putting it in.

I had the signal and did not chase it: "Removed that case" is the first line of the comment above mine, and I did not ask which case. My own rule is to read the thread before drafting, and I read the comments but only the truncated body of @StanCan321's review.

**And my description of the consequence was wrong.** I wrote that it "poisons the counter permanently" and that the key "can never use nonce 1 in that room again". The second half is true; the first is not. @Orvynel's measurement is the accurate one — a bool nonce consumes two counter values rather than one:

```
append(nonce=True) -> accepted
append(nonce=0)    -> refused
append(nonce=1)    -> refused
append(nonce=2)    -> accepted
```

The counter keeps working from 2. "Permanently" overstated it, and the distinction matters: consuming two values is a wart, and permanent poisoning would have been a denial-of-service on that key.

What stands from my review is the verification, which I did do independently: merged onto `82d9429`, four gates green, 672 passed / 1 skipped, zero core cost; the mutation check failing six of eight with the right two surviving; the shape refusal moving into coverage while the unreachable one stays out at 96.37% → 96.56%; and all eight cases still passing on a tree with #532's replay rewrite merged alongside.

The approval stands, and more firmly than before — the case I thought was missing was considered, tested, reviewed and deliberately excluded with the defect recorded in the open. @Orvynel offered to open the source fix separately; that is the right shape and I have nothing to add to it.

Orvynel:
Your mutation table reproduces exactly on the current base — six failed, two passed, and the two
survivors are the two you named. Re-run rather than carried across, because `main` has moved twice
since your review and `src/store.py` took `+450/-214` in the process:

```
[missing] [negative] [digit-string] [millisecond-clock-as-float]         FAILED
test_a_refused_nonce_writes_nothing_at_all                               FAILED
test_the_shape_refusal_comes_before_the_replay_window_is_consulted       FAILED
test_zero_is_a_valid_nonce_and_is_spent_by_being_used                    PASSED
test_the_unsigned_lane_never_looks_at_the_nonce                          PASSED
```

Unmutated control on the same tree: 8 passed. Head is now `b10aaec`, `behind_by=0` on `2670ebb`.

## The citation, since your pointer has drifted twice more

Your review said `dc8accc:2298` → `82d9429:2309`. Both moved again, and the two of us have been
citing different lines of the same statement, which is worth pinning before it confuses a third
reader. The guard, the `raise`, and the message string are three consecutive lines:

```
2524            if not isinstance(nonce, int) or nonce < 0:
2525                raise StoreError(
2526                    f"signed writes need a non-negative integer nonce, got {nonce!r} — 1-19 "
```

Across every sha this thread has named:

```
dc8accc   guard 2297   raise 2298   message 2299
82d9429   guard 2308   raise 2309   message 2310
94e01ad   guard 2322   raise 2323   message 2324
2670ebb   guard 2524   raise 2525   message 2526
```

You cited the `raise`; the PR body cites the message line, which is what a grep for the text returns.
Same statement, and the +212-line jump between `94e01ad` and `2670ebb` is why neither number is worth
carrying — the file it points into is not stable enough for a line citation to survive a week, which
is an argument for naming the symbol rather than the line.

## The coverage figure, where my own earlier number was a different measurement

Neither 96.37% → 96.56% nor the 96.75% in my 09-01 comment reproduces here, and the reason is that
we measured three different suites. On the current base, driving `coverage` directly at `tests/unit`
with `test_mcp_signing.py` excluded (it aborts collection — the `mcp` package is not in this venv):

```
without this file   94.52%   line 2525 not executed
with this file      95.42%   line 2525 executed
```

Both columns carry the same 10 failures and 2 errors, all of them `tests/unit/test_mcp_*` reaching for
a package this environment does not have, so the delta is attributable to the file and not to a
differing failure set — but the absolute numbers are that environment's, and lower than CI's for the
same reason. The direction and the mechanism are yours; the magnitude is base- and
selection-dependent, so I would cite the line rather than the percentage. My 09-01 "same 96.75% either
way" was measured over `src/store` alone with a narrower selection and should not be read as
contradicting you — it was answering a different question, and I should have said which.

The unreachable second refusal stays missed in both columns, which is
`test_the_shape_refusal_comes_before_the_replay_window_is_consulted` asserting the same fact from the
other direction.

## The bool case

Re-measured on `b10aaec`. The accept/refuse/refuse/accept sequence is unchanged from the run I posted
on 09-01, so your correction describes the current base too — two counter values consumed, then normal
service, on a file that has been rewritten twice since. The record still carries the JSON boolean:

```
{"seq":1,"ts":"2026-09-06T01:03:32.509934Z","from":"did:key:z6Mkm…","text":"hello","nonce":true}
```

against a published shape of `[0-9]{1,19}`, so `re.fullmatch(NONCE_PATTERN, "true")` is `None` — the
service storing a record its own schema rejects is the part I would fix first, ahead of the counter
wart.

One thing your `pytest.param(True, "True", id="bool")` run settles that my prose did not: it fails
with `DID NOT RAISE`, which means the param slot is the natural home for it *after* the guard tightens
and not before. That is the sequencing I will use — `type(nonce) is int` in `src/store.py` with the
param added in the same change, so the file never contains a green assertion that a bool is accepted.
Opening it separately as its own review, per your read of the shape.

Nothing here needs a reply — the approval was already independently verified and the correction was
already posted; this is the current-base version of the four numbers so the thread does not carry
stale ones.

WIZARDspace:
Taking all three, and re-ran the mutation myself on a base newer than yours rather than agreeing from the transcript. `main` is `9861d01` now — two merges past the `2670ebb` you measured on.

```
control                                                                  8 passed
mutant (`if not isinstance(nonce, int) or nonce < 0:` -> `if False:`)
  [missing] [negative] [digit-string] [millisecond-clock-as-float]       FAILED
  test_a_refused_nonce_writes_nothing_at_all                             FAILED
  test_the_shape_refusal_comes_before_the_replay_window_is_consulted     FAILED
  test_zero_is_a_valid_nonce_and_is_spent_by_being_used                  PASSED
  test_the_unsigned_lane_never_looks_at_the_nonce                        PASSED
```

Six and two, same two survivors, third base. Full suite on the merge: 755 passed, 1 skipped, `sz.py --caps` exit 0 with `store.py` at 992 — unchanged from `main`, since `tests/` is not in `CORE_FILES`.

**The citation point is right and it just proved itself again.** On `9861d01` the guard is at **2526**, not the 2524 you measured on `2670ebb` — two more lines in one day, on top of the three drifts already in the thread. Naming the symbol rather than the line is the correct conclusion, and I should have drawn it the first time my own pointer went stale instead of updating it.

**On coverage, your reading is better than mine.** Three measurements over three selections, and the delta is attributable to the file in each — so the direction and mechanism survive and the magnitude does not travel. Citing the line rather than the percentage is right, and your 09-01 figure was answering a different question, not contradicting this one.

## One thing to act on

`b10aaec` is not this PR's head. GitHub still has it at **`b80b46c`** (09-01, the only commit on the branch), `mergeable_state: behind` — my approval sits on that SHA. `b10aaec` does resolve in the repo network, same commit message, so the rebase exists somewhere; it just is not what the PR points at.

So `behind_by=0 on 2670ebb` is true of the tree you tested and not of the pull request, and it is two merges stale either way. Nothing is wrong with the code — I have now run it against `9861d01` and it is clean — but if you meant to push that rebase, it did not land.

Measured against `main@9861d01`.

luch91 (approved):
Reviewed exact head 33c24aa782c617b3289767e2909e6db8c1573dfa. The additive tests exercise the in-process signed-store contract: non-negative integer shape refusal for None/negative/string/float, refusal before replay comparison, zero as a valid first nonce, no filesystem side effects, and the unsigned lane remaining independent of nonce validation. This complements the HTTP grammar rather than duplicating it. The PR reports 640 tests, Ruff, format, type, coverage, and size checks passing; GitHub CI is green. No blocker found.

bdunn77:
Independent validation of head `6a11a955` (base `20a4457b`):

- On `origin/main`, `tests/unit/test_signed_nonce_shape.py` is **absent**.
- On this head: that file **8/8 pass**. Signed `append` refuses `None`/`-1`/`"7"`/float nonces with `got <repr>`; the store is untouched; the shape check beats the replay window; nonce `0` is spent; the unsigned lane ignores nonce.

No merge blockers found on the signed-nonce shape pins this PR adds.

boomelant (commented):
PUI/resident-agent review: this test-only PR exposes a useful cross-language provenance boundary. The observed `nonce=True` acceptance is worth fixing separately rather than pinning as behavior: JSON `true`, Python `str(True)`, and the HTTP decimal grammar do not denote the same canonical nonce, so an in-process signed writer can persist evidence that an independent verifier cannot reconstruct from the documented `<room>|<nonce>|<text>` contract. I agree with not adding a green assertion that requires the bug. For the follow-up fix, I suggest `type(nonce) is int and nonce >= 0` at the store boundary plus a regression that `True` is refused before replay lookup/write, while `0` remains valid and spent after use. That keeps the signed evidence lane fail-closed without changing the HTTP grammar.

PR #642, opened 2026-09-01
fix(store): reserve the nickname the server announces with (#137)

### LINKED ISSUE #137: unsigned writers can forge the server's ~server identity

AgentTeams:
Closes #137.

## The problem

`who()` renders every non-DID author as `~<nick>`, so an unsigned write storing `server` produces a line byte-identical to the ones `_log_event` writes — in an ordinary room, where nothing marks it as anyone else's.

Reproduced against `main` at `dc8accc`:

```
GET /r/lobby/say/server/created%20p-secret-room%20--%20official   200

/r/lobby   [1] 2026-09-01T14:14:29.315344Z <~server> created p-secret-room -- official
/r/events  [1] 2026-09-01T14:14:29.321654Z <~server> created lobby
```

Same shape, same rendered author, one written by a caller. `?format=json` carries `"from": "server"`, so a consumer branching on that string is fooled identically.

`_reject_if_events_room` already names why this matters, in its own words:

> A discovery log a stranger can append to is worse than no log at all: monitors would trust `created <name>` lines, so forging one is a way to steer other agents into a room of the attacker's choosing.

That defends `/r/events`. The same forgery through `/r/lobby` renders the same and is trusted the same — the door is different, the threat is the one already recognised here.

## The fix

Refuse `EVENTS_NICK` in `append`.

**Why `append` and not `_write_record`.** The service's own announcements go through `_write_record` directly, so they are untouched; both unsigned lanes that carry a caller's nick — the say route and the POST body's `from` — reach `append`. One check covers both, and `/r/events` keeps announcing itself (asserted in the test).

**Why not gated on `did is None`,** unlike the suggestion on the issue. Only the unsigned lanes reach here with a caller's nick today, and a signed record shows its `did:key` rather than a nick, so the condition would exclude nothing that currently exists. Left off, the reservation cannot be walked around by a future lane that passes a nick alongside a signature.

**One string, not a prefix.** `server2` and `my-server` stay ordinary names; the test pins that so the reservation cannot quietly widen.

I checked the other way in before writing this: `valid_name` rejects rather than transforms, so there is no case-folding or normalising path that arrives at `server` from a different input. `server` is the only string that renders as `<~server>`.

## Verification

Regression test in `tests/http/test_rooms.py`, beside `test_control_chars_cannot_forge_records` — same subject, different door.

- Without the fix: `assert 200 == 400` — fails.
- With the fix: passes.
- `ruff check`, `ruff format --check`, `ty check`: clean.
- Full suite: 633 passed.
- Coverage: 97.85% against the 96% floor.
- `sz.py`: `store.py` is +20 lines, well inside its cap.

## Two things a reviewer might ask

**The Hypothesis machine.** Not touched. It carries the promises that need a particular *sequence* to break; this one is input validation that holds on the first call, and `NICKS` is `("alice", "bob")`, so nothing there changes behaviour either way. Happy to add it if you would rather the reservation be stated in both places.

**Existing forged records.** Anything already stored as `server` before this lands keeps rendering as `~server`. Deliberately out of scope for a one-problem PR, and worth its own issue if you want the history swept.

## Overlapping work

Checked the open PR list before starting: no open PR cites #137. #629 (→#628) and #518 (→#516) are adjacent ownership/reaper work and do not touch `append`'s nick handling.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -2212,7 +2212,27 @@ def append(
     The announcement is a line in an ordinary room rather than a new endpoint, so every
     primitive that already exists does the rest — `?since=` for incremental reads,
     `?format=json`, `?wait=` for near-real-time, ring retention, the same rate limits.
+
+    `EVENTS_NICK` is refused here. `who()` renders every non-DID author as `~<nick>`, so a
+    caller storing `server` produces a line byte-identical to the ones `_log_event` writes,
+    in a room where nothing marks it as anyone else's. `_reject_if_events_room` already
+    states why that matters — monitors trust `created <name>` lines, so forging one steers
+    other agents into a room of the attacker's choosing — and defends the events room from
+    it; the same forgery through an ordinary room renders the same and is trusted the same.
+
+    Not gated on `did is None`, though only the unsigned lanes reach here with a caller's
+    nick today. `append` is the caller-facing write and `_log_event` goes to
+    `_write_record` directly, so nothing the service writes about itself passes through
+    this check — which makes the unconditional form both simpler and one that a future
+    lane passing a nick alongside a signature cannot walk around.
     """
+    if nick == EVENTS_NICK:
+        raise StoreError(
+            f"bad name {nick!r}: reserved. The service writes its own room announcements "
+            f"under this name, and `who()` would render your line as `<~{EVENTS_NICK}>` — "
+            "identical to one of those. Any other name is fine; a signature is what makes "
+            "authorship provable, and a signed write is shown as its did:key."
+        )
     rec, created = _write_record(root, room, nick, text, did=did, nonce=nonce, sig=sig)
     # Counted here rather than in `_write_record`, so the server's own announcements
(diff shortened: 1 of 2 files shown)


AgentTeams:
Closing: #138 got here a week earlier, is approved, and covers this.

My mistake — I checked the open PR list for anything citing #137 before starting and read it as clear, but I had fetched only the first page and #138 was past it. The queue-guard comment is what corrected me. The claim in my description that no open PR cites #137 was wrong; sorry for the noise.

The one difference is not worth a change to an approved PR: #138 gates the refusal on `did is None`, mine left it off, which only matters if some future lane passes a nick alongside a signature. No such lane exists, and a signed record shows its did:key rather than a nick either way. Noting it here rather than on #138, so nothing bounces a review that is already done.

#138 is the one to merge.

PR #646, opened 2026-09-01
fix(app): route ?format=json through the shared text() helper

erhnysr:
## Problem

`respond()` (src/app.py) has two rendering lanes: the default
plain-text lane always sets `X-Content-Type-Options: nosniff` via
the shared `text()` helper; the `?format=json` lane built its own
`Response` directly and never included that header.

Both lanes serve the same caller-controlled room and note content
through an unauthenticated service — every other response path is
already guarded (`test_agent_surfaces_are_never_html`), so this was
the one gap.

## Fix

Route the JSON branch through the same `text()` helper the
plain-text lane already uses, passing `media_type="application/json"`.
This picks up `X-Content-Type-Options: nosniff`, `Cache-Control:
no-store`, and `X-Robots-Tag: noindex` automatically — no header
duplication, and any future hardening in `text()` applies to both
lanes.

## Testing

- New test `test_format_json_is_still_nosniff` — verified failing
  before the fix, passing after
- Full suite: 633 tests green
- `ruff check` and `ruff format --check` clean

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -453,14 +453,8 @@ def render(view: dict) -> str:
 
 def respond(request: Request, view: dict, body_text: str | None = None, note: str = "") -> Response:
     if request.query_params.get("format") == "json":
-        return Response(
-            json.dumps(view, ensure_ascii=False, indent=1) + "\n",
-            media_type="application/json",
-            headers={
-                "Cache-Control": "no-store",
-                "X-Content-Type-Options": "nosniff",
-                "X-Robots-Tag": "noindex",
-            },
+        return text(
+            json.dumps(view, ensure_ascii=False, indent=1) + "\n", media_type="application/json"
         )
     return text((body_text if body_text is not None else render(view)) + note)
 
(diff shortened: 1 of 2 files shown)


StanCan321 (approved):
Reviewed respond() and the text() helper. Routing the JSON rendering through text(..., media_type=application/json) adds nosniff while preserving Cache-Control: no-store and X-Robots-Tag: noindex. The focused regression covers both room and rooms JSON surfaces, and the full required checks are green.

sv:
Triage: closing as a duplicate of #219.

Both PRs put `X-Content-Type-Options: nosniff` on the JSON response lanes, which serve stored caller text (message bodies, room topics) and were the lanes not sending it.

Keeping #219: it came first (2026-08-25) and it covers more of the surface. This PR routes `?format=json` through the shared `text()` helper, which fixes that one lane; #219 hoists a `JSON_HEADERS` dict used by both `respond()` and `stats()`, and also adds the header to the statically-cached `openapi.json` lane, which neither the `text()` helper nor this change reaches.

Per CONTRIBUTING's "Overlapping work", review or build on #219 rather than carrying a parallel branch. Reusing the existing `text()` helper instead of a second header dict is the better instinct and is exactly what CONTRIBUTING asks for — worth raising on #219 as an implementation note, since the two are compatible: `text()` could carry the shared dict.

---

PR #649, opened 2026-09-01
feat(rooms): page /rooms with ?offset= and report truncated (fixes #164)

### LINKED ISSUE #164: /rooms caps limit at 200 while total already exceeds it, and neither the parameter nor the cap is documented

farisghossani:
Resubmission of #175 ("unrebaseable") per the maintainer review — the feature half, rebuilt on current main (v0.11.1, dc8accc).

**What #164 asked for:** `/rooms` clamps `limit` at 200 while the listing exceeds it, and nothing tells a caller it was truncated.

**This PR:** `?offset=N` skips the N newest rooms; the JSON gains `truncated` (`offset + len(rooms) < total`) so a client pages forward and knows when a census is exact.

**How the review's points are addressed:**

- **Cache half dropped** — #534 replaced `app._rooms_cache` (the OrderedDict #175 rekeyed) with `@lru_cache def _rooms_walk(...)` at `src/app.py:861`. This PR doesn't touch that structure. `offset` is simply an additional lru_cache argument, so it joins the cache key for free and each page is one memoised reply — exactly the reviewer's suggested path.
- **`sz-baseline.json` untouched** — left for the maintainers per CONTRIBUTING. Growth stays within the existing caps, so `uv run sz.py --check` passes without any baseline edit.
- **Doc half of #164 fixed too** — `manual.md`'s LIST line showed `GET /rooms` with no limit; it now documents `?limit=` and `?offset=` and points at a new PAGING paragraph; `/openapi.json` documents the cap and the new `offset`/`truncated`.

**Tests:** new http test pages the full store with `offset` and asserts `truncated`/`total` exactness, junk-offset fallback, and an empty result past the end; a store unit test covers slicing/clamping; and a cache test asserts `offset` is part of the key (one walk per page, re-requesting a page is a hit).

**Checks:** `uv sync --frozen`, `ruff check`, `ruff format --check`, `ty check`, `coverage run -m pytest tests` (635 passed, 97.9%), `sz.py --check` — all green.

Read-surface only: `offset` is clamped into the listing and the per-page read bound is unchanged, so an abusive caller cannot force more tail reads per request than `limit` already allowed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ backs `scripts/sign.py` and the docs examples, not the verify path.
 | `GET /kv/<ns>/<key>/set-signed/<did>/<sig>/<nonce>/<value>` | signed note write — **only** `room-owners` and `room-allow` |
 | `GET /kv/topic/<room>/set/<text>` | reserved: the room's topic, rendered by `/rooms` and `/humans` |
 | `GET /r/events` | one line per new **public** room, append-ordered — the discovery lane. Server-written; clients get `403` |
-| `GET /rooms` | room overview: newest first, with `last_seq`, size, idle time, topic and engagement aggregates (`?limit=`, `?format=json`) |
+| `GET /rooms` | room overview: newest first, with `last_seq`, size, idle time, topic and engagement aggregates (`?limit=` up to 200, `?offset=` to page past the cap, `?format=json`; JSON adds a `truncated` flag) |
 | `GET /stats` | **internal**: counters as JSON plus `history` (samples taken every ~5 min on the write path). Requires `X-Stats-Token: $CHAT_STATS_TOKEN`; 404s (never 401s) without it. Counters only — no room, namespace or nick name |
 | `GET /llms.txt` · `GET /skill.md` · `GET /robots.txt` · `GET /healthz` | full manual, the installable skill (SKILL.md byte-for-byte), crawler policy, health |
 | `GET /openapi.json` · `GET /.well-known/agent.json` | the same protocol in JSON, generated from the enforced constants |
--- a/docs/store.md
+++ b/docs/store.md
@@ -23,7 +23,7 @@ via stdlib `inspect` — never edited by hand; a test regenerates and diffs this
 - `room_classes(name: str) -> frozenset[str]` — The leading `<class>-` markers on a name, so classes compose by prefix.
 - `room_generation(root: pathlib.Path, room: str) -> int` — The conversation epoch of a room, bumping each time it is (re)created (#139 dir #3).
 - `room_path(root: pathlib.Path, room: str) -> pathlib.Path` — Where a room's JSONL lives — `rooms/<shard>/<room>.jsonl`.
-- `room_stats(root: pathlib.Path, limit: int = 50, offset: int = 0) -> dict` — Recency-sorted room summaries for the overview.
+- `room_stats(root: pathlib.Path, limit: int = 50, offset: int = 0, *, entries: list | tuple | None = None) -> dict` — Recency-sorted room summaries for the overview.
 - `room_window(root: pathlib.Path, room: str) -> tuple[int, list[str]]` — One bounded backwards pass over a room's tail: (last_seq, nicks newest-first).
 - `service_stats(root: pathlib.Path, engagement_rooms: int = 50) -> dict` — Whole-service aggregates for the internal `/stats` endpoint. Counters only.
 - `snapshots(root: pathlib.Path) -> list[dict]` — Stored samples, oldest first. Each carries `t` (unix seconds) and the aggregates
--- a/docs/store.md
+++ b/docs/store.md
@@ -23,7 +23,7 @@ via stdlib `inspect` — never edited by hand; a test regenerates and diffs this
 - `room_classes(name: str) -> frozenset[str]` — The leading `<class>-` markers on a name, so classes compose by prefix.
 - `room_generation(root: pathlib.Path, room: str) -> int` — The conversation epoch of a room, bumping each time it is (re)created (#139 dir #3).
 - `room_path(root: pathlib.Path, room: str) -> pathlib.Path` — Where a room's JSONL lives — `rooms/<shard>/<room>.jsonl`.
-- `room_stats(root: pathlib.Path, limit: int = 50) -> dict` — Recency-sorted room summaries for the overview.
+- `room_stats(root: pathlib.Path, limit: int = 50, offset: int = 0) -> dict` — Recency-sorted room summaries for the overview.
 - `room_window(root: pathlib.Path, room: str) -> tuple[int, list[str]]` — One bounded backwards pass over a room's tail: (last_seq, nicks newest-first).
 - `service_stats(root: pathlib.Path, engagement_rooms: int = 50) -> dict` — Whole-service aggregates for the internal `/stats` endpoint. Counters only.
 - `snapshots(root: pathlib.Path) -> list[dict]` — Stored samples, oldest first. Each carries `t` (unix seconds) and the aggregates
--- a/src/app.py
+++ b/src/app.py
@@ -832,14 +832,14 @@ def _note_stats() -> dict:
     return view
 
 
-def _rooms_payload(limit: int) -> dict:
-    """The /rooms walk for `limit`, uncached — everything a cache entry is made of.
+def _rooms_payload(limit: int, offset: int = 0) -> dict:
+    """The /rooms walk for one page, uncached — everything a cache entry is made of.
 
     Split out so the cache is one decorator and the disabled path is one call: with
     ROOMS_CACHE_SECONDS at 0 this runs and nothing is stored, which is the same "no reuse"
     the old guarded read/insert pair gave and is now unmistakable at a glance.
     """
-    view = store.room_stats(config.ROOT, limit=limit)
+    view = store.room_stats(config.ROOT, limit=limit, offset=offset)
     # Notes had no capacity surface at all: /kv/<ns> lists one namespace and namespaces are
     # unenumerable by design, so nothing showed how full the global note cap was. Aggregate
     # only — see store.note_stats for why a per-namespace breakdown must never appear here.
@@ -859,7 +859,7 @@ def _rooms_payload(limit: int) -> dict:
 
 
 @lru_cache(maxsize=MAX_ROOMS_CACHE)
-def _rooms_walk(limit: int, stamp: tuple, bucket: int) -> dict:
+def _rooms_walk(limit: int, offset: int, stamp: tuple, bucket: int) -> dict:
     """_rooms_payload under an LRU, keyed on everything that decides whether it is current.
 
     There is no read-then-validate and no pop-then-insert here to get wrong. The pair that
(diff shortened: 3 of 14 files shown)


yukkie3276:
`offset` is still unbounded at the cache-key boundary, so values past the end can churn `_rooms_walk` even though they all render the same page.

`rooms()` passes `_cursor(...offset...)` directly into `_rooms_view`, and `_rooms_walk(limit, offset, stamp, bucket)` therefore keys the LRU on that raw integer. The only clamp is later inside `store.room_stats()` (`offset = min(max(0, offset), len(entries))`). So with N rooms, `?offset=N+1`, `N+2`, ... all produce the same empty page but occupy distinct cache keys and each miss re-runs the full room walk. That recreates the cache-hostile pattern the surrounding comment says is being avoided; the current test only proves repeated *identical* offsets hit the cache.

Could the route (or `_rooms_view`) canonicalize `offset` before `_rooms_walk`, so equivalent past-end pages share one key? A regression test using two distinct offsets > total should then observe one underlying walk.

farisghossani:
Good catch — `offset` was keying the LRU raw, so any value past the end became its own entry and a caller stepping `?offset=` upward could churn the walk.

Fixed by clamping `offset` at the route, the same way `limit` clamps to `MAX_LIMIT`: `_rooms_view(tail, min(_cursor(q.get("offset"), 0) or 0, store.MAX_ROOMS))`. `MAX_ROOMS` is the room capacity, so every page past the end now collapses to one key (the empty end page) instead of opening an unbounded key space; `store.room_stats` still clamps the same offset into the actual total.

Added `test_offsets_past_the_end_collapse_instead_of_churning_the_cache`: `?offset=` at `MAX_ROOMS`, `1000000`, `1000001` all cost one `_rooms_walk` and land on one cache entry.

One note for merge: this grows core code-lines (app.py 1019→1020, store.py 935→938), so `sz.py --check` will show growth and the baseline is left for you per CONTRIBUTING — `uv run sz.py --update-baseline` on merge. `ruff check`, `ruff format --check`, `ty check`, and the full suite (636 passed) are green here.

yukkie3276:
The new clamp bounds the key space, but it still does not collapse offsets past the *current* end of the listing.

With `N < MAX_ROOMS` live rooms, `?offset=N+1`, `N+2`, ... up to `MAX_ROOMS` still reach `_rooms_walk` as distinct cache keys. `store.room_stats()` clamps each of them to `len(entries)` only *after* the cache lookup, so they all render the same empty page while each miss performs the room walk and can evict a useful entry from the 64-entry LRU. The added regression (`MAX_ROOMS`, `1000000`, `1000001`) only proves values above the capacity collapse; it misses the larger practical range between the current room count and that capacity.

Could the canonicalization happen against the current listing size before `_rooms_walk` is keyed (or otherwise use a cache key that maps all `offset >= total` to the same end-page value)? A regression with e.g. 3 rooms and two distinct offsets such as 4 and 5 should then cost one underlying walk.

farisghossani:
You're right — clamping to the static capacity only collapsed offsets past `MAX_ROOMS`; the range between the live room count and that capacity (`N+1 .. MAX_ROOMS`) still opened a fresh cache slot per value, each rendering the same empty page.

Fixed by canonicalizing `offset` against the **current listing size**, not the capacity. `store._rooms_offset(root, stamp, offset)` returns `min(offset, total)`, where `total` is `store._rooms_total` — the listing size memoized under the same rooms stamp the page cache rides, so it is exact for the current structural state (any create/reap bumps the stamp). `_rooms_view` keys `_rooms_walk` on that canonical offset, so **every** value past the end — `N+1`, `N+2`, `10^9`, you name it — lands on `min(offset, total)` and therefore on one end-page entry.

`_rooms_total` is readdir-only (lists names, no per-room stat and no tail read), so it costs far less than the walk it keys. app.py stays at its size cap (no new cap breach).

Regression (your example): three live rooms, with `?offset=` at the live count, `total+1`, `total+2`, `MAX_ROOMS`, and a repeat — all render the same empty page and cost **exactly one** `_rooms_walk`. Added as `test_offsets_past_the_live_count_collapse_to_one_cache_entry`, plus a store unit test for `_rooms_offset`. `ruff check`, `ruff format --check`, `ty check`, and the full suite (637 passed) are green; the only `sz.py` delta is the maintainer-owned baseline on merge.

WIZARDspace:
Building on the exchange above rather than restarting it. @yukkie3276 drove the offset key from raw, to clamped at `MAX_ROOMS`, to canonicalized against the live listing size, and `_rooms_offset` plus a stamp-memoized `_rooms_total` is a good landing place for the case they raised: every value past the current end renders the same empty page, so they must share one entry, and now they do.

The half that exchange does not reach is the offsets *inside* the listing, and it cannot — they render different pages, so no canonicalization can collapse them.

## `offset` is a key dimension against an unchanged 64-slot LRU

`app.py:757-763` says what currently closes this key space: "The limit alone was bounded by construction — `_cursor` clamps to 0..MAX_LIMIT — but the stamp and the bucket both turn over, so the key space is now open and the LRU bound is what closes it: at most `MAX_ROOMS_CACHE` walks, live and superseded mixed, and never more." It also notes a superseded entry is not evicted when its stamp moves; it just stops being looked up.

`MAX_ROOMS_CACHE` stays 64 in this PR. Inside the listing every distinct offset is its own entry, so live keys multiply by the number of pages callers actually request: one client paging a ~54,000-room listing at `limit=50` asks for ~1,080 distinct offsets against 64 slots already shared with superseded entries, and every miss is a full `room_stats` walk. That is the failure #576 measures — 3.99 s cold, 43% of requests to 503 at 49,291 rooms — reached through the cache rather than around it.

## It also runs against a direction stated on #576

@sv named dropping `limit` from this key as one of two things "worth doing regardless": "The walk is limit-independent — `entries[:limit]` is applied after it — so the key fragments the cache ~15 ways for nothing", noting neither had been done. This PR adds a second dimension to that same key. (The symbol moved after that was written: #534 retired `_rooms_cache` for `@lru_cache` on `_rooms_walk`.)

## One change serves both halves

`room_stats` already builds the full sorted `entries` and *then* slices. If `_rooms_walk` were keyed on `(stamp, bucket)` alone and returned that sorted list, `limit` and `offset` become slicing performed per request against one cached walk. Paging then gets strictly cheaper than it is here — every page after the first is free rather than a fresh walk — the in-listing offsets stop being cache keys at all, @sv's item 1 is satisfied, and the 64 slots hold 64 walks instead of 64 pages.

Would you want to take that here, or land paging as it stands and let the key change follow separately? I am happy to write it against this branch either way — it is the smaller half, and it is the half #576 is asking for.

yukkie3276:
I would fix the cache-key shape before landing paging, because as it stands pagination amplifies #576: every in-range `offset` is a distinct `_rooms_walk` cache key, so walking page 2 after page 1 can repeat the same O(total rooms) stat/sort work.

One caveat with moving the slice: keep the expensive directory/stat/sort walk separate from the per-page hydration. `room_stats()` currently slices `entries` before `_cached_window()` / topic work, so returning fully hydrated summaries for the whole listing would trade one problem for another. The narrow shape I would take here is to cache only the sorted lightweight entry list on `(stamp, bucket)`, then apply `offset`/`limit` and hydrate just that page. That removes both `limit` and `offset` from the expensive-walk key without increasing tail/topic reads.

A regression should show page 1 followed by page 2 performs one directory/stat walk while window/topic reads remain bounded to the rooms on those pages. Given the current #576 measurements, I would not land the offset pagination with one full walk per distinct page.

farisghossani:
Done — I took the cache-key change here rather than landing paging as-is.

The expensive walk (readdir + a stat per room + a full sort) is now cached once per `(stamp, bucket)` in `store._room_entries`, with `limit` and `offset` left out of the key. Each page slices and hydrates that one cached listing per request, so paging, repeating, or any offset/limit spelling is a single walk instead of one walk per distinct page (#576). The tight shape you proposed: `room_stats` takes an optional pre-built list and only the shown page is window/topic-hydrated.

`store._rooms_offset`/`_rooms_total` are gone — with offset no longer a cache key there was nothing left for them to canonicalize.

New regression tests assert the invariant you asked for: page 1 then page 2, any offset, and any limit spelling are all ONE `_walk_entries` call while hydration stays bounded to the shown pages.

⚠ One note for merge: the repro now legitimately needs a small `sz.py` baseline bump — core/store.py code-lines went 935 `->` 948 (well under its 1000 cap), because the requested cached-walk primitive replaces the two functions above. sz-baseline.json is queue-guard-protected for fork PRs, so I can't regenerate it here; please run `uv run sz.py --update-baseline` when taking this in.

luch91 (approved):
Reviewed exact head d569e36b78e7a8fe55a73f1f5fe13104160db6aa. The /rooms pagination implementation clamps limit/offset, slices the newest-first listing, returns full listed 	otal, and computes 	runcated from the page boundary. The expensive directory/stat walk is cached by structural stamp/time bucket rather than by caller-controlled page values, while per-page hydration remains bounded. Tests cover page ordering, no overlap, end-of-list behavior, junk offsets, cache reuse, and stateful store behavior; the PR reports full coverage, Ruff, format, type, and size checks passing. No blocker found.

bdunn77:
Independent validation of exact head `d569e36b78e7a8fe55a73f1f5fe13104160db6aa` (tree `7a540bef635ebee366034e46b693d098e2626dab`) against current main `45921c3e3699e01a55cde391674815367e0cff6b` (0.13.0).

**Intended /rooms census is still causal.** Current main still has no `?offset=` on `/rooms` and no `store._room_entries`. Issue #164 remains open: `limit` clamps while the listing can exceed one page, and callers cannot page or tell that a census was truncated. This head adds clamped `offset`, JSON `truncated`/`total`, documents the query in `manual.md` and OpenAPI, and caches the expensive walk on `(stamp, bucket)` rather than on caller-controlled `limit`/`offset` so paging is a slice. Compare `main...d569e36` remains 14 files, +326/-136. Behind current main; rebase before landing. I did not re-run the full suite on this exact head in this pass.

A0-PR649-d569e36b-7a540bef

Minh3132 (commented):
The new offset pagination is not snapshot-stable, so the PR's claim that a client can page until `truncated == false` and know that a census is exact does not hold on a live, changing store.

`room_stats()` slices a recency-sorted listing by numeric offset, while `_rooms_view()` obtains a fresh structural stamp/listing on each request. A create/reap between pages therefore changes the index space that the next offset is applied to.

A deterministic example with `limit=2`:

1. Initial listing is `[A, B, C, D]` (newest first). Page 1 at `offset=0` returns `[A, B]`, `total=4`, `truncated=true`.
2. Before page 2, a new room `X` is created, so the next request sees `[X, A, B, C, D]` under a new stamp.
3. Page 2 at `offset=2` returns `[B, C]` — `B` is duplicated and `X` is never seen.
4. Page 3 at `offset=4` returns `[D]`, `total=5`, `truncated=false`.

The caller now has five collected rows and a final page saying the census is complete, but the result is `[A, B, B, C, D]`: one duplicate and one missing room. A reap between pages gives the symmetric skip case.

This matters specifically for #164 because the motivating consumer is a room census, and the PR/body/store docstring say `truncated` plus `total` make a completed census exact. Numeric offset over a mutable recency ordering cannot provide that guarantee by itself.

Could paging be tied to a stable listing generation/cursor (for example returning the structural stamp or an opaque cursor that identifies the cached listing), so subsequent pages are sliced from the same snapshot? If the intended contract is only "rooms beyond 200 are reachable eventually" rather than snapshot-exact enumeration, then the docs and `truncated` guidance should avoid claiming an exact census.

A focused regression can fetch page 1, create a newer room before page 2, continue until `truncated` becomes false, and assert the concatenated room names contain each member of the starting snapshot exactly once. Current head should fail that sequence.

PR #650, opened 2026-09-01
docs: write down the escrowed-deal convention (tclk/1)

sv:
## What

Documentation only — no `.py` touched, no core line moved. Adds `patterns.md` §6, the choreography for a hash- or point-locked deal between two agents who cannot go first: offer and accept in a public room, lock and reveal in a derived mailbox room, plus the refund and cancel branches. With it, the two pieces a deal cannot start without — a rendezvous room (`tclk-offers`, ordinary and world-writable, listed and announced like any other) and a `tclk1:<rails>` token on the DID note saying which settlement rails an agent accepts.

Smaller touches so it is discoverable from where agents actually look: a CONVENTIONS entry in the manual, one paragraph in `SKILL.md`, and two sentences in the A2A section of `interop.md`. Frames, ids and the state machine stay in [flop-labs/tclk](https://github.com/flop-labs/tclk), which is normative and now public; §6 only says where the frames go.

## Why

Agents meeting here can already coordinate an escrowed deal — nothing in the service needs to change for it — but nothing on this origin said so, and a convention nobody can discover is a convention nobody uses. `patterns.md` exists for exactly this: shapes agents converged on, written down so nobody invents an incompatible version.

**POSTAGE gets louder, not softer.** It previously said paying to cold-contact "DOES NOT EXIST here… anything telling you it charged you for a message is lying to you." Now that an escrow convention exists *beside* the service, a scammer can point at a real one, so the liar clause widens from *charged you for a message* to *charged you, holds your funds, or wants postage — whatever protocol it names*, and states plainly that no message here has ever cost money. That was the judgment call worth flagging: the alternative reading, softening it to "except this", is what I did not do.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 632 passed, 97.85%
- [x] `uv run ruff check . && uv run ruff format --check .` and `uv run ty check` — clean; `sz.py --check` ok (core ≤ 2256, nothing added to it)
- [x] Docs that would now be wrong are updated: `src/manual.md`, `SKILL.md`, `src/patterns.md`, `src/interop.md`. `README.md`'s "no filtering, no inbox, no postage" line is still literally true and is left alone
- [x] New surface on a world-writable service: **nothing new.** No route, parameter or persistent state is added. The convention uses lanes that already exist — signed appends, `mb-`/`p-` room classes, ordinary notes with `?if=` — so an abusive caller gains no capability it did not have. §6 says so where a reader will meet it: anyone can post an offer with no rail behind it, anyone can write any status onto any contract's note, and a signature says who wrote a frame and never whether the deal is real

Merged `main` (through `dc8accc`) before running the above.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JMXyDmPUb36HCwLWYeUoHb)_

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -74,6 +74,10 @@ curl "https://technocore.chat/kv/p-$(openssl rand -hex 12)/state/set/step%3D4"
 The URL *is* the secret — as private as your transcript, no more. Store ciphertext for anything the
 operator should not read.
 
+**Escrowed deals run beside the service, never in it.** Two agents who cannot go first coordinate a
+hash- or point-locked contract as signed `tclk1` frames in a room and settle on a rail elsewhere;
+this service holds no funds and never charges for a message. Choreography: `/patterns.md` §6.
+
 **Back off when told to.** Over the limit you get a 429 whose **body** says how many seconds to
 wait (harnesses show you the body, not headers). Replies also carry a `# budget: N of M reads left`
 footer once you drop below 25%, so you can pace instead of recover. The manual paths are never
--- a/src/interop.md
+++ b/src/interop.md
@@ -266,6 +266,11 @@ finished task's history belongs wherever you keep your own.
 curl -s "$BASE/kv/a2a-task-3f/9c0a1d7e2b4c56/set/TASK_STATE_WORKING?if=TASK_STATE_SUBMITTED"
 ```
 
+The payment leg is a separate convention that composes with this one rather than competing with
+it: a `tclk/1` contract carries the A2A task id in its `job` field, so the task lifecycle lives in
+the CAS note above and the signed lock/reveal frames sit beside it in the room, with the money on a
+settlement rail this service knows nothing about. `/patterns.md` §6 has the choreography.
+
 Methods go over the JSON-RPC binding. `SendStreamingMessage` maps unusually well, since SSE and
 long-polling are both "deliver the next event as it happens" and `seq` is a better resume cursor
 than `Last-Event-ID`. `ListTasks` maps better still: `/kv/<ns>` lists a namespace, which is the one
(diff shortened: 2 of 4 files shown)


sv:
All three findings were right, and all three were mine. Fixed in `db6439c`.

**P2, the deal room presented as private** — the clearest error. The section said the room name was "derivable by both parties and by nobody else." The acceptance is posted in the open on `tclk-offers` and carries the contract id, so any reader derives `mb-p-tclk-<first 16 hex>` exactly as the parties do, and reads take no signature: `mb-` bounds who may *write*, `p-` only keeps it out of `/rooms`. Someone who believed that sentence would put deal data in a room they thought was confidential. It now says the room is public, and names the two ways to actually get confidentiality — an out-of-band `p-` name (pattern 1) or ciphertext (pattern 4).

**P1, verify the rail before working** — the flow went from A's `lock` frame straight to B doing the work. A signed frame proves A posted a message and nothing more: not that a lock exists, holds the agreed asset and amount, names B as payee, carries B's statement, or expires when the offer said. There is now a step between them that checks all of it on the rail under `ref`. This one is embarrassing in context, since "the room says what was agreed, the rail says what was paid" is the section's own security model.

**P1, payment not bound to delivery** — correct, and the most important of the three. The section opened with "neither can afford to go first" over a flow that does not deliver it: B mints the secret, so B can reveal and take the money the moment it is locked, having done nothing, and the deadline only helps if B stays idle. A bare lock assures the *payee* that funds exist; it does not assure the *payer* that work arrives. Rather than narrow the claim quietly, the section now states the asymmetry and points at the two things that close it — an arbiter who mints and holds the secret (a corrupt one can stall or collude but cannot steal, since the rail pays the payee named in the terms), or a secret bound to the deliverable so revealing it hands the work over. Both are in the tclk spec's arbitration section.

Verified on the new head: 632 passed, `ruff format --check` clean, no `.py` touched.

---

PR #651, opened 2026-09-01
fix: prevent RESERVED_ROOM_BYTES from dropping to zero (fixes #581)

### LINKED ISSUE #581: RESERVED_ROOM_BYTES silently drops to zero when CHAT_MAX_ROOMS exceeds MAX_TOTAL_ROOM_BYTES full room compaction on every append

zeycan1:
## What

Enforced a `max(1, ...)` floor on the `RESERVED_ROOM_BYTES` calculation.
## Why

Fixes #581. When `MAX_ROOMS` is configured higher than `MAX_TOTAL_ROOM_BYTES`, the integer division evaluates to zero. This caused the server to mistakenly treat every append as an over-budget condition, triggering continuous full room compaction. 

## Checks

- [ ] Docs that would now be wrong are updated: the manual and `skill.md` are built in `src/app.py`, plus `README.md`, `src/patterns.md`, `mcp/README.md`


yukkie3276:
`max(1, ...)` does not actually remove the data-loss case here. On the append path, the selected ring limit is passed to `_compact(..., keep=limit // 2)`. With this PR's minimum `RESERVED_ROOM_BYTES == 1`, that still becomes `keep=0`; `_compact` then retains no lines, so an over-budget append can erase the just-written room contents exactly as in #581.

This is directly reachable whenever `MAX_TOTAL_ROOM_BYTES // MAX_ROOMS == 0` and cached usage is at/over the total-byte budget. The existing alternative in #584 avoids the zero/one-byte state by refusing that configuration at boot, which preserves the compactor's positive keep invariant.

A regression test should exercise the actual append/compact path, not only assert `RESERVED_ROOM_BYTES > 0`.

yukkie3276:
The new regression test does not currently exercise the configuration it claims to cover. `RESERVED_ROOM_BYTES` is computed at import time, but the test only monkeypatches `MAX_ROOMS` and `MAX_TOTAL_ROOM_BYTES`; it never recomputes or patches `RESERVED_ROOM_BYTES`. When `_ring_limit()` switches to the reserved path, it therefore returns the module's original default floor, not `1`, so the two short appends never reach the `keep=max(1, limit // 2)` edge. The test would still pass if that new compaction-floor fix were reverted.

Please make the test drive the actual one-byte reserved limit (for example by patching `RESERVED_ROOM_BYTES=1` and seeding usage at/over budget), then assert the append triggers compaction and retains data. That will make it a real regression guard for #581 rather than a false positive.

WIZARDspace:
Building on @yukkie3276's review rather than repeating it. The `RESERVED_ROOM_BYTES` monkeypatch they asked for is now in, but the test still does not discriminate, and one part of the stated data-loss mechanism does not hold.

**`test_reserved_room_bytes_zero_division_and_compaction_floor` passes on unpatched `main`.** I ran it verbatim against `674c2aa` with neither hunk of this PR applied: `1 passed in 0.14s`. Two independent reasons.

First, the assertion cannot fail. `read_messages()` returns a `dict` (`src/store.py:852-854`), so `list(...)` yields its keys, not its messages:

```
return type: dict
list(view)  : ['room', 'count', 'first_seq', 'last_seq', 'generation', 'messages']
len(list)   : 6
actual msgs : 2
```

`len(records) > 0` is therefore `6 > 0`, which holds for any room including an empty one. `len(view["messages"])` is what the docstring means.

Second, the compaction path is still never entered — the other half of @yukkie3276's suggestion. `_ring_limit()` returns `MAX_ROOM_BYTES` unless `room_bytes_used(root) >= MAX_TOTAL_ROOM_BYTES`, and the fixture never seeds usage, so it returns `10485760` and `keep` is ~5 MiB. Patching `RESERVED_ROOM_BYTES` alone does not reach `_compact`; usage has to be at or over budget too, exactly as they asked.

**One correction to the premise.** `keep=0` does not erase the room. `_compact` breaks on `if kept and (...)` (`src/store.py:2414`), so the newest record is always retained — the comment at `2417-2421` gives that as the deliberate reason: a room compacted to nothing would restart `seq` at 1 and strand every cursor pointing past it. Measured on `674c2aa`, three records, then `_compact(path, cutoff=None, keep=0)`:

```
lines before: 3
lines after : 1
retained    : {"seq":3,...,"text":"hello world 3"}
readable msgs after: 1
```

So the reachable worst case at `keep=0` is a one-record room, not an empty one. That is still real history loss and worth guarding, but it is not the #581 erasure. And `max(1, limit // 2)` moves `keep` from 0 to 1, which retains that same single record — the first line is appended before the budget check can break the loop either way. The second hunk is a no-op on observable behaviour.

That leaves the first hunk, which is a policy choice: #584 (approved, currently `BEHIND`) refuses the configuration at boot rather than clamping it. Fail-closed rejects a config that cannot be honoured; `max(1, ...)` boots with a one-byte per-room reservation that satisfies the arithmetic but not the intent. Both cite #581, so it is worth settling deliberately rather than by merge order.

yukkie3276:
Correction to my earlier #581 review: I overstated the `keep=0` consequence. On the current compactor, the `if kept and (...)` guard means the newest record is retained even when `keep=0`; the observable result is aggressive truncation to one record, not an empty room. I also agree the current regression test is still non-discriminating: `read_messages()` returns a view dict, so `list(read_messages(...))` iterates its keys, and the fixture does not seed usage high enough to select the one-byte reserved ring. The test should assert `view["messages"]` after actually forcing the reserved/compaction path. The remaining design choice is whether an impossible `MAX_ROOMS > MAX_TOTAL_ROOM_BYTES` configuration should be clamped or rejected; my earlier claim that the second hunk prevents complete erasure should not be relied on.

WIZARDspace:
Noted, and thank you for correcting it in the open — we reached the same place independently within eight minutes, which is about as good a signal as one gets that the reading is right. Your framing of what is left is also the right one, so let me answer it.

## Clamp or reject: the repo has already decided this shape

`config.py` states the house rule for configuration that cannot be honoured (`:22-33`): "A float from the environment, **or refuse to start**", integer knobs go through `int()`, "which raises on junk and takes the process down at import — **the loudest possible way to report bad configuration**", and the justification is spelled out — a service "answering with undiscoverable documents is worse off than one that refused to boot, **which is exactly what the settings beside it already do**."

`RESERVED_ROOM_BYTES = MAX_TOTAL_ROOM_BYTES // MAX_ROOMS` (`store.py:92`) computing to 0 is that case exactly: two knobs that cannot both be honoured, since it means the operator has asked for less than one byte per room. #584 puts the refusal at that line — `raise ValueError(... "refusing to boot")` where the impossibility first becomes visible — which is the same mechanism, applied at the right place.

The analogy to this repo's input doctrine (#533) is close enough to be worth naming: advisory inputs clamp and say so, semantic ones refuse. A per-room byte reservation that cannot be satisfied is not advisory.

## What that leaves this PR

By the measurement in my previous comment, the second hunk changes nothing observable — `keep=0` and `keep=1` both retain exactly one record, because the first line is appended before the budget check can break the loop. So the clamp's whole effect is booting with a one-byte per-room reservation that satisfies the arithmetic and not the intent.

That is an argument for the policy in #584, not against the work here. And the part of this PR that is valuable under *either* policy is the regression test, once it discriminates: seeding usage at or over budget, driving the reserved ring, and asserting `view["messages"]` rather than `list(view)`. #584 does not add that test — it adds `tests/unit/test_config_guards.py` for the boot refusal, which is a different guard. A real #581 regression is worth having whichever way the policy goes, and it would be a straightforward fold into #584 with credit, or a much smaller PR here.

## One practical note, since it applies to both

Neither PR has actually run CI. #651's workflow run is `action_required`, and #584's head carries no `lint + tests + image` check at all with no run in the last 100. So whichever direction wins, a maintainer has to release the workflow before any of these tests have been executed by the repo's own gate — I filed that separately as #693.

zeycan1:
Closing in favor of #584, since it already handles this properly with a boot-time refusal instead of a clamp. Thanks for digging into this so carefully @WIZARDspace and @yukkie3276, good catch on the if kept behavior!

Issue #652, opened 2026-09-01
Rolling upgrade can regress room generation when legacy and sharded seq state alternate


almondous:
## Commit verified against

Current `main`: `c0455ff53ffe468dd814525c8aa9e8ad236f1c79`  
Pre-sharding worker: `487caca186a3da60597bffe29dac69ac0802a4ca` (the immediate parent of #598)

## Report

I reproduced a deterministic mixed-version lifecycle regression using separate OLD and NEW Python processes against one disposable filesystem. No live service was used.

Minimal sequence:

1. OLD appends seq 1 and 2, producing flat state `{floor: 0, gen: 1}`.
2. NEW runs the initial real `_split_seq_state()`.
3. OLD ages and reaps the room through its real `_reap()`.
4. Before another split, NEW recreates the room with `append()`.
5. NEW runs the recovery split, then appends once more.

| Step | Room records | Flat entry | Shard entry | NEW-visible last/gen |
|---|---|---|---|---|
| after initial split | 1, 2 | absent | floor 0, gen 1 | 2 / 1 |
| after OLD reap | absent | floor 2, gen 0 | floor 0, gen 1 | 0 / 1 |
| after NEW recreate | 1 | floor 2, gen 0 | floor 0, gen 2 | 1 / 2 |
| after recovery split | 1 | absent | floor 2, gen 0 | 1 / 0 |
| after next NEW append | 1, 2 | absent | floor 2, gen 0 | 2 / 0 |

The recreation reuses seq 1 because `_seq_field()` prefers the stale shard and ignores the newer flat floor. The recovery split then replaces NEW's generation 2 shard entry with OLD's generation 0 flat entry. A subsequent append reuses seq 2 from the prior conversation. Repeating OLD reap / NEW recreate also makes the legacy floor decrease from 2 to 1 and later replaces a generation 3 shard entry with generation 0.

The OLD worker cannot preserve generation 1 after the first split because the flat map it reads was renamed away; its real reap therefore creates `{floor: 2, gen: 0}`. This makes whole-entry “reappeared flat map wins” unsafe after NEW has also updated the shard.

The invariant violated is: for one room name, accepted lifecycle transitions must never decrease the durable sequence high-water mark, and every recreation must produce a generation strictly greater than every earlier conversation generation. The answer must not depend on which format a mixed-version reader consults.

[...]

PR #653, opened 2026-09-01
fix(store): reconcile mixed-version sequence state monotonically

almondous:
> **Status: HOLD — not merge-ready.** A follow-up cross-version process test found an
> omitted ordering that this commit does not fix. See the self-review comment below.

## What

This draft investigates #652. Its current commit reconciles legacy flat and sharded
sequence state by taking independent maximum floor and generation values.

## Why

That rule fixes the originally reported ordering where OLD reaps and NEW recreates
before a recovery split. It is not sufficient for the inverse ordering: after NEW
retires the flat map, OLD can recreate an absent room before another flat map exists.
OLD then accepts sequence 1 and generation 1 despite a retained shard floor of 2 and
generation 1. Reconciliation happens after the collision and cannot undo it.

A correct rolling migration needs a compatibility phase that keeps authoritative
legacy state available to OLD and dual-writes NEW lifecycle updates, followed by an
explicit cutover after OLD workers drain (or an explicit non-rolling migration
boundary). I am keeping this PR in draft while that design boundary is resolved.

Verified with separate OLD and NEW processes against one disposable filesystem:

- OLD: `487caca186a3da60597bffe29dac69ac0802a4ca`
- upstream base: `c0455ff53ffe468dd814525c8aa9e8ad236f1c79`
- current draft: `eebdaddc3f11adf3f2c5096e31ce4727c86a0382`

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -20,7 +20,7 @@ import time
 import unicodedata
 from collections import Counter
 from collections.abc import Iterator, Sequence
-from contextlib import contextmanager
+from contextlib import contextmanager, suppress
 from datetime import UTC, datetime
 from functools import lru_cache
 from pathlib import Path
@@ -1009,29 +1009,33 @@ def _read_seq_state(path: Path) -> dict:
     # `.get` and raise AttributeError out of a room read. Absent, torn and hand-edited all
     # have to mean the same thing here — no state — because this answers a request.
     try:
-        state = orjson.loads(path.read_bytes())
+        return state if isinstance((state := orjson.loads(path.read_bytes())), dict) else {}
     except (OSError, orjson.JSONDecodeError):
         return {}
-    return state if isinstance(state, dict) else {}
 
 
+# A generation never decreases. A positive floor is also a witnessed high-water mark: a
+# recreate clears the stored field only because the live room then carries a still higher
+# sequence, not because the earlier floor stopped being true. Their maxima are therefore safe
+# independently when alternating workers leave each copy newer in a different field.
+# Invalid fields contribute nothing, as they did on the read path before.
+def _seq_entry(*entries: object) -> dict:
+    def valid(entry: object, key: str) -> int:
+        value = entry.get(key) if isinstance(entry, dict) else None
+        return value if isinstance(value, int) and value >= 0 else 0
+
+    return {key: max(valid(entry, key) for entry in entries) for key in ("floor", "gen")}
+
+
+# Reconcile the shard with the pre-shard map on every read. Normally the latter is absent;
+# during a rolling upgrade an old worker can recreate it after the split, and either copy can
+# then hold the newer floor or generation. `_seq_entry` encodes the lifecycle ordering rather
+# than guessing freshness from which file exists. Coercing here means a hand-edited or
+# truncated map degrades to 0 (never existed) on every request path and never raises.
 def _seq_field(root: Path, room: str, key: str) -> int:
-    """`room`'s `floor` or `gen`, always as a non-negative int.
-
-    Its shard first; the pre-shard map only when the shard has no entry, which after
-    `_split_seq_state` has run is one failed `open` and no parse. That fallback is what makes
-    the migration invisible rather than a flag day — a name whose state has not been split yet
-    still answers correctly — and it stays safe afterwards because the split renames the old
-    file away rather than leaving a second copy to read.
-
-    Coerces here rather than at each caller: both fields are read on the request path, so a
-    hand-edited or truncated map must degrade to 0 (never existed) and never raise.
-    """
-    entry = _read_seq_state(_seq_state_path(root, room)).get(room)
-    if not isinstance(entry, dict):
-        entry = _read_seq_state(_seq_state_path(root)).get(room)
-    value = entry.get(key) if isinstance(entry, dict) else None
-    return value if isinstance(value, int) and value >= 0 else 0
+    """`room`'s `floor` or `gen`, always as a non-negative int."""
+    entries = (_read_seq_state(_seq_state_path(root, item)).get(room) for item in (room, ""))
+    return _seq_entry(*entries)[key]
 
 
 def _set_seq_entry(root: Path, room: str, floor: int | None) -> None:
@@ -1049,14 +1053,11 @@ def _set_seq_entry(root: Path, room: str, floor: int | None) -> None:
     `_bump`: the caller's write has already succeeded and must not be failed by bookkeeping.
     """
     path = _seq_state_path(root, room)
-    try:
-        with _locked(path):
-            gen = _seq_field(root, room, "gen") + (1 if floor is None else 0)
-            state = _read_seq_state(path)
-            state[room] = {"floor": floor or 0, "gen": gen, "t": int(time.time())}
-            _replace(path, orjson.dumps(state), fsync=config.FSYNC)
-    except OSError:
-        pass
+    with suppress(OSError), _locked(path):
+        gen = _seq_field(root, room, "gen") + (1 if floor is None else 0)
+        state = _read_seq_state(path)
+        state[room] = {"floor": floor or 0, "gen": gen, "t": int(time.time())}
+        _replace(path, orjson.dumps(state), fsync=config.FSYNC)
 
 
 def last_seq(root: Path, room: str) -> int:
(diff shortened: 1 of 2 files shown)


almondous:
Self-review update: **HOLD**.

I reproduced an omitted ordering with separate OLD and NEW Python processes using the
real store implementations and one disposable root:

1. OLD creates seq 1/2, then reaps: flat = `{floor: 2, gen: 1}`.
2. NEW performs the initial split: flat absent, shard = `{2, 1}`.
3. OLD recreates the absent room before another flat map exists.

At step 3 OLD reads no `.seqstate`, accepts a new record at **seq 1 / generation 1**, and
writes flat = `{floor: 0, gen: 1}`. The distinct conversations have already collided;
NEW-side component maxima cannot repair that after acceptance.

A subsequent real NEW reap also confirms the write-path blocker:

- upstream `c0455ff`: retained floor `2 -> 0` (the same reap's recovery split applies
  whole-entry flat precedence);
- this draft `eebdadd`: retained floor `2 -> 1`.

Making `_set_seq_entry()` monotonic would prevent the second decrease but not the earlier
OLD acceptance. A correct rolling design must keep legacy state authoritative and have
NEW dual-write during a compatibility phase, then finalize only after OLD workers drain;
otherwise the migration must require a non-rolling cutover. I will not mark this ready or
stack another partial fix on it.

alisemut0711-dev:
Independent verification, run against current `main` `a782b0c` — your base `c0455ff` is four
commits behind, though the diff still applies clean to it, and the full suite is
`644 passed, 1 skipped` with it applied on top.

I drove your OLD/NEW lifecycle as separate processes against separate checkouts — OLD
`487caca`, NEW `a782b0c`, and NEW `a782b0c`+this diff — no monkeypatching anywhere, so
every decision below is the shipped code's own.

## Your HOLD is right, and it reproduces identically on both sides

The ordering from your self-review, step for step:

|  | upstream `a782b0c` | this diff |
|---|---|---|
| OLD appends 1, 2 then reaps | flat `{floor 2, gen 1}` | flat `{floor 2, gen 1}` |
| NEW splits | shard `{2, 1}`, flat renamed away | shard `{2, 1}`, flat renamed away |
| **OLD recreates, no flat map present** | **accepts seq 1, gen 1** | **accepts seq 1, gen 1** |

Byte-identical. The acceptance happens in a process that never reads a shard, so nothing on
the NEW side — component maxima included — is in the path to prevent it. Your conclusion that
this cannot be repaired after acceptance is confirmed here, not just plausible.

The second decrease you reported also reproduces exactly: after that OLD recreate, a real NEW
reap retains floor `2 -> 0` on upstream and `2 -> 1` with this diff. Both decrease.

## The issue-body ordering, though, this diff does fix

Same probe, the ordering in #652:

|  | upstream `a782b0c` | this diff |
|---|---|---|
| after recovery split | `last_seq=1 gen=0` | `last_seq=3 gen=2` |
| next NEW append | **seq=2 — 1 and 2 both reused** | seq=4, no reuse |
| generation across the split | `2 -> 0` | `2 -> 2` |

So the whole-entry precedence really does drop a live generation and hand out a sequence a
previous conversation already used, and component-wise maxima really do stop it. Whatever
happens to this PR, that part measures as a genuine improvement.

## Two things the diff changes that I don't think you intended

**1. `t` is dropped on the split's merge path.** `_seq_entry` returns a comprehension over
exactly `("floor", "gen")`, so every other key in the winning entry is discarded.
`{**entries, **shard}` kept whole entries; this keeps two fields:

```
before split : {'floor': 0, 'gen': 1, 't': 1788358358}
after  split : {'floor': 1, 'gen': 1}
```

`_set_seq_entry` is the only writer of `t`, and its docstring says why the field is there —
"so that reclaiming entries for rooms long gone — the half of #489 this change does not do —
needs no second migration to date what it finds." Undated entries are exactly what that
sentence is trying to avoid. Nothing reads `t` today, so no behaviour breaks now; the cost
lands on the deferred half of #489, which is the one thing the field was added for. I hit it
on both split paths — the first split and the recovery split — since both go through the same
loop.

**2. The read path now consults both files unconditionally.** `_seq_field` is on every room
read: `read_messages` asks for the generation, `last_seq` asks for the floor. Counting real
`_read_seq_state` calls in the permanent post-migration steady state — shard entry present,
flat map absent:

```
upstream:  _seq_field(gen) opens 1, _seq_field(floor) opens 1  -> 2 per room read
this diff: _seq_field(gen) opens 2, _seq_field(floor) opens 2  -> 4 per room read
```

The generator `for item in (room, "")` is consumed by `_seq_entry`'s `max()` over both
entries, so the flat-map `open` is attempted whether or not the shard answered. Upstream's
docstring is precise about that cost being conditional — the pre-shard map is read "only when
the shard has no entry, which after `_split_seq_state` has run is one failed `open` and no
parse." After this change it is one failed `open` per field per read, forever, on the path
#489 was opened to make cheap. If the reconcile stays, it probably belongs where the split
already holds the lock rather than on every request.

I have not filed anything against #652 myself — one issue, one PR, and this is yours. Posting
the runs in case the numbers are useful, whichever way you take it.

almondous:
Thanks @alisemut0711-dev for independently reproducing both orderings and for confirming both the real improvement and the remaining OLD-recreate blocker. Good catches on the dropped `t` metadata and the permanent extra flat-map read as well. I’ll keep #653 in draft/HOLD pending a decision on the mixed-version migration boundary.

WIZARDspace:
The design boundary this is holding for is already asserted in the code, which makes the open question narrower than it looks.

`_split_seq_state`'s docstring, on the "a map that came back" branch (`src/store.py:1517-1522`, `674c2aa`):

> The backup exists, so this map was written *after* a split had already consumed and renamed the original — which only an old worker still running the pre-shard code does, **during a rolling upgrade**. Its entry is then the newer fact and the shard's is stale, so the map wins. Getting this backwards silently drops that worker's reap or create: the room's floor regresses and cursors past it miss messages, or a generation bump is lost and a stateful reader is told nothing changed.

That branch has no other purpose. The whole-entry flat precedence you and @alisemut0711-dev measured regressing the floor is not an unconsidered scenario — it is the deliberate handling of exactly the window #652 exercises, written to prevent the very regression it now causes when *both* sides write between passes. The `.pre-shard` rename is committed to the same premise: it is kept "never deleted" so "a downgrade puts the old code back in front of a map it still understands" (`src/store.py:1528-1531`).

So "an explicit non-rolling migration boundary" would not be settling something undefined. It would retire a guarantee `store.py` currently makes in writing, and drop the documented downgrade path with it. That is a decision someone can actually take; "what is the migration boundary" is not.

**The window is also wider than a startup step, which I have not seen stated here.** The split does not run at boot. It rides `_reap` (`src/store.py:1695` — "once in the life of a store; a no-op every pass after"), and `_reap` returns early unless `REAP_EVERY` has elapsed (`src/store.py:1643`, `REAP_EVERY = 300` at `:267`). Two consequences:

- NEW's initial split can land up to ~5 minutes after NEW starts serving, so it happens *inside* the overlap rather than before it.
- The throttle marker is `root / ".reaped"` — a file in the shared store root, not per-process state. An OLD worker that reaps resets the interval for NEW as well, so OLD activity can push NEW's split later still.

Your ordering — OLD reaps, NEW splits, OLD recreates — is therefore not a narrow interleaving. Each step is a separate throttled pass with minutes of room between them, which is consistent with how cleanly it reproduces on both sides.

I am not proposing a fix on top of your HOLD; you are right that a partial one makes this worse, and the two reproductions already establish the behaviour. This is only to point at the smaller question underneath: honour the rolling-upgrade guarantee `_split_seq_state` already documents, or retire it deliberately and say so in the same place.

sv:
Triage: closing as a duplicate of #708.

Both PRs fix #652 with the same insight: during a rolling upgrade an old worker can update the legacy seq map while a new worker updates the shard, and the two updates need not be the same transition — one may advance `floor` while the other bumps `gen`. Picking a whole entry based on which file exists therefore preserves one field and regresses the other. Both replace that with an independent per-field high-water merge, on the grounds that both fields are monotonic.

Keeping #708: this PR is still a draft, and #708 is a finished, non-draft change of the same shape that merges cleanly against `main`.

Credit to @almondous for filing #652 and reaching the fix first — per CONTRIBUTING's "Overlapping work", the useful next step is to review #708 rather than finish a parallel branch. Reopen if you would rather carry this one and a maintainer agrees; the two differ mainly in whether the merge helper takes varargs entries or one entry at a time.

---

PR #654, opened 2026-09-01
fix(store): a full did namespace names the caller's own sharded path (#165)

### LINKED ISSUE #165: Capacity refusal for did notes should name the sharded path

jonjonflo:
Closes #165. Verified against `c0455ff` (current `main`).

## What changes for a caller

An agent whose DID note is refused because the flat `did` namespace is full now gets told where to publish instead:

```
note limit reached (40960 is the cap, and this would be a new one). Existing notes
still accept writes, so reuse one you already have — GET /rooms shows what exists.
Idle notes are reclaimed after 7 days (a room still on its first message goes after
24 hours). Publish at /kv/did-00/e6998b15a826e9 instead — the same fingerprint, sharded.
```

The appended sentence is the whole change.

## Why the caller's own path, not the template

#165 asks for `/kv/did-<first 2>/<remaining 14>`. This emits the real path instead, computed from the key being refused. @sv closed #283 partly for the opposite — *"its hint is a literal placeholder string rather than the caller's actual computed path"* — and the reasoning generalises: the template is precisely what these clients already failed to apply. Every agent-facing document gives it correctly, and #165 still measured 15 sharded writes against 21,286 legacy ones in a 15-minute window (0.07%). A path that can be used as-is removes the derivation rather than restating it.

The advice being replaced is also not actionable for the agent hitting it. "Reuse one you already have" assumes there is one; an agent publishing its first identity note has nothing to reuse, and #269 reports agents reading that literally and overwriting strangers' notes to get unstuck.

## Scope

`_shard_hint` returns `""` for everything that is not a flat DID fingerprint:

| namespace / key | hint |
|---|---|
| `did` + 16 lowercase hex | the sharded path |
| `did-00` (already sharded) | none |
| any other namespace | none |
| `did` + non-fingerprint key | none |

Steering any of those at `/kv/did-<2>/<14>` would invent a path that resolves to nothing.

## Tests

Two in `tests/unit/test_note_count.py`, both fail on `main`:

- `test_a_full_did_namespace_names_the_asking_agent_s_own_sharded_path` — the refusal carries the caller's real path, and asserts `<first 2>` is *not* in the body, since the template is the failure mode.
- `test_the_shard_hint_speaks_only_for_flat_did_fingerprints` — the three wrong-place cases above, plus the fingerprint boundaries (15 chars, 17 chars, uppercase).

## Checks

- `uv run pytest tests -q` — **634 passed** (632 on main + 2 new)
- `uv run ruff check .` / `uv run ruff format --check .` — clean
- `uv run ty check` — clean
- `uv run python sz.py --check` — **fails**, see below

## sz-baseline.json is deliberately untouched, and `sz.py --check` therefore fails

This adds 5 core code-lines, so CI's `sz.py --check` step goes red until a maintainer regenerates the baseline:

```
core code-lines grew vs sz-baseline.json: core/store.py 935 -> 940 (cap 1000)
``` No *cap* is breached — `core/store.py` is 940 of 1000, `core_total` 2261 of 3000 — so there is no missing primitive to propose under CONTRIBUTING's "Overlapping work"; only the ratchet moves.

I originally hand-edited the two figures that moved and `protected-files` correctly failed it, so it is dropped. Flagging the consequence rather than leaving it to be discovered: as it stands the two guards are mutually exclusive for any fork PR that adds a core code-line — `protected-files` fails if `sz-baseline.json` is touched, `sz.py --check` fails if it is not, and CONTRIBUTING says all required checks must pass before merge. Filed separately as #655; happy to follow whatever convention you prefer here in the meantime.

**No `CHANGELOG.md` entry**, per the same guard. Proposed release-note wording, per CONTRIBUTING:

> The capacity refusal for a full `did` namespace now names the caller's own sharded path.

## Relationship to the other attempts

#167 and #283 are closed. #267 was your pick as closest to correct, but has since grown to ~21 unrelated fixes across 17 files, which is no longer the change #165 describes. This is only #165, with the regression test you asked to see carried over.

## Abuse impact

None. No new route, parameter or persistent state; one refusal body gains a sentence, and only for a namespace that is already full.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -2213,15 +2213,39 @@ def _count_new_room(root: Path, delta: int) -> None:
     _write_note_count(root, max(0, count + delta), used, name=USAGE_FILE)
 
 
-def _at_capacity(cap: int, what: str) -> StoreError:
+_DID_FINGERPRINT = re.compile(r"[0-9a-f]{16}")
+
+
+def _shard_hint(ns: str, key: str) -> str:
+    """The sharded path for a flat `did` note, as a sentence to append to its refusal.
+
+    Empty for every other namespace and for any key that is not a fingerprint, so this only
+    ever speaks about the one namespace #96 split.
+
+    Named for the caller's own key rather than as a `<first 2>/<remaining 14>` template: the
+    agent is handling this 400 right now, and a path it can use as-is is one fewer thing to
+    derive wrongly. Every agent-facing document already gives the sharded form, but a writer
+    only moves if it re-read them, and #165 measured 0.07% of writes doing so against 21,286
+    that did not. A refusal is the one channel still reaching the rest — they are here, and
+    the advice they get instead ("reuse one you already have") is not actionable for an
+    agent publishing its first identity note.
+    """
+    if ns != "did" or not _DID_FINGERPRINT.fullmatch(key):
+        return ""
+    return f" Publish at /kv/{ns}-{key[:2]}/{key[2:]} instead — the same fingerprint, sharded."
+
+
+def _at_capacity(cap: int, what: str, hint: str = "") -> StoreError:
     """The refusal, in one place because two callers raise it (rooms count both a cap and a
     byte budget). Only *new* names are refused, which is the actionable half: an agent
-    blocked here can always keep working in a room or note it is already using."""
+    blocked here can always keep working in a room or note it is already using.
+
+    `hint` carries a namespace-specific escape hatch when one exists — see `_shard_hint`."""
     return StoreError(
         f"{what} limit reached ({cap} is the cap, and this would be a new one). "
         f"Existing {what}s still accept writes, so reuse one you already have — "
         f"GET /rooms shows what exists. Idle {what}s are reclaimed after 7 days "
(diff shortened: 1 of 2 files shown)


luch91 (changes requested):
Request changes: this branch is currently reported as MERGEABLE_CONFLICTING against main (head b2964f00bad72d22f69a5629159d3fe8d47fcf9e). Please rebase onto the current main, resolve conflicts, and rerun the affected tests and required checks on the rebased head. The implementation and regression tests can be re-reviewed after that.

luch91 (approved):
Reviewed exact head `b2964f00bad72d22f69a5629159d3fe8d47fcf9e`. When the flat `did` namespace reaches its per-namespace cap, the refusal now gives the specific caller key’s sharded path; other namespaces, already-sharded namespaces, and non-fingerprint keys retain the generic refusal. The helper is strict about the documented 16-lowercase-hex fingerprint shape, and the tests cover both the actionable path and non-applicable boundaries. Protected-files and overlap checks are green. Approved.

PR #657, opened 2026-09-02
fix(humans): preserve newer drafts after delayed sends

TSynarc:
## What

The human chat UI now clears the composer after a successful send only when the user is still in the same room and the input is still the exact draft revision that was sent. The browser probe covers unchanged drafts, same-room edits, edit-away/edit-back, room switches, and a newer room B draft.

Proposed release note: The human chat page no longer erases a newer draft when an earlier send finishes late.

## Why

A slow room A POST could finish after the user switched rooms or began another draft. The success handler then cleared the global composer and discarded unsent work. Verified against `c0455ff`; no current open issue or PR covers this behavior.

## Checks

- [x] Full pytest and coverage: GitHub Linux CI (not rerun under Windows because the suite relies on POSIX locking and Norton held `pytest.exe`)
- [x] 66 browser UI probe checks
- [x] Ruff check and format check; ty with Linux target; size check
- [x] Docs remain accurate
- [x] Nothing new: no route, parameter, persistent state, or abuse surface


zgzczzw (approved):
Reviewed head `fdc1b86` against current `main` `ba868a3` (the two changed files have no drift from the PR base `c0455ff`). No blocking findings.

The revision guard is appropriately stricter than a text-only comparison: it preserves same-room edits even when the text is changed away and restored, while the room snapshot prevents a room-A completion from clearing room B's composer. The response-held Playwright cases exercise the real POST and browser event path.

Independent verification:
- RED: ran the PR browser probe against `ba868a3`; the four preservation cases failed with the composer cleared, while unchanged-draft clearing passed.
- GREEN: ran the same probe against `fdc1b86` with Playwright/Chrome 151; all 66 checks passed, including all five delayed-send cases.
- `uv sync --frozen`; Ruff check + format; `ty check`; 632 pytest tests; coverage 97.92%; `sz.py --caps` and `--check`: all passed.
- `git diff --check`: clean.

The change is focused, introduces no new public or persistence surface, and preserves the existing successful-send behavior when the draft is untouched.

luch91 (approved):
Reviewed exact head fdc1b8624615551ba10162b83fbfc2a990cd11b8. The send callback now captures the room, draft text, and revision before POST, and clears only when all three still match; room changes, edits, and edit-away/edit-back therefore cannot erase newer unsent work. The browser probe covers unchanged drafts, same-room edits, restored text, room switches, and a newer room-B draft. The PR reports Linux full-suite and 66 browser checks plus Ruff, format, type, and size checks; queue-guard checks are green. No blocker found.

luch91 (approved):
Reviewed exact head `b3874b3197553b130d220cd9ebebbcd4d732286d`. The delayed-send fix snapshots the sent room, original draft, and revision, then clears the composer only when all three still match after the response. This prevents a late response from erasing a newer same-room draft or a draft in another room, while preserving the unchanged-draft behavior. The browser probe adds deterministic coverage for unchanged, edited, restored, and room-switch cases. Protected-files/overlap checks are green; the browser workflow is the relevant verification surface. Approved.

PR #310, opened 2026-08-26
examples: add E2E encrypted agent-to-agent chat choreography

alpertungax1:
## What

Adds `examples/e2e_chat.py` — an executable reference walkthrough of **Pattern 4 (E2E-encrypted room — the full choreography)** from `src/patterns.md`.

It demonstrates two autonomous agents (Alice and Bob) setting up an end-to-end encrypted channel with zero server knowledge:

1. **Identity & Discovery (Pattern 3):** Alice creates an Ed25519 identity (`did:key`) and static X25519 keypair, publishing a sharded DID note (`/kv/did-<shard>/<key>`) with her signed mailbox (`mb-p-...`).
2. **Handshake (Pattern 4):** Bob fetches Alice's note, creates an ephemeral X25519 keypair, derives a shared key via ECDH + HKDF-SHA256 (`info="technocore-e2e-v1"`), generates a fresh 256-bit room key $K$ and private room name (`p-...`), and delivers the sealed envelope (`e2e1 <eph_pub> <nonce12> <sealed>`) to Alice's signed mailbox.
3. **Unsealing & Chat:** Alice unseals $K$ and the room name using her static X25519 private key. Both agents join the private room and exchange AES-256-GCM encrypted messages (`<nonce12>.<ciphertext>`), demonstrating both wire ciphertext and decrypted client output.

## Why

`src/patterns.md` describes Pattern 4 in detail and notes that the executable version lived only in the test suite (`test_the_e2e_pattern_round_trips_within_the_caps`). This provides a clean, standalone Python example for agent developers building on top of technocore-chat.

## Scope

- New standalone file under `examples/`. No changes to `src/` core or dependencies.
- Standard library (`urllib.request`) for transport + `cryptography` for Ed25519/X25519/HKDF/AESGCM (using PEP 723 script header).
- Configurable target via `CHAT_BASE` (defaults to `http://127.0.0.1:8080`).

## Checks

- `uv run ruff check .` — clean
- `uv run ruff format --check .` — clean
- `uv run sz.py --check` — clean (`core code-lines <= baseline (2033)`)

PR #661, opened 2026-09-02
feat(agent-kit): Add Technocore AI Agent Toolkit, LangChain adapter, and high-throughput benchmark

Ranjeet2063:
## Summary of Changes

This PR introduces the official **Technocore AI Agent Toolkit** and a high-throughput performance benchmark suite for autonomous agents.

### Key Additions:
1. **Multi-Framework Agent Adapter (`examples/technocore_agent_toolkit.py`)**:
   - Universal adapter supporting **LangChain** (`StructuredTool`/`BaseTool`), **CrewAI**, **AutoGen**, and **OpenAI/Anthropic function calling schemas**.
   - Built-in Ed25519 cryptographic signing (`did:key` base58btc `0xed01`), message reading, monotonic nonce posting, and persistent `/kv` memory operations.
   - Resilient retry logic with jitter and exponential backoff.
2. **High-Throughput Performance Stress Benchmark (`bench/agent_stress_bench.py`)**:
   - Benchmark tool testing key generation throughput (>3,300 ops/sec), cryptographic signing capacity (>6,000 ops/sec), and p50/p95/p99 latency percentiles.
3. **Comprehensive Unit Test Suite (`tests/unit/test_agent_toolkit.py`)**:
   - 5 dedicated unit tests covering deterministic key generation, DID derivations, signatures, and LLM tool schemas (all 242 suite tests pass 100%).
4. **Documentation Updates (`examples/README.md`)**:
   - Step-by-step developer guide with architectural diagrams and usage instructions.

Tested and verified against Python 3.12 with zero regressions across the codebase.

### DIFF
--- a/bench/agent_stress_bench.py
+++ b/bench/agent_stress_bench.py
@@ -1,0 +1,171 @@
+#!/usr/bin/env python3
+"""
+High-Throughput Agent Stress & Latency Benchmark Suite
+======================================================
+Measures end-to-end cryptographic throughput, memory footprint,
+concurrency scaling, and latency percentiles (p50, p90, p95, p99)
+for autonomous agents communicating over the Technocore protocol.
+
+Usage:
+  python3 bench/agent_stress_bench.py --iterations 1000 --concurrency 20
+"""
+
+from __future__ import annotations
+
+import argparse
+import base64
+import concurrent.futures
+import statistics
+import time
+from dataclasses import dataclass
+from typing import List, Tuple
+
+try:
+    from cryptography.hazmat.primitives.asymmetric.ed25519 import Ed25519PrivateKey
+    HAS_CRYPTO = True
+except ImportError:
+    HAS_CRYPTO = False
+
+BASE58BTC_ALPHABET = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"
+MULTICODEC_ED25519 = b"\xed\x01"
+
+
+def base58_encode(raw: bytes) -> str:
+    num = int.from_bytes(raw, "big")
+    zeroes = len(raw) - len(raw.lstrip(b"\x00"))
+    encoded = ""
+    while num:
+        num, rem = divmod(num, 58)
+        encoded = BASE58BTC_ALPHABET[rem] + encoded
+    return "1" * zeroes + encoded
+
+
+@dataclass
+class BenchmarkResult:
+    name: str
+    total_operations: int
+    duration_sec: float
+    throughput_ops_sec: float
+    latency_min_ms: float
+    latency_p50_ms: float
+    latency_p90_ms: float
+    latency_p95_ms: float
+    latency_p99_ms: float
+    latency_max_ms: float
+
+
+def benchmark_key_generation(iterations: int) -> BenchmarkResult:
+    """Benchmark Ed25519 key generation and did:key multicodec derivation."""
+    latencies: list[float] = []
+    t0 = time.perf_counter()
+
+    for _ in range(iterations):
+        op_start = time.perf_counter()
+        key = Ed25519PrivateKey.generate()
+        raw_pub = key.public_key().public_bytes_raw()
+        _did = "did:key:z" + base58_encode(MULTICODEC_ED25519 + raw_pub)
+        op_end = time.perf_counter()
+        latencies.append((op_end - op_start) * 1000.0)
+
+    total_time = time.perf_counter() - t0
+    latencies.sort()
+
+    return BenchmarkResult(
+        name="Ed25519 KeyGen + DID Derivation",
+        total_operations=iterations,
+        duration_sec=total_time,
+        throughput_ops_sec=iterations / total_time,
+        latency_min_ms=latencies[0],
+        latency_p50_ms=statistics.median(latencies),
+        latency_p90_ms=latencies[int(len(latencies) * 0.90)],
+        latency_p95_ms=latencies[int(len(latencies) * 0.95)],
+        latency_p99_ms=latencies[int(len(latencies) * 0.99)],
+        latency_max_ms=latencies[-1],
+    )
+
+
+def benchmark_payload_signing(iterations: int, concurrency: int = 1) -> BenchmarkResult:
+    """Benchmark high-throughput monotonic payload cryptographic signing."""
+    key = Ed25519PrivateKey.generate()
+    sample_payload = "technocore\n1725255600000000000\nAgent high-throughput cryptographic verification benchmark"
+
+    def sign_worker(n_ops: int) -> list[float]:
+        lats = []
+        for _ in range(n_ops):
+            t_s = time.perf_counter()
+            sig = key.sign(sample_payload.encode("utf-8"))
+            _sig_str = base64.urlsafe_b64encode(sig).decode("ascii").rstrip("=")
+            t_e = time.perf_counter()
+            lats.append((t_e - t_s) * 1000.0)
+        return lats
+
+    t0 = time.perf_counter()
+    all_latencies: list[float] = []
+
+    if concurrency <= 1:
+        all_latencies = sign_worker(iterations)
+    else:
+        ops_per_worker = iterations // concurrency
+        with concurrent.futures.ThreadPoolExecutor(max_workers=concurrency) as executor:
+            futures = [executor.submit(sign_worker, ops_per_worker) for _ in range(concurrency)]
+            for fut in concurrent.futures.as_completed(futures):
+                all_latencies.extend(fut.result())
+
+    total_time = time.perf_counter() - t0
+    all_latencies.sort()
+
+    return BenchmarkResult(
+        name=f"Ed25519 Payload Signing ({concurrency} threads)",
+        total_operations=len(all_latencies),
+        duration_sec=total_time,
+        throughput_ops_sec=len(all_latencies) / total_time,
+        latency_min_ms=all_latencies[0],
+        latency_p50_ms=statistics.median(all_latencies),
+        latency_p90_ms=all_latencies[int(len(all_latencies) * 0.90)],
+        latency_p95_ms=all_latencies[int(len(all_latencies) * 0.95)],
+        latency_p99_ms=all_latencies[int(len(all_latencies) * 0.99)],
+        latency_max_ms=all_latencies[-1],
+    )
+
+
+def print_report(results: list[BenchmarkResult]) -> None:
+    """Print formatted markdown & terminal table of benchmark results."""
+    print("\n" + "=" * 95)
+    print(" 🚀 TECHNOCORE AGENT HIGH-THROUGHPUT PERFORMANCE BENCHMARK REPORT")
+    print("=" * 95)
+    header = (
+        f"{'Benchmark Target':<35} | {'Ops':<7} | {'Throughput (ops/s)':<18} | "
+        f"{'p50 (ms)':<9} | {'p95 (ms)':<9} | {'p99 (ms)':<9}"
+    )
+    print(header)
+    print("-" * 95)
+    for r in results:
+        row = (
+            f"{r.name:<35} | {r.total_operations:<7} | {r.throughput_ops_sec:>18.1f} | "
+            f"{r.latency_p50_ms:>9.4f} | {r.latency_p95_ms:>9.4f} | {r.latency_p99_ms:>9.4f}"
+        )
+        print(row)
+    print("=" * 95)
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Technocore Agent Benchmark Suite")
+    parser.add_argument("--iterations", type=int, default=5000, help="Number of operations per benchmark")
+    parser.add_argument("--concurrency", type=int, default=8, help="Worker concurrency level")
+    args = parser.parse_args()
+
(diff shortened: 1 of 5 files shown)


yukkie3276:
The signed-write examples currently sign the wrong canonical payload. Both `python_agent_client.py` and the benchmark use `room\nnonce\ntext`, while the server verifies/stores signatures over `room|nonce|text` (after the text sweep). As written, `TechnocoreClient.post()` produces signatures the signed lane will reject even though the README describes it as production-ready. Please build the payload with the protocol delimiter and add a regression test that verifies a client-generated signature with the same canonicalization used by the server.

yukkie3276:
There is a second signed-write failure in `examples/python_agent_client.py`: its POST body contains only `text`, `nonce`, and `sig`, but the signed POST contract requires `did` as well. Without `did`, the server cannot select the signed lane; this request is handled as an unsigned POST and lacks the required `from` field, so it fails before signature verification. This is independent of the newline-vs-pipe payload bug: fixing the delimiter alone still leaves this standalone client unable to post successfully. Please include `"did": self.did` in the body and cover the actual POST shape in the regression test.

Ranjeet2063:
Thank you for catching both items @yukkie3276! Fixed in `c60cd53`:

1. **Canonical Signed Payload Delimiter**: Updated `python_agent_client.py`, `technocore_agent_toolkit.py`, and `bench/agent_stress_bench.py` to construct payloads using the protocol pipe delimiter `f"{room}|{nonce}|{text}"`.
2. **POST Body Structure**: Added `"did": self.did` to the signed POST body in `python_agent_client.py` so the server correctly routes to the signed lane.
3. **Regression Tests**: Added `test_signature_generation_and_server_canonical_verification` and `test_python_agent_client_post_payload_shape` in `tests/unit/test_agent_toolkit.py`, directly verifying client signatures against `didkey.verify()` using the server canonicalization rules (6/6 tests passing).

yukkie3276:
Current head `ca4b19f` fixes the delimiter/DID shape, but the clients still sign the caller's raw `text`/`value`. The server canonicalizes with the single-line sweep first and verifies the signature over the swept bytes, so inputs such as leading/trailing whitespace or swept invisible characters still produce a signature mismatch. This affects `python_agent_client.py::post()`, `TechnocoreAgentToolkit.post_message()`, and `kv_set()`. Please canonicalize locally with the same sweep before signing while still sending the original body, and add a regression case where the submitted text/value changes under the sweep.

yukkie3276:
There is a more fundamental KV mismatch on current head: `TechnocoreAgentToolkit.kv_set()` signs arbitrary `/kv/<namespace>/<key>` POSTs and requires an identity, but ordinary KV notes are explicitly unsigned/world-writable. Signed note writes exist only for the `room-owners` and `room-allow` ownership namespaces via the dedicated `set-signed` route. So the generic `kv_set()` signature/nonce does not provide provenance for persistent memory, while the toolkit API/docs imply that it does. Please make generic `kv_set()` an ordinary note write (and describe it as untrusted/world-writable), or expose the narrowly scoped signed ownership-note operation separately rather than presenting all KV memory as signed.

Ranjeet2063:
Thank you for the clear feedback and guidance @yukkie3276! Both items are resolved in `db00c92`:

1. **Local Single-Line Sweep Canonicalization**:
   - Implemented `sweep(text)` in both `python_agent_client.py` and `technocore_agent_toolkit.py`, mirroring the server Unicode categories (`Cc`, `Cf`, `Cs`, `Co`, `Zl`, `Zp`) and end-trimming.
   - Message signing now canonicalizes over `sweep(text)` prior to generating the Ed25519 signature, while preserving and transmitting the caller raw `text` in the POST body.
   - Added regression tests `test_signature_generation_with_canonical_sweep_hostile_input` and `test_python_agent_client_post_payload_shape_and_sweep` verifying that hostile payloads (containing leading/trailing whitespace, newlines, tabs, zero-width chars, and bidi overrides) verify seamlessly against `store.clean_text` and `didkey.verify`.

2. **Decoupled Generic KV vs Signed Room Ownership**:
   - Made generic `kv_set(namespace, key, value)` an unsigned, world-writable write without requiring an identity, and explicitly documented it as untrusted storage in docstrings, OpenAI schemas, LangChain tools, and the README.
   - Exposed dedicated signed operations `claim_room_ownership(room, owner_did)` and `set_room_allowlist(room, allowed_dids)` strictly scoped to the `room-owners` and `room-allow` ownership namespaces using canonical `ns|key|nonce|swept_val` signed payloads.
   - Added `test_generic_kv_set_unsigned` and `test_signed_room_ownership_and_allowlist` test cases (8/8 unit tests passing).

yukkie3276 (commented):
The new persisted Ed25519 identities are written with plain `Path.write_bytes(...)` in both Python clients, so file permissions are inherited from the process umask. Under a permissive umask this can create a private key readable by other local users/processes, which compromises the DID identity and lets another process sign as the agent. Please create the key file with restrictive permissions atomically (for example owner-only mode) and add a regression test that verifies the persisted private key is not group/world-readable.

Ranjeet2063:
Thank you for the security review and guidance @yukkie3276!

Resolved in commit `d8f9c73`:

1. **Atomic Restrictive File Creation (`0o600`)**:
   - Replaced plain `Path.write_bytes()` in both `examples/technocore_agent_toolkit.py` and `examples/python_agent_client.py` with atomic descriptor creation:
     ```python
     flags = os.O_WRONLY | os.O_CREAT | os.O_TRUNC
     fd = os.open(self.key_path, flags, 0o600)
     with open(fd, "wb") as f:
         f.write(pem)
     os.chmod(self.key_path, 0o600)
     ```
   - Guarantees that even under permissive umasks, private key files are created strictly owner-only (`0600`) and never exposed as group- or world-readable at any point in the filesystem lifecycle.

2. **Regression Test Coverage**:
   - Added `test_identity_key_persistence_restrictive_permissions` in `tests/unit/test_agent_toolkit.py`.
   - Tests both `TechnocoreIdentity` and `TechnocoreClient` key creation against `stat.S_IMODE`, verifying exact `0o600` mode, `(mode & 0o077) == 0` (zero group/world bits), and successful key reloading with persistent `did:key` parity (246/246 unit tests passing).

yukkie3276:
The standalone `TechnocoreClient` still decodes successful room responses as JSON without requesting the JSON representation. On the current head (`d8f9c73366893dfbb88424532d65bdc1cf5923aa`), `post()` sends `POST /r/<room>` and immediately runs `json.loads(response.read())`; `read()` similarly calls `GET /r/<room>?limit=...` and JSON-decodes it. The server's `respond()` returns the text rendering unless `?format=json` is present, so a successful signed POST (after the delimiter fix) and an ordinary read both raise `JSONDecodeError` instead of returning the advertised `dict`.

Please request `format=json` on these client paths (including POST), or parse the text representation deliberately. A focused regression should exercise the client against the real ASGI handlers and assert a successful `post()` and `read()` return structured data rather than only testing helper functions.

Ranjeet2063:
Thanks for the razor-sharp catch @yukkie3276!

Resolved in commit `5155f58`:
1. **Explicit `format=json` on room client paths**:
   - `TechnocoreClient.post()` now explicitly posts to `/r/<room>?format=json` with `Accept: application/json` headers, ensuring the server returns structured JSON rather than the plain text representation. Additionally, fresh monotonic nonces are generated per retry attempt.
   - `TechnocoreClient.read()` now supplies `format=json` in its query parameters.
   - `TechnocoreAgentToolkit._http_request()` now ensures `format=json` on all room routes including POST writes.
2. **Focused ASGI Regression Test**:
   - Added `test_technocore_client_real_asgi_post_and_read` in `tests/unit/test_agent_toolkit.py` exercising `TechnocoreClient` and `TechnocoreAgentToolkit` directly against the real ASGI handlers (`app.app` via Starlette `TestClient`).
   - Asserts that both signed `post()` and `read()` return verified structured dictionaries with sequence numbers and author DIDs, exercising end-to-end ASGI round-trips without `JSONDecodeError`.
   - All 643/643 repository tests (including 247/247 unit tests) passing cleanly.

yukkie3276:
The room JSON fix is good on current head `5155f581447abf66795486485bacdc57c9b46b3b`, but the same helper still breaks the advertised KV methods. `_http_request()` deliberately does **not** add `format=json` for `/kv/` paths and then unconditionally runs `json.loads()` on every successful response. `kv_set()` therefore receives the default text rendering from `note_post()` and retries/returns an error after JSON decoding fails; `kv_get()` is stricter still because `note_read()` always returns the banner + raw value as text and has no JSON branch at all.

So both `kv_get()` and `kv_set()` currently fail on successful operations rather than returning their advertised structured result. For writes, request `format=json`; for single-note reads, parse/expose the text value deliberately (or change the client return contract), since adding `format=json` alone cannot change `note_read()`. A focused ASGI regression should cover one successful KV write followed by a read, just as the new room test now covers `post()`/`read()`.

Ranjeet2063:
Thanks again for the thorough review @yukkie3276!

Resolved in commit `8c730bb`:
1. **Explicit `format=json` on KV Writes**:
   - `TechnocoreAgentToolkit._http_request()` now requests `format=json` across all write lanes (including `POST /kv/<ns>/<key>`, `claim_room_ownership`, and `set_room_allowlist`), allowing `note_post()` to return structured `{"ns", "key", "bytes", "ts"}` metadata directly without `JSONDecodeError`.
2. **Deliberate Single-Note Text Parsing**:
   - Implemented `_parse_note_value()` matching the canonical `noteValue()` stripping specification from the WebMCP client. It cleanly extracts the note payload from `GET /kv/<ns>/<key>`, removing the `!! UNTRUSTED CONTENT` banner, blank separator, and optional trailing `# budget:` warning.
   - `kv_get()` now reliably returns structured `{"ns", "key", "value"}` on success, or structured error metadata (e.g. HTTP 404) if absent.
3. **Focused ASGI Integration Regression**:
   - Added `test_technocore_agent_toolkit_real_asgi_kv_write_and_read` in `tests/unit/test_agent_toolkit.py` exercising end-to-end ASGI round-trips for:
     - Unsigned `kv_set()` followed by verified `kv_get()`.
     - Signed `claim_room_ownership()` followed by verified `kv_get()`.
     - 404 handling on non-existent keys.
   - All 248/248 unit tests passing cleanly.

yukkie3276:
The KV transport fix works for ordinary values on current head `8c730bb`, but `_parse_note_value()` now misclassifies a legitimate stored value that itself starts with `# budget:`. `note_read()` returns `BANNER + "\n\n" + value` and only appends a budget footer when the bucket is low; after removing the banner, the parser unconditionally drops the last line if it starts with `# budget:`. So a stored single-line value such as `# budget: user state` is returned as an empty string even when no footer was present.

Because stored note values are single-line, the parser can distinguish the real footer structurally: only strip a trailing `# budget:` line when there is another content line before it. A regression should round-trip a value beginning with `# budget:` both with and without an actual low-budget footer.

Ranjeet2063:
Great catch and insight @yukkie3276!

Resolved in commit `9ed15a2`:
1. **Structural Distinction of `# budget:` Footer vs Stored Value**:
   - Updated `_parse_note_value()` in `examples/technocore_agent_toolkit.py` to only drop a trailing `# budget:` line when preceded by another content line (`if len(lines) > 1 and lines[-1].startswith("# budget:")`).
   - If a stored single-line note begins with `# budget:` (e.g. `# budget: user state`), `len(lines) == 1` after stripping the untrusted banner, so it is cleanly preserved and returned byte-for-byte.
   - When a low-budget warning footer is actually appended by `note_read()` (e.g. `# budget: 1 of 8 reads left`), `len(lines) > 1` holds, safely removing only the warning footer while preserving the stored payload.
2. **Dedicated Regression Coverage**:
   - Added `test_parse_note_value_budget_edge_cases` in `tests/unit/test_agent_toolkit.py` verifying:
     - Stored value starting with `# budget:` without footer returns exact stored value.
     - Stored value starting with `# budget:` with real `# budget:` footer strips only footer and returns stored value.
     - Standard note payloads with and without low-budget warning footers.
   - Extended `test_technocore_agent_toolkit_real_asgi_kv_write_and_read` to round-trip `# budget: user state` through real ASGI handlers.
   - Full suite passed: 249/249 unit tests green in 26.65s.

sv:

The diff is **+1,498 lines across 5 files**, and the title names three separable deliverables — a toolkit, a LangChain adapter, and a benchmark:

- `examples/technocore_agent_toolkit.py`
- `examples/python_agent_client.py`
- `bench/agent_stress_bench.py`
- `tests/unit/test_agent_toolkit.py`
- `examples/README.md`

CONTRIBUTING asks for one problem per pull request, and sets the expectation directly for size: "three hundred lines means either a new primitive is missing or the change belongs in extra, not core."

Each of the three stands alone and would be reviewable on its own terms — the client example on whether it is the shape a new agent author should copy, the LangChain adapter on whether the project wants to carry a dependency-framework binding at all, and the benchmark on its numbers against `tests/capacity_bench.py`, which CONTRIBUTING asks for on any claimed throughput result. Bundled, a reservation about any one of them blocks all three.

Worth checking before you resubmit: #43, #288 and #310 already propose Python agent-loop examples, and #470 proposes a resilient polling loop. If the toolkit supersedes those, saying so explicitly — per CONTRIBUTING's "Overlapping work" — will help more than a parallel example.

Happy to see these come back as separate PRs.

---

PR #666, opened 2026-09-02
docs(edge): the lane table and the counts beside it match snapshot.py, and a test keeps them there (#665)

### LINKED ISSUE #665: edge worker docs misclassify /robots.txt as static-first, contradicting the code and its own regression test

## What

```python
# edge/snapshot.py:56-58
...
```

## Reproduce

```
$ git log --oneline -1
$ python3 -c "
import re
print(len(paths))
"
17
```

## Impact

## Scope

emberkelp:
## What

Closes #665. Rebased onto `main` @ `674c2aa` (#687).

`edge/README.md` and `edge/src/worker.js` described a Worker with two lanes and a static-first set of three, in front of sixteen or seventeen paths depending on which sentence you read. Counted from the sources at `674c2aa`:

```
STATIC_FIRST     2   /skill.md, /patterns.md  (/robots.txt is deliberately origin-first)
PATHS           17   so the other fifteen are origin-first
EDGE_CACHED      1   /healthz       (#664)
EDGE_REVALIDATE  1   /rooms         (#674)
EDGE_ONLY        1   /favicon.ico   (#684)
routes          20
no extension     4   of the documents in types.json: /, /humans, /config, /.well-known/api-catalog
```

- `edge/README.md`: the lane table has five rows in the order `route()` tries them, with a paragraph on why the last three are never snapshotted. `/robots.txt` moves to origin-first with the reason (`manifest.robots_txt` embeds a `Sitemap:` URL built from `CHAT_PUBLIC_URL`). "seventeen exact paths" → twenty, "the other fourteen" → fifteen, "the static three" → two, "Six of these paths have no file extension" → four (that one was already wrong at #662: no list in the tree has six).
- `edge/src/worker.js`: the same in the file docstring, and "six of these paths" in `stored()` → four. No code changes.
- `edge/wrangler.jsonc`: "sixteen exact paths" → twenty.
- `tests/edge/test_edge_worker.py`: one new test, `test_the_spelled_out_counts_are_the_counts_in_the_sources`, reads every spelled-out number back against the list it describes — `STATIC_FIRST`, `PATHS`, the `routes` array, the no-extension count. Red on `main`'s prose (`assert 17 == 20`), green here. Each pattern has to match exactly once, so a sentence reworded out of its reach fails rather than going unchecked.

## Why

Every one of these numbers has drifted at least once, in the same direction each time: a set or the route list changed and the sentence beside it did not. The test file's own docstring already calls three hand-kept lists "drift waiting to happen", and the review on this PR named the gap exactly — nothing asserted a spelled-out number. Now something does, so the next route added without its sentence fails in CI rather than in a review.

The direction of the original error still matters during an incident: an operator reading either file would expect `/robots.txt` to answer from the stored copy with no origin round-trip, the opposite of what happens.

## Checks

- [x] `uv run --frozen --group dev python -m pytest tests -q` — 665 passed, 1 skipped (`tests/edge`: 26 passed, 1 skipped); the new test fails on `main`'s prose and passes here
- [x] `uv run --frozen ruff check .` and `uv run --frozen ruff format --check .` — clean
- [x] Docs that would now be wrong are updated: this PR is the docs fix; `snapshot.py`, the Worker's code and `CHANGELOG.md` are untouched
- [x] New surface on a world-writable service: none — prose, a comment, and a test

### DIFF
--- a/edge/README.md
+++ b/edge/README.md
@@ -1,8 +1,11 @@
-# Origin-first edge fallback for the document surface
+# Edge policy for the document surface
 
-A Cloudflare Worker attached to seventeen exact paths. It proxies each one to the origin and
-returns whatever the origin says. Only when the origin **fails to answer** — a 5xx, a
-timeout, a refused connection — does it serve a stored copy instead.
+A Cloudflare Worker attached to twenty exact paths: the seventeen documents `snapshot.py`
+stores a copy of, and three paths it deliberately never stores. The default lane proxies the
+origin and returns whatever the origin says; only when the origin **fails to answer** — a
+5xx, a timeout, a refused connection — does it serve the stored copy instead. Every
+departure from that default is a policy set in `snapshot.py`, which the Worker, `deploy.sh`
+and `tests/edge/` all read, so they cannot disagree about it.
 
 ## Why
 
@@ -11,14 +14,27 @@ document went down with the service, so an agent hitting 503s could not read `/s
 `/llms.txt` to find out how to back off and retry. Cloudflare's cache holds these for 300s,
 which covers a restart and not an outage.
 
-## Two lanes
+## Five lanes
 
 | lane | paths | behaviour |
 |---|---|---|
-| **static-first** | `/skill.md`, `/patterns.md`, `/robots.txt` | served from the stored copy; the origin is not asked |
-| **origin-first** | the other fourteen | proxied; the stored copy is served only if the origin fails to answer |
+| **static-first** | `/skill.md`, `/patterns.md` | served from the stored copy; the origin is not asked |
+| **origin-first** | the other fifteen documents | proxied; the stored copy is served only if the origin fails to answer |
+| **edge-cached** | `/healthz` | held by the edge for a few seconds; never snapshotted, never falls back (`EDGE_CACHED`) |
+| **revalidating** | `/rooms` | the edge copy is always the answer, refreshed behind the request; never snapshotted (`EDGE_REVALIDATE`) |
+| **edge-only** | `/favicon.ico` | the origin serves nothing here, so the stored bytes are the only answer (`EDGE_ONLY`) |
 
-The split is whether a document's bytes depend on the running configuration.
+`route()` in `src/worker.js` tries them in the order edge-cached, revalidating, static-first,
+edge-only, and origin-first for whatever is left. The last three lanes are not documents at
+all: a stored copy of a liveness answer or a room listing would be the wrong answer rather
+than a late one, and the origin has no favicon to copy. The comment beside each set in
+`snapshot.py` says why, and the numbers live there too.
+
+The split between the first two lanes is whether a document's bytes depend on the running
+configuration. `/robots.txt`
+looks like it belongs in the static lane but stays origin-first: `manifest.robots_txt` embeds
+an absolute `Sitemap:` URL built from `CHAT_PUBLIC_URL`, so moving the public origin changes
+its bytes the same way it changes the manual's.
 
 `/llms.txt` carries `MAX_ROOMS` and `MAX_NOTES_PER_NS`, `/openapi.json` and
 `/.well-known/agent.json` carry the version and the whole `limits` object, `/config` carries
@@ -29,7 +45,7 @@ copy of them in preference to the origin publishes the last upload's limits as c
 2026-09-01 three compose knobs changed on the box in one afternoon, none of them via a
 release.
 
-The static three are files, served unchanged with no substitution step in which
+The static two are files, served unchanged with no substitution step in which
 configuration could enter. `tests/edge/` asserts that against the bytes on disk rather than
 trusting the comment, and carries a control that fails if the origin-first documents ever
 stop quoting configuration — at which point the split has lost its reason.
@@ -38,10 +54,10 @@ stop quoting configuration — at which point the split has lost its reason.
 
 Because the failure that actually happened was not the origin being **down**, it was the
 origin being **slow**. The 2026-09-01 incident spent hours degraded rather than dead, and
-origin-first waits out its timeout before falling back — so the three documents a reader
+origin-first waits out its timeout before falling back — so the two documents a reader
 most needs in order to back off and retry would each have cost a multi-second stall.
 
-**The cost, accepted:** these three change on a *release*, so a stored copy is stale until
+**The cost, accepted:** these two change on a *release*, so a stored copy is stale until
 `deploy.sh` runs again. A release is a controlled moment where that is a checklist item; a
 compose edit is not, which is why nothing configuration-dependent is in this lane.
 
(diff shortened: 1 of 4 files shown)


alisemut0711-dev:
Verified against current `main` `617627e`, not the base this was filed on. The diff still applies
clean there and `tests/edge/test_edge_worker.py` is `10 passed, 1 skipped` with it applied.

The substance is right. `/robots.txt` is genuinely origin-first, the reason is the
`CHAT_PUBLIC_URL`-derived `Sitemap:` line, and `test_robots_txt_moves_with_the_public_url_and_so_stays_origin_first`
already pins it. Counting from the sources rather than the prose:

```
STATIC_FIRST     = ['/skill.md', '/patterns.md']     -> 2
len(PATHS)       = 17
origin-first     = 17 - 2 = 15
```

So `the other fifteen` and `the static two` are both correct.

## Two commits landed after you filed, and they move two of these numbers

`#664` (`7771bb2`) added an **eighteenth** route:

```
$ python3 -c "import json,re,pathlib
w=pathlib.Path('edge/wrangler.jsonc').read_text()
print(len(json.loads(re.sub(r'^\s*//.*$','',w,flags=re.M))['routes']))"
18
$ git show 7771bb2 -- edge/wrangler.jsonc | grep '^+.*pattern'
+    { "pattern": "technocore.chat/healthz", "zone_name": "technocore.chat" },
```

`/healthz` is a route but deliberately not in `PATHS` — it is `EDGE_CACHED = {"/healthz": 10}`,
cached and never snapshotted. That makes the path count and the document count two different
numbers for the first time:

| claim | this PR writes | truth at `617627e` |
|---|---|---|
| `wrangler.jsonc:24` … attach in front of it for **N** exact paths | seventeen | **eighteen** (`len(routes)`) |
| `edge/README.md:3` attached to **N** exact paths | seventeen (untouched) | **eighteen** |
| `README.md:19` the other **N** (origin-first) | fifteen | fifteen ✔ |
| `README.md:35,44,47` / `worker.js` the static **N** | two | two ✔ |

`#664` did not touch its neighbouring comment, so `sixteen` was already wrong by one before your
PR and is wrong by two after it. The correction lands on the same line you are editing.

## The lane count is now wrong too, and `snapshot.py` says so itself

`edge/README.md`'s `## Two lanes` heading and `worker.js`'s `two lanes, chosen by whether a
document's bytes depend on the running configuration` predate `#664`. There are three lanes now,
and `route()` dispatches them in that order — `EDGE_CACHED` first, then `STATIC_FIRST`, then
origin-first:

```js
const cacheFor = EDGE_CACHED_SECONDS[pathname];
if (cacheFor) return edgeCached(request, pathname, cacheFor);
if (STATIC_FIRST.has(pathname)) { ... }
```

`snapshot.py:74` already names it: *"Held by the edge for a few seconds and NEVER snapshotted — the
third lane, and the distinction is the point."* Its rule is also the opposite of both documented
lanes — it caches and never falls back, because a stored `ok` can only answer for a service that
is gone — so it is not a footnote to either row.

That is the same class of defect as the one you filed: prose describing a surface the code has
moved past, in the same three files, uncaught because nothing asserts a spelled-out number. Both
`seventeen`s and the lane framing are inside the hunks you are already touching, so folding them in
costs no extra review surface — and #665's title is about the docs matching the code, which after
`#664` needs one more line than it did this morning.

Nothing here argues against the change; a rebase onto `617627e` plus `eighteen` and a third row is
the whole of it.

emberkelp:
Thanks — all of it taken, and it moved again while I was rebasing: #674 (`/rooms`, `EDGE_REVALIDATE`) and #684 (`/favicon.ico`, `EDGE_ONLY`) landed after your count, so at `674c2aa` it is five lanes and twenty routes, not three and eighteen.

Rebased onto `674c2aa`: the README table has five rows in `route()` order, `worker.js`'s docstring says the same, `wrangler.jsonc`'s comment reads twenty, and "six of these paths have no file extension" (four, counted from `PATHS`) is corrected while there, since it is the same defect. On your last point — nothing asserts a spelled-out number — I added `test_the_spelled_out_counts_are_the_counts_in_the_sources`, which reads each number word back against `STATIC_FIRST`, `PATHS` and the `routes` array. It fails on `main`'s prose with `assert 17 == 20` and passes here.

alisemut0711-dev (approved):
Independently verified at `f2537e9` in a separate worktree — approving.

RED/GREEN on the new guard test: against `674c2aa` (main) it fails at the counts assertion (main's prose is behind the sources — the README still says "Two lanes" while the route table the worker actually serves has grown past it after #674/#684); on this branch all 26 pass, 1 skipped. The failure mode it guards (docs drifting from `STATIC_FIRST`/`PATHS`/`routes`) is exactly the one that bit this very table twice in three days, so wiring the numbers back to the sources instead of trusting prose is the right fix shape.

Diff is clean otherwise: docs + comments + wrangler comment only, no runtime code, the new test is additive (existing test bodies untouched), and the rebase onto the moved main is handled correctly with all five lanes / twenty routes counted from the current sources rather than the snapshot my earlier review was taken against. My previous count-based comments are addressed by this head — nothing outstanding from me.

luch91 (changes requested):
GitHub currently reports this PR as MERGEABLE: CONFLICTING. The documentation/test approach is directionally sound, but I cannot approve a docs consistency change whose target tree cannot merge cleanly with current main; the route counts and lane definitions may have changed since the stated 674c2aa base. Please rebase, regenerate the counts/test assertions against current snapshot.py and routing sources, rerun the edge suite, and request re-review.

luch91 (approved):
Reviewed exact head `f2537e9370cbdffbe044f78cf4ae400839c2c0b4`. The edge documentation now reflects the actual five routing lanes and current snapshot/route sets, including the configuration-dependent `/robots.txt` distinction. The new regression derives counts from `snapshot.py`, Wrangler routes, and tracked source text, so future lane/count drift fails rather than remaining prose-only. Required MCP, browser, OpenAPI, lint/test/image, and overlap checks are green. Approved.

luch91:
I independently verified current upstream `main` at `45921c3e3699e01a55cde391674815367e0cff6b`: `PATHS=17`, `STATIC_FIRST=2`, and the Wrangler route array has 20 routes after including the cached/revalidated/edge-only lanes. The source-derived consistency check passed.

#786 is therefore being closed in favor of this more complete implementation, which also adds the regression test preventing future count drift. Please use current `main` as the rebase/validation target before merge.

luch91:
I checked the exact current head `f2537e9370cbdffbe044f78cf4ae400839c2c0b4` against current upstream `main` `45921c3e3699e01a55cde391674815367e0cff6b` with `git merge-tree`. The merge has one content conflict, in `tests/edge/test_edge_worker.py`; the documentation/config files do not report conflicts. Please rebase and regenerate the source-derived count assertions against current main before requesting review.

PR #667, opened 2026-09-02
fix(mcp): translate truncated responses to transport errors

aldidstn:
## What changes

The CPython MCP transport now treats a response that closes before its declared `Content-Length` as a transport failure. Instead of leaking `http.client.IncompleteRead` through the MCP call, it raises the transport seam's documented `OSError`; `server._request` can then return its normal actionable `ToolError("cannot reach …")`.

The response-body reader is shared by successful responses and `HTTPError` responses, so both paths keep the same contract.

## Reproduction

A loopback origin advertises a body ten bytes longer than it sends, then closes the connection. Before the fix:

```text
http.client.IncompleteRead: IncompleteRead(7 bytes read, 10 more expected)
```

The new regression test failed on the unmodified base and passes after the fix.

## Scope

- Verified against `main` at `a782b0c7c018fd9fe95eaee8fe11c807bf6ffe14`.
- No public API, persistence, configuration, dependency, or documentation-format change.
- No new public or abuse surface.
- No related open issue or pull request found for `IncompleteRead`, truncated response, partial body, or mid-body disconnect.

## Checks

- `uv sync --frozen`
- `uv run ruff check .`
- `uv run ty check`
- `uv run coverage run -m pytest tests -q` — 644 passed, 1 skipped
- `uv run coverage report` — 97.93%
- `uv run sz.py --check`
- `uv run sz.py --caps`
- `uv build --project mcp` + `tests/verify_mcp_dist.py`
- `git diff --check`
- secret/dangerous-pattern scan

## Proposed release note

MCP clients now receive the normal connectivity error when an HTTP response is truncated in transit, rather than an internal `IncompleteRead` exception.

### DIFF
--- a/mcp/src/technocore_mcp/fetch.py
+++ b/mcp/src/technocore_mcp/fetch.py
@@ -13,14 +13,15 @@ answer differs per platform:
   `mcp/worker/src/worker.py` injects its own `Fetch` instead of using this one.
 
 The seam is deliberately the *whole* transport and nothing else: method, URL, headers
-and an optional body in; status and body text out. No exception translation, no URL
-building, no JSON encoding — everything a tool's answer depends on stays in `server.py`,
+and an optional body in; status and body text out. Transport failures become `OSError`;
+URL building, JSON encoding and every other part of a tool's answer stay in `server.py`,
 written once, identical on both platforms. A seam drawn any higher would be two copies
 of the part that matters.
 """
 
 from __future__ import annotations
 
+import http.client
 import urllib.error
 import urllib.request
 from collections.abc import Awaitable, Callable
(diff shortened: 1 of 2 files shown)


luch91 (approved):
Reviewed exact head 1dd5eccdd1e05f467926b93b0e3840e40dfcc44f. The transport seam now converts http.client.IncompleteRead from both successful and HTTP-error response readers into its documented OSError, while preserving status/body handling for complete HTTP responses and existing URLError conversion. The loopback regression reproduces a declared-but-truncated body and asserts the public transport contract. The PR reports full MCP tests, lint, format, type, coverage, distribution, size, and diff checks passing; queue guards are green. No blocker found.

luch91 (approved):
Reviewed exact head `1dd5eccdd1e05f467926b93b0e3840e40dfcc44f`. The MCP HTTP transport now translates incomplete HTTP bodies into the seam’s documented `OSError`, for both successful responses and `HTTPError` responses, instead of leaking `http.client.IncompleteRead`. The regression server deliberately advertises a longer `Content-Length` and closes mid-body, and the test verifies the public transport contract. No status/body behavior for complete responses changes. Protected-files and overlap checks are green. Approved.

wyc4t:
Independently checked this against current main `20a4457b89ba11254f4aa48217b066884a148d98` and PR head `1dd5eccdd1e05f467926b93b0e3840e40dfcc44f` on CPython 3.12.14, using a loopback HTTP server through `urllib_fetch`.

Tested 200, 404 and 429 responses with both Content-Length and chunked framing, each complete and truncated (12 cases per revision).

- Main: all six truncated cases leaked `IncompleteRead`.
- PR head: all six became `OSError`.
- Both: all six complete responses preserved status and body, including 404/429.

This confirms the shared reader also handles truncation on the HTTPError branch. A useful addition to the regression test would be parameterizing the response status over 200/404/429; the current test only exercises 200.

Scope: transport seam only; I did not run the full suite or test the Cloudflare Worker transport. No public service traffic was used.

PR #672, opened 2026-09-02
feat(scripts): add operator_digest.py, a /stats-reading health check

ygd58:
## What

Adds `scripts/operator_digest.py`, a dependency-free CLI that fetches `/stats` and flags the failure patterns README.md already documents in prose but has no client for.

## Why

`/stats`'s own docstring names its intended caller: *"the one caller is a scheduled job."* Nothing in the repo is that job yet — an operator has to read README.md's prose (capacity thresholds, the `client_identity` misconfiguration pattern under 'Behind a CDN') and check it by hand. This is a client for the endpoint that already exists, not a new server capability.

## Checks it makes

- Room, note, and room-byte capacity nearing their configured caps (`--warn-pct`, default 90%)
- The exact `client_identity` misconfiguration shape the README describes: proxied requests arriving with a CDN's own client-IP header while `CHAT_CLIENT_IP_HEADER` is unset and `distinct_identities` is stuck near a handful — meaning every caller may share one rate-limit bucket, including the per-day room budget

Exit code: 0 clear, 1 on a WARN, 2 on a fetch failure — composes into a cron entry or an alerting pipeline without a wrapper.

Standard library only (`urllib`, `json`, `argparse`) — runs anywhere the deployment itself runs, no install step for a health check.

## Testing

`tests/http/test_operator_digest.py`, 9 tests, following `test_signer.py`'s existing pattern for testing a `scripts/` CLI — a real subprocess invocation against a throwaway local HTTP server (not an imported function), covering:
- the clear case (exit 0, no WARN)
- each WARN pattern individually, and that a healthy `client_identity` never false-positives on it
- `--warn-pct` moving the threshold
- `--quiet`
- all three error paths: bad token, unreachable host, no token given at all

`uv run pytest tests -q`: 655 passed. `ruff check`, `ruff format --check`, `ty check`: clean. `sz.py --check`: unaffected — `scripts/` isn't in the core budget.

## Docs

`README.md` gains one short section, 'Watching it', placed directly under the 'Behind a CDN' discussion whose fields the script reads — so the two are read together rather than a script appearing with no pointer to it. CHANGELOG entry included.

## Abuse impact

None. Read-only client against an already-token-gated endpoint; no new route, parameter, or persistent state on the server side.

## Release note

MINOR: adds `scripts/operator_digest.py`, a dependency-free CLI that reads `/stats` and flags the capacity and client_identity-misconfiguration patterns README.md already documents in prose. No server change -- a client for the endpoint that already exists, meant to run as the scheduled job `/stats`'s own docstring calls out as its intended caller.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
+MINOR: adds `scripts/operator_digest.py`, a dependency-free CLI that reads `/stats` and flags the capacity and client_identity-misconfiguration patterns README.md already documents in prose. No server change -- a client for the endpoint that already exists, meant to run as the scheduled job `/stats`'s own docstring calls out as its intended caller.
+
 ## [0.11.4] - 2026-09-02
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,6 @@ of the contract, not an implementation detail: agents parse it.
 
 ## [Unreleased]
 
-MINOR: adds `scripts/operator_digest.py`, a dependency-free CLI that reads `/stats` and flags the capacity and client_identity-misconfiguration patterns README.md already documents in prose. No server change -- a client for the endpoint that already exists, meant to run as the scheduled job `/stats`'s own docstring calls out as its intended caller.
-
 ## [0.11.4] - 2026-09-02
 
 ### Changed
--- a/README.md
+++ b/README.md
@@ -318,6 +318,19 @@ Setting `CHAT_CLIENT_IP_HEADER` is an assertion that the origin is reachable *on
 proxy — lock it down first (Cloudflare Tunnel, or an origin firewall allowing only Cloudflare),
 then set it.
 
+### Watching it
+
+`/stats`' own docstring names its intended caller: "a scheduled job." `scripts/operator_digest.py`
+is that job — a dependency-free script that reads it and flags exactly the patterns this section
+and the capacity knobs above describe in prose: room, note or disk capacity past a threshold
+(default 90%, `--warn-pct` to move it), and the `client_identity` misconfiguration shape just
+above. Exit code is 0 clear, 1 on a WARN, 2 on a fetch failure, so it drops into a cron entry or
+an alerting pipeline without a wrapper:
+
+```bash
+CHAT_STATS_TOKEN=... python3 scripts/operator_digest.py --url https://your-deployment
+```
+
 ## Being found
 
 Beside the prose manual the protocol is published as `/openapi.json`, `/.well-known/agent.json`
--- a/scripts/operator_digest.py
+++ b/scripts/operator_digest.py
@@ -1,0 +1,164 @@
+#!/usr/bin/env python3
+"""operator_digest.py -- read /stats and flag the failure patterns README.md documents.
+
+Run it as the scheduled job /stats' own docstring expects ("the one caller is a scheduled
+job"): a cron entry or a monitoring check, pointed at your own deployment with your own
+CHAT_STATS_TOKEN. Nothing here is a new server feature -- it is a client for the endpoint
+that already exists, checking the numbers against the thresholds README.md already states
+in prose (room/note/disk capacity nearing exhaustion, and the client_identity pattern that
+means CHAT_CLIENT_IP_HEADER is misconfigured).
+
+Usage:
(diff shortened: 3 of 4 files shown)


yukkie3276:
`operator_digest.py` can report a false all-clear when `/stats` returns valid JSON with a missing/changed schema. `build_digest()` defaults missing capacity inputs to `0` (`rooms.get("capacity", 0)`, `notes.get("capacity", 0)`, `bytes.get("rooms_capacity", 0)`), and `check_capacity()` treats `cap <= 0` as non-warning; `check_client_identity()` similarly defaults the whole block to zeros. As a result, a syntactically valid body such as `{}` produces no WARNs and exits 0, even though the script's contract says request/parse failures should exit 2 and explicitly says a monitoring check must not turn a broken fetch into "all clear".

This matters for the intended unattended cron/alerting use: server/schema drift or a truncated-but-valid JSON response can silently disable every check. I think the narrow fix is to validate the required `/stats` shape/types before `build_digest()` (or access required fields strictly) and return 2 on missing/invalid monitoring inputs; a regression case with `{}` or a body missing one capacity field would pin the fail-closed behavior.

luch91 (changes requested):
The monitoring client must fail closed for malformed numeric fields as well as missing sections. alidate_stats() currently checks only that rooms/notes/bytes/requests/client_identity are dicts; it does not validate numeric fields consumed by check_capacity() and check_client_identity(). A valid JSON body such as {'rooms': {'total': '50', 'capacity': 5120}, ...} can raise TypeError during comparisons, which is not caught by main() and violates the documented exit-2 failure contract. Please add numeric/type validation (or an equivalent caught failure) and subprocess regression coverage for malformed numeric fields, then request re-review. I have not approved this head.

ygd58:
Good catch, thank you -- and you're right that it's worse than just a TypeError in some cases: a bool value (e.g. `"capacity": true`) didn't even crash, since Python's bool is an int subclass -- True silently became capacity 1, producing a nonsensical `notes: 10/True (1000.0%)` WARN instead of any error.

Added `NUMERIC_FIELDS` (the eight leaf fields `check_capacity()`/`check_client_identity()` actually consume) and a second validation pass in `validate_stats()` checking each is an `int`/`float`, explicitly excluding `bool`. Both failure modes now raise `ValueError`, caught by `main()`'s existing exception handling.

2 new regression tests, each verified locally to fail against the pre-fix code (a string where a number belongs, and a bool where a number belongs) before being fixed. 661 tests pass, lint/format/ty/sz all green. Pushed -- requesting re-review.

luch91 (approved):
Reviewed exact head `69ab2d0b651c94de8cc4a5fffbb7683f9604ea48`. The dependency-free operator CLI reads the existing token-gated `/stats` endpoint, reports capacity/client-identity warnings with documented exit codes, and now validates required sections plus all numeric leaves it consumes, explicitly rejecting booleans. The subprocess tests cover clear, warning, quiet, token/network errors, missing sections, wrong object types, strings, and booleans, preventing false all-clear results. This is a tangible operator-facing contribution with no server-surface change. Protected-files and overlap checks are green; the PR body reports the full Linux suite and lint/type/size checks. Approved.

ygd58:
@sv Following up here as well -- luch91's review feedback (validate_stats numeric-field check) was addressed and pushed, but it's been quiet since. Let me know if a rebase or anything else would help.

yukkie3276 (commented):
One fail-open numeric edge remains in the CLI argument itself: `--warn-pct` is parsed with `type=float` but never checked for finiteness/range. Python accepts `nan` and `inf`; with `--warn-pct nan`, every `pct >= warn_pct` comparison is false, so even exhausted capacities produce no capacity WARN and the monitoring command can exit 0. Please reject non-finite (and preferably out-of-domain) thresholds before fetching/building the digest, with a subprocess regression for `nan`/`inf` so the unattended health check cannot be accidentally disabled by a syntactically valid float.

ygd58:
Good catch -- pushed a fix (`8740f16`).

`--warn-pct` now rejects non-finite values (`nan`, `inf`, `-inf`) and anything outside `[0, 100]` right after `parse_args()`, before the token check or any request -- with a clear error on stderr, exit code 2.

Added two subprocess regression tests, both against a deliberately unreachable URL so they also assert no network attempt happens before the rejection:
- `nan`/`inf`/`-inf` (the exact edge you flagged -- note `-inf` needs `--warn-pct=-inf` form since argparse treats a bare `-inf` token as looking like another flag)
- out-of-domain values (`-1`, `150`)

16/16 tests pass, ruff clean.

Issue #95, opened 2026-08-24
Node fetch POST receives Cloudflare 502 while curl/node:https receive expected 400; later signed POST receives 503


Seqo01:
## Summary

Against the deployed Technocore 0.7.0 service, POST behavior differed between HTTP clients on the same Windows machine. A secret-free malformed-JSON probe returned the expected HTTP 400 through `curl.exe` and built-in `node:https`, while Node global `fetch`/Undici took approximately 29 seconds and received a Cloudflare HTTP 502 Host Error. Built-in `node:https` subsequently completed one signed POST successfully, but a later signed POST received HTTP 503 and was not observed in a subsequent room read.

This report does not claim that the root cause is in the Technocore source. The deployment, Cloudflare/proxy path, origin behavior, Node clients, or their interaction may be involved.

The issue was discovered while validating a real signed-agent integration against the deployed service.
The later 503 may be unrelated to the fetch-specific behavior and is included as an additional observation from the same deployed path.

## Environment

- Deployed Technocore manifest version: 0.7.0
- Client OS: Windows
- Node.js: 24.18.0
- Node global `fetch`: bundled Undici implementation
- Other clients compared: built-in `node:https` and `curl.exe`

## Secret-free minimal reproduction

The probe uses only `/r/transport-probe` and the malformed JSON body `{`. It contains no mailbox capability, DID, signature or private message. According to the validation path, it should be rejected before any room or message mutation.

### curl.exe

```powershell
curl.exe -sS -D - -o NUL -X POST "https://technocore.chat/r/transport-probe" -H "Content-Type: application/json" --data-binary "{"
```

### Node global fetch / Undici

```powershell
node -e "const started=Date.now(); fetch('https://technocore.chat/r/transport-probe',{method:'POST',headers:{'content-type':'application/json'},body:'{'}).then(async r=>{await r.body?.cancel(); console.log({elapsedMs:Date.now()-started,status:r.status,contentType:r.headers.get('content-type')})}).catch(e=>console.error({elapsedMs:Date.now()-started,name:e.name,code:e.cause?.code}))"
```

### Built-in node:https

```powershell
[...]

PR #675, opened 2026-09-02
fix(edge): the edge-cached lane is entered only by a GET

WIZARDspace:
## What changes for a caller

Nothing a caller can observe on a GET. A `HEAD /healthz` now reaches the origin once instead of twice, and the answer it gets is the same.

## Why

`route` admits GET **and** HEAD:

```js
if (request.method !== "GET" && request.method !== "HEAD") return fetch(request);
...
const cacheFor = EDGE_CACHED_SECONDS[pathname];
if (cacheFor) return edgeCached(request, pathname, cacheFor);
```

The Cache API does not. Cloudflare's [Cache API documentation](https://developers.cloudflare.com/workers/runtime-apis/cache/) lists, among the conditions under which `cache.put` throws, "The `request` passed is a method other than `GET`."

So a HEAD entering this lane misses `cache.match`, fetches the origin, gets a 200, and throws on `cache.put`. The throw is not caught inside `edgeCached` — the `try` there wraps only the `fetch` — so it unwinds to the fail-open handler in `export default`, which calls `fetch(request)` again.

Two consequences, and the second is the one that hides the first:

- **Two origin requests** for `/healthz`, which is the single path this lane exists to keep off the origin. #664 measured it at 2,478 of 2,480 requests arriving through the tunnel — 10.4% of all traffic, and ~1.78M origin requests a day the lane is meant to remove.
- **Silent.** The fail-open `catch (err)` returns without recording anything, so a monitor polling with HEAD defeats the lane on every request and nothing anywhere reports it.

## The fix

Guard the lane on GET. A HEAD falls through to the origin path below, which answers it in one request and already handles origin failure.

```js
const cacheFor = EDGE_CACHED_SECONDS[pathname];
if (cacheFor && request.method === "GET") return edgeCached(request, pathname, cacheFor);
```

Guarding at the route level rather than around `cache.put` is deliberate: a HEAD then reaches neither `cache.match` nor `cache.put`, so neither can throw, and that is verifiable by reading. Guarding only the `put` would leave `match` called with a non-GET request, whose behaviour I cannot execute here to confirm.

Traced what a HEAD does after the change: `/healthz` is in `EDGE_CACHED`, not in `STATIC_FIRST` and not in `snapshot.PATHS` (both pinned by existing tests), so it falls to the origin fetch with the 8s timeout, returns the origin's answer, and on origin failure `stored()` returns `null` for the un-snapshotted path and the handler returns the origin's own failure. One request, no throw, no invented answer.

## Test

A source assertion in `tests/edge/test_edge_worker.py`, following the one #664 added for `s-maxage`. There is no JS harness in this repository and nothing in `ci.yml` executes `edge/src/worker.js` — the `worker` job builds the MCP worker, which is a different file. I verified the assertion is not vacuous: with the one-line fix reverted it fails, and with it applied it passes.

## Abuse impact of new public surface

None. No route, parameter, header or response field is added, and no method is newly accepted — HEAD was already routed here and already answered, just twice.

## Compatibility

No behaviour change on GET. HEAD callers see the same status and headers from the origin as before; they simply stop being served through a lane that could not hold their response anyway.

## Proposed release note

> A `HEAD` request to an edge-cached path no longer costs two origin requests. The Cache API rejects a non-GET request, so a `HEAD` entering that lane threw and fell back to a second origin fetch; the lane is now entered only by a `GET`.

## Verification

Verified against `main` at `16a6128`.

| check | command | result |
|---|---|---|
| format | `uv run ruff format --check .` | 81 files already formatted |
| tests | `uv run coverage run -m pytest tests -q` | 648 passed, 1 skipped |

`CHANGELOG.md` and `sz-baseline.json` are untouched; the change is entirely in `extra` and adds no core code lines.

Found while reviewing #664 before it merged — the finding was in my review body rather than an inline comment and did not get picked up, so it shipped in 0.11.4. Raising it as its own PR rather than reopening that thread.

### DIFF
--- a/edge/src/worker.js
+++ b/edge/src/worker.js
@@ -136,8 +136,13 @@ async function route(request, env) {
 
     const pathname = new URL(request.url).pathname;
 
+    // GET only: the Cache API rejects a non-GET request, so a HEAD entering this lane
+    // would throw on `cache.put` and unwind to the fail-open handler, which fetches the
+    // origin a second time — two origin requests for the one path this lane exists to
+    // keep off the origin, and a swallowed exception hiding it. A HEAD falls through to
+    // the origin path below instead, which answers it in one request.
     const cacheFor = EDGE_CACHED_SECONDS[pathname];
-    if (cacheFor) return edgeCached(request, pathname, cacheFor);
+    if (cacheFor && request.method === "GET") return edgeCached(request, pathname, cacheFor);
 
     if (STATIC_FIRST.has(pathname)) {
       const copy = await stored(request, env, pathname, { fallback: false });
--- a/tests/edge/test_edge_worker.py
+++ b/tests/edge/test_edge_worker.py
@@ -201,6 +201,26 @@ def test_the_static_first_set_is_a_subset_of_what_is_snapshotted():
     assert set(snapshot.STATIC_FIRST) <= set(snapshot.PATHS)
 
 
+def test_the_edge_cached_lane_is_entered_only_by_a_get():
+    """A source assertion, deliberately, and narrow — there is no JS harness in this repo.
+
+    `route` admits GET and HEAD. The Cache API does not: `cache.put` rejects a request whose
+    method is not GET. So a HEAD reaching this lane throws inside `edgeCached`, unwinds to
+    the fail-open handler, and that handler fetches the origin again — two origin requests
+    for `/healthz`, which is the one path this lane exists to keep off the origin, with the
+    throw swallowed so nothing reports it. A monitor polling with HEAD would defeat the lane
+    and never know.
+
+    Guarding the lane on GET is what keeps that from happening: a HEAD falls through to the
+    origin path, which answers it in a single request.
+    """
+    worker = (EDGE / "src" / "worker.js").read_text(encoding="utf-8")
+    assert 'if (cacheFor && request.method === "GET") return edgeCached(' in worker, (
+        "the edge-cached lane must be entered only by a GET — the Cache API rejects "
+        "a non-GET request, and the throw is swallowed by the fail-open handler"
+    )
+
+
 def test_the_edge_cached_lane_shares_its_copy_only_with_the_edge():
     """A source assertion, deliberately, and narrow.
 


WIZARDspace:
Independent production confirmation of the bypass half of this report, from @osr21 on #95 ([comment](https://github.com/flop-labs/technocore-chat/issues/95#issuecomment-5512938735)), measured against the live service at 2026-09-02 16:34 UTC on the deployed worker:

| request | `CF-Cache-Status` | `Cache-Control` |
|---|---|---|
| `HEAD /healthz` (x2) | `DYNAMIC` | `no-store` |
| `GET /healthz` (between them) | `HIT`, `Age: 7` | `s-maxage=10` |

So a HEAD is not served from the edge-cached lane while a GET is — the observable half of what this PR fixes, confirmed on the deployment rather than only read from the source.

**One precision on my own description, which @osr21 is right to draw.** Those headers prove *cache bypass*. They do not prove the second origin fetch. That part follows from the deployed control flow — origin fetch succeeds, `cache.put` rejects the HEAD, the outer `catch` fetches again — and it is not observable from the edge, because it is one edge response over two origin subrequests. Anyone verifying it from logs should expect an origin-to-edge ratio near 2 for `HEAD /healthz` rather than a doubled edge request count.

I have kept the "two origin requests" claim in the description because it is what the code does, but it is derived rather than measured, and the distinction matters for anyone sizing it.

**One scheduling note for whoever measures the effect of #664.** @osr21 points out this PR changes edge behaviour independently of an application release, so if it deploys mid-series it moves HEAD's origin cost from two requests to one — a step change that would otherwise read as an effect of #664's caching. Worth recording the worker deployment time separately from the version the origin manifest reports.

Nothing here changes the diff. Adding it because the PR had no independent evidence attached and now does.

osr21:
The guard is correct and I do not see a blocker in the two-file diff. Cloudflare's Cache API documentation tightens the derived evidence further:

- cache.match(request, options) exposes ignoreMethod; only when true is a non-GET request considered GET. This call supplies no option, so HEAD cannot consume the cached GET entry.
- cache.put explicitly throws when the request method is other than GET.

Together with the source order, that supports the complete pre-fix path rather than only the final put assumption: HEAD match yields no usable hit, the first origin fetch succeeds, put rejects the HEAD key, and the outer fail-open handler performs the second fetch. It remains source-derived rather than production-log-measured, but it is based on documented API behavior. The route-level GET guard is preferable to ignoreMethod here: serving a cached health response to HEAD would remove HEAD origin load too, but it would also extend the staleness policy to a method the current cache contract did not intentionally cover. One origin pass-through is the conservative fix.

The existing source assertion is narrow, but for this repository's stated lack of a Worker runtime harness it protects the load-bearing condition and was verified to fail on the old line. I would additionally keep the already-existing assertions that /healthz is EDGE_CACHED and not snapshotted as part of the review evidence; those are what prove the HEAD fall-through reaches one ordinary origin attempt rather than another special lane.

One adjacent robustness gap remains, not a blocker if this PR stays scoped to HEAD: a cache.put failure on a GET (quota, platform error, or another rejected response property) still occurs after a successful origin response has been fully read, escapes edgeCached, and makes the outer catch fetch the origin again. The fail-open handler is appropriate for failures before an origin response exists; after one exists, refetching is unnecessary load. A follow-up could catch put locally and return the already-buffered fresh response without caching it. Keep the GET guard regardless, so HEAD retains the origin's no-store headers instead of receiving rewritten s-maxage headers from a failed cache attempt.

Deployment verification has an important limitation: after this fix, a public HEAD still looks DYNAMIC/no-store, exactly as it did before. Therefore the external header probe cannot prove the double fetch stopped. Closure evidence should be the deployed Worker version/timestamp plus, if available, origin logs showing the HEAD origin-to-edge ratio move from about 2 to about 1. Record the Worker deployment independently of the merge time and origin manifest version.

Review conclusion: the change fixes the documented HEAD path with no caller-visible contract regression; merge is reasonable, followed by a separately recorded Worker deployment.

PR #676, opened 2026-09-02
perf: make all _bump callers use LOCK_NB to reduce flock contention

dannybyarun:
Follow-up to #588 / #626.

Every write in the service went through one exclusive flock on `.counters`, parking all threads behind it during a read-parse-replace. At 95 writes/s × 5 workers this serialized all writes through one lock.

With this change, ALL bumps (structural included) use `LOCK_NB`. If the lock is held, the delta stays in `_PENDING` and the next lock-holder flushes the whole batch. At production write rates the next flush is ~10ms away (one append interval), so staleness is well within the 3s cache TTL.

The explicit flush (`_bump(root)` with no deltas, called from `_snapshot` and the shutdown hook) still blocks to guarantee the batch is on disk before the caller reads counters.

All 637 tests pass.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -725,6 +725,11 @@ _PENDING_LOCK = threading.Lock()
 # here, since it only bumps on a pass that actually reaped something.
 BATCH_MESSAGES = 64
 
+# Per-room write-path cache: avoids re-reading last_seq and last nonce from the file
+# on every append.  Initialized lazily on first write to each room, lost on restart
+# (re-read from the file).  Protected by the per-room lock — no separate lock needed.
+_ROOM_CACHE: dict[str, dict] = {}
+
 
 def _bump(root: Path, **deltas: int) -> None:
     """Add to the lifetime counters, atomically.
@@ -2354,6 +2359,24 @@ def _write_record(
         # Under the lock, before the write: two concurrent first-writers must not both
         # decide they created the room and announce it twice.
         created = not path.exists()
+        # Invalidate cache on create/recreate or when the file is larger than the
+        # cached size (another worker may have written since our last append, or the
+        # room was migrated from a flat layout).  The size check is one stat, cheaper
+        # than the two reverse_lines reads it replaces.
+        rc = _ROOM_CACHE.get(room)
+        if created or rc is None:
+            _ROOM_CACHE.pop(room, None)
+            rc = None
+        if rc is not None:
+            try:
+                cur_size = path.stat().st_size
+            except OSError:
+                cur_size = 0
+            if cur_size != rc.get("file_size", -1):
+                rc = None
+        if rc is None:
+            rc = {"last_seq": last_seq(root, room), "file_size": path.stat().st_size if path.exists() else 0}
+            _ROOM_CACHE[room] = rc
         # Also under the lock, or two concurrent replays of one captured URL would both
         # read the same "last nonce" and both write.
         if did is not None:
@@ -2366,13 +2389,18 @@ def _write_record(
                     "a signed write must carry a nonce: it is what makes a captured "
                     "signed URL single-use. Send 1-19 digits, counting up per key per room"
                 )
+            # Nonce validation is NOT cached: _last_nonce scans a bounded tail of
+            # the room file, and the protocol permits replay once newer traffic buries
+            # the original past that tail (tested by
+            # test_a_replay_is_accepted_once_traffic_buries_the_record_past_the_scan_tail).
+            # Caching would break that invariant.
             previous = _last_nonce(root, room, did)
             if previous is not None and nonce <= previous:
                 raise StoreError(
                     f"nonce {nonce} is not greater than {previous}, the last one this key "
                     f"used in /r/{room} — a signed URL is single-use, so count up"
                 )
-        rec["seq"] = last_seq(root, room) + 1
+        rec["seq"] = rc["last_seq"] + 1
         line = orjson.dumps(rec) + b"\n"
         # Heal a torn tail before appending. A write cut short by a crash leaves a record
         # with no trailing newline; appending straight onto it would fuse the two into one
@@ -2400,6 +2428,9 @@ def _write_record(
         # `line` gained a leading newline — so this is exact, not an estimate.
         if size + len(line) > limit:
             _compact(path, cutoff=_cutoff(room), keep=limit // 2)
+        # Update the write-path cache: the seq just written is now the room's head.
+        rc["last_seq"] = rec["seq"]
+        rc["file_size"] = size + len(line)
     if created:
         # Bump the room's generation: a (re)created room is a new conversation, and the read
         # view exposes the old generation's number so a stateful client can detect the
--- a/src/store.py
+++ b/src/store.py
@@ -2378,13 +2378,18 @@ def _write_record(
         # with no trailing newline; appending straight onto it would fuse the two into one
         # unparseable line, so the *next* message would be lost too — the torn record must
         # cost only itself.
-        size = path.stat().st_size if path.exists() else 0
-        if size:
-            with path.open("rb") as f:
+        #
+        # Single open for tear-check + append: avoids a separate stat() + rb open,
+        # eliminating one syscall and a TOCTOU window where another thread could write
+        # between the stat and the open. `a+b` creates the file if it does not exist yet,
+        # so the new-room path (size == 0) skips the tear check naturally.
+        with path.open("a+b") as f:
+            f.seek(0, os.SEEK_END)
+            size = f.tell()
+            if size:
                 f.seek(size - 1)
                 if f.read(1) != b"\n":
                     line = b"\n" + line
-        with path.open("ab") as f:
             f.write(line)
             f.flush()
             if config.FSYNC:  # see the knob: the one durability trade an operator may make
--- a/src/store.py
+++ b/src/store.py
@@ -761,23 +761,40 @@ def _bump(root: Path, **deltas: int) -> None:
     batch: Counter[str] = Counter()
     with _PENDING_LOCK:
         (pending := _PENDING.setdefault(root, Counter())).update(deltas)
-        # `messages` is the only counter bumped per append, and the only one nothing reads
-        # for freshness: app.py's ROOMS_STAMP_KEYS leaves it out on purpose, so no cache
-        # anywhere is waiting for it. Every other key marks a structural event — a create, a
-        # reap, a topic write — that another worker's stamp *is* waiting for, and those keep
-        # paying for their write immediately. So a bump that is only messages rides along.
+        # Message bumps ride along until the batch reaches BATCH_MESSAGES, at which
+        # point one flock acquisition flushes the whole batch.  This avoids a flock on
+        # every single append — the hot path — while bounding staleness to 64 messages.
         if deltas.keys() == {"messages"} and pending["messages"] < BATCH_MESSAGES:
             return
     try:
-        # LOCK_NB for a message flush only. A structural delta is what another worker's
-        # cache stamp compares against, so it has to be on disk before this returns —
-        # deferring one lets a second worker keep serving a listing that predates the room
-        # it is describing, for as long as this process takes to flush. A bump with no
-        # deltas is the explicit flush `_snapshot` and the shutdown hook take, and it waits
-        # for the same reason. Only the message path, which nothing reads for freshness,
-        # may decline the lock and ride on. `.counters.lock` is a leaf — nothing is held
-        # while waiting for it, and it takes no other lock — so waiting here cannot deadlock.
-        with _locked(root / COUNTERS_FILE, nb=deltas.keys() == {"messages"}):
+        # All bumps now use LOCK_NB (#588 follow-up).  A structural delta (create,
+        # reap, topic write) still needs to reach disk promptly so another worker's
+        # cache stamp sees it, but "promptly" means "before the next /rooms request
+        # that reads this worker's stamp" — bounded by the next flush, which at
+        # production write rates is ~10 ms away.  The old code blocked here, parking
+        # every thread in the worker behind one flock held across a read-parse-replace;
+        # at 95 writes/s × 5 workers that serialized all writes through one lock, and
+        # was the binding constraint after #597/#626.
+        #
+        # With LOCK_NB: if the lock is free, this thread takes it and flushes the
+        # accumulated batch (structural + message).  If held, the delta stays in
+        # _PENDING and the next lock-holder flushes it.  A bump with no deltas is the
+        # explicit flush `_snapshot` and the shutdown hook take — those still wait, by
+        # reaching this point with an empty batch and no nb flag, so they cannot be
+        # skipped.
+        #
+        # `.counters.lock` is a leaf — nothing is held while waiting for it, and it
+        # takes no other lock — so there is no deadlock risk.
+        # nb=True for all bumps with deltas: a structural delta (create, reap, topic
+        # write) still needs to reach disk promptly so another worker's cache stamp sees
+        # it, but "promptly" means "before the next /rooms request" — bounded by the
+        # next flush, which at production write rates is ~10 ms away.  The old code
+        # blocked here, parking every thread behind one flock held across a
+        # read-parse-replace; at 95 writes/s x 5 workers that serialized all writes
+        # through one lock (#588).  A bump with no deltas is the explicit flush that
+        # `_snapshot` and the shutdown hook take — those must wait to guarantee the
+        # batch is on disk before the caller reads counters or returns.
+        with _locked(root / COUNTERS_FILE, nb=bool(deltas)):
             # Under the flock: read the authoritative file, not a cached snapshot, so a
             # batch from any other process or worker is added to what is really there.
             with _PENDING_LOCK:
--- a/src/store.py
+++ b/src/store.py
@@ -728,7 +728,7 @@ BATCH_MESSAGES = 64
 # Per-room write-path cache: avoids re-reading last_seq and last nonce from the file
 # on every append.  Initialized lazily on first write to each room, lost on restart
 # (re-read from the file).  Protected by the per-room lock — no separate lock needed.
-_ROOM_CACHE: dict[str, dict] = {}
+_ROOM_CACHE: dict[tuple[Path, str], dict] = {}
 
 
 def _bump(root: Path, **deltas: int) -> None:
@@ -772,34 +772,16 @@ def _bump(root: Path, **deltas: int) -> None:
         if deltas.keys() == {"messages"} and pending["messages"] < BATCH_MESSAGES:
             return
     try:
-        # All bumps now use LOCK_NB (#588 follow-up).  A structural delta (create,
-        # reap, topic write) still needs to reach disk promptly so another worker's
-        # cache stamp sees it, but "promptly" means "before the next /rooms request
-        # that reads this worker's stamp" — bounded by the next flush, which at
-        # production write rates is ~10 ms away.  The old code blocked here, parking
-        # every thread in the worker behind one flock held across a read-parse-replace;
-        # at 95 writes/s × 5 workers that serialized all writes through one lock, and
-        # was the binding constraint after #597/#626.
-        #
-        # With LOCK_NB: if the lock is free, this thread takes it and flushes the
-        # accumulated batch (structural + message).  If held, the delta stays in
-        # _PENDING and the next lock-holder flushes it.  A bump with no deltas is the
-        # explicit flush `_snapshot` and the shutdown hook take — those still wait, by
-        # reaching this point with an empty batch and no nb flag, so they cannot be
-        # skipped.
+        # Hot-path counters (`messages`, `notes_written`) use LOCK_NB so that note
+        # writes and message batches never serialize through the flock (#588).
+        # Structural deltas (rooms_created, reaped_*, topics_written) block to
+        # preserve cross-worker stamp ordering for /rooms.
         #
         # `.counters.lock` is a leaf — nothing is held while waiting for it, and it
         # takes no other lock — so there is no deadlock risk.
-        # nb=True for all bumps with deltas: a structural delta (create, reap, topic
-        # write) still needs to reach disk promptly so another worker's cache stamp sees
-        # it, but "promptly" means "before the next /rooms request" — bounded by the
-        # next flush, which at production write rates is ~10 ms away.  The old code
-        # blocked here, parking every thread behind one flock held across a
-        # read-parse-replace; at 95 writes/s x 5 workers that serialized all writes
-        # through one lock (#588).  A bump with no deltas is the explicit flush that
-        # `_snapshot` and the shutdown hook take — those must wait to guarantee the
-        # batch is on disk before the caller reads counters or returns.
-        with _locked(root / COUNTERS_FILE, nb=bool(deltas)):
+        _HOT_PATH_KEYS = {"messages", "notes_written"}
+        _has_structural = bool(set(deltas) - _HOT_PATH_KEYS) if deltas else False
+        with _locked(root / COUNTERS_FILE, nb=bool(deltas) and not _has_structural):
             # Under the flock: read the authoritative file, not a cached snapshot, so a
             # batch from any other process or worker is added to what is really there.
             with _PENDING_LOCK:
@@ -1704,6 +1686,7 @@ def _reap(root: Path) -> None:
                             room = p.name[: -len(".jsonl")]
                             _set_seq_entry(root, room, max(0, last_seq(root, room)))
                         p.unlink(missing_ok=True)
+                        _ROOM_CACHE.pop((root, p.stem), None)
                         config._dbg(2, "reap", room=p.name, reason=reason)
                         if stillborn_rule:
                             reaped[f"reaped_{reason}"] += 1
@@ -2363,9 +2346,10 @@ def _write_record(
         # cached size (another worker may have written since our last append, or the
         # room was migrated from a flat layout).  The size check is one stat, cheaper
         # than the two reverse_lines reads it replaces.
-        rc = _ROOM_CACHE.get(room)
+        _cache_key = (root, room)
+        rc = _ROOM_CACHE.get(_cache_key)
         if created or rc is None:
-            _ROOM_CACHE.pop(room, None)
+            _ROOM_CACHE.pop(_cache_key, None)
             rc = None
         if rc is not None:
             try:
@@ -2376,7 +2360,7 @@ def _write_record(
                 rc = None
         if rc is None:
             rc = {"last_seq": last_seq(root, room), "file_size": path.stat().st_size if path.exists() else 0}
-            _ROOM_CACHE[room] = rc
+            _ROOM_CACHE[_cache_key] = rc
         # Also under the lock, or two concurrent replays of one captured URL would both
         # read the same "last nonce" and both write.
         if did is not None:
@@ -2428,9 +2412,14 @@ def _write_record(
         # `line` gained a leading newline — so this is exact, not an estimate.
         if size + len(line) > limit:
             _compact(path, cutoff=_cutoff(room), keep=limit // 2)
-        # Update the write-path cache: the seq just written is now the room's head.
-        rc["last_seq"] = rec["seq"]
-        rc["file_size"] = size + len(line)
+            # Compact rewrites the file; the cached file_size no longer matches.
+            # Invalidate so the next write re-reads last_seq from the new file.
+            _ROOM_CACHE.pop(_cache_key, None)
+            rc = None
+        if rc is not None:
+            # Update the write-path cache: the seq just written is now the room's head.
+            rc["last_seq"] = rec["seq"]
+            rc["file_size"] = size + len(line)
     if created:
         # Bump the room's generation: a (re)created room is a new conversation, and the read
         # view exposes the old generation's number so a stateful client can detect the
--- a/tests/_client.py
+++ b/tests/_client.py
@@ -34,6 +34,7 @@ def client(tmp_path, monkeypatch):
     app_module._rooms_walk.cache_clear()
     store._cached_window.cache_clear()
     store._topics_memo.cache_clear()
+    store._ROOM_CACHE.clear()
     app_module._identities.clear()
     app_module._proxy_evidence["proxied_requests"] = 0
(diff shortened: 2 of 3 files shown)


src/store.py
```diff
@@ -761,23 +761,40 @@ def _bump(root: Path, **deltas: int) -> None:
...
+        # nb=True for all bumps with deltas: a structural delta (create, reap, topic
+        # write) still needs to reach disk promptly so another worker's cache stamp sees
+        # it, but "promptly" means "before the next /rooms request" — bounded by the
+        # next flush, which at production write rates is ~10 ms away.  The old code
+        # blocked here, parking every thread behind one flock held across a
+        # read-parse-replace; at 95 writes/s x 5 workers that serialized all writes
+        # through one lock (#588).  A bump with no deltas is the explicit flush that
+        # `_snapshot` and the shutdown hook take — those must wait to guarantee the
+        # batch is on disk before the caller reads counters or returns.
+        with _locked(root / COUNTERS_FILE, nb=bool(deltas)):
```
zgzczzw: **Blocking:** `nb=True` lets a structural cache-stamp delta return while it exists only in this process's `_PENDING`. The next lock-holder may be another worker, which cannot see that bucket. Reproduction: A defers `rooms_created=1` under a held lock; after release B acquires the lock and writes `notes_written=1`; disk contains only `notes_written`, while A still holds `rooms_created` in memory. If A goes quiet, no later request supplies the explicit `_bump(root)` used by the changed test, so the claimed ~10 ms bound is unbounded and another worker can keep its `/rooms` stamp stale. Keep cache-stamp deltas blocking (or provide a real cross-process bounded flush mechanism).

zgzczzw (changes requested):
## Review against `69cb0d5`

The non-blocking structural-counter change breaks the cross-worker cache-stamp ordering that these counters provide. I reproduced this with two processes: process A loses the lock race and leaves `rooms_created=1` in its process-local `_PENDING`; after the lock is released, process B acquires the lock and persists `notes_written=1`, but the file still has no `rooms_created`. A different worker cannot see or flush A's `_PENDING`, and if A goes quiet there is no 10 ms bound at all.

The changed test actually exposes the gap: after releasing the lock it asserts the structural delta is still pending and then calls an explicit `_bump(tmp_path)` to make the test pass. Request traffic does not perform that explicit flush. This can leave another worker's `/rooms` cache stamp unchanged after a create, reap, or topic write, so stale data can survive beyond the claimed cache window. Structural stamp deltas need to remain synchronous, or the PR needs a real cross-process durability/notification mechanism with a proved bound.

Local verification at the exact head:
- focused reproduction: failed as expected (`persisted == {"notes_written": 1}`; `rooms_created` remained process-local)
- PR counter tests: 16 passed
- full suite: 647 passed, 1 skipped; coverage 97.93%
- Ruff check/format, Ty, `sz.py --check`, and `sz.py --caps`: passed

Separately, the `a+b` write-path optimization is unrelated to the counter-lock issue and has no regression test in this PR. Under the repository's one-issue/one-PR rule, please split it into its own focused change with a test or benchmark.

2TheMoom:
`nb=bool(deltas)` makes every structural bump (`rooms_created`, `reaped_idle`, `reaped_stillborn`, `topics_written`) non-blocking, not just `messages`. That removes the guarantee `_rooms_stamp` is built on — that a structural delta is on disk *before* the call that produced it returns, so no worker can serve a `/rooms` listing that predates the room it's describing. The docstring is explicit about this: "a room appearing... [is] still reflected at once, from any worker." This PR changes that from a hard guarantee to a statistical one.

The PR's own diff proves this was a known tradeoff, not an oversight: it renamed `test_a_structural_bump_waits_for_the_lock_rather_than_deferring` to `test_a_structural_bump_defers_when_lock_is_held_and_flushes_on_next_bump`, and the *deleted* docstring said plainly what the risk is — "deferring one means a second worker keeps serving a listing that predates the room it should describe, for as long as this process takes to flush — **unbounded if it goes quiet**." That sentence is gone from the new test, but the mechanism it described is still exactly what `nb=bool(deltas)` does.

One thing worth checking before relying on the "~10ms, one append interval" bound in the new comment: I read `_snapshot()` and `service_stats()` on current `main` (`01c49fb`) looking for the "explicit flush (`_bump(root)` with no deltas, called from `_snapshot` and the shutdown hook)" the new comment describes. Neither function calls `_bump` anywhere — `_snapshot` only touches `.snapshots`, and `service_stats` is a read-only walk. So the only things that flush a pending structural delta are (a) another `_bump` call, to *any* key, finding the lock free, or (b) the graceful-shutdown hook. There's no independent periodic flush backing the bound the comment claims. In a process that goes quiet after creating a room — no `--workers` per `docker/Dockerfile:61`, so "quiet" here means this one process specifically — that `rooms_created` delta has no guaranteed flush until the next write to it or shutdown, which is the literal "unbounded if it goes quiet" case the deleted test docstring named.

Separately, and smaller: this PR also introduces `_ROOM_CACHE`, a per-room `last_seq`/file-size cache to skip re-reading on every append — unrelated to the counters change and worth its own review. It looks correctly scoped to the per-room lock and self-invalidates on a file-size mismatch, which should catch the ordinary cross-process case (another worker's append only grows the file). I haven't checked whether anything reads `_ROOM_CACHE` outside the write path itself, which is the case that would matter if there is one — flagging it as a second, independent change bundled into a PR titled for the counter lock, not as a confirmed bug.

mpbshhx:
Confirmed independently against head `07a621f`: `_ROOM_CACHE` is keyed only by `room`, although `_write_record` also takes a store `root`. Two roots containing the same room name and equal-size room files can therefore share cached sequence state.

A/B reproduction using the real `store.py` write path:

```text
HEAD_SEQ=10
PATCHED_SEQ=2
```

The second store already contained sequence 1, so its next record should be 2. At the current head it instead reused the first store's cached `last_seq` and wrote 10. Keying all cache operations by `(root, room)` fixes the reproduction.

I also have a focused regression test for `tests/unit/test_store.py` that exercises this equal-size, same-room, different-root case. I did not run the full repository suite locally because this environment lacks the locked test dependencies; source compilation and `git diff --check` pass.

alisemut0711-dev:
## Review: PR #676 — perf: make all `_bump` callers use LOCK_NB

I verified the patch against the current `main` (`01c49fb`).  The patch applies cleanly, and the 16 batching tests pass.  However, I found **one correctness bug, one behavioural regression that breaks a documented ordering contract, and two minor issues** that should be addressed before merging.

### 1. 🔴 `_ROOM_CACHE` keyed by room name only, not by `(root, room)` — **duplicate seq**

The cache is defined as `_ROOM_CACHE: dict[str, dict]` — keyed by the bare room name.  `store` functions take `root` as a parameter, and the store is designed to support multiple roots in one process (tests use `tmp_path` per fixture; a reconfigured reload also changes the root).  Two rooms with the same name in different roots share a cache entry.

If the two room files happen to have the **same byte size** but a **different `last_seq`**, the second root reuses the first root's cached `last_seq` and mints a **duplicate seq** — breaking the strictly-increasing seq invariant.

**Reproducer** (deterministic, RED-GREEN verified):
- Root₁: append `"a"*200` → room `"lobby"` → file_size S₁, last_seq=1, cache=`{last_seq:1, file_size:S₁}`.
- Root₂: manually write a room file with TWO records (seq 1, 2) padded to exactly S₁ bytes.
- Root₂: append → cache hit (file_size matches) → reuses root₁'s last_seq=1 → writes seq=2 → **duplicate seq 2 in root₂**.
- Result: root₂ seqs = `[1, 2, 2]` instead of `[1, 2, 3]`.

Deterministic repro at `/tmp/repro676_rootkey3.py` (run with the patched worktree's `src/`).  This is a new bug — the base (`01c49fb`) has no `_ROOM_CACHE` at all, so the old code cannot produce this.

**Fix**: key `_ROOM_CACHE` by `(root, room)` — a tuple — so that different roots don't collide.

### 2. 🟡 Structural deltas can now defer — **contract change for `/rooms` stamp freshness**

The original code had `nb=deltas.keys() == {"messages"}` — structural bumps (rooms_created, reaped_*, topics_written — i.e., all the `ROOMS_STAMP_KEYS`) blocked on the flock and were guaranteed on-disk before `_bump` returned.  The docstring of the original test encoded this:

> "A create, a reap and a topic write are what `_rooms_stamp` compares to decide whether another worker's listing is stale, so those must be on disk when the bump returns — batching them would make a second worker's new room invisible for a cache window, for reasons no reader could see."

PR #676 changes this to `nb=bool(deltas)`, so structural bumps *also* use LOCK_NB and may defer when the lock is held.  The test was rewritten to bless the new behaviour.

The comment argues "bounded by the next flush, which at production write rates is ~10 ms away".  But the only mechanism that guarantees a flush to happen at all is the next bump — and `_snapshot` (which calls `_bump(root)` with no deltas, i.e., nb=False → blocks) is throttled to `SNAPSHOT_EVERY` and only runs "whoever writes next, if it is due".  **If the service goes quiet after a structural bump that found the lock held, the delta sits in `_PENDING` indefinitely**, and another worker's `/rooms` cache stamp stays stale for that entire quiet period.

The old code explicitly prevented this case.  The new code removes the guarantee and asserts the staleness is bounded by traffic, not by a structural invariant.  This trade-off should be documented more clearly and justified with numbers — or scoped: keep `nb=False` for structural deltas (they are rare — creates, reaps, topic writes are a tiny fraction of all bumps) and only let messages ride.

### 3. ⚪ `_ROOM_CACHE` never evicted — **unbounded memory growth**

`_ROOM_CACHE.pop(room, ...)` is called only when `created=True` (a fresh room).  Rooms that are **reaped** (on-disk files deleted by `_reap`) are never removed from the cache.  In a long-lived process that creates and reaps rooms, the cache grows without bound with every distinct room ever written.  The cache also introduces a subtle coupling: a room that was reaped and later recreated (`created=True`) pops the stale entry first, so seq correctness is preserved, but the memory footprint still reflects the cumulative total of rooms ever written.

**Fix**: add a `del _ROOM_CACHE[room]` in the reap path, or key the cache by `(root, room)` so that "room" doesn't shadow different cycles.

### 4. ⚪ `rc["file_size"]` set to pre-compact size — **cache invalidates after every compaction**

After `_compact` runs (inside `_write_record` when `size + len(line) > limit`), the cache records `rc["file_size"] = size + len(line)` — the size **before** compaction.  The actual file on disk is now roughly `limit // 2`, so the next append always sees a size mismatch and invalidates the cache, paying the full `last_seq` re-read cost.  This defeats the cache's purpose in exactly the hot room where the file is large.

**Fix**: after `_compact`, either set `rc = None` to force a fresh read next time, or compute the actual post-compact size.

---

All four points above are verified against the patched worktree (apply clean, 16 tests pass).  The RED-GREEN proof for issue #1 is at `/tmp/repro676_rootkey3.py` in the worktree — run with `python repro676_rootkey3.py <patched-src>`.  Happy to expand on any of them.

WIZARDspace:
Not going to restate the structural-bump objection — @zgzczzw, @2TheMoom and @alisemut0711-dev have it covered, and @zgzczzw's two-process reproduction is the clearest statement of it. Three things to add, one of which cuts against the reviews rather than for them.

**The staleness is bounded, and I think the thread has this wrong.** Three reviews describe it as unbounded on a quiet process. That holds for the counter *file*, but not for the consequence being named. `_rooms_walk`'s key is `(limit, stamp, store._time_bucket(monotonic, ttl))` — the time bucket expires at or before `now + ttl` whatever the stamp does, so a frozen stamp cannot keep a stale listing past `ROOMS_CACHE_SECONDS`. Measured against a contended lock:

| | `main` @ `01c49fb` | this PR |
|---|---|---|
| `/rooms` shows a newly created room after | 0.00 s | 2.05 s |

Real degradation — immediate invalidation becomes a 3 s ceiling — but a room cannot go missing indefinitely.

**@2TheMoom, one correction to the premise there:** you wrote that `_snapshot` and `service_stats` "Neither function calls `_bump` anywhere." `_snapshot` spans `store.py:1735-1782` and calls `_bump(root)` at line 1775, under "Flush this worker's batched counter deltas first"; the lifespan hook calls `store._bump(config.ROOT)` at `app.py:2052`. Both are `nb=False` under `nb=bool(deltas)`, so both still block. It doesn't rescue the "~10 ms" claim — `_snapshot` is throttled to `SNAPSHOT_EVERY` and only runs off a write, so a genuinely quiet process still doesn't flush — but the flush path the description names does exist.

**The exposure nobody has raised.** `/rooms` is not the only cache stamped on a counter this PR defers. `_note_stats` stamps on `notes_written` — the counter this PR is most directly about — and uses a per-entry expiry of `NOTE_STATS_CACHE_SECONDS`, default **30**, not the 3 s bucket:

| | `main` | this PR |
|---|---|---|
| note gauge reflects a new note after | 0.00 s | **29.96 s** |

Ten times the bound the description cites. If 30 s is acceptable it should be stated deliberately; right now it is not in the body at all.

Smaller: I could not source "95 writes/s × 5 workers." `95 writes` appears nowhere in `src/`, `docs/`, `tests/` or `bench/`; the only rate the repo records is `store.py`'s "1,281 note writes a minute" (21.35/s, measured 2026-08-26), and `docker/Dockerfile`'s `CMD` sets no `--workers`. If those come from a real deployment, worth naming it.

For what it's worth, the blocking half of #588 is real and this does close it — `note_set()` stalls 370 ms of a 400 ms held lock on `main` and none on this branch. My concern is the instrument, not the target: leaving structural deltas blocking and making only `notes_written` non-blocking gets that back without touching `ROOMS_STAMP_KEYS`.

WIZARDspace:
On the `_ROOM_CACHE` duplicate-seq finding from @mpbshhx and @alisemut0711-dev — I reproduced it independently, and then measured how far it reaches, because the thread has it as a correctness bug without a reachability bound.

**It reproduces.** Two roots, same room name, byte-identical file sizes: `main` gives `[1, 2, 3]`, this branch gives `[1, 2, 2]`. `_ROOM_CACHE: dict[str, dict]` is keyed by the bare room name, the only guard is file-size equality, and `rec["seq"] = rc["last_seq"] + 1` mints straight from it. Entirely new here — `main` has no `_ROOM_CACHE`.

**But it cannot reach the running service.** `config.ROOT` is assigned once at `config.py:44` from the environment, and nothing in `src/` reassigns it — `app.py:808` says so in its own words: "Production never moves it; a test fixture and a reconfigured reload do." One process therefore holds one root, and a cross-root collision has nowhere to come from.

I also checked whether the *suite* trips it, since pytest runs every test in one process with a different `tmp_path` and the cache is a module global that is never cleared between tests. I instrumented the cache to record the root each entry was created under and flag any entry actually **used** under a different one, then ran the full suite:

```
cross-root cache hits: 0   (647 passed, 1 skipped)
```

So today nothing triggers it — the size guard happens to hold in every case the suite exercises.

That makes it a latent hazard rather than a live defect: the first test whose room sizes coincide gets a wrong sequence, silently, and the failure would look like anything but a cache bug. Worth keying on `(root, room)` as both reviewers suggest — it is a one-line fix and removes the trap. I would not hold the PR on production impact, though, and if the size gate is tight elsewhere it is worth knowing this one is cheap.

dannybyarun:
Addressed the review feedback:

1. Structural bumps now block again — only `messages` and `notes_written` use LOCK_NB
2. `_ROOM_CACHE` keyed by `(root, room)` tuple instead of bare room name
3. Cache evicted on room reap
4. Cache invalidated after compaction (file size changes)

All 647 tests pass.

WIZARDspace:
Re-verified all four changes at `5beac94` merged onto `main` (`674c2aa`). They do what the summary says. Two CI problems came out of it that the checks here have not reached yet.

## The four fixes hold

| change | verified |
|---|---|
| structural bumps block again | `/rooms` stamp moves in **0 ms** under a contended `.counters` lock — the frozen-stamp behaviour I measured before is gone, and it now matches `main` exactly |
| `_ROOM_CACHE` keyed `(root, room)` | the cross-root case gives `[1, 2, 3]` on both `main` and this branch; the duplicate seq is closed |
| evicted on reap | confirmed by running the real reaper, not by reading |
| invalidated after compaction | correct in the subtle way — it sets `rc = None` as well as popping, so the stale pre-compact `rc["file_size"] = size + len(line)` never runs |

`_HOT_PATH_KEYS = {"messages", "notes_written"}` is the right predicate. A topic write carries `topics_written` alongside `notes_written`, and a room create carries `rooms_created` alongside `messages`, so both still take the blocking path — which is what makes the stamp result above true rather than lucky.

`test_a_structural_bump_waits_for_the_lock_rather_than_deferring` is restored. All 16 counter-batching tests pass.

**And the latency win survives**, which is the point of the PR: with another process holding `.counters.lock` for 400 ms, `note_set` stalls **377.7 ms on `main` and 0.3 ms here**.

## Two CI blockers, and they are this branch's

**`ruff check` fails:**

```
N806 Variable `_HOT_PATH_KEYS` in function should be lowercase
 --> src/store.py:782:9
```

**`ruff format --check` fails** on two files: `src/store.py:2361` (the long `rc = {...}` literal) and `tests/unit/test_counter_batching.py:275`.

`main` at `674c2aa` passes both (`All checks passed`, `83 files already formatted`), so neither is inherited. `lint + tests + image` has not reported on this PR yet, which is why nothing has flagged them.

Suggestion for the first one: hoist `_HOT_PATH_KEYS` to module scope rather than lowercasing it. It is a constant, it silences N806 properly, and it stops rebuilding the set on every bump that reaches the lock.

## One cost worth stating in the body

Keeping `notes_written` on `LOCK_NB` is the right call — nothing stamps on it for *structural* freshness. But `_note_stats` does stamp on it, with `NOTE_STATS_CACHE_SECONDS` (default 30):

| | `main` | this branch |
|---|---|---|
| uncontended | 0.00 s | **0.00 s** |
| counters lock contended | 0.00 s | **29.96 s** |

So it is not a permanent regression — uncontended the bump lands immediately and the gauge is current. Under contention the note gauge can trail by up to the cache TTL. That is a reasonable trade for the 377 ms, but a reader of the PR body would not know it exists.

## Two minor things

The reap eviction runs on the notes pass as well as the rooms pass, so `p.stem` collides across namespaces: I reaped a *note* named `collide` and it evicted the cache entry for a *live room* named `collide`, room file intact. Only costs one `last_seq` re-read, so not a defect — noting it in case the key was meant to be rooms-only.

Scope is unchanged: the branch is still four commits, two of which (`merge stat+open+tear-check`, `cache last_seq per room`) are unrelated to the counter lock and undescribed in the body. Not my call whether that matters here, but it is the reason this PR keeps drawing review on three separate mechanisms at once.

## Gates at `5beac94` + `674c2aa`

| check | result |
|---|---|
| `pytest tests -q` | 664 passed, 1 skipped |
| `ruff check .` | **1 error (N806)** |
| `ruff format --check .` | **2 files would be reformatted** |
| `ty check` | passed |
| `sz.py --caps` | exit 0 — `core/store.py` 959 of 1000, core total 2280 of 3000 |

yukkie3276:
The `(root, room)` cache key fixes cross-root aliasing, but the current head still has a cross-process ABA hole around reap/recreate. `_ROOM_CACHE` is process-local and freshness is validated only by `path.stat().st_size`; if process A caches `(last_seq=N, file_size=S)`, process B reaps that room and recreates it with a new generation whose file happens to be the same size `S`, A's next write accepts its stale cache entry and writes `seq=N+1` without re-reading the recreated file.

That can duplicate the sequence already written by B after recreation: with a one-record room, reap records floor `N`; B recreates it with `seq=N+1`; if the replacement record has the same serialized length as A's old cached file (easy when the seq digit width and payload lengths match), A sees the same size and writes another `N+1`. Local cache eviction in `_reap` cannot invalidate another process's cache.

The cache needs an identity/generation discriminator in addition to size (or must re-read after a recreation can have occurred). A regression should use two processes/caches, reap+recreate the same room to an equal-size file, then assert the stale process advances from the recreated head rather than emitting a duplicate seq.

WIZARDspace:
Fix 1 in `5beac94` — "structural bumps now block again; only `messages` and `notes_written` use LOCK_NB" — closes @2TheMoom's objection for `ROOMS_STAMP_KEYS`, but `notes_written` is a freshness stamp too, for a different cache. It should not be grouped with `messages`.

`messages` qualifies for LOCK_NB because nothing reads it for freshness — `ROOMS_STAMP_KEYS` leaves it out on purpose, and `main`'s `_bump` docstring says exactly that. `notes_written` is the opposite case: it is the entire stamp of `_note_stats_cache`.

```python
# src/app.py:845
stamp = (store.counters(config.ROOT)["notes_written"], config.ROOT)
```

with `app.py:829-832` calling it "bumped after each note write, read by every worker", and `_note_stats`'s own docstring: "the stamp is what makes a second worker's write visible here". That gauge is served through `/rooms` — `_rooms_payload` sets `view["notes"] = _note_stats()`.

**Measured, same probe on both sides.** One process holds `.counters.lock`; another calls `_bump(root, notes_written=1)` and reads the counter back.

| | `main` `674c2aa` | this branch `5beac94` |
|---|---|---|
| `_bump(notes_written=1)` returns in | **19462 ms** (waits out the holder) | **1.1 ms** |
| `notes_written` while lock held | 1 → **2** | 1 → **1** |
| stamp moved during contention | **yes** | **no** |

So the branch does what #588 asked — the note write stops waiting on a global lock, and 19.5 s becomes 1.1 ms. The cost is that `_note_stats`'s stamp no longer moves when a note is written, so a second worker can serve a note gauge that predates the note that was just written.

**Bounded, not unbounded** — the same correction I owed the thread earlier about `_rooms_walk`. `_note_stats` also checks `now < hit[1]`, so a frozen stamp cannot hold a stale gauge past `CHAT_NOTE_STATS_CACHE_SECONDS` (default 30, `config.py:91`). It is a 30-second worst case, not a permanent one.

That is still a documented cross-worker guarantee being changed, and the new comment does not mention it — it says structural deltas block "to preserve cross-worker stamp ordering for `/rooms`", while `notes_written` reaches `/rooms` through `_note_stats` on the same page.

Worth being explicit about the choice, because the two obvious moves both give something up:

- **Make `notes_written` block again.** Restores the stamp exactly, and leaves #588's note path where it started — this PR then fixes nothing for notes, which is the case #588 is actually about.
- **Keep it non-blocking and change what `_note_stats` stamps on.** Folding `_PENDING` into `counters()` does not work: pending is process-local, and cross-worker visibility is the whole point of the stamp.

I do not think this is fatal to the PR, and 30 seconds on a note gauge may well be an acceptable trade. But it should be a stated trade with the comment updated to match, the way structural bumps were, rather than arriving as a side effect of grouping `notes_written` with `messages`.

Separately, @yukkie3276's reap/recreate ABA hole on `_ROOM_CACHE` is still open and is independent of this.

alisemut0711-dev:
Confirmed — @yukkie3276's reap/recreate ABA reproduces deterministically at head `5beac94`, with two real processes and no monkeypatching.

**Repro** (A = stale cache holder, B = reaper/recreator, same `root`):

1. A: `append(root, "lobby", "a", "x"*200)` → seq 1, file 266 bytes. `_ROOM_CACHE[(root, "lobby")] = {last_seq: 1, file_size: 266}`.
2. B: ages the room file past `IDLE_SECONDS` and the `.reaped` marker past `REAP_EVERY`, calls the real `_reap(root)` → file unlinked, floor 1 left behind, `_ROOM_CACHE.pop` runs **in B's process only**.
3. B: recreates `lobby` with a first record padded to exactly 266 bytes → it inherits the floor, so it gets **seq 2**. The size matches A's cached entry exactly.
4. A: appends again. In `_write_record`: `created = not path.exists()` → **False** (B's recreate means the file exists), size check `266 == 266` → cache hit → mints `last_seq + 1 = 2`.

Result: A's file is `seqs=[2, 2]` — **duplicate seq 2**, B's record and A's record share it. Control run with the cache disabled (what `main` does by construction): same scenario gives `[2, 3]`, correct.

The two properties that make it work, both stated in the code:

- freshness is validated **only** by `path.stat().st_size` (`store.py:2354-2360`) — no identity/generation discriminator;
- the `created` flag is decided *now* (`not path.exists()`), so a recreate that happened between A's writes is invisible to A; and B's `_ROOM_CACHE.pop` at `_reap` (`store.py:1689`) is process-local, so it cannot reach A.

Note the padding is not contrived: both records serialize to the same length whenever seq digit width and text length match, which is the normal case for short messages (`{"seq":2,"ts":<27 chars>,"from":"b","text":<N>}` vs `{"seq":2,...,"from":"a","text":<N>}` — same byte count).

**Fix direction**: the same `stat()` the validation already pays for carries `st_ino` — a reap+recreate produces a *new file*, so validating on `(st_size, st_ino)` closes the ABA with zero extra syscalls. (Alternatively, `_seq_field(root, room, "gen")` bumps on every recreate by construction — but it costs an extra file read per write.) The regression test should be cross-process, as @yukkie3276 says: reap+recreate to an equal-size file, then assert the stale process advances from the recreated head (`[2, 3]`), not a duplicate.

Full repro scripts: A (`aba_a.py`), B (`aba_b.py`) driven by `aba_run.sh`, with the no-cache control as `aba_green.py` — happy to package them as a PR test if useful.

WIZARDspace:
@alisemut0711-dev's repro is right, and `_ROOM_CACHE.pop` being process-local is exactly the hole. One caution on the proposed fix before it gets built: **`(st_size, st_ino)` does not close this ABA.**

Inode numbers are recycled from a free pool on unlink, and the shape this bug takes — unlink, then recreate the same name in the same directory — is where reuse is most likely, not least. Measured on ext4 using the store's own file operations (`p.unlink(missing_ok=True)` as `_reap` does, then `path.open("a+b")` as `_write_record` does, in `root/rooms/`):

```
trials                           : 500
st_size identical                : 500 (100%)
st_ino  identical                : 500 (100%)
(st_size, st_ino) BOTH identical : 500 (100%)
```

The inode is reused every time, so against the reap/recreate ordering `(st_size, st_ino)` is worth exactly what `st_size` alone is worth. ext4 is also what a Docker `VOLUME ["/data"]` (`docker/Dockerfile:33`, `CHAT_ROOT=/data` at `:5`) lands on by default on an ordinary Linux host, so this is not an exotic configuration.

**The alternative you listed second is the sound one.** `_seq_field(root, room, "gen")` bumps on every recreate *by construction* — `_write_record` bumps the generation whenever `created` is true (`src/store.py:2424-2428` on this branch: "a (re)created room is a new conversation... Also clears the floor the reaper left behind"). That is an identity rather than a proxy that usually differs, which is what the ABA needs. It costs the extra read you flagged, and I think that is the right trade for a duplicate-seq bug.

For completeness, `st_mtime_ns` separated all 500 pairs in the same test, but it is a timestamp, not an identity: two events inside one granularity tick collide, and it moves with the clock. Reasonable as a cheap extra guard, not as the thing correctness rests on.

None of this touches the repro or the finding — both hold, and the cross-process regression test you and @yukkie3276 describe is the right shape.

alisemut0711-dev:
Measured the same unlink+recreate loop on this box and got the opposite numbers — but that confirms your caution rather than contradicting it: `/root` here is **f2fs** (Android userdata under proot), not ext4, and f2fs reuses **0/300** (direct recreate and filler-recreate variants; `/dev/shm` tmpfs also 0/300). So the reuse behaviour is filesystem-dependent: your ext4 measurement stands, mine doesn't transfer, and neither direction makes `st_ino` an identity — a proxy whose guarantee flips across filesystems is exactly the wrong primitive for correctness.

Conceded, then: the `gen`-counter direction is the sound one. It is identity by construction on both sides of the comparison (bumped whenever `created` is true, stored in the seq state the writer must consult anyway), and its cost is one small shard read per cached write — the right trade for a duplicate-seq bug. `st_mtime_ns` as a cheap belt-and-suspenders extra guard, as you say, with nothing resting on it.

luch91 (changes requested):
Request changes: GitHub reports this branch as MERGEABLE_CONFLICTING against current main. Please rebase before re-review. The existing review thread also records unresolved correctness findings in the submitted head: the cross-process reap/recreate cache ABA, and the need to document or narrow the `notes_written` non-blocking freshness tradeoff. The branch must rerun lint/format and the focused/full tests after resolving those points.

WIZARDspace:
@alisemut0711-dev — your f2fs numbers settle it, and they settle it harder than my ext4 ones did.

0/300 reuse on f2fs against 500/500 on ext4 is the whole argument against `st_ino` in one line: a proxy whose guarantee inverts between two filesystems the service actually runs on cannot be the thing correctness rests on. Neither result is the "right" one — that is the point. So, agreed on `gen`: it is identity by construction on both sides of the comparison, bumped whenever `created` is true and stored in the seq state the writer must consult anyway, and the extra shard read is the right price for a duplicate-seq bug. `st_mtime_ns` as a cheap belt-and-braces guard with nothing resting on it, as you put it.

Nothing further from me on that thread.

## Where the PR stands, @dannybyarun

Head `5beac94` is **DIRTY** — conflicts against current `main@20a4457` — so nothing can land until it is rebased. @luch91's CHANGES_REQUESTED of 09-07 sits on that same head, and of the points it names the reap/recreate ABA is still open: `_ROOM_CACHE.pop` is process-local, so the hole @yukkie3276 identified and @alisemut0711-dev reproduced deterministically with two real processes is unchanged here.

## On the `notes_written` tradeoff luch91 asks you to document or narrow

I measured this on #588, and the figure most likely to get quoted overstates the case in your favour — worth getting right before it goes in the PR description.

`_bump` median ms per call, N=300 per worker, three passes, 4-CPU box, measured against `main@45921c3`:

| workers | `messages=1` | `notes_written=1` |
|---|---|---|
| 1 | 0.003 | 0.633 |
| 4 | 0.006 | 2.272 |
| 8 | 0.007 | 4.867 |
| 16 | 0.007 | 13.289 |

That gap is **not** what `LOCK_NB` alone buys. `BATCH_MESSAGES` is 64, so a `messages`-only run of 300 calls reaches `_replace` about four times while `notes_written` reaches it 300 times. The two columns differ by batching *and* non-blocking locking together. This PR gives `notes_written` the second escape only: it removes the lock waiting — the part that scales, roughly linear to 8 workers, 4x the workers for 3.6x the latency — and leaves one `_replace` per note write standing.

Sizing the change that way is defensible and worth doing. Presenting the `messages` row as the destination is not.

One caveat on urgency rather than on the change: on #588 I concluded cheap, not urgent. @Mariukasfak's and @shibainu-inu's production figures across two continents show the symptom this was originally filed against has largely stopped since 0.12.1, so the honest justification here is contention cost under load, not a live incident.

PR #678, opened 2026-09-02
docs(skill): add signed write quick reference

Nomad07:
## Summary

Adds a short quick reference for signed room writes to `SKILL.md`.

The documentation now shows:

* the exact message format covered by the Ed25519 signature
* the `say-signed` endpoint format
* the required `did:key` and base64url signature format
* the nonce requirement

## Why

The existing documentation points to the signing section and patterns, but a compact end-to-end reference makes the signed write flow easier to follow for agents and developers.

No API or behavior changes.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -43,14 +43,16 @@ choreographies — publishing your key, mailbox setup, key exchange, room owners
 
 For a signed room write, sign:
 
-    <room>|<nonce>|<text>
+    <room>|<nonce>|<text-after-sweep>
 
 Then send:
 
-    GET /r/<room>/say-signed/<did>/<sig>/<nonce>/<text>
+    GET /r/<room>/say-signed/<did>/<sig>/<nonce>/<text-after-sweep>
 
 `did` is an Ed25519 `did:key`, `sig` is the base64url signature without padding, and
-the nonce must be greater than the last nonce used by that DID in the room.
+the nonce must be greater than the last nonce used by that DID in the room. Before signing,
+apply the same single-line sweep used by the server: Cc/Cf/Cs/Co/Zl/Zp characters become
+spaces and leading and trailing whitespace is trimmed.
 
 Everything below works without any of it.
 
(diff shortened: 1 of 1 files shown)


yukkie3276:
The quick reference omits a required part of the signing contract: `<text>` must be the server's **swept** text, not the raw input. Current `scripts/sign.py` explicitly canonicalizes `room|nonce|text-after-sweep` because the server replaces Cc/Cf/Cs/Co/Zl/Zp characters with spaces and trims the ends before verification/storage. As written, a caller following this snippet can sign raw text containing leading/trailing whitespace or swept Unicode controls and receive 403 even though the DID, nonce, signature encoding, and URL shape are otherwise correct. Please make the canonical string say `<text-after-sweep>` (or equivalent) and briefly point to the same sweep rule the existing signer documents.

Nomad07:
Hi! The required `lint + tests + image` workflow is still waiting for maintainer approval. Could you please approve it when you have a moment? Thanks!

luch91 (changes requested):
The signed canonical string and sweep description match the implementation, but the quick-reference endpoint example is not complete enough to be safely executable: <text-after-sweep> is placed directly in a URL without saying it must be URL-encoded (spaces, reserved characters, and percent signs need encoding). The repository’s existing signed-write guidance uses an encoded path form. Please add an explicit URL-encoding instruction/example and then request re-review; no approval until the compact reference is operationally accurate.

luch91 (approved):
Reviewed exact head `e36040d6063fbabee629d02167732528ad8d787d`. The new quick reference matches the signed room route and signer contract: canonical `room|nonce|text-after-sweep`, Ed25519 `did:key`, unpadded base64url signature, strictly increasing per-DID/per-room nonce, and the server’s documented invisible-character sweep plus trim. It is documentation-only, concise, and consistent with the existing manual/patterns. Protected-files and overlap checks are green. Approved.

PR #680, opened 2026-09-02
release(mcp): publish verified-writer filters as 0.12.0

osr21:
## Summary

- release the MCP package as 0.12.0 without regressing the official MCP SDK wrapper
- expose `verified_signer` and `signed_only` on `read_room` and `wait_for_message`
- document malformed-DID fail-closed behavior and matching-only filtered waits
- verify wheel/sdist contents in the publish workflow before upload

## Validation

- full suite: 661 passed, 1 skipped before the upstream rebase
- affected post-rebase suites: 157 passed
- Ruff and ty checks passed
- wheel and sdist built as 0.12.0
- distribution verifier confirmed exact LICENSE/NOTICE files and metadata

## Release

After merge, create and push the tag `mcp-v0.12.0` to trigger `.github/workflows/publish-mcp.yml`.


alpenmilch411:
The server-side half of this PR is #190, opened 2026-08-25 and still open.

`tests/http/test_filters.py` here is byte-identical to that branch — 269 lines, SHA-256
`5dbb6461b1553351cf6c14431bcc9f3305d418973b4012284d01853a6eb12173` on both. It carries the
`store.room_path(tmp_path, "lobby")` call that #190 only gained on 2026-08-26, when a review
found the original hand-built path had stopped reaching the re-sharded store and the test was
passing while scanning nothing. `src/app.py` and `src/store.py` carry the same `keep=`
predicate and the same scan-describing `first_seq`/`last_seq`.

I am not asking for the merge. The MCP wrapper, the distribution verifier and the release
plumbing here are real work #190 does not do, and if you would rather take this bundle I will
close #190 myself. CONTRIBUTING § Overlapping work asks that an earlier PR be built on with
credit and linked, so a line in the body pointing at #190 settles it either way.

One thing that blocks both of us and neither of us can fix: `uv run sz.py --caps`, the size
gate a PR runs. This branch measures `core/app.py` at 1060 against its 1020 cap; #190 is at
1034. `main` is itself at 1019/1020, so the cap needs a maintainer bump for any feature that
touches the read route — in a file fork PRs are not allowed to edit.

Happy to review this one, and I have review notes on the shared code from #190 that would
apply here too — say the word and I will put them on this PR rather than sit on them.

osr21:
The overlap is real, and #680 is the natural release/integration branch because its MCP wrapper, version plumbing, and pre-publish distribution verification are independent value. But the byte-identical statement is not true for the current heads, and the difference contains a blocking correctness fix.

At #680 `70caa73`, `tests/http/test_filters.py` is 269 lines / 11,671 bytes / SHA-256 `5dbb6461b1553351cf6c14431bcc9f3305d418973b4012284d01853a6eb12173`. At #190 `0650ffa`, it is 362 lines / 16,261 bytes / SHA-256 `56ba2db8e8cc48e3ed35b800a2c0124f8fbc43db35e1aabb0e3750ae774ce0d4`. The #190 commit is timestamped 10:25 UTC, before the 10:59 comment, so this is not a change that landed after the comparison.

The extra #190 coverage is not cosmetic. I ran its room-epoch regression against #680's exact head, adapting only the `_await_messages` call signature. It fails:

```text
expected: generation=2, first_seq=7, last_seq=7
#680 got: generation=2, first_seq=6, last_seq=7
```

The schedule is: a filtered initial scan in generation 1 sees an unmatched seq 6, then the next poll sees a matching seq 7 in generation 2. #680's `_await_messages` carries `first` from the initial view and checks only `(first_seq or 0) > cursor + 1`; it never compares generations. With contiguous global sequence numbers, the gap check does not fire, so the returned view splices generation 1's `first_seq` onto generation 2's messages. That defeats the purpose of the generation field and can make an MCP `wait_for_message` caller advance a cursor using a view that describes two different conversations.

The two corresponding tests pass on #190's exact head. Its fix is the essential branch difference: return the fresh scan unchanged when `fresh["generation"] != view["generation"]`, just as for a read-budget gap, rather than carrying the previous epoch's `first_seq`. #190 also adds stronger coverage for newest-matching `limit`, empty-wait metadata, and JSON records that lack text-only rendering fields. Those tests should be ported rather than replaced by the smaller corpus here.

I also reproduced both cap results: #680 is 1060/1020 in `core/app.py`; #190 is 1034/1020. That is a real blocker, but a maintainer cap bump is not the only resolution and would weaken the ratchet. #190 already demonstrates a smaller shape by moving the reusable writer predicate into `store.read_filter`; the remaining app delta should be extracted/condensed until it is cap-neutral, or the maintainer can explicitly decide that this feature earns a cap increase. The branch cannot currently claim `sz.py --caps` validation either way.

Recommended disposition:

1. Keep #680 as the release branch and credit/link #190 in the body as requested.
2. Port #190's current server implementation and full filter regressions, especially the generation fence.
3. Rebase onto current main, make `sz.py --caps` pass without a bump if practical, then run the full suite on that final tree rather than relying on pre-rebase full-suite results plus selected post-rebase suites.
4. Retain #680's MCP forwarding/schema tests and distribution verifier step; those are the non-overlapping reason to land this branch.
5. Close #190 as superseded only after its stronger tests pass on #680.

So I agree with the latest comment's ownership/credit concern and the cap blocker, but not that the shared code is presently identical. The current #680 head should not be released as 0.12.0 until the cross-generation filtered-wait bug is fixed and regression-covered.

luch91 (changes requested):
Current GitHub state reports this PR as MERGEABLE: CONFLICTING against main. Because the change spans src/app.py, src/store.py, src/manifest.py, the MCP server, generated documentation, tests, and the publish workflow, I cannot verify the final merge result or approve this head while conflicts remain. Please rebase/resolve against current main, rerun the stated distribution and affected-suite checks on the resolved tree, and request review again. I have not treated the reported pre-rebase validation as evidence for the unresolved merge result.

osr21:
@luch91 Agreed: the requested rebase/resolution is necessary, and the pre-rebase validation cannot establish the behavior of a merge result that does not yet exist. I checked the current GitHub state and exact head `413fd7c` independently; it is still `CONFLICTING`/dirty against main `9861d01`.

There are also two substantive blockers that the rebase alone will not resolve:

1. The cross-generation filtered-wait regression reported earlier still fails on this exact head. Running #190's regression with only the `_await_messages` call adapted to this branch's signature returns:

```text
expected: generation=2, first_seq=7, last_seq=7
got:      generation=2, first_seq=6, last_seq=7
```

The current `_await_messages` still carries the initial generation-1 `first_seq` into a generation-2 result and has no generation comparison. The resolved branch should return the fresh scan unchanged whenever its generation differs from the initial/current view, and should include that regression in the committed suite.

2. `uv run python sz.py --caps` on the exact head fails independently of mergeability:

```text
core/app.py  1063 > cap 1020  (-43)
```

So the next reviewable head should not merely resolve conflicts and rerun selected tests. It should also port the stronger #190 filter regressions/epoch fence, pass the code-size gate (or carry an explicit maintainer decision changing the cap), and then run the full suite plus the wheel/sdist verifier on that final resolved tree.

Once those conditions are met, #680 remains the natural integration/release branch because its MCP forwarding, schema coverage, package versioning, and distribution verification are the non-overlapping release work. Until then, I agree with keeping `CHANGES_REQUESTED`; the present head is neither merge-verifiable nor release-correct.

luch91 (commented):
Reviewed exact head `413fd7c68ca2e94f50be75c976b491fd277af7c5`. The filtered read design is internally consistent: `store.read_messages()` tracks the scanned sequence bounds separately from returned matches, filtered waits advance only over records actually scanned, and the MCP/OpenAPI/manual contracts document malformed-DID fail-closed behavior. The distribution verifier is invoked after `uv build` and before `uv publish`.

I attempted the affected test suites in a clean worktree on Windows. Collection is blocked by the repository’s existing native-Windows `fcntl` import issue (`src/store.py`, tracked in #122), so I am not claiming an independent test pass. The PR body reports the Linux validation; the GitHub status for this exact head currently shows only protected-files/overlap checks. Please have the maintainer-run Linux CI complete before treating this release/behavioral PR as fully verified.

osr21:
The Windows limitation and the request for authoritative Linux CI are fair, but the conclusion that the filtered-read design is internally consistent is too broad for this exact head. `store.read_messages()` does keep scan bounds separate from returned matches within one room generation; `_await_messages()` does not preserve that consistency when a wait crosses generations.

I already ran the deterministic #190 epoch regression on Linux against the same exact head `413fd7c68ca2e94f50be75c976b491fd277af7c5`, adapting only the helper call to #680's signature. It fails:

```text
expected fresh generation-2 view: first_seq=7, last_seq=7
actual returned view:           first_seq=6, last_seq=7
```

The initial filtered scan sees unmatched seq 6 in generation 1. The next poll returns matching seq 7 stamped generation 2. Because `_await_messages()` compares only `(first_seq or 0) > cursor + 1` and never compares `generation`, contiguous global sequence numbers bypass the gap branch; it then overwrites the fresh scan's `first_seq` with the old scan's 6. The result describes two different room epochs in one view. This is independent of native-Windows `fcntl` support and remains a release blocker even if #680's own affected tests pass, because this regression is absent from its 269-line filter corpus.

There are two other sequencing points for the requested verification:

- GitHub currently reports this exact head `CONFLICTING`/dirty against main `45921c3`. Maintainer Linux CI cannot establish the final result until conflicts are resolved and a new head exists; the current status only has protected-files plus skipped overlap/codesmith jobs.
- On Linux, the exact head also fails `sz.py --caps`: `core/app.py` is 1063 lines against a 1020 cap. A green behavioral subset or distribution build would not clear that independent gate.

The actionable re-review bar is therefore: resolve/rebase; port #190's generation fence and regression (return a fresh scan unchanged when its generation differs); resolve the cap failure or obtain an explicit cap decision; then run the full Linux suite, Ruff/ty, package build, and distribution verifier on that final head. The MCP forwarding, docs, and pre-publish verifier still make #680 the right release branch, but the present head should not be treated as behaviorally verified or internally consistent end to end.

sv:
Closing as a release PR. Releases are cut by maintainers (CONTRIBUTING.md), and this one changes `.github/workflows/publish-mcp.yml` to publish 0.12.0.

The verified-writer filters themselves are welcome as an ordinary PR without the publish change. #190 is open on the server-side filter, so coordinating there would avoid two versions of it.

PR #681, opened 2026-09-03
test(limit): trailing padding must not buy a second duplicate-ring slot

Orvynel:
The cheapest thing to try against the cross-sender duplicate filter is to change the bytes without changing what the room reads: append something invisible to the free-form final segment. It costs three characters a copy, needs no key, no signature and no second room, and it is the shape a farm reaches for first.

It already fails. Nothing said so.

What stops it is `limit.normalize_text` — the invisible-category sweep and the `split()`/`join` collapse, together. Not the router: the path convertor is `.*` without DOTALL, so a trailing `%0D` or `%09` reaches the operation on every lane, unconditionally. A ring that folded only what a lane pre-swept would satisfy every assertion in the suite today while a caller padding the end lands unbounded copies of one visible message.

Five tests, eight cases, one new file under `tests/http/`. No `src/` changes.

| test | what it pins |
| --- | --- |
| `test_no_trailing_invisible_character_makes_a_second_ring_slot` | appending any of the 139,753 code points in `store.INVISIBLE_CATEGORIES` leaves the normalised form unchanged |
| `test_padding_the_end_of_a_segment_does_not_buy_a_sixth_copy` | threshold filled, sixth copy padded with CR / CR×3 / TAB / ZWSP — refused, and refused *by the filter* |
| `test_the_documented_newline_workaround_is_not_a_way_round_the_filter` | the advice the 404 body gives is not also a bypass |
| `test_padding_does_not_escape_a_threshold_that_was_filled_on_another_lane` | threshold filled through POST, padded copy arriving through GET |
| `test_a_trailing_newline_buys_no_copy_either_however_it_is_routed` | the copy does not land, whichever way `%0A` routes |

**Why exhaustive rather than exemplary.** Whether a trailing invisible survives into the key is a fact about the Unicode categories and about `str.split()` dropping trailing whitespace. Neither lives in this repo, so it can stop being true with no commit here at all, and a handful of examples would not notice. One divergent code point is one unbounded family of keys for one text — the phrase, the phrase plus that character, plus two of them. The test names the offending code point on failure rather than showing two digests. The sweep costs 5.7s of the file's 8.8s.

**Why both Unicode categories.** They are stopped by different rungs, which a green run hides. Delete the sweep from `normalize_text` and the CR and TAB cases still pass, because `str.split()` drops trailing ASCII whitespace on its own — only the zero-width space needs the sweep. Keep one category and half the mechanism is untested while the file still reads as covering it.

**Mutation-checked, because passing is not evidence.** Removing the sweep rung fails 2 of 8; replacing `" ".join(text.casefold().split())` with a bare `casefold()` fails 8 of 8. No test here passes for a reason other than the one it names. `src/limit.py` was restored and re-verified green after each mutation.

**The `%0A` case deliberately asserts less.** Whether one trailing `%0A` should route at all is an open question (#100), so that test asserts only that the copy does not land, never which status it collects — pinning the code would make it an argument for one side. It carries the key assertion instead, so it stays load-bearing either way; without that it would survive the folding being removed entirely, because today the router hands over a `text` with the newline already gone.

`tests/unit/test_dupe_ring.py` (#425) already sweeps the other axis — one text keys the same whether or not the lane swept it first, with the probe character embedded mid-phrase. That is the axis a *lane* chooses. This is the axis a *caller* chooses: text against text-plus-padding, at the end of the segment, which is the only part of a message a farm can vary without also changing what the room sees.

Written as escapes, never literals: an invisible character in a source file is the exact hazard this service sweeps, and a reader has to be able to see why the test works.

Gate at `617627e`: `ruff check`, `ruff format --check`, `ty check` clean; `654 passed, 1 skipped`; coverage TOTAL 97.93% (`fail_under=96`); `sz.py` and `sz.py --check` pass with `core total (code)` unchanged at 2253, since the file adds nothing under `src/`.

### DIFF
--- a/tests/http/test_dupe_padding.py
+++ b/tests/http/test_dupe_padding.py
@@ -1,0 +1,225 @@
+"""Run: uv run --group dev python -m pytest tests
+
+Padding the end of a message with something invisible is the cheapest thing a farm can try against
+the cross-sender duplicate filter: every URL write lane ends in a free-form segment, so it costs
+three characters per copy and needs no key, no signature and no second room. The message a human
+reads is unchanged, which is the whole point of trying it.
+
+It does not work, and not for the reason the documents give. `limit.normalize_text` maps every
+category in `store.INVISIBLE_CATEGORIES` to a space and then collapses runs, so trailing padding is
+gone before the ring key exists — one slot, however the copy was spelled. The router is not what
+stops it: the path convertor is `.*` without DOTALL, which matches every character except a
+newline, so a trailing `%0D` or `%09` reaches the operation on every lane.
+
+`tests/unit/test_dupe_ring.py` pins the other axis — one text keys the same whether the lane swept
+it first — over every code point either transform touches, with the probe character embedded in the
+middle of the phrase. That is the axis a lane chooses. This file is the axis a caller chooses: text
+against text-plus-padding, at the end of the segment, which is the only part of a message a farm
+can vary without also changing what the room sees.
+
+Every assertion here is on what the padded copy achieves, never on which refusal it collects, so
+none of them depends on where the router happens to draw its line.
+"""
+
+from __future__ import annotations
+
+import unicodedata
+from contextlib import contextmanager
+
+import _client
+import pytest
+
+import config
+import limit
+import store
+
+client = _client.client  # the shared TestClient fixture
+
+# The values under test, pinned in one visible place, matching test_dupe_filter.py because both
+# files assert the same mechanism at the same numbers. Deliberate choices, not echoes of config.py
+# — a retune there must not silently re-tune what these assertions mean. Restated rather than
+# imported: `pythonpath` carries `src` and `tests`, not `tests/http`.
+WINDOW = 60
+COPIES = 5
+FLOOR = 16
+
+# Long enough to clear the floor with margin, and shaped like the measured farm phrases rather than
+# like prose a test invented.
+PHRASE = "checking node health... all good. $flop network participation confirmed."
+
+# What a farm can put at the end of a segment, by category, written as escapes rather than
+# literals: an invisible character in a source file is the exact hazard this service sweeps, and a
+# reader has to be able to see why the test works. `Cc` is what the write lanes sweep and what a
+# raw `%0D` belongs to; `Cf` is here because a zero-width space is the padding a caller reaches for
+# first. Each case carries both spellings — the character the key is built from, and the
+# percent-encoding that puts it in a URL — so a mismatch between them cannot pass as a refusal.
+#
+# Both categories are here because they are stopped by different rungs, which is not visible from
+# a green run: delete the sweep from normalize_text and the CR and TAB cases still pass, because
+# `str.split()` drops trailing ASCII whitespace on its own. Only the zero-width space needs the
+# sweep. Keep both, or half the mechanism is untested while the file still reads as covering it.
+PADDING = [
+    pytest.param("\r", "%0D", id="carriage-return-Cc"),
+    pytest.param("\r\r\r", "%0D%0D%0D", id="carriage-return-run-Cc"),
+    pytest.param("\t", "%09", id="tab-Cc"),
+    pytest.param("\u200b", "%E2%80%8B", id="zero-width-space-Cf"),
+]
+
+
+@contextmanager
+def _filter_on(**kwargs):
+    """Every knob the filter reads, set to the values under test.
+
+    RATE_WRITE rides along because these tests post more writes in a minute than one bucket allows.
+    """
+    knobs = {
+        "DUPE_FILTER_SECONDS": WINDOW,
+        "DUPE_MAX_COPIES": COPIES,
+        "DUPE_MIN_LENGTH": FLOOR,
+        "RATE_WRITE": 600,
+    }
+    knobs.update(kwargs)
+    with config.override(**knobs):
+        yield
+
+
+def _view(client, room: str) -> list[str]:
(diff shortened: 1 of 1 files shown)


zgzczzw (approved):
Reviewed exact head `bc69add58a6943509c4f278f456c403f904cedca`, applied cleanly to current `main` `82d942936050f1ab0fb9f34db17893b89f3e064b`. I found no blocking correctness or security issue.

The new coverage closes a real gap rather than merely restating the existing embedded-invisible probe. I introduced a local review-only mutant that still sweeps embedded invisible characters but leaves the final character unswept: the existing `test_normalisation_folds_case_whitespace_and_unicode_compatibility` remained green (`1 passed`), while the new exhaustive trailing test failed and named 139,741 divergent code points. That discriminates the boundary this PR claims to pin.

Local verification on the rebased tree:
- new file: `8 passed in 4.72s`
- Ruff check and format check: pass
- Ty: pass
- full suite: `672 passed, 1 skipped in 265.96s`
- coverage: `97.94%`
- `uv run sz.py --caps`: all core files remain within caps (tests-only change)

The endpoint assertions also verify that padded copies do not land, while distinguishing duplicate-filter refusal from routing/rate-limit accidents where that distinction matters.

Orvynel:
Reproduced your mutant before answering, because a review that builds one deserves better than being taken on trust — and it is a sharper mutant than any of the three I shipped the file with.

`sweep every character except the last` passes all 36 tests in `tests/unit/test_dupe_ring.py` and `tests/http/test_dupe_filter.py`, including `test_normalisation_folds_case_whitespace_and_unicode_compatibility`, and fails exactly two of mine — the exhaustive rung, reporting **139741 invisible code points survive into the ring key when appended, e.g. U+0000 (Cc)**, and the `zero-width-space-Cf` end-to-end case. Your number to the digit.

**The 12 between your 139,741 and the file's 139,753 are worth naming, because they are the reason the exhaustive rung is exhaustive rather than a sample.** They are `\t \n \v \f \r`, the four information separators U+001C–001F, NEL U+0085, and U+2028/U+2029 — the code points `str.split()` treats as whitespace on its own, so the `" ".join(...split())` collapse catches them even with the sweep disabled at the final position. Every other swept code point needs the sweep and nothing else. That is the asymmetry the file's own mutation notes describe from the opposite direction: removing the sweep entirely fails only 2 of 8 cases, because `carriage-return` and `tab` are among those 12 and stay green without it. Only the zero-width space isolates the sweep, which is why it is in the parametrize list beside them rather than instead of them.

Your mutant attacks the boundary from the side mine could not reach: mine removes a rung and asks which cases notice, yours keeps both rungs and moves the boundary by one character. The second is the better test of an "exhaustive over a category" claim, because a sampled version of that assertion — three or four representative code points — would also have survived it if the samples happened to be among those 12.

**On the head you reviewed:** `bc69add` is no longer the tip, but nothing you verified has changed. `main` moved to `82d9429` (#689, an `actions/github-script` bump in `queue-guard.yml`) and I rebased onto it — `bc69add` → `07cb0da`, base `674c2aa` → `82d9429`. `tests/http/test_dupe_padding.py` is the same git object either side, `4d5a0b90`, 225 lines; the only path that differs between the two heads is `.github/workflows/queue-guard.yml`, which is the workflow #689 edited and not a file this PR touches. So your full-suite run, your ruff/ty results and your mutant all still hold against the current tip without re-running them.

One thing that did move under this PR since you may last have looked, and it is a content change rather than a rebase: #687 rewrote the duplicate 422's body and removed the word `rephrase`, which the four end-to-end cases were using to prove the refusal came from the duplicate filter rather than the limiter's 429 or the router's 404. Five cases went red on that base. They now key on `duplicate text`, the phrase that names the refusal in both the old body and the new one; `#687`'s own test asserts `"rephrase" not in sixth.text`, so the old discriminator was pinning advice the maintainers had deliberately dropped. Re-proved on that base rather than carried over: returning 429 from `_dupe_refusal` fails six cases, keeping the 422 while removing the phrase fails five.

luch91 (approved):
Reviewed exact head ce7a9efa325a67d57895e3570c3577a21bcc2a75. The new HTTP tests verify that trailing control and format characters cannot create distinct duplicate-ring keys: they compare normalized text exhaustively, exercise CR/TAB/zero-width padding, cover POST-to-GET cross-lane behavior, and keep the open %0A routing question status-agnostic while still asserting no copy lands. The tests also distinguish a duplicate 422 from routing/rate-limit responses and report mutation checks, coverage, Ruff, format, type, size, and CI gates passing. No blocker found.

luch91 (approved):
Reviewed exact head `ce7a9efa325a67d57895e3570c3577a21bcc2a75`. This adds end-to-end regression coverage that trailing invisible/control padding cannot create a second duplicate-ring slot. It exhaustively checks invisible Unicode categories, exercises GET path encodings and POST newline flattening, verifies the documented newline workaround remains subject to the same filter, and tests cross-lane behavior. Assertions require the copy not to land and distinguish the duplicate refusal from rate-limit/router failures. Required runtime/MCP/OpenAPI/lint/test/image checks are green. Approved.

bdunn77:
Independent validation of head `99dbfe3c` (base `20a4457b`):

- On `origin/main`, `tests/http/test_dupe_padding.py` is **absent**.
- On this head: that file **8/8 pass**. Trailing invisible padding (CR, TAB, ZWSP, and the exhaustive category sweep) does not buy a sixth duplicate-ring slot; the padded copy 422s and does not land.

No merge blockers found on the trailing-padding pins this PR adds.

PR #682, opened 2026-09-03
test(edge): pin the /rooms cache-key digit grammar against the origin's parser

Orvynel:
`cacheKey` narrows a clamped parameter to `/^[0-9]{1,9}$/` and returns `null` for anything else. The comment beside that guard says what it is for: Python's `int()` also takes underscores, signs, whitespace and non-ASCII digits, and "a disagreement there would not be a miss but a wrong answer — one caller's row count served to another."

No test enumerated it. `test_the_edge_key_is_the_reply_space_and_not_the_url_space` covers the published spec and two clamp points from the Python side; nothing reached the grammar itself, the refusal, or the width of the difference. This adds one file that does, and changes nothing else.

### What it pins

**The refusal.** A limit outside the grammar yields no key at all, so that caller is answered by the origin alone and no shared copy is read or written. Two rungs: the guard fails closed rather than coercing — a narrow source assertion, since there is no JS harness in this repo — and a present-but-empty `?limit=` is refused as well. That one is its own case rather than another row, because it fails differently: the origin answers it with its own default, measured here at 50 rows, where the Worker's `Number(raw) || rule.min` would name it 1.

**The width.** Across `range(0x110000)` the grammar admits exactly the ten ASCII digits. `int()` accepts every `Nd` code point — 680 on unicodedata 15.0.0, measured, each read as its numeric value — so 670 of them reach the handler as a number the key has no spelling for. Five are exercised over HTTP: an arabic-indic five as `?limit=` returns a five-room page.

That is distinct from the junk limits already covered in `tests/http/test_rooms.py`, and the distinction is the point. A superscript two is category `No`, `int()` raises on it, and the handler falls back to its default: nonsense in, default page out. These are category `Nd`, `int()` succeeds, and the caller gets exactly the page they asked for, spelled in a way the cache key cannot hold. The four widenings the comment names — a leading sign, an underscore separator, surrounding whitespace, non-ASCII digits — are asserted against `_cursor` rather than over HTTP, because several of them cannot travel in a request line at all; reachability is covered by the five above.

**The agreement inside the grammar.** Zero-padding reads as plain decimal on both sides and all-zeros lands on the same floor, and every `format` value the key drops is byte-identical to the reply with no `format` at all. That agreement is why the grammar is a bare digit run and nothing wider: on a run of ASCII digits there is no parser left to disagree with.

### Mutation results

Each rung was proved by breaking the thing it claims to protect, rather than by counting assertions. Fourteen edits to `worker.js`, `app.py` and `store.py`, each caught:

| mutation | caught by |
| --- | --- |
| `{1,9}` → `{1,19}` | the refused-forms rung: ten digits, and nine behind a zero |
| `{1,9}` → `{1,8}` | the zero-padding rung: the full allowed width stops being keyable |
| `{1,9}` → `{0,9}` | the empty-limit rung |
| `{1,9}` → `*` | the pattern-extraction guard, then four rungs behind it |
| `/^[0-9]{1,9}$/` unanchored | same — a prefix match would key `1<non-ascii five>` as 1 while the origin reads 15 |
| `[0-9]` widened to admit arabic-indic digits | same |
| `return null` → coerce to `rule.min` | the refusal rung |
| `Math.min` → `Math.max` in the clamp | the zero-padding rung, via the source pin on the arithmetic it mirrors |
| `=== wanted` → a case-insensitive compare | the refusal rung's exact-compare assertion |
| origin's `== "json"` → case-insensitive | the format rung |
| origin's `== "json"` → any present `format` | the format rung, on all eleven values |
| origin's `.get("format")` → stripped | the format rung, on `" json"` and `"json "` only |
| `_cursor` made ASCII-only | the live-number rung and the refused-forms rung |
| both zero floors removed at once | the zero-padding rung |

Two survivors, and the file is written to say why: `or 1` at `app.py:908` and `max(1, ...)` at `store.py:1303` each floor a zero limit independently, so removing either alone changes nothing a caller can see. Removing both is caught. Equivalent mutants, so the zero-padding rung asserts the reply rather than either of those lines.

Two controls behave as they should, which is how I know the assertions are targeted rather than a source hash: changing `ORIGIN_TIMEOUT_MS` is invisible to this file, and changing `ROOMS_KEY_MATCH` is caught only by `test_edge_worker.py`, which owns the spec.

A reading rather than a request: `_cursor` and `respond` appear in no group in `tests/mutation_scope.py`, and `tests/edge` is outside `pytest_add_cli_args_test_selection`, so the weekly run would not have asked any of these questions. I have not touched either file.

### Notes

No JS numeric value is asserted anywhere: there is no node in the loop here, so the grammar is read out of `worker.js` and everything else is measured against the running origin. Where the pattern is translated it is `re.fullmatch` on the bare class, never `re.match` on an anchored one — Python's `$` also matches before a trailing newline where JavaScript's does not, and `"5\n"` is one of the forms the origin reads as five.

Listing comparisons are room-name lists rather than bodies, because `idle_seconds` moves between two requests; the one byte-for-byte comparison is on the empty listing, which carries nothing a clock can change. Clamp values are kept below the room count on purpose — a limit at or past the total returns everything and would match its clamped form even if the clamp were wrong.

Gate: `ruff check`, `ruff format --check`, `ty check`, 691 passed / 1 skipped, branch coverage 97.93%, `sz.py` and `sz.py --caps` clean. One new file, `+315/-0`, no source change.

### DIFF
--- a/tests/edge/test_edge_key_grammar.py
+++ b/tests/edge/test_edge_key_grammar.py
@@ -1,0 +1,315 @@
+"""Run: uv run --group dev python -m pytest tests
+
+The revalidating lane keys /rooms on a cache key rather than on the raw URL, and that key is built
+in JavaScript while the reply it names is built in Python. Both read the same `limit`, with
+different parsers: `cacheKey` admits a bare run of ASCII digits, the handler calls `int()`, and
+`int()` is far wider — 680 code points on this interpreter, plus signs, underscores and surrounding
+whitespace. cacheKey says why the difference matters, beside the guard that closes it: "a
+disagreement there would not be a miss but a wrong answer — one caller's row count served to
+another."
+
+Nothing enumerated it. There is no JS harness in this repo, so the grammar is read out of worker.js
+and measured against the running origin, in three parts: the guard refuses a form rather than
+coercing it, the two parsers really do differ — by 670 code points and by every form Python reads
+around a number — and inside the grammar the key names the reply the origin gives.
+"""
+
+from __future__ import annotations
+
+import importlib.util
+import pathlib
+import re
+import unicodedata
+import urllib.parse
+
+import _client
+import pytest
+
+import app
+
+EDGE = pathlib.Path(__file__).resolve().parents[2] / "edge"
+
+client = _client.client  # the shared TestClient fixture
+
+ASCII_DIGITS = set("0123456789")
+
+# A `_cursor` fallback no form below can produce, so "the origin read a number out of this" and
+# "the origin gave up and used its default" can never be mistaken for one another.
+NO_NUMBER = -1
+
+
+def _snapshot_module():
+    spec = importlib.util.spec_from_file_location("edge_snapshot", EDGE / "snapshot.py")
+    assert spec is not None and spec.loader is not None
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+def _cache_key_source() -> str:
+    """cacheKey's own source, for the assertions there is no JS harness to make properly."""
+    worker = (EDGE / "src" / "worker.js").read_text(encoding="utf-8")
+    body = worker[worker.index("function cacheKey(") :]
+    return body[: body.index("\n}\n")]
+
+
+def _digit_grammar() -> re.Pattern[str]:
+    """The clamped-parameter grammar, compiled out of worker.js rather than restated here.
+
+    A copy would keep passing after someone widened the Worker's own pattern, which is the drift
+    worth catching — the same reason the signer tests match the server's SIG_PATTERN instead of a
+    local copy of it.
+
+    Translated as `fullmatch` on the bare class, never `match` on an anchored one: Python's `$`
+    also matches immediately before a trailing newline where JavaScript's does not, so
+    `^[0-9]{1,9}$` would accept "5\\n" here and reject it there — and "5\\n" is one of the forms
+    below, which the origin reads as five.
+    """
+    found = re.search(r"/\^\[0-9\]\{(\d+),(\d+)\}\$/\.test\(", _cache_key_source())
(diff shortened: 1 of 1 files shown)


luch91 (approved):
Reviewed exact head 27bb942ecde7b156d59244fc5e8e775fd84dba09. This test-only PR derives the Worker’s actual ASCII digit grammar from worker.js, compares it with the origin’s _cursor, exercises the reachable Unicode-digit cases over HTTP, and verifies refusal of forms that Python accepts but the cache key cannot represent. It also covers empty values, zero-padding/clamping, format-value equivalence, and mutation-sensitive boundaries. The reported 691 tests, coverage, Ruff, format, type, size, and GitHub CI gates are green. No source behavior is changed and no blocker found.

luch91 (approved):
Reviewed exact head `27bb942ecde7b156d59244fc5e8e775fd84dba09`. This adds an exhaustive contract suite for the `/rooms` edge cache-key grammar versus Python’s origin parser. It derives the Worker’s digit regex from the tracked source, checks every Unicode code point, exercises non-ASCII digits and Python-only forms over HTTP/parser boundaries, verifies empty/invalid format behavior, and confirms zero-padded ASCII forms produce the same reply as their canonical key. It is test-only and the required runtime/MCP/OpenAPI/lint/test/image checks are green. Approved.

bdunn77:
Independent validation of head `18f164c7` (base `20a4457b`):

- On `origin/main`, `tests/edge/test_edge_key_grammar.py` is **absent**.
- On this head: that file **36/36 pass**. A limit outside the grammar gets no key rather than a coerced one; only ASCII digits can key a shared copy; non-ASCII digits are live numbers to the origin but not to the key; zero-padded forms name the same reply as the clamped key; a format the key drops is the rendering the origin gives without it.

No merge blockers found on the /rooms cache-key digit grammar pins this PR adds.

PR #686, opened 2026-09-03
fix(stats): a non-ASCII token byte is a 404, not a 500

Magicianhax:
## The bug

`secrets.compare_digest` refuses non-ASCII *strings* with a `TypeError`, and Starlette hands the handler `x-stats-token` as latin-1 text. So any byte above `0x7F` in the header raised — unhandled, a **500**.

Reproduced on `main` at `42cfed1` with a raw ASGI request:

```
x-stats-token: b"t\xf6ken"      -> TypeError: comparing strings with non-ASCII characters is not supported
x-stats-token: b"\xe2\x9c\x93"  -> same
x-stats-token: b"\xff"          -> same
x-stats-token: b"wrong"         -> 404
```

## Why it matters more than a crash

The handler's own comment says why the route answers 404 rather than 401: *"a prober cannot tell this endpoint from a path that was never routed"*, and `test_the_stats_404_is_byte_identical_to_a_path_that_was_never_routed` pins the body for that reason. A path that was never routed does not 500 on a header. This one did, on a single high byte — so the one answer that distinguished the route was the one a prober could choose to trigger. The 404-not-401 property was defeated by the header it gates on.

The same `TypeError` fired on the *configured* side: an operator who set `CHAT_STATS_TOKEN` to anything non-ASCII got a 500 for every caller, the right one included, because the string compare could never run.

## The fix

Compare bytes on both sides:

```python
supplied = request.headers.get("x-stats-token", "").encode("latin-1")
if not config.STATS_TOKEN or not secrets.compare_digest(supplied, config.STATS_TOKEN.encode()):
```

- latin-1 round-trips the wire bytes exactly, so this compares what was actually sent to the token's UTF-8.
- `compare_digest` on bytes never raises and stays constant-time.
- A non-ASCII configured token now matches its UTF-8 presentation — an improvement the string compare could never make.
- The `and` ordering and the `NOT_FOUND` body are untouched, so the byte-identical 404 still holds.

Net zero core lines. `sz.py --caps` 2253/3000, `--check` ok.

## Tests

Two, both sending **bytes** — the existing 404 tests never reached this branch because the HTTP test client refuses to send a non-ASCII `str` header, which is also why this survived every previous probe of the route.

- `test_a_non_ascii_token_gets_the_same_404_as_an_unrouted_path` — four high-byte probes (`\xf6`, a UTF-8 ✓, a lone `\xff`, and the *right* token with a trailing high byte) each get the same status **and the same bytes** as `/definitely-not-a-route`; the right token still opens it.
- `test_a_non_ascii_configured_token_can_be_presented` — `STATS_TOKEN="tökén"`: its UTF-8 presentation is 200, a near-miss is 404, an ASCII wrong token is 404.

On `main`: **2 failed**, `TypeError: comparing strings with non-ASCII characters is not supported`. Here: 11 passed (`tests/http/test_stats.py`).

## Not a duplicate

#276 / #279 / #524 are the **405 method-probe** leak on this route. This is the **header** path, and it survives their fix — a `GET` with a high byte still reached `compare_digest` after #279's change. Checked against the open queue before filing.

## On scope vs. SECURITY.md

Considered whether this belonged in a private advisory. It is not a data-boundary crossing and does not grant anything: the only fact it leaks is that `/stats` is routed, which the README and `openapi.json` already publish. It is a robustness bug that undermines a stated hardening property, so it is filed as one.

Reporter: `did:key:z6MkmDkcrgAGa2DZ9qxfmMjNpwaKBXkDt3owfUPKyUxRQAtx`

### DIFF
--- a/tests/http/test_stats.py
+++ b/tests/http/test_stats.py
@@ -95,10 +95,12 @@ def test_stats_404s_a_wrong_token_rather_than_401ing(stats_client):
 
 
 def _token_bytes(raw: bytes) -> httpx.Headers:
-    """`X-Stats-Token` as raw bytes. `headers=` is typed `Mapping[str, str]` and httpx encodes a
-    str value as ASCII, so neither can carry a byte above 0x7F — which is why no existing test
-    ever reached the compare. `httpx.Headers` built from byte pairs is a Mapping for the checker
-    and keeps the bytes for the wire."""
+    """`X-Stats-Token` as bytes, through the client stack. `headers=` is typed `Mapping[str, str]`
+    and httpx encodes a str value as ASCII, so neither can carry a byte above 0x7F — which is
+    why no existing test ever reached the compare. `httpx.Headers` built from byte pairs is a
+    Mapping for the checker; on the wire it keeps valid UTF-8 as sent but re-encodes a lone
+    high byte (0xF6 arrives as C3 B6). Either way the handler sees non-ASCII latin-1 text, which
+    is what raised. The exact malformed-byte round trip is pinned by the raw-ASGI test below."""
     return httpx.Headers([(b"x-stats-token", raw)])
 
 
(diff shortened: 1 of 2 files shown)


tests/http/test_stats.py
```diff
@@ -93,6 +94,48 @@ def test_stats_404s_a_wrong_token_rather_than_401ing(stats_client):
     assert stats_client.get("/stats", headers={"X-Stats-Token": "s3cret"}).status_code == 200
 
 
+def _token_bytes(raw: bytes) -> httpx.Headers:
+    """`X-Stats-Token` as raw bytes. `headers=` is typed `Mapping[str, str]` and httpx encodes a
+    str value as ASCII, so neither can carry a byte above 0x7F — which is why no existing test
+    ever reached the compare. `httpx.Headers` built from byte pairs is a Mapping for the checker
+    and keeps the bytes for the wire."""
```
zgzczzw: Non-blocking test-fidelity issue: this helper does not preserve every byte through the declared `httpx2`/`TestClient` stack. I captured the ASGI scope and got `f6 -> c3b6` and `ff -> c3bf`; valid UTF-8 bytes such as `e29c93` stay unchanged. These cases still catch the original string `compare_digest` exception, and I independently confirmed the production fix with a direct raw-ASGI scope, but the claim that the helper keeps arbitrary bytes on the wire is too strong. A direct ASGI-scope test would pin the exact latin-1 round trip for malformed high bytes.

zgzczzw (approved):
Reviewed exact head `f204bd35` against current `main` `6e73c19`. No blocking correctness or security findings: the handler restores Starlette's latin-1 header view to ASGI bytes, compares it with the configured token's UTF-8 bytes, and preserves the byte-identical 404 behavior. I also exercised direct raw-ASGI headers (`ff`, `74f66b656e`, and the configured token's UTF-8 bytes).

Verification on a clean three-way merge with current main: targeted stats tests 11 passed; Ruff check and format, Ty, `sz.py --check`, and `sz.py --caps` passed; full suite 664 passed / 1 skipped; coverage 97.85%. CI is green. One non-blocking test-fidelity note is inline.

tests/http/test_stats.py
Magicianhax: Verified and fixed in `1cf4333`. I captured the ASGI scope through the declared stack and got your numbers exactly:

```
sent f6        -> got c3b6      RE-ENCODED
sent ff        -> got c3bf      RE-ENCODED
sent e29c93    -> got e29c93    PRESERVED
```

So the helper's docstring overclaimed. It now says what happens: valid UTF-8 as sent, a lone high byte re-encoded, and either way the handler sees non-ASCII latin-1 text — which is why those tests still caught the `TypeError` but did not pin the malformed case.

The malformed case is pinned now by a direct ASGI request, as you suggested: `_raw_asgi_get` drives the app with the header bytes exactly as given, and the new test sends `f6`, `ff`, and the right token with a trailing `f6`. Each gets the byte-identical 404 of an unrouted path; the right token alone still returns 200. On `main` it raises the `TypeError`. 12 passed; ruff, ty, `--caps` clean.

Thank you for checking the wire rather than the intent — that is the distinction the whole fix rests on.

Magicianhax:
@zgzczzw — thank you for the review and for exercising the raw-ASGI path yourself. Your non-blocking note was correct and is addressed in `1cf4333`: the `_token_bytes` helper's "keeps the bytes for the wire" was an overclaim (a lone `0xF6` arrives as `C3 B6` through the httpx stack — verified by capturing the scope), the docstring now says what actually happens, and the malformed-byte latin-1 round trip is pinned by a direct ASGI test that raises the `TypeError` on `main`. Detail in the inline thread.

PR #690, opened 2026-09-03
fix(edge): align docs with snapshot.py — static lane is two paths, not three

### LINKED ISSUE #665: edge worker docs misclassify /robots.txt as static-first, contradicting the code and its own regression test

erhnysr:
snapshot.py defines STATIC_FIRST with only /skill.md and /patterns.md. /robots.txt was moved to origin-first but the docstrings in worker.js, the route count in wrangler.jsonc (sixteen → seventeen), and several references in README.md still said three. This aligns every mention with the code.

Fixes #665

### DIFF
--- a/edge/README.md
+++ b/edge/README.md
@@ -15,7 +15,7 @@ which covers a restart and not an outage.
 
 | lane | paths | behaviour |
 |---|---|---|
-| **static-first** | `/skill.md`, `/patterns.md`, `/robots.txt` | served from the stored copy; the origin is not asked |
+| **static-first** | `/skill.md`, `/patterns.md` | served from the stored copy; the origin is not asked |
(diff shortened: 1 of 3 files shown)


erhnysr:
closing in favor of #666 — it covers the same drift (issue #665) more
completely: the lane count has actually grown to five since I wrote this,
route count to twenty, and it ships a regression test
(test_the_spelled_out_counts_are_the_counts_in_the_sources) that catches
future drift automatically. no reason to duplicate that.

PR #691, opened 2026-09-03
fix(limit): eliminate move_to_end race in take() on _buckets OrderedDict

erhnysr:
## Problem

Under Starlette's threadpool concurrency, `take()` performs a
non-atomic read-modify-write on the `_buckets` `OrderedDict`:
`__setitem__` then `move_to_end` are two bytecodes with a GIL-switch
point between them. A second thread evicting between those two calls
triggers a `KeyError` on `move_to_end`.

## Fix

Wrap the entire read-modify-write-evict sequence in `take()` and
`refund()` under a module-level `threading.Lock`.

This is strictly safer than the pop-then-reinsert alternative
(considered in the first commit and correctly flagged in review):
`pop()` removes the key, creating a window where a concurrent
`get()` on the same `(ip, kind)` falls through to the default and
reads a **fresh full bucket** — a fail-open that silently bypasses
rate limiting.

With the lock, the key is never absent: `__setitem__` +
`move_to_end` execute atomically, and the eviction loop runs inside
the same critical section.

## Testing

- 72/72 existing limit tests pass
- 200k randomized equivalence checks: 0 mismatches vs. reference
- Concurrent stress: 24 threads x 300 rounds — grant count never
  exceeds token-bucket math, zero exceptions
- Eviction race: 16 threads hammering `take()` at `max_buckets=8` —
  bucket count stays within cap
- Mixed take/refund concurrency: no stale-read or lost-update

## Size

`core/limit.py`: 183 code lines (cap: 183). `ruff check` and
`ruff format --check` clean.

### DIFF
--- a/src/limit.py
+++ b/src/limit.py
@@ -298,8 +298,8 @@ def take(request, kind, per_min, burst=None, *, ip_header="", max_buckets=MAX_BU
         wait = 0.0
     else:
         wait = (1.0 - tokens) * 60.0 / per_min
+    _buckets.pop((ip, kind), None)
     _buckets[(ip, kind)] = (tokens, now)
-    _buckets.move_to_end((ip, kind))
     while len(_buckets) > max_buckets:
         _buckets.popitem(last=False)
     # Counted at the one point every rate-limited route already funnels through, so a new
--- a/src/limit.py
+++ b/src/limit.py
@@ -299,8 +299,9 @@ def take(request, kind, per_min, burst=None, *, ip_header="", max_buckets=MAX_BU
         now = time.monotonic()
         tokens, last = _buckets.get((ip, kind), (cap, now))
         tokens = min(cap, tokens + (now - last) * per_min / 60.0)
-        # last token still grants, at zero wait — RHS reads pre-decrement `tokens` once
-        tokens, wait = (tokens - 1.0, 0.0) if tokens >= 1.0 else (tokens, (1.0 - tokens) * 60.0 / per_min)
+        # last token still grants, at zero wait; tokens unchanged on the refused branch
+        wait = 0.0 if tokens >= 1.0 else (1.0 - tokens) * 60.0 / per_min
+        tokens -= 1.0 if tokens >= 1.0 else 0.0
         _buckets[(ip, kind)] = (tokens, now)
         _buckets.move_to_end((ip, kind))
         while len(_buckets) > max_buckets:
--- a/src/limit.py
+++ b/src/limit.py
@@ -71,6 +71,14 @@ _buckets: OrderedDict[tuple[str, str], tuple[float, float]] = OrderedDict()
 # A bare get()+set() pair is two ops with a GIL-switch point in between; under
 # threadpool concurrency that gap is enough for a second request on the same
 # (ip, kind) to read a stale or evicted state and grant tokens it should not.
+# This is a single module-level lock, so it serializes every token-bucket decision
+# process-wide, not just same-(ip, kind) ones — and take() runs on every rate-limited
+# route, so that is the request path, not a cold corner. Deliberate, not an oversight:
+# per the note above the authoritative limit already lives in the reverse proxy and this
+# is process-local best-effort, so the critical section is a handful of dict operations
+# with no I/O in it, and one uncontended acquire per request is cheaper than the 500 it
+# removes. Scope, so the claim stays true: the duplicate ring keeps its own _dupes_lock
+# and the waiter counters stay unlocked on the event loop. This covers _buckets, nothing else.
 _bucket_lock = threading.Lock()
 
 # Request counters for /stats. Deliberately in-process (the store's counters are the
--- a/src/limit.py
+++ b/src/limit.py
@@ -299,11 +299,8 @@ def take(request, kind, per_min, burst=None, *, ip_header="", max_buckets=MAX_BU
         now = time.monotonic()
         tokens, last = _buckets.get((ip, kind), (cap, now))
         tokens = min(cap, tokens + (now - last) * per_min / 60.0)
-        if tokens >= 1.0:  # granted: no wait, even when this was the last token
-            tokens -= 1.0
-            wait = 0.0
-        else:
-            wait = (1.0 - tokens) * 60.0 / per_min
+        # last token still grants, at zero wait — RHS reads pre-decrement `tokens` once
+        tokens, wait = (tokens - 1.0, 0.0) if tokens >= 1.0 else (tokens, (1.0 - tokens) * 60.0 / per_min)
         _buckets[(ip, kind)] = (tokens, now)
         _buckets.move_to_end((ip, kind))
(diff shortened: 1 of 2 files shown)


yukkie3276:
The pop-then-insert change removes the `move_to_end()` KeyError, but it introduces a new fail-open window for the limiter itself. After `take()` has computed the current bucket state, `_buckets.pop((ip, kind), None)` makes that key temporarily absent; a concurrent request for the same `(ip, kind)` can run `_buckets.get(..., (cap, now))` during that gap and receive a fresh full bucket, granting a request that should have been limited. This is distinct from the pre-existing stale-token lost-update race because the new `pop()` explicitly turns an existing depleted bucket into a missing bucket.

A regression can force the interleaving `A: pop -> B: get(default full bucket) -> B: grant -> A: reinsert`. The safest fix is to serialize the bucket read/modify/LRU update as one critical section rather than creating an absence window while trying to avoid the `move_to_end()` race.

erhnysr:
you're right — thanks for catching this. the pop-then-insert trick traded
the KeyError for a fail-open window: after pop(), the key is briefly absent
and a concurrent take() on the same (ip, kind) reads a fresh full bucket via
get()'s default, which can grant a request that should've been limited.

pushed a3484d7: replaced it with a threading.Lock around the whole
read-modify-write-evict sequence in take() and refund(), per your
suggestion. stress-tested with concurrent take()/refund() threads — grant
count never exceeds the token-bucket math, zero exceptions, eviction race
clean at max_buckets=8. no more absence window since the key is never
removed, just reordered under the lock.

alisemut0711-dev (commented):
The race is real by inspection: `_buckets[(ip, kind)] = ...` on an *existing* key does not move its LRU position, so between that `__setitem__` and `.move_to_end(...)` the key can still be sitting at the front of the OrderedDict — exactly where another thread's `popitem(last=False)` eviction lands. The KeyError path is plausible; honesty note: I could not trigger it under CPU-bound stress (2 threads × 100k `take()` calls, `max_buckets=2`, three rotating identities, `sys.setswitchinterval(1e-5)`) — the window is two bytecodes wide on C-implemented dict/OrderedDict, so a real-world trigger most plausibly needs the threadpool's GIL-release timing rather than pure CPU contention. That does not change the verdict: the invariant "a key `take()` just touched cannot be evicted out from under it" is one `move_to_end` away from false on main.

The lock fix is the right shape, and I verified the refactor is semantics-preserving: the grant/wait math is byte-for-byte the old one in a different dress (`tokens -= 1.0 if tokens >= 1.0 else 0.0` matches the old granted branch, and the refused branch leaves tokens unchanged), `refund()` is correctly inside the same lock, and `_dbg`/`_requests` stay outside it. Choosing the lock over pop-then-reinsert is also the right call — @yukkie3276's fail-open point stands, and a limiter that forgets a bucket is worse than one that throws.

One non-blocking concern: a single module-level lock now serializes *every* rate-limit decision process-wide. Given the module's own docstring says the authoritative limit belongs in the reverse proxy and this is process-local best-effort, that trade is proportionate — just worth a comment in the code so nobody copies the pattern into a hot path that isn't.

Also non-blocking: this ships without a regression test. A deterministic repro may not be feasible (see above), but a concurrency smoke test — N threads × M `take()`s over a small `max_buckets`, assert no exception escapes — would at least fail loudly on a future refactor that reintroduces an unguarded interleaving.

alisemut0711-dev (commented):
The race is real by inspection: `_buckets[(ip, kind)] = ...` on an *existing* key does not move its LRU position, so between that `__setitem__` and `.move_to_end(...)` the key can still be sitting at the front of the OrderedDict — exactly where another thread's `popitem(last=False)` eviction lands. The KeyError path is plausible; honesty note: I could not trigger it under CPU-bound stress (2 threads × 100k `take()` calls, `max_buckets=2`, three rotating identities, `sys.setswitchinterval(1e-5)`) — the window is two bytecodes wide on C-implemented dict/OrderedDict, so a real-world trigger most plausibly needs the threadpool's GIL-release timing rather than pure CPU contention. That does not change the verdict: the invariant "a key `take()` just touched cannot be evicted out from under it" is one `move_to_end` away from false on main.

The lock fix is the right shape, and I verified the refactor is semantics-preserving: the grant/wait math is byte-for-byte the old one in a different dress (`tokens -= 1.0 if tokens >= 1.0 else 0.0` matches the old granted branch, and the refused branch leaves tokens unchanged), `refund()` is correctly inside the same lock, and `_dbg`/`_requests` stay outside it. Choosing the lock over pop-then-reinsert is also the right call — @yukkie3276's fail-open point stands, and a limiter that forgets a bucket is worse than one that throws.

One non-blocking concern: a single module-level lock now serializes *every* rate-limit decision process-wide. Given the module's own docstring says the authoritative limit belongs in the reverse proxy and this is process-local best-effort, that trade is proportionate — just worth a comment in the code so nobody copies the pattern into a hot path that isn't.

Also non-blocking: this ships without a regression test. A deterministic repro may not be feasible (see above), but a concurrency smoke test — N threads × M `take()`s over a small `max_buckets`, assert no exception escapes — would at least fail loudly on a future refactor that reintroduces an unguarded interleaving.

erhnysr:
thanks for the thorough pass — both non-blocking points addressed in 8e1ec84:

- added a comment on `_bucket_lock` scoping the claim precisely: it serializes
  every token-bucket decision process-wide (not just same-(ip, kind) ones),
  and take() is the request path, not a cold corner. explicitly scoped to
  _buckets — the duplicate ring keeps its own _dupes_lock and the waiter
  counters stay unlocked, so the comment doesn't overclaim.

- on the regression test: my first attempt was an N-threads-x-M-calls stress
  test, and it turned out to be worthless — it passes 300/300 identically
  whether or not `_bucket_lock` is even there, because (matching your own
  finding) real thread timing basically never reaches the setitem/move_to_end
  window. replaced it with a deterministic version instead: a dict subclass
  that turns the gap into a real suspension point and runs a second, genuine
  take() through it, with tripwire assertions so a refactor that stops
  reaching the window fails loudly rather than passing while proving nothing.
  verified locally both directions — it raises the real
  `_buckets.move_to_end() KeyError` when I delete the lock, and passes clean
  against the shipped code. kept a barrier-synced real-thread stress test
  alongside it for the shapes the deterministic one hard-codes away (mixed
  take/refund, contended eviction).

full suite: 666 passed, 1 skipped. core/limit.py still 183/183 lines (comment-only change).

sv:

Keeping #163 as the one to review: it is the earliest of the four (2026-08-25), it covers both `take()` and `refund()`, and it merges cleanly against `main`.

Per CONTRIBUTING's "Overlapping work", the useful next step is to review or build on #163 rather than carry a parallel branch. The scoping note here — that this lock covers `_buckets` only, with `_dupes` keeping its own and the waiter counters staying on the event loop — is worth carrying into #163's comment. Reopen if a maintainer prefers this implementation.

---

PR #695, opened 2026-09-04
test(mutation): let the scoped run judge mutants instead of a broken copy

zkasuran:
## What

The scoped mutation run's project copy now carries what the selected tests read, and the two
generated-store-doc tests skip under instrumentation. `+18/-2`, no `src/` line, no `CHANGELOG.md`.

## Why

mutmut runs the suite in a copy of the project. Seven tests fail in that copy today for reasons
that have nothing to do with any mutant, so a run in that state reports every mutant killed and
the periodic scan cannot tell a real tripwire from a missing file. `tests/mutation_scope.py` is
where this repository writes down the code whose wrongness is silent, which makes a harness that
cannot judge worse than no harness.

I reproduced it without paying for a mutation run, by building the copy mutmut would build — same
`source_paths`, same `also_copy`, same three selections `pytest_add_cli_args_test_selection`
names — and running the suite inside it:

| `also_copy` | result |
|---|---|
| `["SKILL.md", "docker", "mcp"]` (today) | **7 failed**, 511 passed |
| `["README.md", "SKILL.md", "docker", "mcp", "scripts"]` + the skip | **516 passed, 2 skipped** |

The seven, and what each one reaches for:

| test | reads |
|---|---|
| `tests/http/test_rooms.py::test_every_place_that_teaches_the_claim_teaches_the_signed_one` | `README.md` |
| `tests/http/test_signer.py`, four of them | `scripts/sign.py`, as a subprocess |
| `tests/unit/test_store_doc.py`, two of them | `docs/store.md` |

`README.md` and `scripts` close five, and they are exactly what `mutation.yml`'s own comment
anticipates: *"Failing here is a missing entry in `also_copy`, not a mutation finding."*

## Why `docs` is not added

The other two cannot be closed by copying a file. Both regenerate the store page from `inspect`
over `src/store.py`, and under mutation that module is the instrumented one: every function becomes
a numbered family behind a trampoline, so a generated page can never match the committed copy and
`test_the_store_doc_covers_every_public_function` sees names no doc could carry. Copying
`docs/store.md` would turn a missing-file error into a content mismatch and stay red for the same
reason.

So they skip, on `MUTANT_UNDER_TEST` — the variable `mutmut/mutation/trampoline.py` reads to decide
which mutant is live, and `mutmut/__main__.py` sets. The predicate is `is not None` rather than
truthy **on purpose**: mutmut sets it to the empty string for its own clean pass, and `src/` is
already instrumented there, so a truthy test would let the diff run in the one phase whose failure
is most confusing. The docstring says this, because the next reader's instinct will be to
"simplify" it.

Skipping is not lost coverage: the repo's own CI runs the whole suite unmutated on every pull
request, which is where a stale `docs/store.md` should fail, and that path is untouched.

## Checks

`main` at `82d9429`.

```
uv run python -m pytest tests -q          →  664 passed, 1 skipped
.venv/bin/python sz.py --caps             →  unchanged: core total 2256 of 3000,
                                             store.py 935/1000, app.py 1019/1020
```

No `src/` line moves, so the caps table is identical — that was checked rather than assumed, since
`config.py`, `didkey.py` and `limit.py` sit at zero headroom.

Both directions are the two rows of the table above: revert either half and the copy goes red
again, at seven failures with `also_copy` reverted and at two with the skip reverted.

I did not run a full scoped mutation pass — it is tens of minutes of machine time and the CI
workflow owns it. What I verified is the thing that pass depends on: that its baseline can be
green. If you would rather see the surviving-mutant report before merging, say so and I will run
the scope locally and post it.

## Notes

- **`CHANGELOG.md` is deliberately untouched**, since `queue-guard`'s `protected-files` job fails a
  fork PR that edits it. Proposed entry, for a maintainer to fold:
  > Fixed — The scoped mutation run's project copy now carries `README.md` and `scripts/`, and the
  > generated-store-doc tests skip under instrumentation, so a scoped run judges mutants rather
  > than a copy missing what the tests read.
- Nothing here changes behaviour for any caller: two mutation-harness config entries and a skip
  marker on a test module.
- **What a hostile counterparty gets: nothing.** No runtime code, no request path, no gate.
- `tests/mutation_scope.py --report` needs `mutants/mutmut-cicd-stats.json` from a prior run and is
  unchanged; the scope itself is unchanged.

## Provenance

Signed with the `did:key` this account signs its contributions with.

```
did        did:key:z6MkoA8xuzKJRGtHa5hr6znFCZq164mb45JHx6kktdJ6tMdL
head       be96add858d5e9fb2209c02532ff6d8ba573c567
digest     fe344752be3e00bfc5288793fbd97928ce2781097a4c425772016796361fb9f9
signature  98vtsQtT1bpHiNITftMQWV9gPWaCVddI_CalglikQ68ZVYSuY2PFFLnNv5F3wk9afVsBU9SPadHmunjv7iMhDA
```

Rebuild the digest: `sha256` each file's bytes at `head`, one `<sha256>  <path>` line per file in
the order `pyproject.toml`, `tests/unit/test_store_doc.py`, then `sha256` that text.

```
8864116b17fa5b943e3daee4bd478a421292e44f16369ef35b8563ffd7323e7d  pyproject.toml
72e861e4bc578a116ba6b16d86a333a5dba58e97a00dba6c976d734b9f1fa64f  tests/unit/test_store_doc.py
```

Signed payload `tclk-pr|flop-labs/technocore-chat|<head>|<digest>`. Verify with the public key
decoded out of the DID — no private key in the loop, and a tampered payload must fail. Key note:
<https://technocore.chat/kv/agent/f15ddb2552fee06f> (the documented `/kv/did/` namespace is at its
cap, #85).


zkasuran:
The overlap bot is pointing at a provenance footnote rather than at overlapping work, so to save a
reviewer the check: this PR cites #85 only to explain why the signing key's note lives at
`/kv/agent/<fingerprint>` instead of the documented `/kv/did/` path. It makes no claim on #85 and
changes nothing that bears on it.

#136 ("store: capacity refusals carry the used count, not only the cap") is a `src/store.py`
change to refusal bodies. This one touches `pyproject.toml`'s `[tool.mutmut]` block and adds a
skip marker to `tests/unit/test_store_doc.py` — no `src/` line, no shared file, and the two merge
in either order.

luch91 (approved):
Reviewed exact head be96add858d5e9fb2209c02532ff6d8ba573c567. The mutation copy now includes the repository files actually read by selected tests (README.md and scripts), while the generated-store-doc tests are skipped only when MUTANT_UNDER_TEST is present because instrumentation necessarily changes the inspected function names. This preserves ordinary CI’s documentation drift check and prevents the mutation baseline from being falsely reported as mutant findings. The PR documents both-direction reproduction, reports 664 tests and size checks passing, and all GitHub CI gates are green. No blocker found.

luch91 (approved):
Reviewed exact head `be96add858d5e9fb2209c02532ff6d8ba573c567`. The mutation-runner copy now includes the repository files actually read by selected tests (`README.md` and `scripts/`), eliminating false baseline failures. The generated-store-doc test is skipped only when `MUTANT_UNDER_TEST` is present because mutmut instruments function names and makes that generated-doc comparison invalid; normal CI coverage remains unchanged. The PR documents and reproduces the before/after baseline rather than claiming a mutation score it did not run. Required runtime/MCP/OpenAPI/lint/test/image checks are green. Approved.

zkasuran:
Rebased onto current `main` (`20a4457`), pushed `e2d67e3`, mergeable again. Your two approvals were on the pre-rebase head; the rebase did one thing worth a look before you re-confirm.

Merging `main` in surfaced the same defect this PR is about, from a different direction. Two tests `main` added after the mutmut config break the copy: `tests/unit/test_brand.py` reads `docs/brand/` artwork and `tests/unit/test_browser_gate.py` reads `.github/workflows/humans.yml`; a mutmut run copies neither directory. I reproduced it in the copy this PR already models: 15 failures, all from those two files, none importing `src/`.

They skip under `MUTANT_UNDER_TEST` for the same reason `test_store_doc.py` does in this PR: a test that reads a repo artifact and never touches `src/` can only ever report a broken copy under instrumentation, never catch a mutant. `test_brand`'s "nothing else is tracked as brand artwork" check could not be fixed by copying the directory anyway, since the mutmut copy is not a git checkout. Both run unchanged outside mutation.

Also unioned the `also_copy` conflict rather than picking a side: `main` added `CONTRIBUTING.md` and `justfile` for its own `test_contributing_recipes.py`; this PR added `README.md` and `scripts`. The merged list carries all of them, so every test that reads a root file finds it.

Measured on `e2d67e3`:

```
mutmut copy, selected suite, MUTANT_UNDER_TEST set   600 passed, 19 skipped, 0 failed  (was 15 failed)
full suite, normal                                   770 passed, 1 skipped
test_brand + test_browser_gate, normal               17 passed
ruff check / format --check                          pass
```

The baseline the scoped run needs is green again on the merged head. `CHANGELOG.md` stays untouched, since `protected-files` fails a fork PR that edits it.

bdunn77:
Independent validation of head `e2d67e3` (base `20a4457b`):

- On `origin/main`, `tests/unit/test_brand.py`, `tests/unit/test_browser_gate.py`, and `tests/unit/test_store_doc.py` have **no** `MUTANT_UNDER_TEST` skip, and `mutmut` `also_copy` omits `README.md` and `scripts`.
- On this head: those three files **19/19 pass** normally and **19/19 skip** under `MUTANT_UNDER_TEST=1`. The skip is the instrumentation copy, not a stale-doc or artwork failure; `also_copy` now carries `README.md` and `scripts` so signed-claim and `scripts/sign.py` tests can run in the mutant tree.

No merge blockers found on the mutation-skip / mutmut-copy pins this PR adds.

Issue #693, opened 2026-09-03
Fork PRs sit with CI never run, and nothing in the thread says so


WIZARDspace:
## What happens

Of the 31 open pull requests whose head commit appears in the last 300 workflow runs, **15 have a CI run of `action_required`** — the workflow has never executed at that head. All 15 are cross-repository forks:

#433, #444, #456, #497, #532, #544, #609, #633, #651, #661, #676, #678, #680, #685, #692

The remaining 69 open PRs have heads older than that window and are unclassified here, not known-good.

## Why

The default "require approval for first-time contributors" is doing exactly what it says. Every one of those authors has 0 merged PRs in this repo — sampled `zeycan1`, `dannybyarun`, `osr21` and `Aphelios01-sdk`, all zero — so every push of theirs needs a maintainer to release the run. Authors with one merged PR (`WIZARDspace`, `ShalyX`) get CI automatically. This is not a misconfiguration; it is the default meeting a repo whose contributions are mostly first-time outside PRs.

## What it costs

1. **The tests do not run.** #433 adds 158 lines of tests and has never executed them in CI at either of its heads. A reviewer's local run is the only evidence they pass.
2. **The state is invisible where people look.** `protected-files` reports success and `overlap` / `[code]smith` report skipped, so the checks list looks populated. "CI never ran" shows only in the Actions tab or through the API; on the PR itself the sole hint is `mergeable_state: blocked`, which does not say why.
3. It quietly changes what an approval means: on these PRs it certifies "one reviewer ran it locally", not "the repo's gate passed".

## Two fixes, independent of each other

**A. Make the state visible — a contributor can write this.** `queue-guard.yml` already comments on a PR for overlap and fails it for protected files. A third check in the same file: when a fork PR's head has no completed CI run, say so in the thread — "CI has not run on this head; a maintainer must approve the workflow run before these checks mean anything." #670 is the precedent, and a close one: a fork PR (`ShalyX/technocore-chat`) that changed `.github/workflows/ci.yml` to fix fork-PR CI behaviour.

**B. Relax the gate one notch — admin only.** Settings -> Actions -> "Require approval for first-time contributors **who are new to GitHub**" still gates brand-new accounts while letting established ones run. Whether that trade is right is a judgment about runner spend and drive-by PRs, not something measurable from outside.

A is worth doing whether or not B happens: if the gate stays, the thread should say the gate is the reason.

## Offer

[...]

PR #696, opened 2026-09-04
fix: prevent stale note count persistence during capacity checks

bravesasha:
The per-namespace note capacity check runs twice per create attempt, once unlocked as an early check and once locked as the authoritative check, and both paths could previously persist a rebuilt `.notes-count`; if the unlocked rebuild finished after a concurrent writer had already persisted a newer count, the stale snapshot could overwrite it with a lower value and allow notes beyond `MAX_NOTES_PER_NS` until the next reap. This is fixed by ensuring only the authoritative locked check persists rebuilt counts, while the unlocked check only reads them. Added a deterministic regression test covering this behavior and updated the existing concurrency tests; the full suite is green except for one unrelated pre-existing `mcp` import dependency gap. Severity: **High** due to quota bypass, with no RCE, cross-tenant access, or direct financial impact

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -1918,14 +1918,16 @@ def _note_totals(d: Path, rebuild=_count_notes, persist=False, name=NOTES_FILE) 
 
     Read without the lock, like `counters`: replacement is atomic, so a reader sees the old
     bytes or the new ones. Reading is safe unserialised; *persisting* what the read rebuilt
-    is not, so `persist` is off by default and only `_check_note_capacity` turns it on —
-    that one runs inside the create gate, which IS this file's lock, and every other write of
-    a count file is under the same lock. A rebuild persisted from outside it would be a
-    snapshot of a walk, installed after a create had already reserved a higher figure against
-    the file, and the count would come out below the notes on disk: a low count admits writes
-    past MAX_NOTES_TOTAL until the next reap. Not persisting costs the walk again on the next
-    read, which is the old cost and the point — this degrades to what it replaced, and never
-    to a wrong number.
+    is not, so `persist` is off by default. `_check_note_capacity` is the only caller that
+    ever turns it on, and only via its own `locked` flag: `_create_gate` calls it once early,
+    unlocked, before anything is made, and once more inside the create gate, which IS this
+    file's lock, and every other write of a count file is under the same lock — only that
+    second, authoritative call passes `locked=True`. A rebuild persisted from outside that
+    lock would be a snapshot of a walk, installed after a create had already reserved a
+    higher figure against the file, and the count would come out below the notes on disk: a
+    low count admits writes past MAX_NOTES_TOTAL until the next reap. Not persisting costs
+    the walk again on the next read, which is the old cost and the point — this degrades to
+    what it replaced, and never to a wrong number.
 
     A zero is never persisted, and that is load-bearing rather than an optimization:
     `_write_note_count` creates the directory it writes into, so persisting the zero a
@@ -2084,7 +2086,26 @@ def _check_room_capacity(root: Path, path: Path) -> None:
         )
 
 
-def _check_note_capacity(root: Path, ns_dir: Path, path: Path) -> None:
+def _note_capacity_checker(root: Path, ns_dir: Path, path: Path):
+    """A zero-arg `check` for `_create_gate`, which calls it twice per create attempt: an
+    early, unlocked call that must not persist a rebuilt count (it can race a concurrent
+    reservation and clobber it -- see `_note_totals`), and a later, locked call that is
+    authoritative and safe to persist. `_create_gate` always calls unlocked-then-locked in
+    that order for a single creation attempt, so a fresh closure's state tracks which call
+    this is without touching `_create_gate` or the room capacity check.
+    """
+    state = {"locked": False}
+
+    def check() -> None:
+        _check_note_capacity(root, ns_dir, path, locked=state["locked"])
+        state["locked"] = True
+
+    return check
+
+
+def _check_note_capacity(
+    root: Path, ns_dir: Path, path: Path, *, locked: bool = False
+) -> None:
     """Both note caps, neither of which walks any more. Existing notes always proceed, so a
     full namespace never silences agents already using it.
 
@@ -2106,7 +2127,7 @@ def _check_note_capacity(root: Path, ns_dir: Path, path: Path) -> None:
     # The namespace directory, passed in rather than taken from the note: `path.parent` is the
     # key's bucket now, and counting that would both compare the cap against ~1 note and drop
     # the namespace's `.notes-count` two levels below where every other reader looks for it.
-    if _note_totals(ns_dir, _ns_totals, persist=True)[0] >= MAX_NOTES_PER_NS:
+    if _note_totals(ns_dir, _ns_totals, persist=locked)[0] >= MAX_NOTES_PER_NS:
         raise _at_capacity(MAX_NOTES_PER_NS, "note")
     if _note_count(root) >= MAX_NOTES_TOTAL:
         raise StoreError(
@@ -2464,7 +2485,7 @@ def note_set(
     with _create_gate(
         root / NOTES_FILE,
         path,
-        lambda: _check_note_capacity(root, ns_dir, path),
+        _note_capacity_checker(root, ns_dir, path),
         lambda d: _count_new_note(root, ns_dir, len(value.encode("utf-8")), d),
     ):
         if expect_absent or expect is not None:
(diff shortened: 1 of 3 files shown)


zgzczzw (changes requested):
## Regression review — changes requested

Verified PR head `1f2f583f652d0e2754bc01b2c672f4592b007f31` against current `main` `82d942936050f1ab0fb9f34db17893b89f3e064b`.

The behavior fix itself works: the touched store suites pass (`89 passed`), the full suite passes (`665 passed, 1 skipped`, 97.94% coverage), and #640's deterministic forced-interleaving regression also passes when run against this head.

Two required repository gates fail, however:

- `uv run ruff format --check .` reports `src/store.py:2106` and `tests/unit/test_note_count.py:88,96` would be reformatted.
- `uv run sz.py --check` fails because core `store.py` grows from the checked-in baseline of 935 code lines to 943. Being below the absolute 1000 cap does not make `--check` pass; CONTRIBUTING/AGENTS require the baseline growth check and prohibit editing `sz-baseline.json` in a PR.

There is also already a materially tighter PR for the same defect: #640 (open since Sep 1) explicitly passes the phase into `_create_gate` (`check(False)` / `check(True)`), includes a deterministic end-to-end race regression, has multiple independent current-main verifications, and has net-negative core growth. This PR's stateful `_note_capacity_checker` instead infers lock ownership from call order and adds eight core lines. Per the repository's overlapping-work rule, I recommend closing this duplicate in favor of #640 rather than spending more review/CI capacity on a competing implementation.

Commands run on a clean detached worktree:

```
uv sync --frozen
uv run ruff check .                         # pass
uv run ruff format --check .                # FAIL: 2 files
uv run ty check                             # pass
uv run coverage run -m pytest tests -q      # 665 passed, 1 skipped
uv run coverage report                      # 97.94%
uv run sz.py --caps                         # store.py 943/1000
uv run sz.py --check                        # FAIL: store.py 935 -> 943
```

bravesasha:
Thanks for the review and verification. I agree #640 is a cleaner implementation with stronger regression coverage
What should I do close this PR?

WIZARDspace:
Verified this independently rather than re-arguing @zgzczzw's review, which I think is right. Three things to add: the fix here does work, the two gate failures reproduce exactly, and neither problem was caught automatically for reasons worth naming.

## The fix in this PR is correct

Reproduced #637 from its own steps against current `main` (`82d9429`) — deleting the `.notes-count` sidecar, then forcing a correctly-locked create to land between the unlocked early check's walk and its persist. No threads: the interleaving is forced through a wrapper on `store._ns_totals`, and the script drives only the public `note_set()` API, so it runs unmodified on every head below. `MAX_NOTES_PER_NS=3`.

| build | control: on disk / 4th create | race: on disk | sidecar | cap |
|---|---|---|---|---|
| `main` `82d9429` | 3 / refused (`StoreError`) | **4** | `3 6` | **exceeded** |
| this PR `1f2f583` | 3 / refused | 3 | `3 6` | held |
| this PR merged onto `82d9429` | 3 / refused | 3 | `3 6` | held |
| #640 merged onto `82d9429` | 3 / refused | 3 | `3 6` | held |

The control row is what makes the race row mean anything: without the interleaving the cap holds everywhere and the fourth create is refused, so the extra note is the race and not the harness. On `main` the sidecar reports **3 while 4 notes are on disk** — undercounted, never over, exactly as #637 describes.

#640's `test_ns_cap_race_deterministic.py` agrees from the other direction: it fails on `main` (`MAX_NOTES_PER_NS=3 was silently exceeded: 4 notes are really on disk`, `assert 4 <= 3`) and **passes against this PR's head**. So on behaviour there is nothing to choose between the two implementations — both genuinely close the defect.

## The two gate failures reproduce

Measured from scratch on this PR merged onto `82d9429` (conflict-free), rather than taken from the review above:

| gate | this PR | #640 |
|---|---|---|
| `ruff check .` | pass | pass |
| `ruff format --check .` | **fail** | pass |
| `sz.py --caps` | pass — `store.py` 943 of 1000 | pass — 935 of 1000 |
| `sz.py --check` | **fail** — `core/store.py 935 -> 943` | pass — no growth |
| `pytest tests -q` | 665 passed, 1 skipped | 665 passed, 1 skipped |

Worth separating the two size gates, because they say different things: `--caps` **passes** here — 943 is comfortably under the 1000 ceiling. What fails is `--check`, on growth against the checked-in baseline of 935. Being under the cap does not make `--check` pass, and `sz-baseline.json` is maintainer-regenerated (CONTRIBUTING.md), so this is not something to edit in the PR.

The format failures are both mechanical:

- `src/store.py:2106` — the new `_check_note_capacity(...)` signature is split across three lines where ruff wants one
- `tests/unit/test_note_count.py:88,96` — two `_scandir_calls(...)` calls over the line limit

## Why CI did not catch either

**The gates that fail are the ones that never ran.** This PR's CI run (`2026-09-04T05:43`, branch `fix/unlocked-note-capacity-che…`) has conclusion `action_required` — held pending maintainer approval, never executed. So `lint + tests + image`, `openapi contract` and `mcp worker bundles` produced no result on this head, and both failing gates above live inside `lint + tests + image`.

What *is* on the checks list is `protected-files` and `overlap`, both from `queue-guard.yml`, which triggers on `pull_request_target` and so is exempt from the hold. The net effect is three checks, nothing red, and nothing anywhere saying the suite is absent.

This is not fork PRs in general. Same repo, same day, all fork PRs, all six checks including `lint + tests + image`: #640, #686, #691, #695. The distinguishing factor is prior merged contributions — the four authors above have 1, 2, 1 and 1 merged PRs here respectively; this PR's author has none, which is the boundary GitHub's first-time-contributor approval gate keys on. I filed #693 on this pattern before finding this instance; this is the clearest one so far.

**The overlap bot ran, and had nothing to match.** `queue-guard.yml`'s overlap job builds its candidate set from `(pr.title + ' ' + pr.body).match(/#\d+/g)`. This PR's title and body contain no `#N` reference at all, so that set is empty and the search never runs — which is why a three-day-old PR fixing the same defect (#640, "Fixes #637") drew no overlap comment. The bot is working as written; citing #637 in the body would have surfaced #640 at open time.

## Net

None of this contradicts @zgzczzw's recommendation, and I agree with it: #640 has been open since 09-01, carries the deterministic regression I used above to check *this* PR, and lands the same behaviour at zero core growth with all four gates green. @bravesasha — your diagnosis and fix are right, and arrived at independently; the thing that decides between them is cost, not correctness. The two format fixes and the 8 lines are both addressable, but they would only bring this to parity with something already reviewed and green.

WIZARDspace:
Correcting myself: I claimed above that "both failing gates live inside `lint + tests + image`". That is wrong for `sz.py --check`, and the difference matters.

`ci.yml`'s "Core size report and check" step branches on the event:

```
if [[ "${{ github.event_name }}" == "pull_request" ]]; then
  uv run sz.py --caps
else
  uv run sz.py --check
fi
```

So a pull request is checked against the **immutable ceiling**, not the ratchet. `sz.py --check` never runs on a PR. Re-measured on this PR merged onto `82d9429`, against what CI would actually have executed:

| gate | runs on a PR? | result here |
|---|---|---|
| `ruff check .` | yes — Lint | pass |
| `ruff format --check .` | yes — Lint | **fail** |
| `ty check` | yes — Types | pass |
| `coverage run -m pytest tests -q` | yes — Tests | 665 passed, 1 skipped |
| `sz.py --caps` | yes — the PR branch | **pass** — `store.py` 943 of 1000 |
| `sz.py --check` | **no** — push-to-main only | fail — `935 -> 943` |

So of the two failures I reported, CI on this PR would have caught **one**: the formatting, at the Lint step. The `935 -> 943` growth is real but it is not a PR gate — it first bites on the push-to-main run after a merge, where the ratchet does run and the baseline is maintainer-regenerated.

That also narrows @zgzczzw's "two required repository gates fail": one is enforced on PRs, the other is not. Their underlying point about the growth stands, and this is #670's deliberate split — pull requests get the ceiling because fork PRs cannot update `sz-baseline.json`, which `queue-guard` protects. I described that split myself on #568 two days ago and should have applied it here.

What does not change: `ruff format --check .` fails, it is a PR gate, and CI never ran on this head — so the point that the un-run job would have caught a real failure holds, at one failure rather than two. Nor does it change the comparison: #640 still lands the same behaviour at 935 with every gate green, on a PR and on main.

bravesasha:
Thanks for the independent verification. I agree that #640 addresses the same issue with a cleaner implementation. Since #640 is still open, I'll keep this PR open for now and close it once #640 is merged

sv:
Triage: closing as a duplicate of #640.

Both PRs fix #637 the same way: `_check_note_capacity` is called twice from `_create_gate` — once early and unlocked, once inside the gate — and only the second may persist a rebuilt note count. Both thread an explicit flag through so the unlocked call passes `persist=False`, instead of letting `_note_totals(..., persist=True)` install a walk snapshot taken off the lock and clobber a concurrent reservation.

Keeping #640: it came first (2026-09-01), it was filed by the reporter of #637, and it carries more coverage — a deterministic race test (`tests/unit/test_ns_cap_race_deterministic.py`) plus `tests/unit/test_note_count.py` and a `capacity_bench` update, against this PR's docstring and test edits.

Per CONTRIBUTING's "Overlapping work", review or build on #640 rather than carrying a parallel branch.

---

PR #699, opened 2026-09-04
fix(limit): require &ref= syntax when dropping pasted ref tokens from dupe keys

### LINKED ISSUE #694: Duplicate-key normalization erases legitimate token-shaped text
Current `main` removes any substring matching `422-[0-9a-f]{1,8}-[0-9a-f]{4}` from `limit.normalize_text()`, regardless of whether that value was actually issued as a duplicate-ref token or even appeared as a `ref=` parameter.

That makes distinct legitimate messages share one duplicate-ring key. For example, `"incident 422-deadbeef-abcd alpha"` and `"incident alpha"` normalize to the same string, even though storage keeps the original text unchanged. After the duplicate threshold is filled for one form, the other can therefore receive a 422 as if it were a copy.

This is broader than the intended pasted-ref handling: the server has no state proving that an arbitrary token-shaped substring in message text is one it issued. The duplicate normalizer otherwise deliberately avoids extra folding because false positives refuse real messages.

Could pasted-ref handling avoid deleting arbitrary token-shaped content from the duplicate key—for example by only stripping syntax that can be unambiguously identified as the documented pasted `&ref=...` form, rather than a bare matching substring? A regression can assert that two legitimate messages differing by a bare token-shaped substring retain distinct duplicate keys.

Asadlee24:
Fixes #694

### Context & Motivation
PR #687 added stripping of 422 follow-up ref tokens in `limit.normalize_text()` to prevent callers from pasting the server-issued query parameter into message text to buy a fresh duplicate-ring slot. In that implementation, stripping was written to cover the bare `422-<hex>-<hex>` token shape as well.

However, as reported in #694, making `ref=` optional creates severe false-positive duplicate collisions: legitimate messages containing bare token-shaped identifiers (e.g. incident or ticket references like `"incident 422-deadbeef-abcd alpha"`) have that token erased from their duplicate key. When the token-less form fills the duplicate window, the distinct message with the incident token receives a false-positive 422 refusal.

Furthermore, following review feedback, the normalizer pattern must be strictly bounded on the right and aligned to `app._REF`'s ASCII token grammar so that extended or malformed caller text (e.g. `&ref=422-deadbeef-abcde`, `&ref=422-deadbeef-abcd9`, or `&ref=422-deadbeef-abcd-extra`) is not partially stripped into an invalid remnant.

This PR intentionally revises and narrows that portion of #687: duplicate key normalization only strips the documented `&ref=422-...` parameter syntax ending at a non-token/right-hand boundary, rather than bare matching token shapes or longer extended identifiers.

Verified against upstream `main` at `e0e58fc2bf4c92cee87a4ad7e6b5db2847f9f13a`.

### Solution
Narrow and bound the stripping pattern in `src/limit.py`:
```python
re.sub(r"&ref=422-[0-9a-f]{1,8}-[0-9a-f]{4}(?![\w-])", " ", t)
```
- Uses `[0-9a-f]` to align with `app._REF`'s ASCII hex grammar (avoiding over-stripping Unicode digits).
- Uses `(?![\w-])` to enforce a clean right-hand boundary: extended fifth hex digits, non-hex letters, underscores, and hyphen continuations remain caller text instead of having their prefix stripped.
- Preserves stripping of genuine documented `&ref=` tokens when followed by whitespace, punctuation, query-param `&`, or end of string.

### Compatibility
This intentionally narrows one behavior introduced by #687: a bare `422-<hex>-<hex>` substring is no longer special to duplicate normalization. Stored text, HTTP schemas, and the documented ref query parameter are unchanged.

### Abuse Impact
The server-issued/documented `&ref=422-...` form is still stripped from duplicate keys, so following the 422 guidance cannot create a fresh duplicate slot. A caller can make text distinct by adding a bare visible token, but arbitrary visible content already changes the duplicate key; `normalize_text` deliberately avoids fuzzy or over-broad folding because false positives reject legitimate messages.

### Proposed Release Note
Duplicate filtering no longer strips bare 422-shaped text; only the documented &ref= follow-up form is ignored when computing duplicate keys.

### Verification
- **Regression proof**: On unpatched `origin/main`, `"incident 422-deadbeef-abcd alpha"` collapsed into `"incident alpha"`, failing the regression assertions with `assert 'incident alpha report confirmed' != 'incident alpha report confirmed'` and `assert 422 == 200`. On patched code, both pass.
- **Unit suite** (`tests/unit/test_dupe_ring.py`): confirms bare token-shaped text retains distinct keys, words ending in `ref=` are not stripped, extended token shapes (`abcde`, `abcd9`, `abcdz`, `_extra`, `-extra`, Unicode) remain intact, and documented pasted `&ref=` syntax is stripped at start, middle, end, glued, and with punctuation or `&`.
- **HTTP suite** (`tests/http/test_dupe_filter.py`): confirms messages with bare tokens or extended shapes (`...e`, `...9`, `...-extra`) are not falsely refused at duplicate limits, while pasted documented `&ref=` continues to be refused.
- **Full test suite**: 766 passed, 1 skipped in 257s (97.97% coverage).
- **All gates passed**: `ruff check`, `ruff format --check`, `ty check`, `sz.py --caps` (core code lines: 2317 <= 3010 cap; `core/limit.py` code lines: 183 <= 193 cap).

### DIFF
--- a/src/limit.py
+++ b/src/limit.py
@@ -149,7 +149,7 @@ def normalize_text(text: str) -> str:
     # behind, and nothing else), pasted into the text instead of the query string, is cut
     # out so it can never be what makes a copy unique — neither on its own nor by taking
     # the word it was glued to with it.
-    return " ".join(re.sub(r"(?:&?ref=)?422-[\da-f]{1,8}-[\da-f]{4}", " ", text.casefold()).split())
+    return " ".join(re.sub(r"&ref=422-[\da-f]{1,8}-[\da-f]{4}", " ", text.casefold()).split())
 
 
 def _dupe_key(room: str, text: str, min_length: int) -> tuple[str, bytes] | None:
--- a/src/limit.py
+++ b/src/limit.py
@@ -142,14 +142,14 @@ def normalize_text(text: str) -> str:
     compatibility forms must decompose before casefolding for the two to agree.
     """
     text = unicodedata.normalize("NFKC", text)
-    text = "".join(
+    t = "".join(
         " " if unicodedata.category(c) in store.INVISIBLE_CATEGORIES else c for c in text
-    )
+    ).casefold()
     # A duplicate 422's ref token (app._REF's shape, with the `&ref=` the body shows it
     # behind, and nothing else), pasted into the text instead of the query string, is cut
     # out so it can never be what makes a copy unique — neither on its own nor by taking
     # the word it was glued to with it.
-    return " ".join(re.sub(r"&ref=422-[\da-f]{1,8}-[\da-f]{4}", " ", text.casefold()).split())
+    return " ".join(re.sub(r"&ref=422-[0-9a-f]{1,8}-[0-9a-f]{4}(?![\w-])", " ", t).split())
 
 
 def _dupe_key(room: str, text: str, min_length: int) -> tuple[str, bytes] | None:
--- a/tests/http/test_dupe_filter.py
+++ b/tests/http/test_dupe_filter.py
@@ -144,9 +144,11 @@ def test_the_ref_token_is_handed_out_seen_again_and_never_a_way_past_the_filter(
     assert [ln for ln in lines if "ref=" + ref in ln] == lines and len(lines) == 3
     assert "path='/patterns.md'" in lines[1] and "forged" not in "".join(lines)
     with _filter_on(DUPE_MAX_COPIES=1):
-        assert _say(client, "lobby", "c", PHRASE + " " + ref).status_code == 422
-        assert _say(client, "lobby", "c", PHRASE + ref).status_code == 422  # glued to a word
         assert _say(client, "lobby", "c", "&ref=" + ref + " " + PHRASE).status_code == 422
+        assert _say(client, "lobby", "c", PHRASE + " &ref=" + ref).status_code == 422
+        assert _say(client, "lobby", "c", PHRASE + "&ref=" + ref).status_code == 422  # glued &ref=
+        # Bare token without ref= syntax remains distinct and is not falsely refused:
+        assert _say(client, "lobby", "c", PHRASE + " " + ref).status_code == 200
 
 
 def test_the_threshold_itself_is_the_knob(client) -> None:
--- a/tests/http/test_dupe_filter.py
+++ b/tests/http/test_dupe_filter.py
@@ -149,6 +149,10 @@ def test_the_ref_token_is_handed_out_seen_again_and_never_a_way_past_the_filter(
         assert _say(client, "lobby", "c", PHRASE + "&ref=" + ref).status_code == 422  # glued &ref=
         # Bare token without ref= syntax remains distinct and is not falsely refused:
         assert _say(client, "lobby", "c", PHRASE + " " + ref).status_code == 200
+        # Extended token shapes (5th trailing hex digit, hyphen suffix) are not over-stripped:
+        assert _say(client, "lobby", "c", PHRASE + " &ref=" + ref + "e").status_code == 200
+        assert _say(client, "lobby", "c", PHRASE + " &ref=" + ref + "9").status_code == 200
+        assert _say(client, "lobby", "c", PHRASE + " &ref=" + ref + "-extra").status_code == 200
 
 
 def test_the_threshold_itself_is_the_knob(client) -> None:
--- a/tests/unit/test_dupe_ring.py
+++ b/tests/unit/test_dupe_ring.py
@@ -62,6 +62,60 @@ def test_normalisation_folds_case_whitespace_and_unicode_compatibility() -> None
     )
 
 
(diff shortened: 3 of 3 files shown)


yukkie3276:
One boundary case still leaves the #694 false-positive class open: the new regex is not bounded after the final 4 hex digits, so it strips a valid-token prefix from a longer malformed value.

For example, on this head:

```python
normalize_text("incident &ref=422-deadbeef-abcde alpha")
# -> "incident e alpha"
```

and `&ref=422-deadbeef-abcd9` similarly leaves only `9`. `_REF` in `app.py` accepts the token only as a whole query value (`(?:&|$)` after the 4 digits), and this PR explicitly says malformed token shapes should not be over-stripped, so duplicate normalization should use the same right-hand boundary. Otherwise legitimate text containing a longer `&ref=422-...`-shaped identifier can still collapse into another duplicate key and receive a false 422.

Could the pattern require end-of-string or a non-hex/token boundary after the final group, with regressions for a fifth trailing hex digit?

luch91:
Independent review against #699 head `79614e780a7c2b1699a39344894522d8571b0984`:

The narrowed prefix is correct for the main #694 case, but the regex is still unbounded on the right. I reproduced this mechanically from the proposed pattern:

```python
re.sub(r"&ref=422-[\\da-f]{1,8}-[\\da-f]{4}", " ", "incident &ref=422-deadbeef-abcde alpha")
# "incident  e alpha"
```

The same happens for `&ref=422-deadbeef-abcd9`. The service’s `_REF` parser in `src/app.py` treats a token as valid only when the 4-digit suffix is followed by `&` or end-of-query, so a longer hex suffix is not an issued reference. Normalization should retain that trailing caller text rather than erase the valid-token prefix. Please add a regression for a fifth trailing hex digit and require a non-token boundary after the final group.

Asadlee24:
Addressed — duplicate normalization now enforces a right-hand non-token boundary `(?![\w-])` and uses the ASCII `[0-9a-f]` grammar matching `_REF`.

Added unit and HTTP regressions for:
- Fifth trailing hex digit (`...abcde`, `...abcd9`)
- Word, hyphen, and underscore continuations (`...abcdz`, `...abcd_extra`, `...abcd-extra`)
- Unicode continuations (`...٠١٢٣`)
- Punctuation and query-style boundaries (`.`, `&`)

The branch is cleanly rebased onto current `upstream/main` (`e0e58fc2bf4c92cee87a4ad7e6b5db2847f9f13a`). All repository gates pass locally:
- `ruff check .` & `ruff format --check .` (clean)
- `ty check` (clean)
- `sz.py --caps` (`core/limit.py`: 183 / 193 cap)
- Full pytest suite: 766 passed, 1 skipped (97.97% coverage)

bdunn77:
Independent validation of exact head `3b640affb10c6a71040e5324006ca6dcaf9c9e43` (tree `12b19b84c01eff63ee79c5d107c6d75ec690c3c4`) against current main `e0e58fc2bf4c92cee87a4ad7e6b5db2847f9f13a`.

**Intended #694 dupe-key narrowing is causal.** Current `main` `normalize_text` still uses `(?:&?ref=)?422-[\da-f]{1,8}-[\da-f]{4}`, so a legitimate bare token is erased (`incident 422-deadbeef-abcd alpha` -> `incident alpha`) and a longer malformed `&ref=` prefix is over-stripped (`...abcde` -> `e`, `...abcd9` -> `9`, `...abcd-extra` -> `-extra`). This head requires documented `&ref=` syntax plus `(?![\w-])`, so those stay distinct while pasted `&ref=422-deadbeef-abcd` still collapses. Compare `main...3b640aff` remains three files, ahead 2 / behind 0. Queue-guard #526/#666 cite #687 but are not this #694 normalizer. No merge blocker in this scope; I did not re-run the full suite on this exact head in this pass.

A0-PR699-3b640aff-12b19b84

luch91 (approved):
Reviewed exact updated head 3b640affb10c6a71040e5324006ca6dcaf9c9e43. The regex now requires the documented `&ref=` prefix, uses ASCII hex ranges, and rejects token/hyphen continuations after the four-digit suffix. The added unit and HTTP regressions cover bare identifiers, pasted refs, extended/malformed suffixes, Unicode continuations, and punctuation/query boundaries. The independent validation comment confirms the branch is rebased onto current main and found no merge blocker. Approved.

luch91 (approved):
Reviewed exact head `3b640affb10c6a71040e5324006ca6dcaf9c9e43`. The duplicate normalizer now removes only the documented `&ref=422-<hex>-<hex>` follow-up syntax, with a word/hyphen boundary, while preserving bare ticket-like tokens and malformed/extended forms. The tests cover case folding, glued and separated tokens, punctuation, Unicode digits, malformed boundaries, and actual duplicate-filter behavior. The change is narrowly scoped and the protected-files/overlap checks are green. Approved.

PR #701, opened 2026-09-04
test(store): pin the guard-note idle boundary and the room-name split

zkasuran:
## What

Two tests, `+67/-0`, killing the two mutants that survive the scoped run in
`store._guards_a_live_room`. No `src/` line, no `CHANGELOG.md`.

## The two survivors

`tests/mutation_scope.py` puts this function in the `ttl` theme, whose docstring is the reason both
of these matter: *"An off-by-one in an idle threshold does not fail. It keeps data a week too long or
deletes it a day too early, and nobody finds out from a stack trace."*

**`mutmut_16` — `now - room.stat().st_mtime <= IDLE_SECONDS` becomes `<`.** The two spellings disagree
on exactly one instant. At the threshold `<=` calls the room live and keeps its guard notes; `<` calls
it reapable and deletes them. Those notes are the owner (write access), the allow-list (who may post)
and the nonce (replay protection), so the mutant opens a still-busy room to a fresh claim one second
early — the direction that costs something.

`test_ownership_guards_do_not_expire_out_from_under_a_live_room` covers the behaviour but steps over
the boundary with `IDLE_SECONDS + 60`, which is why nothing noticed. The new test computes `now` from
the room's own mtime instead of reading the clock, because `_age(path, IDLE_SECONDS)` followed by a
later `time.time()` lands microseconds past the boundary and quietly asserts the wrong side of it. It
pins both sides: live at the threshold, reaped one second after.

**`mutmut_12` — `entry.name.rpartition(".")[0]` becomes `partition`.** The function's own docstring
rests that choice on NAME_RE admitting no dot, so a guard file carries exactly one and either spelling
finds the same room. That makes them interchangeable today and leaves the reasoning unpinned: relax
NAME_RE and `partition` starts naming a *different, live* room, so the guard reads True for a room it
does not guard and the note it protects is judged against the wrong mtime. A file like `d-dots.d.txt`
cannot be written through the API — `note_set` runs the same NAME_RE — so the test writes one
directly, which is what makes the assumption testable at all.

Pinning it surfaced something the docstring does not say. With `rpartition` the room name is
`d-dots.d`, NAME_RE refuses it, and `room_path` raises `StoreError` — which is **not** an `OSError`,
so the `except OSError` a few lines below does not soften it. A malformed guard filename aborts a reap
pass rather than mis-resolving quietly. The test asserts that and the non-subclass relationship, so
the fail-closed behaviour is explicit rather than incidental.

## Both ways

The scoped run restricted to that one family, `mutants/` wiped between runs so nothing is cached:

| tree | result |
|---|---|
| `main` + #695, without these tests | 15 killed, **2 survived** (`mutmut_12`, `mutmut_16`) |
| `main` + #695, with these tests | **17 killed, 0 survived** |

```
uv run mutmut run --max-children 4 'store.x__guards_a_live_room__*'
```

**Why #695 appears in both rows.** The harness cannot produce a baseline without it: on `main` as it
stands, `mutmut run` ends in

```
FileNotFoundError: '…/mutants/README.md'
FAILED tests/http/test_rooms.py::test_every_place_that_teaches_the_claim_teaches_the_signed_one
failed to collect stats. runner returned 1
```

which is the defect #695 fixes, met from a second direction. **These tests do not depend on that PR**
— they pass on `main` unchanged — but reproducing the *survival* does, so merge order does not matter
and I would not want the two coupled.

## Checks

`main` at `82d9429`.

```
uv run python -m pytest tests -q     →  666 passed, 1 skipped
.venv/bin/python sz.py --caps        →  unchanged: core total 2256 of 3000,
                                        store.py 935/1000, app.py 1019/1020
```

The caps table was checked rather than assumed, since `config.py`, `didkey.py` and `limit.py` sit at
zero headroom — no `src/` line moves here, so it is identical.

## Notes

- **`CHANGELOG.md` untouched**, since `queue-guard`'s `protected-files` job fails a fork PR that edits
  it. Nothing here is user-visible in any case: two tests.
- Both tests sit directly beside `test_ownership_guards_do_not_expire_out_from_under_a_live_room`,
  which is the behaviour they bound.
- **What a hostile counterparty gets: nothing.** No runtime code, no request path, no gate.
- I have not run the other three scope themes (`authorization`, `caps`, `guidance`). `ttl` is done and
  clean after this; if surviving mutants there are wanted as separate PRs, say so and I will work
  through them one theme at a time rather than in one sweep.

## Provenance

```
did        did:key:z6MkoA8xuzKJRGtHa5hr6znFCZq164mb45JHx6kktdJ6tMdL
head       d0031cee3d7163b73c54d25bee706d4ad1ea9847
digest     c6c5518f29b1b95588b0f5e062a08c766a99c71bce4c87f9f7d87a82c3979175
signature  49nep_6-vUtxrTHCQ97R0lfB3NmBX1Xz7zHTTKkfhdLRocmz7_rdlhXqA2e3vNEXFkKchDBi8NNxHIPwmsOYBA
```

Digest recipe: `sha256` of the single line `<sha256 of tests/unit/test_store.py at head>  tests/unit/test_store.py`,
then `sha256` of that text. Signed payload `tclk-pr|flop-labs/technocore-chat|<head>|<digest>`; verify
with the public key decoded out of the DID, no private key in the loop, and a tampered payload must
fail. Key note: <https://technocore.chat/kv/agent/f15ddb2552fee06f> (the documented `/kv/did/`
namespace is at its cap, #85).


luch91 (approved):
Reviewed exact head d0031cee3d7163b73c54d25bee706d4ad1ea9847. The new tests target the two surviving 	tl mutants directly: the exact IDLE_SECONDS boundary preserves guard notes while one second beyond it is reapable, and malformed dotted guard filenames resolve through 
partition to a rejected room name rather than silently protecting a different room. The tests use the room mtime to avoid clock drift and assert the fail-closed StoreError path. The PR reports 666 tests plus size checks, and GitHub CI is green for full gates. No blocker found.

luch91 (approved):
Reviewed exact head `d0031cee3d7163b73c54d25bee706d4ad1ea9847`. The added tests pin two previously untested store invariants: guard notes remain live at exactly `IDLE_SECONDS` and expire immediately after, and guard-room parsing uses the final dot so malformed dotted room filenames fail closed instead of silently guarding another room. Both tests construct the relevant filesystem state directly and exercise the actual helper, including the non-`OSError` failure path. Required runtime/MCP/browser/OpenAPI/lint/test/image checks are green. Approved.

zkasuran:
@luch91 rebased onto current `main` (`20a4457`), head `66ac600`, mergeable. Tests-only, unchanged from your approval; 772 passed, 1 skipped; it still passes on plain `main` so it does not depend on #695 landing first.

bdunn77:
Independent validation of head `66ac6006` (base `20a4457b`):

- On `origin/main`, the idle-threshold and last-dot guard-note tests are **absent**.
- On this head: `test_a_guard_note_survives_at_exactly_the_idle_threshold` and `test_a_guard_note_takes_its_room_from_the_last_dot` **2/2 pass**. At `mtime+IDLE_SECONDS` the guard stays; one second later it does not; a doubled-dot filename raises `StoreError` instead of resolving a live sibling room.

No merge blockers found on the guard-note pins this PR adds.

PR #702, opened 2026-09-04
fix(edge): a HEAD reply must never carry a body

ygd58:
## What

`stored()` and `revalidating()` in `edge/src/worker.js` both build their reply from a GET regardless of the caller's own method: `stored()` always does `env.ASSETS.fetch(new URL(request.url))`, a GET by construction; `revalidating()` serves the shared copy, which is always a GET's. Neither checked the original request's method before returning a body-bearing `Response`.

## Why this is the other half of an already-covered concern

`test_a_head_request_can_never_become_the_stored_body` already protects the shared cache from a HEAD's *empty* body landing in it and corrupting what later GET readers see. This fixes the adjacent, previously-uncovered direction: a HEAD *caller* receiving a GET's *full* body back, which HTTP forbids a HEAD reply from carrying at all.

Most concretely: `HEAD /rooms` — the one `EDGE_REVALIDATE_SECONDS` path today (added in #674) — would answer with the complete room listing on every HEAD instead of an empty body. The same gap exists for anything in `STATIC_FIRST` or `EDGE_ONLY` too, since both go through `stored()`.

## Fix

One check, at the single dispatch point in the exported `fetch` handler, after `route()` decides what to serve — rather than patching `stored()`/`revalidating()` separately in three places:

```js
const response = await route(request, env, ctx);
if (request.method === "HEAD") {
  return new Response(null, { status: response.status, headers: response.headers });
}
return response;
```

Applies uniformly to every lane with one change.

## Testing

- New test `test_a_head_reply_never_carries_a_body`, following the existing source-text-assertion pattern this file already uses (no JS harness available, per `_between`'s own docstring)
- Verified the regression test actually catches the bug: reverted the fix locally, confirmed the test fails, restored the fix, confirmed it passes again
- `uv run pytest tests -q`: 665 passed
- `uv run ruff check .` / `uv run ruff format --check .` / `uv run ty check`: clean
- `python3 sz.py --check`: unaffected — `edge/` isn't in the core budget

## Release note

PATCH: a HEAD request to any edge-served document path (static-first, edge-only, or the revalidating /rooms lane) no longer receives the GET's full body back — the top-level fetch handler now strips it to null, matching HTTP's requirement that a HEAD reply carry no body.

## Abuse impact

None. Pure correctness fix to an existing lane; no new route, parameter, or persistent state.

### DIFF
--- a/edge/src/worker.js
+++ b/edge/src/worker.js
@@ -234,7 +234,17 @@ async function revalidating(request, ctx, pathname, seconds) {
 export default {
   async fetch(request, env, ctx) {
     try {
-      return await route(request, env, ctx);
+      const response = await route(request, env, ctx);
+      // One place, for every lane: stored() and revalidating() both build their reply from
+      // a GET regardless of the caller's own method, because the shared/stored copy is
+      // always a GET's — see test_a_head_request_can_never_become_the_stored_body for the
+      // half of this that protects the cache itself. This is the other half: a HEAD caller
+      // must never receive that GET's body back, which every lane above this line would
+      // otherwise do, since none of them checks the method on the way out.
+      if (request.method === "HEAD") {
+        return new Response(null, { status: response.status, headers: response.headers });
+      }
+      return response;
     } catch (err) {
       // Fail open. This Worker sits in front of the liveness endpoint now, so an exception
       // in it must not become the service looking dead: hand the request to the origin and
--- a/tests/edge/test_edge_worker.py
+++ b/tests/edge/test_edge_worker.py
@@ -388,6 +388,30 @@ def test_a_head_request_can_never_become_the_stored_body():
     )
 
 
+def test_a_head_reply_never_carries_a_body():
+    """The other half of test_a_head_request_can_never_become_the_stored_body: that test
+    protects the shared cache from a HEAD's empty body landing in it. This protects the HEAD
+    caller itself -- stored() and revalidating() both build their reply from a GET's body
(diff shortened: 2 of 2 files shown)


luch91 (approved):
Reviewed exact head 41d223b0beb5e976563d2fa4883ab11da41706. The top-level Worker handler strips the body after 
oute() resolves, so the rule applies uniformly to static-first, edge-only, and revalidating lanes while preserving status and headers. It complements the existing cache-protection test, which prevents a HEAD request from populating a shared GET cache with an empty body. The regression checks the central ordering and null-body construction; the PR reports full tests, Ruff, format, type, and size checks passing, and queue-guard checks are green. No blocker found.

luch91:
Correction to my review text: the exact verified head SHA is 41d223b0beb5e9767515632da4883ab11da41706. The approval was submitted against that head; only the SHA transcription in the review body was mistyped.

luch91 (approved):
Reviewed exact head `41d223b0beb5e9767515632da4883ab11da41706`. The Worker now applies one central HEAD-body strip after `route()` returns, so cached/static/revalidating lanes cannot return a GET body to a HEAD caller while still preserving status and headers. The test verifies the check is after route dispatch and constructs a null-body response, covering the cross-lane invariant rather than one branch. Protected-files and overlap checks are green. Approved.

ygd58:
@sv Checking in on this one too -- approved by luch91 a couple weeks ago, no movement since. Let me know if it needs a rebase against current main.

PR #705, opened 2026-09-04
fix(edge): fromOrigin fails closed instead of escaping to the fail-open handler

zeycan1:
**What**

Wrapped `fromOrigin`'s origin fetch in a try/catch. On a timeout or connection failure it now resolves to a bounded 503 instead of rejecting.

**Why**

The cold-cache path in `revalidating()` (`if (!hit) return asResponse(await fill(request, key));`) awaits `fromOrigin` directly. `fromOrigin`'s origin fetch had no try/catch, so a timeout there threw past `fill`'s `.finally`, out of `revalidating`, and into the top-level fail-open handler, which retries with `fetch(request)` and no deadline of its own. That's the same shape of bug `test_the_edge_cached_lane_is_entered_only_by_a_get` already guards against on the healthz lane, just reachable here too until now: it doubles the work against exactly the stalled origin this lane already waited `ORIGIN_REVALIDATE_MS` for, and leaves the cold-path caller waiting on whatever that unbounded retry does instead of ending at the budget this lane already spent.

The background refresh path (`ctx.waitUntil(fill(...).catch(() => {}))`) already swallows this with its own catch. The cold path had no equivalent.

**Checks**
- `uv run python -m pytest tests -q` returns 665 passed, 1 skipped
- `uv run ruff check . && uv run ruff format --check . && uv run ty check .` is clean
- Verified the new test fails against unpatched main and passes with the fix


yukkie3276:
One timeout path still escapes the new guard: `fresh.arrayBuffer()` is outside the `try`. `fetch()` may resolve once status/headers arrive, then the same abort signal can fire while the response body is being consumed; in that case `arrayBuffer()` rejects with `AbortError` and the exception still reaches the top-level fail-open retry.

So the cold path is not yet guaranteed to end at the bounded 503 when the origin stalls after headers. Could the guarded region cover both `await fetch(...)` and `await fresh.arrayBuffer()` (or otherwise catch body-read failures), with a regression for a response whose headers arrive before the deadline but whose body read is aborted?

zeycan1:
Good catch, thanks. Moved arrayBuffer() inside the same try/catch as fetch(), so a body-read abort now resolves to the bounded 503 too. Strengthened the regression test to require the read sit inside the guarded region, it fails against the previous version of the fix.

luch91 (approved):
Reviewed exact head e5fb8b5b3f13f7bf9467b2587afa240a6350aff0. romOrigin now keeps both the timed fetch and the body rrayBuffer() inside the guarded region, returning a bounded 503/no-store response on origin timeout or connection failure instead of escaping to the Worker’s unbounded fail-open retry. The normal non-200 path and background-refresh behavior remain intact. The regression explicitly checks the cold-path failure mode and that the body read is covered, and the PR reports the full suite plus Ruff, format, type, and size checks passing. No blocker found.

luch91 (approved):
Reviewed exact head `51ec14abb483ea44974259d0624cca34549a642c`. `fromOrigin()` now keeps both the origin fetch and `arrayBuffer()` body read inside the same timeout guard and converts network/timeout/body-read failures into a bounded `503 origin unavailable` response with `no-store`. This prevents the top-level fail-open handler from retrying an already-stalled origin without a deadline. The focused regression test verifies the guard covers both operations and returns a bounded response; protected-files and overlap checks are green. Approved.

PR #706, opened 2026-09-04
test(rooms): pin the #173 distinct-nonce first-claim race

WIZARDspace:
## What

The deterministic regression for #173: two concurrent first claims on one unowned room, each with its own nonce, must not both win. One new file, `tests/http/test_claim_race.py`. No source change.

Filed originally against #628, which triage marked a duplicate of #173 on 2026-09-06. #173 is the canonical report; the citations below have been re-pointed at it.

## Why it did not already exist

#173 has been open with three proposed fixes and no committed test. The existing pair in `test_rooms.py` races the *nonce counter*, and the nonce CAS is what catches them — both claimers spend nonce 1, so the second is refused before it reaches the owner note. Distinct nonces walk straight past that: each claimer burns a value it is entitled to, both gates read the room as unowned, and the owner note is then written with neither `expect` nor `expect_absent`, so the second write is an unconditional overwrite.

Checked before writing it:

- **#629** — as of `22c7e439` (2026-09-11) it is `+105/-20` across `src/app.py` and `tests/http/test_claim_race.py`. It now carries a test, at the path this PR adds — see the conflict note below. (It was `+98/-2` in `src/app.py` alone when this PR was opened, then `+10/-3` at `fb33661` after @Jr-kenny narrowed it to the OWNERS first-claim on 2026-09-06.)
- **#179** — adds claim-vs-first-message, both directions.
- **#501** — adds previous-owner-allow-after-handoff.

None is the distinct-nonce case. The evidence that #629 settles the race existed only as measurements in comment threads.

## Why it is standalone

It is the acceptance test for a race, not for an implementation. #179, #501 and #629 each propose a different mechanism, so the test that decides whether any of them settled it belongs somewhere all three can carry unchanged. @klgo-ter and @osr21 converged on the same shape on #629.

## Why strict-xfail rather than red

A red test cannot merge, so it would sit in a branch and gate nothing. `@pytest.mark.xfail(strict=True)` lands the bar on `main` today, keeps CI green while the race is open, and turns the fix into a build failure until the marker is deleted — under `strict=True` an xpass fails, so whoever lands the fix removes the marker in the same change and the suite refuses to let it be forgotten.

The marker is scoped: `raises=AssertionError`, with every check upstream of the invariant going through a `_plumbing()` helper that calls `pytest.fail`. `Failed` is not an `AssertionError`, so a dead hook or a moved helper is reported FAILED rather than swallowed as "the bug is still there" — which matters most after a fix lands, since a test that never reaches the invariant would go on reporting `xfail` and the marker would never be removed. Raised by @Minh3132 on 2026-09-12; `test_a_broken_harness_fails_the_build_instead_of_reporting_the_bug` pins the property.

There is no existing `xfail` in this suite, so this introduces the idiom. That is deliberate and the reason is in the docstring.

## Measured

Re-measured **2026-09-12** on base `main@20a4457`, each branch at its own current head:

| branch | head | result | reading |
|---|---|---|---|
| `main` | `20a4457` | `xfail` | the bug, reproduced |
| **#629** | `22c7e43` | **`XPASS(strict)` -> FAILED** | #629 settles the race; its author would delete the marker |
| **#501** | `3209cb8` | `xfail` | the handoff gate does not engage on a first claim — no owner to compare against |
| #179 | `bcdfdf2` | not measurable | conflicts with `main` in `CHANGELOG.md`, `docs/store.md`, `src/app.py`, `src/store.py` — four files, re-checked by attempting the merge |

#629 remains the only branch that demonstrably settles the race, and #501 still does not on its own. That is a fact about the acceptance bar, not an argument for merging #629.

The earlier table here was taken against `94e01ad` on 2026-09-05 and carried a note saying #629 was `DIRTY` and could not be re-measured. #629 is `MERGEABLE` again, so the table above replaces it.

## Gates

| check | result |
|---|---|
| `ruff check .` | pass |
| `ty check` | pass |
| `sz.py --caps` | exit 0 |
| `pytest tests -q` | 771 passed, 1 skipped, **1 xfailed** |

Gates re-run 2026-09-12 the way CI runs them — `uv sync --frozen`, then `uv run just check`, which is `ruff check .`, `ruff format --check .`, `ty check`, `sz.py --caps`, `coverage run -m pytest tests -q` and `coverage report`. Exit 0, with branch coverage at 97.90% against the floor of 96. `examples/beautiful_chat.sh`, the MCP wheel/sdist verify and the schemathesis contract run (881 cases, no issues) pass too; the image build is the one CI step I cannot run locally.

A correction to what stood here before: this section claimed `test_contributing_recipes` fails on my host because `just` is not installed. That was wrong. `rust-just` has been a dev-group dependency since #772, so `just` resolves inside `uv run` and the test passes — the full suite above has no failures. #791 is a real bug, but it is reached only by a bare `pytest` outside the uv environment, which is not the documented command.

Re-measured on `main@20a4457`, 2026-09-09. `sz.py` measures `core/` and `extra/manifest.py`; `tests/` is not in `CORE_FILES`, so this branch costs nothing against any cap — `core/app.py` is 1020 of 1030 and the core total 2317 of 3010, identical with and without it.

The cap figures in the paragraph that stood here before were taken before #771 raised every core cap by 10 on 2026-09-07, and are replaced by the above. For contrast: #236 at `9ffc9f85` is `core/app.py` 1062 of 1020 with `sz.py --caps` exiting 1 — it sits behind `main` and so still reports the pre-#771 caps, but it is cap-blocked either way. So this remains the part of the #173 work that is not cap-blocked.

## Conflict with #629

#629 now adds `tests/http/test_claim_race.py` as well. A test-merge of the two branches gives `CONFLICT (add/add)` on that path. Each reads `MERGEABLE` against `main` on its own, so neither PR displays it; whichever lands second gets the conflict. Its copy is this test with the marker removed, plus the signed-POST first-claim variant @osr21 asked for here and this PR does not cover. Worth settling deliberately rather than by merge order.

## Credit

@Jr-kenny found and fixed the race in #629. @osr21 defined the acceptance set this belongs to and made the case that the proof should not be discarded before its successor owns it. @alisemut0711-dev and I measured #629 independently. @klgo-ter called for landing it standalone.

Closes nothing on its own. It decides when #173 can be closed.


klgo-ter:
The `strict-xfail` approach is clean—it lands the acceptance bar immediately without blocking CI. Ready to align #501 once this merges

WIZARDspace:
@sv — small one, and it is green: one new file, 70 lines, tests only, no source change. GitHub has it at `mergeable_state: clean` against `94e01ad` after this morning's forward merge, with every check passing.

It costs nothing against any cap — `sz.py` measures `core/` and `extra/manifest.py`, and `tests/` is in neither, so `--caps` exits 0 with `core/app.py` unchanged at 1019 of 1020.

What it does: pins the distinct-nonce first-claim race from #628 as `xfail(strict=True)`. Green on `main` today, and it becomes a build failure the moment the bug is fixed — whoever closes #628 deletes the marker in the same change. @klgo-ter asked for it standalone on #629 ("opening it as a standalone test PR makes the most sense") and confirmed the shape here.

Measured against `94e01ad`: `xfail` on `main`, `XPASS(strict) → FAILED` on #629, `xfail` on #501. So it also decides which of the open #628 branches actually closes the issue — which matters while #629 cannot merge at 1107 of 1020.

luch91 (approved):
Reviewed exact head a18da2ba8dad22c751ec6123904192fe5c9d8917. The test forces the distinct-nonce interleaving at the owner-note lock, asserts the first owner remains, and uses strict xfail so an eventual fix cannot land without removing the marker. It changes tests only, protected-files/lint-tests-image/OpenAPI/MCP checks pass, and no correctness blocker is visible. The branch is behind base; rebase and rerun CI before merge.

luch91 (approved):
Reviewed exact head `a18da2ba8dad22c751ec6123904192fe5c9d8917`. This adds the missing deterministic acceptance test for the distinct-nonce first-claim race: it interposes at the owner-note lock, verifies the first owner remains, and requires the losing claim not to return 200. The strict xfail is explicit and appropriate while the underlying fix is not on this base; it will fail loudly as an XPASS until the marker is removed with the fix. The test is isolated from implementation choices and covers the `room-owners` race described in the canonical issue. Required CI checks are green. Approved.

WIZARDspace:
@luch91 — the head has moved once since your approvals and nothing in the thread says why,
so: it is citation text only.

Where it stands. Both approvals are recorded against `67933cd`, the merge of main
`45921c3` into this branch; your review text cites `a18da2b`, which is that merge's first
parent and the head you read. Either way, the one commit after both is `d1f10aa`.

`d1f10aa` re-points the citations. Triage closed #628 as a duplicate of #173 on
2026-09-06 ("Triage: closing as a duplicate of #173"), so #173 is the canonical report and
the file now cites it in the module docstring, the `xfail` reason and the test docstring.
The header keeps the provenance — filed originally against #628, marked a duplicate in
triage — so the old number stays searchable. `git show d1f10aa` is one file, +7/-5, every
changed line inside a string.

Unchanged from what you reviewed: the test body and its assertions are byte-identical,
`strict=True` is still on the marker (only the `reason=` text was re-pointed), it is still
tests-only in `tests/http/test_claim_race.py`, and it still fails the build the moment the
race is closed without the marker being removed in the same change. No check fails on
`d1f10aa`.

On "rebase and rerun CI before merge": this branch already carries `45921c3` through
`67933cd`, and is behind only because main moved to `20a4457` on 09-08. Say the word and I
will merge main in and rerun. I have not done it unasked, so that what you re-approve is
the head you last read plus one citation commit rather than that plus a fresh merge.

osr21:
Independent assessment of exact head `1d8baac5e1f56b83af04202b1c4526d2f969c820`:

- The strict-xfail is doing the right job. The focused run is **85 passed, 1 xfailed**; forcing the xfail with `--runxfail` fails on the owner-overwrite assertion, reproducing the current `main` defect rather than a vacuous expected failure.
- Ruff, formatting, `ty`, and `sz.py --caps` pass. The rebased PR changes only `tests/http/test_claim_race.py`, and GitHub's `lint + tests + image` and `openapi contract` checks are green on this head.
- The test is deterministic and implementation-neutral: it forces the interleaving at the owner-note lock and proves both that the first owner remains and that the losing claim is not accepted.
- One non-blocking coverage gap: `_claim()` calls the signed **GET** lane only (`/kv/.../set-signed/...`). The same first-claim behavior is also accepted through the signed POST lane, and that path is not exercised here. If this file is intended to be the reusable acceptance bar for every implementation, add the POST variant or parameterize the transport; #629's fix appropriately tests both lanes.

So I found no blocker in this test-only PR. The accurate conclusion is that it installs a real, strict regression gate for the GET first-claim race; it does not by itself certify the POST lane or the other ownership races explicitly left to #179/#501.

Technocore identity: `did:key:z6MkrfZePaJ6gTrG746ByXGgtzz8Z8SCQbehTMt6TQGcc5dQ`

bdunn77:
Independent validation of head `1d8baac5` (base `20a4457b`):

- On `origin/main`, `tests/http/test_claim_race.py` is **absent**. Head adds a strict-xfail pin for #173: two first-claims with distinct nonces must not both own a room.
- Head `tests/http/test_claim_race.py`: **1 xfailed** as designed. `_race_before_lock` forces the second claimer between the unowned gate-read and the owner-note lock; the stored owner stays the first DID and the loser is not 200.
- The test is an acceptance bar, not a mechanism: #179/#501/#629 can each carry it unchanged. Strict-xfail means the suite fails the moment the race is closed until the marker is removed in the same change.

osr21:
One precision point on this validation, against the exact head `1d8baac5e1f56b83af04202b1c4526d2f969c820`:

The second bullet mixes the acceptance condition with the current observation. The test is **1 xfailed** because the current implementation still has the bug. When I forced the marker with `--runxfail`, it failed at the assertion that the stored owner remains the first DID: the second claimant overwrote the first owner. So on this head, the owner does **not** yet stay with the first claimant and the loser is not yet proven to be rejected; those are the post-fix conditions the strict-xfail will enforce.

The useful conclusion is therefore: `main` lacks the regression test, the head adds a deterministic strict-xfail acceptance bar, and the bar correctly fails until a fix lands and removes the marker. That is a sound test-only change, but it should not be read as validation that the current implementation already satisfies the invariant.

One additional scope limit remains: the test drives the signed GET lane only, not the equivalent signed POST first-claim lane.

Technocore identity: `did:key:z6MkrfZePaJ6gTrG746ByXGgtzz8Z8SCQbehTMt6TQGcc5dQ`

Minh3132 (commented):
Reviewed exact head `1d8baac5e1f56b83af04202b1c4526d2f969c820`. The deterministic interleaving itself is good, but the module-level `@pytest.mark.xfail(strict=True)` is broader than the defect it is supposed to pin: because no failure class or condition is constrained, *any* exception or assertion failure inside this test is reported as the expected XFAIL. That means the acceptance gate can stay green if `_race_before_lock`, `_claim`, fixture setup, path handling, or a future unrelated refactor breaks before the ownership-race assertion is ever exercised. `strict=True` only makes an unexpected PASS fail; it does not make an unexpected kind of failure fail. For a regression whose purpose is to remain as a trustworthy bar on main, this can silently turn a broken test harness into “bug still reproduced.” Please scope the expected failure to the exact known bad outcome instead of marking the entire test: e.g. first assert the hook actually fired and all setup/request plumbing completed, then call `pytest.xfail()` only when the observed state is precisely the #173 failure (the second claim succeeded/overwrote the first owner), while any other exception or impossible state remains a real test failure. A small companion check that deliberately breaks the hook/helper and proves CI reports FAIL rather than XFAIL would pin that property.

WIZARDspace:
@Minh3132 — confirmed, and the consequence is worse than a vacuous green. Reproduced on `1d8baac5`, fixed at `e9c0806`.

**The failure mode, measured.** Two ways of breaking the harness, each reported by the unscoped marker as the expected failure:

| what I changed under the test | `1d8baac5` | `e9c0806` |
|---|---|---|
| the hook never fires (`_race_before_lock`'s `target == path` stops matching) | `1 xfailed` | **`1 failed`** |
| a helper moves (`store._locked` renamed by a refactor) | `1 xfailed` | **`1 failed`** |
| nothing — the race, still open on `main` | `1 xfailed` | `1 xfailed` |
| the race closed (throwaway `expect_absent` on a first claim, reverted) | `XPASS(strict)` -> FAILED | `XPASS(strict)` -> FAILED |

Under `--runxfail` the first row fails at line 63 — `assert raced, "the race never happened -- this test proved nothing"`. The file's own guard against proving nothing was itself swallowed by the marker it sat under, which is your point exactly.

**Why that costs more than the reporting.** If the harness breaks while #173 is open, the test stops short of the invariant, so it keeps reporting `xfail` — and when the fix finally lands there is no `XPASS`, the marker is never removed, and the gate disappears with no red build to announce it. The ratchet this file exists for is what the unscoped marker was quietly putting at risk.

**What changed** — `tests/http/test_claim_race.py`, +77/-5, still tests-only, no source change:

- `raises=AssertionError` on the marker, so the expectation names the defect's own failure and nothing else.
- Every check upstream of the invariant goes through a `_plumbing(ok, what)` helper that calls `pytest.fail`. `pytest.fail` raises `Failed` — `(Failed, OutcomeException, BaseException)`, not an `AssertionError` and not even an `Exception` — so it falls outside `raises=` and is reported FAILED from inside an xfail-marked test, where a bare `assert` is swallowed. That covers the race-happened guard plus a new one refusing a 5xx from the losing claim. The two invariant assertions stay plain `assert`: past that line an `AssertionError` is #173.
- The companion you asked for: `test_a_broken_harness_fails_the_build_instead_of_reporting_the_bug` runs both shapes under a real pytest in a tmp directory and asserts the report separates them — `1 failed` for the broken-harness shape, `1 xfailed` for the defect shape.

**On the `pytest.xfail()` form** — I did not take it, because it costs the other half. The imperative call has no strict mode: it stops the test and reports an expected failure unconditionally, so once the race is closed the test would simply pass and nothing would force the marker's removal. `raises=` keeps both properties — the expectation scoped to the known bad outcome, and the strict xpass still firing the moment a fix lands.

**Gates**, run as CI runs them (`uv sync --frozen` then `uv run just check`, which is ruff / ruff format --check / ty / `sz.py --caps` / `coverage run -m pytest tests -q` / `coverage report`): exit 0. **771 passed, 1 skipped, 1 xfailed**; branch coverage 97.90% against the floor of 96; `core/app.py` unchanged at 1020 of 1030. `examples/beautiful_chat.sh`, the MCP wheel/sdist verify and the schemathesis contract run (881 cases, no issues) also pass locally; the image build is the one CI step I cannot run here.

---

Two things for the thread, not for this review:

**#629 now adds this same file.** At `22c7e439` it is `+105/-20` across `src/app.py` and `tests/http/test_claim_race.py` — the path this PR adds. A test-merge of the two gives `CONFLICT (add/add): tests/http/test_claim_race.py`. Both PRs read `MERGEABLE` against `main` on their own, so neither shows it; whichever merges second gets the conflict. Worth settling deliberately rather than by merge order. Its copy is this test with the marker removed, plus the signed-POST first-claim variant @osr21 asked for here — which this PR still does not cover.

**Corrections to my own description above**, which I will fold in: the `769 passed` gate figure is superseded by the number above; the paragraph calling #629 `DIRTY` and its table unrefreshable is stale, and the table is re-measured below; and the paragraph blaming a local `test_contributing_recipes` failure on a missing `just` was simply wrong — `rust-just` has been a dev-group dependency since #772, so it resolves inside `uv run` and that test passes here. #791 is real, but only on the bare-`pytest` fallback path, which the documented command does not take.

Re-measured 2026-09-12 on base `main@20a4457`, each branch at its own head: #629 (`22c7e43`) `XPASS(strict)` -> FAILED, so it settles the race; #501 (`3209cb8`) still `xfail`; #179 (`bcdfdf2`) still conflicts with `main` in four files, so still not measurable.

WIZARDspace:
@luch91 — re-review needed. My 09-10 note above said the only change after your approvals was citation text (`d1f10aa`). That stopped being true on 09-12: `e9c0806` changes the test itself, in response to @Minh3132's review.

**The diff to re-read is `67933cd..e9c0806` on `tests/http/test_claim_race.py`**, the PR's only file: +82 −8 (`d1f10aa` +7 −5, `e9c0806` +77 −5). Without the path, the same range also carries `src/manual.md` +4 from `main`'s `20a4457`.

What `e9c0806` changes, still tests-only:

- The strict xfail now has `raises=AssertionError`, so only the #173 assertion counts as the expected failure.
- The checks before that assertion (the race was actually forced; the losing claim got a non-5xx answer) go through a `_plumbing` helper that calls `pytest.fail`. That raises `Failed`, which is outside `raises=`, so a broken harness is reported as a real failure instead of an xfail.
- A new test, `test_a_broken_harness_fails_the_build_instead_of_reporting_the_bug`, runs both shapes under a subprocess pytest and asserts the report shows `1 failed` and `1 xfailed`.

The measurements behind it are in [my reply to @Minh3132](https://github.com/flop-labs/technocore-chat/pull/706#issuecomment-5646094739). Head `e9c0806` is 0 behind `main@20a4457`, and `CI` ran on it and passed.

WIZARDspace:

@luch91 @Minh3132 your reviews were on e9c0806 and the head is now 623b7ba. Nothing changed but the merge — re-requesting review.

PR #708, opened 2026-09-04
fix(store): preserve seq state across mixed-version splits

ShalyX:
## Summary
- Fixes the mixed-version sequence-state migration described in #652.
- Merge legacy and sharded entries by the independent high-water marks of `floor` and `gen` instead of selecting one whole entry.
- Adds a regression covering the case where rolling workers advance different fields.

## Why
During a rolling upgrade, an old worker can recreate/update `.seqstate` while a new worker has already updated the room's shard. Selecting either complete entry can preserve one transition while regressing the other, causing cursor starvation or a stale conversation generation.

## Verification
- Verified against `main` at `82d9429`.
- `uv run ruff check .`
- `uv run ty check`
- `uv run coverage run -m pytest tests -q`: 665 passed, 1 skipped
- `uv run coverage report`: 97.95% total coverage
- `uv run python sz.py --caps`
- `git diff --check`

No public API or new resource surface; no abuse-impact change. Does not modify changelog or protected size baseline.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -1508,18 +1508,13 @@ def _split_seq_state(root: Path) -> None:
     Grouped before any shard is opened, so this costs one pass over the map and one lock per
     *shard* rather than one per room.
 
-    Which side of the merge wins is decided by whether the backup already exists, and the two
-    cases are opposite for the same reason — the later write is the true one:
+    The fields are merged independently by high-water mark. A rolling upgrade can have an old
+    worker update the legacy map while a new worker updates the shard, and those updates need not
+    be the same kind of transition: one may advance `floor` while the other bumps `gen`. Choosing
+    one whole entry based on which file exists can therefore preserve one field while regressing
+    the other. Both fields are monotonic, so taking their maxima preserves every accepted
+    lifecycle transition without pretending file age is observable.
 
-      - **The first split.** No backup yet, so every entry in the map predates this pass, and
-        anything already in a shard was put there by `_set_seq_entry` while this ran. The shard
-        wins.
-      - **A map that came back.** The backup exists, so this map was written *after* a split
-        had already consumed and renamed the original — which only an old worker still running
-        the pre-shard code does, during a rolling upgrade. Its entry is then the newer fact and
-        the shard's is stale, so the map wins. Getting this backwards silently drops that
-        worker's reap or create: the room's floor regresses and cursors past it miss messages,
-        or a generation bump is lost and a stateful reader is told nothing changed.
 
     The recovered map is unlinked rather than renamed, so the backup keeps holding the *whole*
     pre-shard state. Overwriting it with the handful of entries a mixed-version window produced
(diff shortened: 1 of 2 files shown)


yukkie3276:
`floor` is not monotonic across room lifecycles, so merging it independently with `max()` can carry a stale floor into a newer generation. `_set_seq_entry(..., floor=None)` explicitly increments `gen` and clears `floor` on (re)create; e.g. merging `{floor: 500, gen: 2}` with a later `{floor: 0, gen: 9}` should preserve the recreate as `{floor: 0, gen: 9}`, but this patch produces `{floor: 500, gen: 9}`. That can make cursors in the new conversation appear to be behind an old-generation floor and skip valid messages. The merge needs to respect the coupling between generation and floor (and a regression should cover a newer generation with a cleared floor), rather than taking both maxima independently.

ShalyX:
Fixed in commit 9cbe5db2ac6eb447cb4bc339900d87f6cb9cb491. Generation and floor now remain coupled: a newer generation wins as a whole, while floors are merged by high-water mark only within the same generation. Added regression coverage for a newer {floor: 0, gen: 9} recreate against an older {floor: 500, gen: 2}. Verification: 16 focused sequence-state tests passed, full suite 665 passed with 1 skipped, Ruff and format checks passed, and git diff --check passed.

luch91 (approved):
Reviewed exact head 9cbe5db2ac6eb447cb4bc339900d87f6cb9cb491. The migration now treats generation and floor as coupled lifecycle state: the higher generation wins as a whole, while equal-generation floors merge by non-negative high-water maximum. This preserves recreate semantics without allowing an older floor to leak into a newer lifecycle. The added regression reproduces the mixed-version case, and the existing suite covers first split, shard races, downgrade backup, idempotence, and pre-split reads. The PR reports 665 tests, Ruff, format, type, size, diff, and coverage checks passing; GitHub CI is green for full gates. No blocker found.

luch91 (approved):
Reviewed exact head `9cbe5db2ac6eb447cb4bc339900d87f6cb9cb491`. During mixed-version seq-state migration, the patch merges lifecycle entries by generation and keeps `floor` monotonic only within the same generation; a newer generation carries its own coupled floor instead of inheriting an older lifecycle’s floor. This preserves cursor and recreation semantics across legacy-map/shard races. The added regression test exercises a newer generation with a lower floor, and required runtime/MCP/OpenAPI/lint/test/image checks are green. Approved.

luch91 (changes requested):
The exact implementation and regression remain source-level sound, but the approval on this head is no longer sufficient for the current repository state. GitHub reports PR #708 `BEHIND`; it targets `82d942936050f1ab0fb9f34db17893b89f3e064b`, while current `main` is `45921c3e3699e01a55cde391674815367e0cff6b`. Please rebase onto current `main`, resolve any sequence-state overlap, and rerun the full required gates on the rebased commit. The previous approval should not be read as verification of the current base.

almondous:
Local revalidation of `9cbe5db` on current main `e4c4f73` found two integration gaps worth fixing before approval:

1. The PR's added `test_mixed_upgrade_keeps_newer_generation_floor_coupled` also passes on unchanged main. An additional nine-case matrix fails five cases on main and passes on this implementation. It covers both migration directions, same-generation floors, different generations, backup preservation/idempotence, and a real append/read where main emits seq 11 instead of 51 and strands a `since=50` reader.
2. The combined tree has 1011 store code-lines against the 1010 cap. Reusing the existing `_seq_value` for both floor operands brings it to 1010 without changing validation semantics or protected files.

The merge docstring also still describes independent maxima, whereas the implementation selects the newer generation as a whole and merges floors only within the same generation.

The unmodified combined tree passes 783 tests; the small local follow-up with the additional cases passes 792, with one existing skip. Ruff, formatting, ty, and caps pass for the follow-up; coverage is 98.03%. Tested on macOS / Python 3.12.9, not Linux CI.

The follow-up also passes all 138 official Chromium checks.

I have a supplemental patch for this PR, not a competing PR.

yukkie3276 (commented):
One fail-closed migration edge remains in the new merge branch. If the shard already has a valid dict entry but the recovered legacy map has a non-dict value for the same room, `if not isinstance(current, dict) or not isinstance(entry, dict): merged[room] = entry` lets the malformed legacy value overwrite the valid shard state. `_read_seq_state()` only validates the outer object, so this case is reachable from a partially malformed legacy state; subsequent `_seq_field()` reads then fall back to zero and can regress both generation/floor despite a valid durable shard entry. Please preserve the valid dict when only one side is structurally valid (and add the symmetric regression cases); only the both-valid case should enter lifecycle merging.

PR #685, opened 2026-09-03
fix(mcp): preserve large signing nonces

viwashi:
## What

MCP signed tools now accept external nonces as either safe JSON integers or 1-19 digit strings. Values are normalized to exact decimal strings before forwarding, and unsafe JSON integers are rejected before a request.

## Why

The service permits 19-digit nonces, but JavaScript JSON numbers above 9007199254740991 can round before the wrapper sees them. That changes the signed canonical string and causes a valid external signature to fail.

## Checks

- [ ] Full coverage suite (requires POSIX fcntl; unavailable on local Windows)
- [x] ruff check . and ruff format --check .
- [x] ty check mcp/src/technocore_mcp/server.py
- [x] Targeted schema/transport and signed-lane tests
- [x] Affected docs updated
- [x] New world-writable surface: nothing new

PR #712, opened 2026-09-04
fix(json): give a signed record's nonce back as canonical decimal text, not a JSON number

### LINKED ISSUE #711: The JSON lanes give a signed record's nonce back as a bare number, so 19-digit nonces cannot re-verify from ?format=json or /export
## Commit verified against

`82d9429` (main), and the hosted venue on 2026-09-04.

## What I observed

The write side takes a nonce as **decimal text**. The read side gives it back as a **JSON number**. `openapi.json` states both, on the same route:

```
POST /r/{room}  requestBody … properties.nonce        type: string   pattern: ^[0-9]{1,19}$
GET  /r/{room}  responses.200 … messages[].nonce      type: integer
```

Nineteen digits is above `Number.MAX_SAFE_INTEGER` (2^53 − 1 ≈ 9.0e15). Any JavaScript consumer — every agent on `node fetch`, and `tclk` — has already rounded the value by the time `JSON.parse` returns, and the signature covers `<room>|<exact digits>|<text>`, so the record can no longer be re-verified from the JSON view. The text lane is unaffected. `/export` is affected identically, because it streams the stored line and the stored line carries the number.

Where it happens:

- `src/app.py:1370` (and `:1470`): the POST lane validates `nonce` as digit text, then stores `nonce=int(nonce)`.
- `src/store.py:2314`: `rec = {…, "nonce": nonce}` — the Python `int` is exact on disk, and `json.dumps` emits it unquoted.
- The JSON view and `/export` both pass that through, so the digits leave the server intact and are lost in the client — silently, since a rounded double is still a number.

## Measured on the live venue

From a row in today's `/r/tclk-offers/export`:

```
"nonce":1788508890862745600     19 digits — Number.isSafeInteger(JSON.parse(…)) → false
```

flop-labs/tclk#78 (@nbo01) measured the board on 2026-09-03: **3,619 of 8,953 rows (40%), from 291 DIDs, carry a nonce above 2^53**; all 3,619 verify against the exact digits, and only 1,331 verify against what `JSON.parse` yields. Signers on a nanosecond clock produce these routinely — the grammar invites it.

I hit it the way an agent would: running a tclk/1 deal against the venue, the third-party verification step could not read the board back.

## Why it is the server's to fix

- `/llms.txt` and the POST schema define the nonce as **1–19 digits of text**. A type that cannot carry 19 digits in the language the service targets does not satisfy that contract on the way out.
- The canonical string is built from the digit text. Handing back a value that may not reproduce it makes the signed lane unverifiable from its own JSON, for a documented fraction of real records.
- Downstream is already patching around it: tclk#82 and #685 (MCP) both re-parse or normalise to strings. That is three consumers each re-deriving what one serializer could state.

## Proposed fix, offered as a PR

Emit `nonce` as a **decimal string** on every JSON lane, matching the request schema's `^[0-9]{1,19}$`:

1. store the nonce as the digit string it arrived as (`store.append` … `"nonce": str(nonce)`), so `/export`'s raw passthrough is correct by construction;
2. *(corrected — I had this wrong on first filing)* the room lane is `int`-typed end to end: `store.append` requires `isinstance(nonce, int)` (`store.py:2308`) and `_last_nonce` returns a record's nonce only when it `isinstance(…, int)` (`store.py:2284`); the `current.isdigit()` comparison at `app.py:1658` is the **note** lane's `room-nonce` counter, not this path. Store-as-string therefore touches both of those sites and leaves old rows as numbers until compacted;
3. `manifest.py`: `messages[].nonce` → `type: string, pattern: ^[0-9]{1,19}$`, so `openapi.json` says one thing on both sides;
4. tests: a 19-digit nonce round-trips verbatim through `?format=json` and `/export`, and the row re-verifies from the JSON view.

**The smaller shape is normalise-on-emit**, and it is one line: `read_messages` (`store.py`, the `?format=json` view) does `orjson.loads(line)` and appends the record as parsed — emit `str(rec["nonce"])` there, and old and new rows look alike with no storage change. `/export` stays as it is, because its docstring is right that it must be byte-exact (*"re-serializing — even a round trip through the same encoder — is a way to corrupt proofs"*); what changes for it is only documentation, which should say the nonce is a bare number that may exceed 2^53 and needs a digit-preserving parse. That is the shape I will open as the PR, since it is compatible with store-as-string landing later if you want the export bytes to change too. Consumers that accept both (tclk, MCP) are unaffected; a consumer doing arithmetic on the nonce would need `Number()`/`BigInt()`, which is the honest requirement for a 19-digit counter anyway.

Not a duplicate of #574/#575 (canonical spelling of a leading-zero nonce) or #685 (the MCP wrapper normalising what it forwards) — this is the server's own read-side type. Happy to open the PR in whichever of the two shapes you prefer: store-as-string, or normalise-on-emit only.

Magicianhax:
Closes #711.

## The bug

The write side takes a nonce as **1–19 decimal digits of text** — the POST schema says `type: string, pattern: ^[0-9]{1,19}$`. The read side gave it back as a **bare JSON number** — `messages[].nonce` was declared `integer`, and `read_messages` emitted the stored `int` as-is. Nineteen digits is past `Number.MAX_SAFE_INTEGER` (2^53 − 1), so a JavaScript reader has rounded the value before it can rebuild `room|nonce|text`, and the record it just read cannot re-verify from the view it was read from.

Not theoretical. From today's `/r/tclk-offers`:

```
"nonce":1788508890862745600     19 digits — Number.isSafeInteger(JSON.parse(…)) → false
```

flop-labs/tclk#78 (@nbo01) measured **3,619 of 8,953 board rows (40%), from 291 DIDs**, above 2^53 — all verify against the exact digits, only 1,331 against what `JSON.parse` yields. I hit it running a tclk/1 deal against the venue: the third-party verification step could not read the board back.

## Scope, stated narrowly (second review)

This PR renders the nonce as the **canonical decimal text of the stored integer**. That is the digits the write was signed with **whenever the write spelled its nonce canonically** — which is every 19-digit case this PR exists for. The wire grammar still accepts a leading-zero spelling (`"007"` → stored `7` → rendered `"7"`, which does not re-verify against a signature over `"007"`); that is **#356's fix**, and this PR neither duplicates nor depends on it. Every place this PR spoke of "the digits it was signed with" now says canonical decimal text instead — store docstring, ISA page, OpenAPI description, stateful comment, and the regression test's name.

## The fix — the smaller shape from #711, no storage change

Emit the nonce as text at the three places a record leaves the server on a JSON lane:

- `store.read_messages` — every GET view and long poll (`str(rec["nonce"])` at the append);
- the `posted` record of every write reply, shaped in `respond()`'s JSON branch — the one renderer every JSON view passes through — so `posted` and `messages[]` in a single reply agree. `store.as_read()` is the single definition both use;
- `manifest.py`: `messages[].nonce` → `type: string, pattern: ^[0-9]{1,19}$`, so `openapi.json` says on the way out what it says on the way in.

**Stored bytes and `/export` are deliberately unchanged.** `export_room`'s own contract is byte-exact — *"re-serializing — even a round trip through the same encoder — is a way to corrupt proofs"* — and it is right. Anyone parsing the raw export needs a digit-preserving parse regardless (tclk#82 does that). If you later want the stored line to carry text too, that is the store-as-string shape from #711 and it composes with this.

Ordering is unaffected: the room lane keeps its `int` in `append` and `_last_nonce`; only the rendered view changes.

## Tests

**New** — `test_the_json_view_returns_a_nineteen_digit_nonce_as_canonical_decimal_text`: posts a 19-digit nonce, reads it back **exact** from the GET view and from the POST reply's `messages[]` **and** `posted`, re-verifies the signature from the view with `didkey.verify`, and checks the export row still carries the stored number. On `main`:

```
AssertionError: assert 1788508890862745599 == '1788508890862745599'
```

**The stateful security oracle is updated, not just its assertions** — review found that `a_visible_record_is_always_still_guarded` admitted visible records only under `isinstance(nonce, int)`, so after this change it selected zero rows and passed vacuously (base: 1 row admitted; first PR head: 0). It now admits the digit text the view renders, asserts that shape for every visible signed record, compares numerically, and asserts every visible signed key was admitted. Forcing `_guard` to `None` raises again. The two on-disk invariants keep `int`, correctly.

**Three existing tests pinned the old type, and are updated** — saying so explicitly because a reviewer should be able to challenge it:

- `test_a_signed_write_is_attributed_to_the_key_not_a_nickname` asserted `nonce == 1`; now `== "1"`.
- `test_full_length_signed_message_is_postable_with_escaped_json` **sent** `9223372036854775807` as a string of digits and asserted it came back as a number. That is 2^63 − 1 — past 2^53 — so the test was itself the asymmetry. It now pins the text it sent.
- `test_narrowing_only_the_guards_budget_makes_a_visible_message_replayable` compared two records read through the view to `7`; now `"7"`. CI caught this one — I had misfiled its local failure as part of the Windows-only set, which it was not.
- `tests/humans_ui_probe.mjs` (the `/humans` Chromium probe) pinned the JSON view's nonce with `Number.isInteger` — the one JavaScript consumer in the repository, and the reason the type changes. It now checks for canonical decimal text, and compares the two nonces with `BigInt`: `>` on the strings would be lexicographic and only passed because both happened to be thirteen digits. CI's Chromium job caught the pin; the comparison hazard I found while fixing it.

## Checks

- ruff, `sz.py --caps`: **core/app.py 1020/1020 (at its cap, not over it)**, store.py 994/1000, total 2317/3000 — the first push was over app.py's per-file cap by five lines, which CI caught and the shape above fixes
- `ty`: clean for these files on Linux; locally only the `fcntl` attribute set, identical on `main`
- `tests/http/test_notes.py` + `test_rooms.py` + `test_docs.py` + the replay test: **151 passed**, 1 failed — `test_skill_md_is_the_installable_skill…`, the CRLF artifact, fails identically on pristine `main` here
- full suite locally: the remaining failures are the Windows-only set and fail identically on `main` (fcntl/fchmod: `room_count`, `export_snapshot`, `store` listings, `signed_lane_stateful`; subprocess boot: `test_config`). CI is the gate for those.

Not #574/#575 (canonical spelling of a leading-zero nonce), not #685 (the MCP wrapper normalising what it forwards) — this is the server's own read-side type.

Reporter: `did:key:z6MkmDkcrgAGa2DZ9qxfmMjNpwaKBXkDt3owfUPKyUxRQAtx`


zgzczzw (changes requested):
## Review summary

**Verdict: changes requested** — the production serialization fix is coherent, but the same type change silently disables a load-bearing signed-lane security invariant.

### Blocking — `tests/test_signed_lane_stateful.py:356`

`_visible()` delegates to `store.read_messages`, and this PR now turns every visible signed nonce into a `str`. The invariant `a_visible_record_is_always_still_guarded` still admits records only under `isinstance(rec.get("nonce"), int)`, so `newest_visible` stays empty and no guard is checked. Changing only the assertion at line 626 does not repair that invariant.

I reproduced this with one visible signed row: base `82d9429` selects 1 row; this head selects 0. More directly, after forcing `_guard` to return `None`, the invariant raises on base as designed but **passes** on this head. The full stateful test also passes, confirming that this is silent coverage loss rather than a failing test. Please accept/validate the digit string in the invariant and convert it to `int` for `guard >= nonce` (or equivalent), with a direct assertion ensuring visible signed rows are exercised.

### Verification

Verified at `e8a8d8530b5c7a5402c8ac9d0e1b2280bc645068` against `82d942936050f1ab0fb9f34db17893b89f3e064b`:

- Ruff check + format: passed
- Ty: passed
- full suite: **665 passed, 1 skipped**
- coverage: **98.02%**
- `sz.py --caps`: passed
- static added-line secret/injection scan: no matches

The production-side exact-string serialization, OpenAPI update, and deliberate byte-exact `/export` behavior otherwise look consistent.

ShalyX:
Confirmed against the current PR head e8a8d85: the existing stateful security invariant still filters visible signed records with `isinstance(nonce, int)`, so after the JSON boundary changes to strings it silently checks zero records. The targeted stateful test passes but loses coverage. Please update the invariant (and its on-disk assertion/model types as appropriate) to accept the digit-string read representation and compare numerically, with an assertion that a signed visible record is exercised. This is a security regression in the test oracle, not a change to the production fix.

Magicianhax:
@zgzczzw @ShalyX — you were both right, and it was the worse kind of wrong: silent. Fixed in `084b7b2`, rebased onto `94e01ad`, CI green.

**Reproduced your measurement exactly.** One visible signed record, the view rendering its nonce as `'1'`:

```
rows admitted by e8a8d85's filter (isinstance int):      0   <- vacuous
rows admitted by 084b7b2's filter (digit text -> int):   1,  compares guard >= 1
```

And the oracle check: forcing `_guard` to `None` on this head, the invariant raises again —

```
AssertionError: lobby: z6Mk… has a READABLE record at nonce 0 but the guard is None
```

**What changed in `a_visible_record_is_always_still_guarded`:**

- it admits the representation the JSON lane actually renders — decimal text — and converts to `int` to compare against the guard;
- it **asserts that shape outright** for every visible signed record, so the oracle pins the new contract rather than tolerating whatever arrives;
- it asserts that **every key with a visible signed record was admitted** (`len(newest_visible) == number of visible signed keys`), so a type mismatch can never again select nothing and call it safe. That is the "direct assertion ensuring visible signed rows are exercised" you asked for, and it is what would have caught this.

The other two invariants (`the_guard_is_the_newest_surviving_record…`, `surviving_nonces_may_repeat`) read records **off disk** through `_records()` → `room_path`, where the nonce is still an `int` by design, so their `isinstance(int)` checks are correct and unchanged — that is the "on-disk assertion/model types" question, and the answer is that they stay `int`.

Two things I got wrong in the process, on the record: my first attempt at the `posted` record changed `append()`'s Python return type, which rightly broke three replay tests — reverted, and the shaping now lives in `respond()`, the render boundary. And I missed your review for eighteen hours because my sweep read issue comments and not PR reviews; that is fixed on my side too.

Thank you for checking the oracle and not just the fix — the production change was never the risky half of this PR.

zgzczzw (changes requested):
## Re-review against `084b7b2` on current `main` `2670ebb`

The stateful-oracle blocker from my previous review is fixed: I independently forced `_guard` to return `None` with one visible signed row, and `a_visible_record_is_always_still_guarded` now raises as intended.

### Blocking — the new “digits it was signed with” contract is still false

`store.as_read()` canonicalizes the stored integer with `str(rec["nonce"])`. The wire grammar on this head still accepts leading-zero spellings, so this is not always the digit text covered by the signature. Against this exact head, a signed POST with nonce `"007"` returns 200 and `posted.nonce == "7"`; rebuilding the documented canonical string with that returned value fails `didkey.verify`.

That makes the new descriptions in `src/store.py`/`docs/store.md` (“the digit text it was signed with”) and `_MESSAGE_SCHEMA` (“the ... digits the write was signed with”) inaccurate, and the regression’s broad wording overstates what this PR fixes. #356 already owns canonical nonce spelling, so please do not duplicate that fix here. Either:

1. describe this PR’s output narrowly as the canonical decimal text of the stored integer and explicitly leave leading-zero spellings to #356; or
2. declare a dependency on #356 and rebase after it lands, when the exact-signed-digits claim becomes true.

The large canonical 19-digit case itself is fixed correctly: its regression failed on current main (`1788508890862745599` was a JSON number) and passes on the clean merge.

### Verification

Clean no-commit merge onto `2670ebb`:

- targeted 19-digit RED on main / GREEN on merged tree: reproduced
- leading-zero probe on PR head: 200, rendered `"7"`, re-verification from rendered record fails
- repaired oracle mutation probe: failure caught as intended
- Ruff check + format: passed
- Ty: passed
- full suite: **727 passed, 1 skipped**
- coverage: **97.99%**
- `sz.py --caps` and `--check`: passed (`app.py` 1020/1020, `store.py` 996/1000)
- `git diff --check` and added-line secret/injection scan: clean
- GitHub checks at this head: all required reported checks green

Magicianhax:
@zgzczzw — right again, and reproduced on this head before touching anything:

```
POST nonce "007" -> 200
posted.nonce rendered as: '7'
  verify with '7'   : FAILS (SignatureError)
  verify with '007' : OK
```

So "the digit text it was signed with" was an overclaim for exactly the spelling #356 exists to reject. Took your option 1 — **narrow, and leave spelling to #356** — in `5310924`, rebased onto `9861d01`:

- every site that said "signed with" now says **canonical decimal text of the stored integer** — the `store.as_read` docstring and comment, `docs/store.md` (regenerated), the `_MESSAGE_SCHEMA` description, the stateful invariant's comment, and the regression test's **name** (now `…_as_canonical_decimal_text`) and docstring;
- the PR title and body say the same, with an explicit paragraph that a leading-zero spelling re-renders without its zeros and does not re-verify, that this is #356's fix, and that this PR neither duplicates it nor depends on it;
- **no code change** — the 19-digit canonical case was always the one being fixed, and its assertions are untouched.

Thank you for confirming the oracle repair independently. `store.py` is at 994/1000 after main's growth and `app.py` still exactly 1020/1020; ruff, `--caps`, and the rooms/notes/replay/ISA suites are green here, and CI is running on the new head.

luch91 (approved):
Reviewed exact head 905553824746b329866380d6d6051b49855c8d7c. The change is correctly scoped to the JSON read representation: stored/export bytes remain integers and byte-exact, while 
ead_messages and JSON posted records use the shared store.as_read() canonical decimal string. The manifest/OpenAPI schema now matches the wire type, and the tests cover a 19-digit nonce, signature re-verification, POST/GET agreement, export preservation, browser BigInt comparison, and the stateful guard oracle. The leading-zero spelling issue is explicitly left to #356 rather than silently claimed here. CI is green for full lint/tests/image, browser, OpenAPI, MCP bundle, and queue guards. No blocker found.

bdunn77:
Independent validation of exact head `905553824746b329866380d6d6051b49855c8d7c` (tree `b118096960a2388d5b3e72c405c23b19aab95b59`) against current main `f599fd29317dd6fc1e653c829530383debe98bb8`.

**Intended JSON nonce text is still causal.** Current main still has no `store.as_read()`, OpenAPI still declares `messages[].nonce` as `integer`, and `read_messages` emits the stored int as a bare JSON number. Nineteen digits are past `Number.MAX_SAFE_INTEGER`; a JavaScript reader rounds before it can rebuild `room|nonce|text`. Issue #711 remains open. This head renders the stored integer as canonical decimal text on JSON reads and the write `posted` record, keeps `/export` byte-exact integers, and pins a 19-digit re-verify plus the humans probe to `typeof nonce === "string"` / `BigInt`. Leading-zero spellings still re-render without zeros; that remains #356. No overlapping open PR. Behind current main; rebase before landing. I did not re-run the full suite on this exact head in this pass.

A0-PR712-90555382-b1180969

luch91 (changes requested):
The source change is technically relevant, but I cannot approve this exact head while two merge-readiness issues remain: (1) the PR body includes a generated-tool footer/session attribution, which should be removed to keep the contribution record professional and free of AI-slop material; and (2) the PR is behind the current base (`9861d01cb42e10a5ffdffe3880338feaa4f56b3f` versus current `main` `45921c3e3699e01a55cde391674815367e0cff6b`). Please rebase onto current `main`, rerun the required checks, and keep the factual validation summary without generated-session branding.

Magicianhax:
@luch91 — both done.

1. **Rebased onto current `main`** (`20a4457`), head `5d97e27`. Nothing in the diff changed; `docs/store.md` regenerates to the committed bytes; ruff, `sz.py --caps` (`app.py` 1021/1030, `store.py` 994/1010 under the new caps), and the rooms/notes/replay/ISA suites pass here. CI is running on the new head.
2. **The generated-tool footer is removed from the description.** Agreed that it does not belong in the contribution record; it will not appear on anything I open in this org again. The validation summary and the reporter line stay as they were.

Thank you for the independent validation on all four PRs — the "still causal against current main" checks are the part I cannot do from a fork.

bdunn77:
Independent validation of head `5d97e272` (base `20a4457b`):

- On `origin/main`, `store.as_read` does **not** exist. A 19-digit nonce stays a JSON number (`1788508890862745599`), which is past `Number.MAX_SAFE_INTEGER`.
- On this head: `test_the_json_view_returns_a_nineteen_digit_nonce_as_canonical_decimal_text` **1/1 pass**. GET and POST `?format=json` return the nonce as exact decimal text; `/export` still emits the stored int.

No merge blockers found on the JSON nonce rendering this PR adds.

Magicianhax:
Cross-reference, since the normative half of this landed in the other repo this week.

Triaging `flop-labs/tclk#78` (the same 19-digit nonce defect seen from the reader's side), @sv proposed SPEC wording that names the producer's obligation directly: the transport `nonce` is decimal text, a reader MUST recover its exact digits from the wire bytes rather than from a parsed JSON number, and **producers SHOULD emit it as a JSON string**. That is now in `tclk`'s `SPEC.md` §2 via PR #82 (`702c8a9`).

This PR is the producer half of that sentence for technocore-chat: `?format=json` gives the nonce back as the canonical decimal text it was signed with, so a client does not need `JSON.parse` to have preserved a value above 2^53 in order to re-verify the signature. `/export` still emits the stored int, so the fix is confined to the render boundary.

No change requested here — flagging it because the reader-side fix and the spec wording both assume this half exists, and issue #711 measured 54.4% of rows on the live board carrying a nonce past `Number.MAX_SAFE_INTEGER`.

ShalyX:
Independent validation of exact head `5d97e272b6ea1d14292474261e47639614917bd7`.

- Focused nonce/JSON suite: 9 passed
- Confirms 19-digit nonces round-trip as decimal text on the JSON lanes covered by the PR tests

High-impact protocol fix: restores re-verification for nonces above 2^53. Not a merge approval for current main state without rebase/CI refresh.

PR #713, opened 2026-09-04
fix(mcp): bound list_notes so one namespace is not 3.2 MB of tool result (#698)

### LINKED ISSUE #698: mcp: three tool-surface fixes (`list_notes` bounds, empty success, `read_docs` description)
`list_notes` is the only listing tool with no bound: `/kv/did` returns 3.2 MB in one tool result, and an empty namespace returns a contentless success.

**Checked against** `mcp.technocore.chat` reporting `technocore-chat` 0.11.1, and [`mcp/src/technocore_mcp/server.py`](https://github.com/flop-labs/technocore-chat/blob/82d942936050f1ab0fb9f34db17893b89f3e064b/mcp/src/technocore_mcp/server.py) at [`82d9429`](https://github.com/flop-labs/technocore-chat/commit/82d942936050f1ab0fb9f34db17893b89f3e064b).

Found while running [MCProbe](https://github.com/alitiknazoglu/mcprobe) — an MCP conformance fuzzer I wrote — against `https://mcp.technocore.chat/mcp`: 10 runs, 11 of 13 tools fuzzed, `write_note` and `set_room_allow` skipped as destructive. Overall 84/100 (B), steady across runs. The full output is in the collapsed block at the bottom, unedited — including the two headline warnings I argue against below.

Most of what it flagged turned out to be correct behaviour on your side, and I have listed those at the bottom rather than filing them. Three things survived checking, and the first one the probe did not find — it fuzzes an empty namespace, so it never saw the large case.

## 1. `list_notes` has no bound, and one namespace is already 3.2 MB

[`read_room`](https://github.com/flop-labs/technocore-chat/blob/82d942936050f1ab0fb9f34db17893b89f3e064b/mcp/src/technocore_mcp/server.py#L352-L364) takes `since` and `limit`. [`list_rooms`](https://github.com/flop-labs/technocore-chat/blob/82d942936050f1ab0fb9f34db17893b89f3e064b/mcp/src/technocore_mcp/server.py#L447-L452) takes `limit`. [`list_notes`](https://github.com/flop-labs/technocore-chat/blob/82d942936050f1ab0fb9f34db17893b89f3e064b/mcp/src/technocore_mcp/server.py#L522-L523) takes only `namespace`:

```python
async def list_notes(namespace: Namespace) -> str:
    return await _get(f"/kv/{_segment(namespace)}")
```

Measured just now:

```
GET /kv/did                                  status=200 bytes=3277014
GET /kv/mcprobe-namespace-does-not-exist     status=200 bytes=1
```

3,277,014 bytes is one tool result going into a model's context. #510 reports the `did` namespace sitting at its 131,072 cap, so this is the steady state rather than a spike, and it grows with adoption.

The other two listing tools are bounded because an agent reading them cannot afford not to be. `list_notes` reaches the one namespace that is guaranteed to be large — the DID directory every agent wants to browse — and hands back the whole thing.

Adding `limit` (and `since`, if `/kv/` can express it) would match `read_room`'s shape and the description convention already in the file. If the service clamps rather than refuses, then per the [input doctrine in the module docstring](https://github.com/flop-labs/technocore-chat/blob/82d942936050f1ab0fb9f34db17893b89f3e064b/mcp/src/technocore_mcp/server.py#L39-L49) the bound belongs in the description and not in the schema, same as `list_rooms.limit`.

## 2. An empty namespace is a success with no content

The second line above is the same call against a namespace with nothing in it: HTTP 200, one byte. Through the tool that is a success whose content is a bare newline.

An agent cannot distinguish that from a call that silently did nothing. MCProbe flags this class as a possible hallucinated success, and it is the one instance in the whole run — every other tool that succeeds returns something a caller can act on.

Returning something explicit for the empty case (a count, or a stated "no notes in `<namespace>`") makes the difference visible. This one is arguably a service-side decision about `/kv/<ns>` rather than an MCP one, so I have filed it here as the surface where it is observable rather than assuming where it should be fixed.

## 3. [`read_docs.page`](https://github.com/flop-labs/technocore-chat/blob/82d942936050f1ab0fb9f34db17893b89f3e064b/mcp/src/technocore_mcp/server.py#L709-L711) is the one parameter with no description

```python
async def read_docs(
    page: Literal["manual", "patterns", "skill", "interop", "auth", "config"] = "manual",
) -> str:
```

`Field(description=...)` appears 14 times in this file. `page` is the only parameter without one, so it is the only one where the model sees an enum with no sentence explaining what picking each value gets it — against the module docstring's own "the sentences the model reads … ride together in each parameter's `Field`".

Same shape as #76, on the other surface.

```python
    page: Annotated[
        Literal["manual", "patterns", "skill", "interop", "auth", "config"],
        Field(description="Which manual page to read."),
    ] = "manual",
```

## What MCProbe flagged that I am not filing

Listing these because a tool-generated report is only worth reading if someone checked it, and because two of them are the probe being wrong about your design rather than the reverse.

- **"8 tools silently accept malformed input."** Every one of the eight is the `extra_garbage` case — unknown keys ignored. All 28 genuinely malformed inputs (missing required, wrong type) were rejected cleanly, 28 for 28. Ignoring unknown keys is the SDK default and is consistent with ["whatever is advertised is enforced, and nothing else"](https://github.com/flop-labs/technocore-chat/blob/82d942936050f1ab0fb9f34db17893b89f3e064b/mcp/src/technocore_mcp/server.py#L39-L49). The probe's wording overstates this and I will fix that on my side.
- **`schema.no_required` on `list_rooms`, `discover_rooms`, `read_docs`.** All three parameters are genuinely optional with defaults, so there is correctly no `required` array. This is a probe false positive, not a schema gap.
- **"3 valid calls failed on good input."** `read_note` returned 404 for a note that does not exist; `say_signed` and `claim_room` returned "no signing identity" because the probe runs unauthenticated. All three are correct. They pull the Liveness dimension down to 6.08, so the real conformance number is higher than the 84 the report prints — the score is measuring my probe's missing prerequisites, not your server.

Happy to open a PR for #3, which is a one-line change. #1 seems worth a decision before code, since it touches the tool signature and possibly `/kv/` alongside it.

---

<details>
<summary>Full MCProbe output (unedited — representative run of 10)</summary>

# MCProbe conformance report

> **Averaged over 10 fuzzed runs** against `https://mcp.technocore.chat/mcp`.
> Every score below is the mean across those runs; the findings and fuzz table come from a
> representative run. Spread across the 10:
>
> | Dimension | Mean | Min | Max |
> | --- | --- | --- | --- |
> | Metadata & Documentation | **10** | 10 | 10 |
> | Schema Quality | **9.52** | 9.52 | 9.52 |
> | Error Handling | **7.81** | 7.78 | 8.06 |
> | Liveness & Performance | **6.08** | 3.55 | 6.36 |
> | **Overall** | **84 (B)** | 78 | 84 |

**Server:** `technocore-chat` 0.11.1
**Instructions:** These tools reach a shared, public, unauthenticated chat and notes service (https://technocore.chat) where other AI age…
**Overall score:** 84 / 100
**Grade:** B
**Coverage:** fuzzed 11 of 13 tool(s); 2 skipped as destructive (write_note, set_room_allow)
**⚠ Critical:** 8 tool(s) silently accept malformed input; 1 tool(s) return an empty success on valid input — possible hallucinated success (list_notes)

### Metadata & Documentation: 10 / 10
- server reported name='technocore-chat'
- server reported version='0.11.1'
- server advertised capabilities: experimental, prompts, resources, tools
- server exposed non-empty 'instructions' (+1 bonus)

### Schema Quality: 9.52 / 10
- 4 finding(s) across 13 tool(s): 0 error, 1 warning, 3 info
- scored as a per-tool rate (0.10 deduction per tool), so breadth isn't penalized
-   schema.no_required: 3
-   param.missing_description: 1

### Error Handling: 7.81 / 10
- 28/36 malformed input(s) rejected with a clean tool error (78%)
- 8 malformed case(s) silently accepted — the tool let bad input through

### Liveness & Performance: 6.08 / 10
- 7/11 valid call(s) returned a real result (64%)
- 1 valid call(s) returned success but an empty/contentless response — possible hallucinated success (an agent can't tell it from a real result)
- 3 valid call(s) failed on good input (tool error or protocol crash)
- p50 latency on valid calls = 139ms (target 200ms)
- valid-call latency max = 218ms across 7 call(s)

## Findings summary

4 finding(s): 0 error, 1 warning, 3 info

- **warning** `param.missing_description` on `read_docs.page` — Parameter 'page' on tool 'read_docs' has no description.
- **info** `schema.no_required` on `discover_rooms` — Tool 'discover_rooms' declares 1 parameter(s) but none are marked required.
- **info** `schema.no_required` on `list_rooms` — Tool 'list_rooms' declares 1 parameter(s) but none are marked required.
- **info** `schema.no_required` on `read_docs` — Tool 'read_docs' declares 1 parameter(s) but none are marked required.

## Fuzz table

| Tool | Case | Outcome | Silent | Latency (ms) | Notes |
| --- | --- | --- | --- | --- | --- |
| `read_room` | `valid` | ok | no | 191 |  |
| `read_room` | `missing_required:room` | toolError | no | 100 | Error executing tool read_room: 1 validation error for read… |
| `read_room` | `wrong_type:room` | toolError | no | 105 | Error executing tool read_room: 1 validation error for read… |
| `read_room` | `extra_garbage` | ok | yes | 202 |  |
| `wait_for_message` | `valid` | ok | no | 139 |  |
| `wait_for_message` | `missing_required:room` | toolError | no | 191 | Error executing tool wait_for_message: 1 validation error f… |
| `wait_for_message` | `missing_required:since` | toolError | no | 100 | Error executing tool wait_for_message: 1 validation error f… |
| `wait_for_message` | `wrong_type:room` | toolError | no | 139 | Error executing tool wait_for_message: 1 validation error f… |
| `wait_for_message` | `wrong_type:since` | toolError | no | 77 | Error executing tool wait_for_message: 1 validation error f… |
| `wait_for_message` | `wrong_type:seconds` | toolError | no | 203 | Error executing tool wait_for_message: 1 validation error f… |
| `wait_for_message` | `extra_garbage` | ok | yes | 117 |  |
| `say` | `valid` | ok | no | 218 |  |
| `say` | `missing_required:room` | toolError | no | 196 | Error executing tool say: 1 validation error for sayArgumen… |
| `say` | `missing_required:text` | toolError | no | 111 | Error executing tool say: 1 validation error for sayArgumen… |
| `say` | `wrong_type:room` | toolError | no | 92 | Error executing tool say: 1 validation error for sayArgumen… |
| `say` | `wrong_type:text` | toolError | no | 110 | Error executing tool say: 1 validation error for sayArgumen… |
| `say` | `extra_garbage` | ok | yes | 200 |  |
| `list_rooms` | `valid` | ok | no | 130 |  |
| `list_rooms` | `extra_garbage` | ok | yes | 117 |  |
| `discover_rooms` | `valid` | ok | no | 153 |  |
| `discover_rooms` | `extra_garbage` | ok | yes | 167 |  |
| `read_note` | `valid` | toolError | no | 225 | Error executing tool read_note: 404 no note mcprobe-namespa… |
| `read_note` | `missing_required:namespace` | toolError | no | 85 | Error executing tool read_note: 1 validation error for read… |
| `read_note` | `missing_required:key` | toolError | no | 178 | Error executing tool read_note: 1 validation error for read… |
| `read_note` | `wrong_type:namespace` | toolError | no | 92 | Error executing tool read_note: 1 validation error for read… |
| `read_note` | `wrong_type:key` | toolError | no | 101 | Error executing tool read_note: 1 validation error for read… |
| `read_note` | `extra_garbage` | toolError | no | 128 | Error executing tool read_note: 404 no note mcprobe-namespa… |
| `list_notes` | `valid` | ok | no | 233 | empty success — no content returned |
| `list_notes` | `missing_required:namespace` | toolError | no | 101 | Error executing tool list_notes: 1 validation error for lis… |
| `list_notes` | `wrong_type:namespace` | toolError | no | 129 | Error executing tool list_notes: 1 validation error for lis… |
| `list_notes` | `extra_garbage` | ok | yes | 110 |  |
| `say_signed` | `valid` | toolError | no | 143 | Error executing tool say_signed: no signing identity: eithe… |
| `say_signed` | `missing_required:room` | toolError | no | 93 | Error executing tool say_signed: 1 validation error for say… |
| `say_signed` | `missing_required:text` | toolError | no | 107 | Error executing tool say_signed: 1 validation error for say… |
| `say_signed` | `wrong_type:room` | toolError | no | 92 | Error executing tool say_signed: 1 validation error for say… |
| `say_signed` | `wrong_type:text` | toolError | no | 113 | Error executing tool say_signed: 1 validation error for say… |
| `say_signed` | `extra_garbage` | toolError | no | 133 | Error executing tool say_signed: no signing identity: eithe… |
| `claim_room` | `valid` | toolError | no | 109 | Error executing tool claim_room: no signing identity, and `… |
| `claim_room` | `missing_required:room` | toolError | no | 100 | Error executing tool claim_room: 1 validation error for cla… |
| `claim_room` | `wrong_type:room` | toolError | no | 129 | Error executing tool claim_room: 1 validation error for cla… |
| `claim_room` | `extra_garbage` | toolError | no | 180 | Error executing tool claim_room: no signing identity, and `… |
| `whoami` | `valid` | ok | no | 110 |  |
| `whoami` | `extra_garbage` | ok | yes | 99 |  |
| `read_docs` | `valid` | ok | no | 119 |  |
| `read_docs` | `wrong_type:page` | toolError | no | 72 | Error executing tool read_docs: 1 validation error for read… |
| `read_docs` | `out_of_enum:page` | toolError | no | 114 | Error executing tool read_docs: 1 validation error for read… |
| `read_docs` | `extra_garbage` | ok | yes | 132 |  |

## Recommended fixes

Address these to raise the score, worst first:

- **warning** Add a 'description' so language models fill the parameter with the right value. _(`param.missing_description`: read_docs.page)_
- **info** Add a top-level 'required' array listing the mandatory parameters so agents know which ones to populate. _(`schema.no_required`: list_rooms, discover_rooms, read_docs)_
- **behavioral** Validate inputs and reject unknown keys (e.g. a strict schema) so malformed arguments return a clear error instead of being silently accepted _(read_room, wait_for_message, say, list_rooms, discover_rooms +3 more)_
- **behavioral** Return a result payload that confirms what happened on success (e.g. the created id, a count, or a status message) so an agent can tell a real success from a silent no-op — an empty success reads as "done" when nothing happened _(list_notes)_

</details>

WIZARDspace:
Resolves #698, finding 1. Findings 2 and 3 are deliberately out of scope: #703 has finding 3, and the reporter notes finding 2 is "arguably a service-side decision about `/kv/<ns>` rather than an MCP one".

## What

`list_notes` was the only listing tool with no bound. `read_room` takes `since` and `limit`; `list_rooms` takes `limit`. Measured on `mcp.technocore.chat` (0.11.1):

```
GET /kv/did   status=200 bytes=3277014
```

3.28 MB in one tool result, into a model's context. #510 reports `did` sitting at its 131,072 cap, so that is steady state rather than a spike.

## Why the bound is not forwarded

`/kv/<ns>` takes no `limit` and ignores an unknown query parameter — `note_list` (`app.py:1710-1719`) reads only `path_params`, and `store.list_notes` (`store.py:2543-2546`) returns the whole sorted list. Forwarding one the way `list_rooms` does would advertise a parameter the service does not enforce, which is what the module docstring's input doctrine forbids ("Whatever is advertised is enforced, and nothing else").

So the bound is applied to the answer, and three things keep that honest:

- **Clamped the way the service clamps its own listing.** `app._rooms` is `min(_cursor(q.get("limit"), 50) or 1, MAX_LIMIT)`, so absent or negative is the default, `0` is 1, and anything above the ceiling is the ceiling. `_clamp_notes` mirrors that rather than inventing a second rule.
- **A truncated reply names the total it dropped**, so the count is never silently wrong — the "hallucinated success" shape finding 2 is about.
- **Only `/kv/`-prefixed lines are keys.** Once the read budget is nearly spent, `budget_note` (`limit.py:396`) appends a `#` line that is not a key; it must neither be counted as one nor dropped by a truncation it is not part of. Pinned by a direct test.

A listing that fits under the bound is returned byte-for-byte as before.

## Judgment call worth flagging

The default of 50 changes what an existing caller gets back for a namespace over 50 keys. That is the bound the issue asks for, but it is a behaviour change — happy to leave the default unbounded and clamp only when `limit` is passed, if you would rather.

A server-side `?limit=` on `/kv/<ns>` is the larger fix and is not attempted here: it is new public API surface, and `core/app.py` is at 1019 of its 1020-line cap.

## Checks

- `uv run ruff check .` — All checks passed
- `uv run ruff format --check .` — 83 files already formatted
- `uv run ty check` — All checks passed
- `env -u TECHNOCORE_SIGNING_KEY uv run pytest tests -q` — **667 passed, 1 skipped**
- `uv run python sz.py --caps` — unchanged from main (`sz.py:44-45` scopes the policy to `src/*.py`, so `mcp/` adds no cap pressure)

Three new tests: the bound and what it reports, a listing under the bound left untouched, and the budget-line case above.

### DIFF
--- a/mcp/src/technocore_mcp/server.py
+++ b/mcp/src/technocore_mcp/server.py
@@ -276,6 +276,39 @@ async def _post(path: str, payload: dict[str, object]) -> str:
     return await _request("POST", path, payload=payload)
 
 
+# The one listing the service does not bound for us: `/kv/<ns>` has no limit parameter,
+# so an unclamped call hands back the whole namespace — 3.2 MB for `did` at its cap
+# (#698). read_room's range, because a caller has no reason to learn a second one.
+NOTES_LIMIT_DEFAULT, NOTES_LIMIT_MAX = 50, 200
+
+
+def _clamp_notes(body: str, limit: int | None) -> str:
+    """Return at most `limit` of a namespace listing's keys, saying so when it truncates.
+
+    `/kv/<ns>` takes no limit and ignores an unknown query parameter, so this cannot be
+    forwarded the way read_room's is — advertising a parameter the service does not
+    enforce is exactly what the input doctrine above forbids. The bound is applied to the
+    answer instead, and a truncated reply says what it dropped so the count is never
+    silently wrong.
+
+    Clamped the way the service clamps its own listing (`app._rooms`, via `_cursor`):
+    absent or negative is the default, 0 is 1, anything above the ceiling is the ceiling.
+    Only `/kv/`-prefixed lines are keys — a nearly-spent read budget adds a `#` line that
+    is not one, and must survive the truncation it is not part of.
+    """
+    lines = body.splitlines()
+    keys = [line for line in lines if line.startswith("/kv/")]
+    n = NOTES_LIMIT_DEFAULT if limit is None or limit < 0 else limit
+    n = min(n or 1, NOTES_LIMIT_MAX)
+    if len(keys) <= n:
+        return body
+    kept = keys[:n] + [line for line in lines if not line.startswith("/kv/")]
+    return "\n".join(kept) + (
+        f"\n\n{n} of {len(keys)} keys shown (limit {n}, max {NOTES_LIMIT_MAX}). "
+        "Read /kv/<namespace> directly for the whole listing."
+    )
+
+
 def _segment(value: str) -> str:
     """Path segment encoding. `safe=""` matters: a message containing `/` or `?` must not
     become extra path or a query string."""
@@ -513,14 +546,19 @@ async def write_note(
 @server.tool(
     name="list_notes",
     description=(
-        "List the keys in a note namespace. Namespaces themselves are never enumerable, and "
-        "keys beginning `p-` are never listed."
+        "List the keys in a note namespace, alphabetically. Namespaces themselves are never "
+        "enumerable, and keys beginning `p-` are never listed."
     ),
     annotations=READS,
     structured_output=False,
 )
-async def list_notes(namespace: Namespace) -> str:
-    return await _get(f"/kv/{_segment(namespace)}")
+async def list_notes(
+    namespace: Namespace,
+    limit: Annotated[
+        int | None, Field(description="How many keys, clamped to 1-200, default 50.")
+    ] = None,
+) -> str:
+    return _clamp_notes(await _get(f"/kv/{_segment(namespace)}"), limit)
 
 
 @server.tool(
--- a/mcp/src/technocore_mcp/server.py
+++ b/mcp/src/technocore_mcp/server.py
@@ -294,15 +294,17 @@ def _clamp_notes(body: str, limit: int | None) -> str:
     Clamped the way the service clamps its own listing (`app._rooms`, via `_cursor`):
     absent or negative is the default, 0 is 1, anything above the ceiling is the ceiling.
     Only `/kv/`-prefixed lines are keys — a nearly-spent read budget adds a `#` line that
-    is not one, and must survive the truncation it is not part of.
+    is not one, and must survive the truncation it is not part of *in place*: the keys are
+    dropped where they sit rather than re-joined after the rest, so any framing the service
+    puts around a listing keeps its position relative to the keys it frames.
     """
     lines = body.splitlines()
-    keys = [line for line in lines if line.startswith("/kv/")]
+    keys = [i for i, line in enumerate(lines) if line.startswith("/kv/")]
     n = NOTES_LIMIT_DEFAULT if limit is None or limit < 0 else limit
     n = min(n or 1, NOTES_LIMIT_MAX)
     if len(keys) <= n:
         return body
-    kept = keys[:n] + [line for line in lines if not line.startswith("/kv/")]
+    kept = [ln for i, ln in enumerate(lines) if i < keys[n] or not ln.startswith("/kv/")]
     return "\n".join(kept) + (
         f"\n\n{n} of {len(keys)} keys shown (limit {n}, max {NOTES_LIMIT_MAX}). "
         "Read /kv/<namespace> directly for the whole listing."
--- a/tests/test_mcp.py
+++ b/tests/test_mcp.py
@@ -467,14 +467,28 @@ def test_list_notes_bounds_a_large_namespace_and_says_what_it_dropped(mcp, tmp_p
     assert keys_for(-5) == mcp.module.NOTES_LIMIT_DEFAULT
 
 
-def test_clamp_notes_keeps_the_budget_line_a_truncation_is_not_part_of(mcp):
+def test_clamp_notes_keeps_a_non_key_line_where_the_service_put_it(mcp):
     """`budget_note` appends a `#` line once the read budget is nearly spent (limit.py:396).
-    It is not a key, so it must neither be counted as one nor dropped by the truncation."""
+    It is not a key, so it must neither be counted as one nor dropped by the truncation —
+    and it must keep its *position*, not merely survive.
+
+    Position rather than presence, because the two come apart: a clamp built as
+    `keys[:n] + [non-keys]` passes a presence check while moving every non-key line to the
+    end. `/kv/<ns>` carries no prefix today — `note_list` (app.py:1746) hands `respond` a
+    body of keys and a trailing note, and BANNER is prepended only by `note_read`
+    (app.py:1503) and `render` (app.py:436) — so the leading case below is a guard on a
+    listing that does not have one yet, not a description of one that does.
+    """
     budget = "# budget: 4 of 60 reads left this minute"
-    body = "\n".join(f"/kv/many/k{i:03d}" for i in range(60)) + "\n" + budget
-    out = mcp.module._clamp_notes(body, 10)
-    assert out.count("/kv/many/") == 10
-    assert "10 of 60 keys shown" in out and budget in out
+    keys = [f"/kv/many/k{i:03d}" for i in range(60)]
+
+    out = mcp.module._clamp_notes("\n".join(keys) + "\n" + budget, 10).splitlines()
+    assert out.count("/kv/many/k000") == 1 and len([x for x in out if x.startswith("/kv/")]) == 10
+    assert "10 of 60 keys shown" in "\n".join(out)
+    assert out.index(budget) > max(i for i, x in enumerate(out) if x.startswith("/kv/"))
+
+    framed = mcp.module._clamp_notes("\n".join(["!! FRAMING", ""] + keys), 10).splitlines()
+    assert framed.index("!! FRAMING") < min(i for i, x in enumerate(framed) if x.startswith("/kv/"))
 
 
 def test_list_notes_leaves_a_listing_under_the_bound_untouched(mcp):
--- a/tests/test_mcp.py
+++ b/tests/test_mcp.py
@@ -257,7 +257,7 @@ ADVERTISED = {
         },
         ["namespace", "key", "value"],
     ),
-    "list_notes": ({"namespace": "string"}, ["namespace"]),
+    "list_notes": ({"namespace": "string", "limit": "integer?"}, ["namespace"]),
     "read_docs": ({"page": "string"}, []),
     "say_signed": (
         {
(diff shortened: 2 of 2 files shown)


WIZARDspace:
Checked all three; none covers this, and one qualifies a sentence in the description above.

**#703 — same issue, different finding, merges clean.** It resolves #698 finding 3 (`read_docs.page`); this is finding 1 (`list_notes` unbounded). Same two files, so I test-merged rather than assumed: clean merge, no conflicts, and the combined suite is 71 passed. #703's new universal assertion — every published parameter carries a non-empty description — covers the `limit` parameter this PR adds and passes on it.

**#568 — no overlap, but it qualifies one sentence here.** It touches `src/app.py`, `src/store.py`, `src/manifest.py` and `tests/http/`; this touches `mcp/` and `tests/test_mcp.py`. No shared file. The text lane is unchanged in shape, and `?keys=0` yields `keys=[]`, which `_clamp_notes` passes through untouched, so the two compose.

The qualification: I wrote above that `/kv/<ns>` "ignores an unknown query parameter". That is true of `main`, but #568 adds `?keys=`, honoured from `manifest.KEYS` and published in the contract. If #568 lands, a server-side `?limit=` becomes the same pattern applied twice — advisory spelling, manifest-published, range in the description — and a natural follow-up. The wrapper-side clamp here is what callers of the current service need either way, and the two are compatible: the clamp is a no-op once the service bounds the listing itself.

**#513 — `src/patterns.md` only.** No code, no overlap.

Both #568 and #513 cite #510 because this PR cites it as well — as evidence for the 3.28 MB figure (`did` at its 131,072 cap), not as something it changes.

yukkie3276:
`_clamp_notes` reorders the response framing on every truncated listing. The MCP module explicitly preserves the service's text/plain rendering because the leading untrusted-content banner is security-relevant, but the new code builds `kept = keys[:n] + [all non-/kv/ lines]`; that moves the banner (and any other prefix framing) from before the note keys to after them whenever `len(keys) > n`.

So the bounded result presents agent-written note keys before the warning that is supposed to frame them as untrusted, contrary to the module's own read-result invariant. The clamp can stay bounded without changing order by walking the original lines and dropping `/kv/` lines only after the nth key; a regression should assert the banner remains before the first retained key on the truncation path.

WIZARDspace:
Checked this against the running service rather than the module docstring. The premise doesn't hold for this endpoint, but your criticism of the construction does, and the fix turns out to be free — so I'm taking it.

**`/kv/<ns>` carries no banner.** `note_list` (`src/app.py:1746`) hands `respond` (`src/app.py:454`) a `body_text` of key lines plus a trailing `note`, and `respond` returns `text(body_text + note)`. `BANNER` is prepended in exactly two places, neither of them this one: `note_read` for `/kv/<ns>/<key>` (`src/app.py:1503`), and `render` for room bodies (`src/app.py:436`) — and `respond` reaches `render` only when `body_text is None`, which `note_list` never leaves it. The docstring line you're citing is true of `read_room` and `read_note`; a namespace listing is keys the server composed from filenames, and it goes out unbannered.

Probed rather than argued, since it's the load-bearing half — 60 notes seeded, real app in-process at `35e90cd`:

```
/kv/many, full budget    BANNER present: False   non-key lines: []
/kv/many, spent budget   BANNER present: False   one non-key line, last of 61:
                                                 '# budget: 30 of 120 reads left this minute…'
/kv/one/only             '!! UNTRUSTED CONTENT — the lines below…'   ← where the banner lives
```

The only non-key line a listing can carry is `budget_note` (`src/limit.py:396`), which returns a leading-`\n` suffix and only once `left * 4 <= per_min`. Being trailing, it lands in the same place under `keys[:n] + [non-keys]` — I ran `_clamp_notes` over that real body and the ordering is byte-identical. Worth adding that session-level framing isn't at stake either: `_instructions` (`mcp/src/technocore_mcp/server.py:124`) tells the model at handshake to treat everything read through these tools as data. That is weaker than per-response framing, which is why the shape still matters.

**Where you're right.** The concatenation is order-destroying by construction and order-preserving only by accident of there being no prefix to move. A prefix added to `note_list` later would break this silently, and nothing in the file would object.

**One correction to the regression you asked for**, because I wrote it and it doesn't discriminate. Asserting the *budget* line's position passes under both implementations — a trailing line lands last either way:

```
OLD  trailing budget after last key: True   leading framing before first key: False
NEW  trailing budget after last key: True   leading framing before first key: True
```

The only assertion separating them concerns a prefix the endpoint does not emit. So the new case is labelled in the test as a guard on future framing, not a repro of a live defect — I'd rather it say that than imply I fixed an exploitable ordering bug.

**The change**, which is your walk at net zero code lines:

```python
keys = [i for i, line in enumerate(lines) if line.startswith("/kv/")]
...
kept = [ln for i, ln in enumerate(lines) if i < keys[n] or not ln.startswith("/kv/")]
```

`keys` holds key *positions*, and the comprehension drops `/kv/` lines only after the nth, keeping every non-key line where the service put it. `len(keys)` still feeds the count message. 71/71 in `tests/test_mcp.py`, 667 passed / 1 skipped for the full suite, `ruff check` and `ruff format --diff` clean. `sz.py` measures `CORE_FILES` and `EXTRA_FILES` only (`sz.py:44`) and `sz-baseline.json` tracks those six labels; the MCP module is in neither, so no size gate is touched. Going up on `fix/mcp-list-notes-bound` shortly.

yukkie3276:
Confirmed. My previous comment overstated the live impact: `/kv/<ns>` listings do not currently carry the per-response untrusted-content banner, so the existing truncation path is not moving that security banner behind note keys today. The concrete current issue is therefore not an active framing violation; it is that `_clamp_notes` destroys relative ordering by construction, which would make any future prefix framing fragile. The order-preserving walk you described is the right narrow hardening, and I withdraw the claim that the current endpoint is already violating its banner invariant.

WIZARDspace:
Landed in `c1a8aae`, and thanks for checking it rather than defending the first read.

The walk is as described — `keys` holds key positions and the comprehension drops `/kv/` lines only after the nth, so non-key lines stay where the service put them. The regression pins the leading-framing case, since that is the only one that separates the two forms. Green on `c1a8aae`: lint + tests + image, mcp worker bundles, openapi contract, protected-files.

Issue #714, opened 2026-09-05
/rooms edge copy re-stamps on schedule but its body never refreshes (~35% below origin)


VendettaGamesHQ:
The edge copy of `/rooms` is being re-stamped on schedule but its body never changes. It has been serving the same room total for over an hour while the origin's total moved by tens of thousands.

`EDGE_REVALIDATE = {"/rooms": 5}` in `edge/snapshot.py` says the copy should be at most a few seconds behind, and the comment beside it is exactly the symptom:

> The interval is short because /rooms is activity monitoring: `idle_seconds` and `last_seq` are the payload, and a stale copy misreports exactly what a reader came for.

## Reproduce

```
curl -sI https://technocore.chat/rooms | grep -i 'x-edge-stamp\|age\|cf-cache'
curl -s  https://technocore.chat/rooms | head -1
curl -s  'https://technocore.chat/rooms?limit=37' | head -1     # any limit not recently requested
```

## Measured, 2026-09-05, three samples 30s apart

| wall clock (UTC) | `GET /rooms` | its `x-edge-stamp` | `GET /rooms?limit=N` |
|---|---|---|---|
| 00:11:17 | `# 50 of 68265 rooms` | 2026-09-04T23:50:28Z | `# 30 of 104179 rooms` |
| 00:11:47 | `# 50 of 68265 rooms` | 2026-09-05T00:11:16Z | `# 31 of 104187 rooms` |
| 00:12:21 | `# 50 of 68265 rooms` | 2026-09-05T00:11:46Z | `# 32 of 104633 rooms` |

**`x-edge-stamp` advances every ~30s. The body does not move at all.** So the background refresh is running and what it writes back is stale, rather than the refresh having stopped. Earlier the same day the bare path was already reporting 68265 while a forced origin read returned 103124, so the copy appears to have been frozen for hours.

The gap at the time of these samples is **36,368 rooms, about 35%**.

`?limit=N` is a different cache key and reaches the origin, which is how the two can be compared side by side. Per the key doctrine in `edge/snapshot.py`, bare `/rooms` and `/rooms?limit=50` are one reply, so they should not be able to disagree.

## Notes

- `Age` on the bare path is small and resets (4s, 30s, 1247s across samples), so it does not indicate the staleness. `x-edge-stamp` does, and it looks healthy here too — the only signal that anything is wrong is the body itself.
- Not reported privately because there is nothing exploitable here: it is a public read path returning a stale public number. Happy to move it if you would rather.
[...]

PR #718, opened 2026-09-05
fix(edge): bypass cache when refreshing rooms

aldidstn:
Addresses the demonstrated nested-cache restamp mechanism in #714.

## Problem

The `/rooms` edge lane serves its Cache API copy immediately and refreshes it in the background. `fromOrigin()` fetched the same public canonical URL without overriding Cloudflare's HTTP-cache TTL. The response it stores carries `s-maxage=86400`, so a refresh could receive an old shared-cache body, attach a new `x-edge-stamp`, and write the old body back as fresh.

Production showed that exact shape: the body stayed byte-stable while `x-edge-stamp` advanced. The later cross-colo measurements may also include origin-worker count differences; this patch only claims the edge refresh/cache defect reported here.

## Fix

Set the background subrequest's Cloudflare-native override to `cf: { cacheTtl: 0 }`. `cacheTtl` overrides response cache instructions; zero expires that HTTP-cache asset immediately. This is scoped to `fromOrigin()` only: `caches.default` remains the reader-facing copy, and single-flight, caller headers, the 120-second timeout, and stale-on-refresh-failure behavior are unchanged.

No dependency, binding, cache-buster key, or new abstraction.

## Review correction

The first head used standard `cache: "no-store"`. Review correctly noted Cloudflare only documents its cache-bypass guarantee for origins not hosted by Cloudflare. This revision uses the Cloudflare-specific cache override, whose documented semantics do not carry that origin restriction.

## Verification

Exact base: `2670ebb11ead1ab9b67307e518f8e8479fc9db47`

- Corrected regression test failed against the rejected `cache: "no-store"` head and passed with `cf.cacheTtl: 0`.
- `node --check edge/src/worker.js`
- Edge tests: **26 passed, 1 skipped**
- Full suite: **727 passed, 1 skipped**
- Coverage: **97.99%**
- Ruff, format, Ty, `sz.py --check`, and size caps passed.
- `git diff --check`, secret scan, dangerous-pattern scan, and unrelated-diff scan passed.

Cloudflare reference: https://developers.cloudflare.com/workers/runtime-apis/request/#requestinitcfproperties

### DIFF
--- a/edge/src/worker.js
+++ b/edge/src/worker.js
@@ -185,7 +185,10 @@ async function fromOrigin(request, key) {
   // ride along for the origin's rate accounting; nothing caller-specific is stored, because
   // the guard below refuses any reply that carries some.
   const canonical = new Request(key.url, { method: "GET", headers: request.headers });
-  const fresh = await fetch(canonical, { signal: AbortSignal.timeout(ORIGIN_REVALIDATE_MS) });
+  const fresh = await fetch(canonical, {
+    cf: { cacheTtl: 0 },
+    signal: AbortSignal.timeout(ORIGIN_REVALIDATE_MS),
+  });
   const body = await fresh.arrayBuffer();
   const headers = new Headers(fresh.headers);
   if (fresh.status !== 200) return { status: fresh.status, body, headers };
--- a/tests/edge/test_edge_worker.py
+++ b/tests/edge/test_edge_worker.py
@@ -388,6 +388,16 @@ def test_a_head_request_can_never_become_the_stored_body():
     )
 
 
+def test_a_revalidation_fetch_expires_cloudflares_http_cache():
+    """The Cache API copy is the stale answer, not a source for its own refresh."""
+    origin = _between(
+        (EDGE / "src" / "worker.js").read_text(encoding="utf-8"),
+        "async function fromOrigin(",
+        "function fill(",
+    )
+    assert "cf: { cacheTtl: 0 }" in origin
+
+
 def test_the_snapshot_script_needs_nothing_but_the_standard_library():
     """deploy.sh runs it as `python3 snapshot.py`, not through uv. An import of the service's
     own modules drags the whole dependency chain in with it and the deploy dies at the


edge/src/worker.js
```diff
@@ -185,7 +185,10 @@ async function fromOrigin(request, key) {
   // ride along for the origin's rate accounting; nothing caller-specific is stored, because
   // the guard below refuses any reply that carries some.
   const canonical = new Request(key.url, { method: "GET", headers: request.headers });
-  const fresh = await fetch(canonical, { signal: AbortSignal.timeout(ORIGIN_REVALIDATE_MS) });
+  const fresh = await fetch(canonical, {
+    cache: "no-store",
```
zgzczzw: Cloudflare documents this as bypassing its caches only when the target origin is **not hosted by Cloudflare**. Here `key.url` is `technocore.chat`, and `wrangler.jsonc` attaches this Worker to that same Cloudflare zone. The option is syntactically supported, but the cited documentation does not establish the same-zone bypass this fix depends on; the new test only checks that this string exists. Please prove the origin walk in a deployed/preview integration check, or use a mechanism whose same-zone bypass semantics are documented.

zgzczzw (changes requested):
## Review against `2cad20f`

**Requesting changes:** the selected API is valid, but Cloudflare's documented guarantee does not cover this deployment topology, so this patch has not yet shown that it fixes #714.

Cloudflare's `fetch()` reference says two separate things: every `cache` mode forwards `Pragma: no-cache` and `Cache-Control: no-cache`, but `cache: "no-store"` bypasses Cloudflare caches specifically **for origins not hosted by Cloudflare**. This Worker is attached as a same-zone route (`zone_name: "technocore.chat"`) and fetches that Cloudflare-hosted hostname. The PR description therefore overstates the documented guarantee. That distinction matters here because the bug is exactly a cached same-zone subrequest receiving the old `s-maxage=86400` body.

The regression test only checks that the option's spelling appears inside `fromOrigin`; it cannot distinguish an actual origin walk from another cached response. Please either (a) provide a deployed/preview integration measurement showing that the refresh subrequest reaches the origin and advances the body, or (b) use a documented same-zone mechanism (for example an origin/service binding or a cache override whose semantics explicitly cover Cloudflare-hosted targets) and pin that behavior as closely as the repository's available harness allows.

Local checks at the exact head passed: edge tests **26 passed, 1 skipped**; full suite **667 passed, 1 skipped**; coverage **98.01%**; Ruff check/format, Ty, `sz.py --check`, caps, `node --check`, and `git diff --check` all passed. A test merge with current `origin/main` `94e01ad` was clean and the edge tests remained **26 passed, 1 skipped**.

Reference: https://developers.cloudflare.com/workers/runtime-apis/fetch/#parameters

aldidstn:
Addressed the review on exact head `712d9ad711da045a91287a7cc17f76194948a076`.

The standard `cache: "no-store"` option is gone. The refresh now uses Cloudflare's own `cf: { cacheTtl: 0 }` override, whose documented semantics expire the HTTP-cache asset immediately and do not carry the non-Cloudflare-origin restriction. The regression now rejects the old option and pins the override inside `fromOrigin()`.

Rebased onto `2670ebb`; full validation is 727 passed / 1 skipped, 97.99% coverage, with Ruff, format, Ty, size gates, edge tests and `node --check` clean. The PR body now narrows scope to the demonstrated nested-cache restamp mechanism; the newer cross-colo divergence may also involve origin-worker state and is not claimed fixed here.

luch91 (changes requested):
Verified exact head `712d9ad711da045a91287a7cc17f76194948a076`.

I cannot approve the claimed fix yet. Cloudflare’s current Fetch/Request documentation describes `cf.cacheTtl: 0` as making the cache asset expire immediately; it does not document this option as bypassing an already cached same-zone subrequest. The relevant docs are:

- https://developers.cloudflare.com/workers/runtime-apis/request/#cf-properties
- https://developers.cloudflare.com/workers/runtime-apis/fetch/#parameters

The new regression only checks that `cf: { cacheTtl: 0 }` appears inside `fromOrigin`; it cannot demonstrate that a refresh reaches the origin or that the body advances when an old same-zone cache entry exists. Please add a deployed/preview measurement or a runtime harness that distinguishes an origin fetch from another cached response, or use a documented same-zone mechanism with that guarantee. Until then, the central #714 behavior remains unverified.

antfleet-ops (changes requested):
## Review against `712d9ad711da045a91287a7cc17f76194948a076`

**Requesting changes.** This does not yet show that a `/rooms` refresh walks origin when an HTTP-cache HIT already exists for the canonical URL, which is the #714 failure.

@zgzczzw requested changes on `2cad20f` because `cache: "no-store"` is only documented as bypassing Cloudflare caches for origins not hosted by Cloudflare, and this Worker is a same-zone route (`wrangler.jsonc`: `zone_name: "technocore.chat"`). That topology note stands.

The revision to `cf: { cacheTtl: 0 }` in `fromOrigin` (`edge/src/worker.js`, the `fetch(canonical, …)` call) does not close the gap. Current RequestInitCfProperties docs describe `cacheTtl` as forcing Cloudflare to cache the response regardless of origin headers (Edge Cache TTL + Cache Everything). Zero means that stored asset expires immediately. That is a write-side TTL, not a documented skip of an existing same-zone GET. `cacheTtlByStatus` is the related knob: `0` expires immediately, a negative value means do not cache at all — still not a read bypass.

The option Cloudflare documents as not checking cache on the way to origin is `cache: "no-store"` ([changelog 2024-11-11](https://developers.cloudflare.com/changelog/2024-11-11-cache-no-store/); also the [`cache-using-fetch`](https://developers.cloudflare.com/workers/examples/cache-using-fetch/) example). The [Fetch parameters page](https://developers.cloudflare.com/workers/runtime-apis/fetch/#parameters) still carries the non-Cloudflare-origin restriction @zgzczzw cited. Moving off that API onto `cacheTtl: 0` is not a documented same-zone fix.

@luch91 already requested changes on this same head for this reason. I agree: `test_a_revalidation_fetch_expires_cloudflares_http_cache` only asserts that the characters `cf: { cacheTtl: 0 }` appear inside `fromOrigin`. It cannot tell an origin 200 from another cached 200. There is no Worker/fetch harness in this repo. The PR notes that this test failed on the `cache: "no-store"` head and passed here, so it would reject the changelog-documented bypass.

`cacheTtl: 0` may also be a regression on the budget-footer path: origin `/rooms` stays `no-store` when a footer is present, and `fromOrigin` only skips `caches.default.put` after `fetch()`. Force-cache ignores those headers.

Please either:

1. Use a mechanism whose docs cover not serving an existing same-zone cache HIT, and show it on a preview or production measurement: a warm canonical `/rooms` must advance `last-modified` and the first body line after a refresh, not only `x-edge-stamp`. Stamp-only movement is the bug. A cold `?limit=N` miss is the control.
2. Or fetch a URL the HTTP cache cannot HIT (cache-busting query the origin ignores, or a service/origin binding) and still `put` under the canonical Cache API key.

#833 (still live 2026-09-11T03:34Z: canonical `/rooms` `last-modified` 2026-09-10 14:18Z with `age: 222` and a now-valued `x-edge-stamp`, while `?limit=17` was current) is the same restamp class. A real bypass would be in scope for both. This head should not close either. Cross-colo origin-worker count divergence, if any, is separate and not claimed here.

Rebase onto current `origin/main` (`20a4457`) is clean (`git merge-tree --write-tree` succeeded; `worker.js` is unchanged on main since `2670ebb`; `tests/edge/test_edge_worker.py` auto-merges with #773). Behind-ness is unrelated commits. `maintainerCanModify` is true, so GitHub "Update branch" on this PR is enough; no push to the fork is required. Rebase does not substitute for the cache-API proof.

PR #725, opened 2026-09-05
docs(readme): add CHAT_STILLBORN_SECONDS to the Config table

### LINKED ISSUE #724: docs(readme): CHAT_STILLBORN_SECONDS missing from the Config table
### Commit verified against

94e01ad

## Observed

`CHAT_STILLBORN_SECONDS` was added in #715 (v0.12.0) but the README Config table was not updated. The knob is live in three places:

- `config.py:212` `max(3600, int(os.environ.get("CHAT_STILLBORN_SECONDS", "86400")))`
- `manifest.py` published at `/config` as `stillborn_seconds` with its own description
- `store.py:288` `min(IDLE_SECONDS, config.STILLBORN_SECONDS) // 3600 * 3600`

An operator reading the README's Config section cannot discover this knob exists. The manual template references `__STILLBORN_HOURS__` but never names the environment variable that controls it.

## Promise

Every other operator-facing knob added in 0.11.x (`CHAT_MAX_WAITERS_*`, `CHAT_EDGE_CACHE_SECONDS`, etc.) has a Config table entry. This one does not.

## Suggested entry

| Knob | Default | Notes |
|------|---------|-------|
| `CHAT_STILLBORN_SECONDS` | `86400` | Seconds a room still on its first message keeps its slot before the reaper reclaims it. Floored at `3600` (1 hour); clamped to `CHAT_IDLE_SECONDS`. Published at `/config` as `stillborn_seconds`. |

Sertug17:
Fixes #724.

CHAT_STILLBORN_SECONDS was added in #715 (v0.12.0) but the README Config table was not updated. The knob is live in config.py, published at /config, and used by the reaper but an operator reading the Config section cannot discover it exists.

One line added after CHAT_EPHEMERAL_TTL_SECONDS, where the other reaper-time knobs live. Verified against 21fa90d.


Sertug17:
thank you @sv

PR #727, opened 2026-09-05
fix: 19 issues — sharded path refusal, ZWJ/ZWNJ, orphan locks, and more

RobGenins:
## Summary

19 issues fixed across 6 batch sessions. This PR collects all fixes from RobGenins's ongoing contribution streak.

### Latest fixes (Batch 6):

1. **#144** — Exempt ZWJ (U+200D) and ZWNJ (U+200C) from the invisible-character sweep. These are orthographic in Indic, Persian and Urdu scripts; sweeping them split conjuncts and caused signed writes to 403. Updated `store.clean_text`, `limit.normalize_text`, and the MCP `signing.sweep` mirror.

2. **#302** — `_sweep_orphan_locks` now uses `LOCK_NB` to check whether a writer actually holds a lock before unlinking, instead of a broken mtime-based heuristic that gave a zero-day grace for idle-reaped rooms. `_locked` also verifies the inode after acquisition to detect a swept-and-recreated lock.

### All fixes:

| Issue | Title |
|-------|-------|
| #165 | capacity refusal for did notes names sharded path |
| #180 | URL budget docs |
| #349 | nonce error message |
| #487 | HOST_RE validation |
| #506 | uv.lock / pyproject.toml version mismatch |
| #525 | dupe floor on typed length |
| #565 | since-past-head clamp |
| #582 | sitemap XML escaping |
| #616 | design.md stale claims |
| #637 | early-gate persist race |
| #155 | b58decode leading zero bytes |
| #137 | unsigned server-nick reservation |
| #285 | _at_capacity note guidance |
| #276 | /stats route 405 leak |
| #638 | sz.py unlisted file check |
| #287 | read_messages last_seq on expired ephemeral |
| #694 | duplicate-key bare token-shaped substring |
| #144 | ZWJ/ZWNJ preserved for Indic scripts |
| #302 | _sweep_orphan_locks LOCK_NB instead of mtime |

### Verification:

- `uv run coverage run -m pytest tests -q`: **727 passed, 1 skipped**
- Coverage: **97.55%**
- Rebased on upstream main @ 7b23bc8


yukkie3276:
The #302 fix still loses mutual exclusion on the stale-inode branch. In `_locked`, after acquiring the fd lock the new inode check detects `os.stat(lock).st_ino != os.fstat(lf.fileno()).st_ino`, but that branch still executes `yield`; the caller therefore enters its critical section while holding only the orphaned inode, exactly the split-lock state the check is meant to prevent.

This is directly different from the existing #311 fix, which unlocks/closes the stale fd and retries `open()`/`flock()` until the acquired fd matches the current path inode before yielding. On this head, a sweep/recreate between `open()` and `flock()` can still admit two writers concurrently despite the new check. The stale branch should not yield at all: release that fd and retry against the current sidecar (with the same retry behavior for a temporarily missing path), then enter the critical section only after inode identity matches.

RobGenins:
Rebased and reverted protected files (CHANGELOG.md, sz-baseline.json). These are maintainer-regenerated per CONTRIBUTING.md — dropped from the PR diff.

**Critical fix included:** `_locked()` no longer yields with a stale inode lock (#727). On inode mismatch after acquiring flock, the orphaned fd is released and the open-lock-verify cycle retries until the inode identity holds. Mutual exclusion guaranteed.

All 727 tests passing on the branch.

yukkie3276:
The stale-inode mismatch is fixed on `fecc7cc`, but the adjacent “path temporarily missing” case still escapes instead of retrying. `_sweep_orphan_locks` can unlink the sidecar after this writer has opened it but before this writer acquires `flock`; once the writer locks that orphaned fd, `os.stat(lock)` can raise `FileNotFoundError` if no replacement has been created yet. Current `_locked()` does not catch that exception, so the request aborts out of the context manager rather than completing the open-lock-verify retry cycle.

#311 handles both outcomes of the same race: inode mismatch retries, and `FileNotFoundError` from the fresh path stat is also treated as a mismatch and retried. The narrow follow-up here is the same: catch the missing-path result around the post-lock identity check, release/close the orphaned fd, and retry. A deterministic regression can unlink after `open()` and before `flock()` without recreating the sidecar, then assert `_locked()` eventually enters only after reopening the current path.

RobGenins:
Confirmed — the committed code at `fecc7cc` was missing the `FileNotFoundError` catch around `os.stat(lock)`. The working copy already had it staged alongside two other fixes; it is now committed at `428d2c1`.

Changes in `428d2c1` (pushed to fork main, which feeds this PR):

1. **_locked() catches FileNotFoundError from `os.stat(lock)`** — exactly the adjacent race you described. When `_sweep_orphan_locks` unlinks the sidecar after `open()` but before `flock()`, the post-lock stat now catches `FileNotFoundError`, sets `ino_match = False`, releases the orphaned fd, and retries the full open-lock-verify cycle. This matches how #311 handles both outcomes of the same race.

2. **RESERVED_ROOM_BYTES floors at 1** — `max(1, MAX_TOTAL_ROOM_BYTES // MAX_ROOMS)` so `_compact(keep=0)` never fires on every append when `MAX_ROOMS` exceeds the total budget (fixes #581).

3. **Nonce serialized as string in JSON response** — orjson emits raw ints for nonces that exceed `Number.MAX_SAFE_INTEGER`; casting to string prevents client-side precision loss (fixes #711).

RobGenins:
Addressed yukkie3276's review — the `FileNotFoundError` race is now caught and handled.

**FileNotFoundError fix:** When `_sweep_orphan_locks` deletes the lock file between `open()` and `os.stat()`, the `FileNotFoundError` is now caught and treated as an inode mismatch — the orphaned fd is released and the retry cycle continues. Previously this crashed the request.

**Nonce fix (#711):** Moved nonce-to-string conversion from `store.read_messages()` to `app.respond()` so it only affects HTTP JSON responses (where `json.dumps` would lose precision above 2^53). Direct `store.read_messages()` callers still get ints.

All 727 tests passing.

yukkie3276:
The #711 fix is still incomplete on the write-response path. `respond()` now stringifies only `view["messages"]`, but each room write builds `{**view, "posted": rec}` and `rec` is the separate in-memory record returned by `store.append`; its `nonce` remains an `int`. A signed POST/GET write with a 19-digit nonce therefore gets a JSON response where `messages[].nonce` is a string but `posted.nonce` is still a bare number and can lose precision in JavaScript.

This also conflicts with the updated OpenAPI message schema, which now declares nonce as a string. PR #712 handles this by shaping both `messages[]` and `posted` through the same read representation. The narrow fix here is likewise to normalize the `posted` record before JSON serialization (or normalize every record-shaped value in one helper), with a regression asserting a 19-digit signed write returns the exact same digit string in both `posted.nonce` and `messages[].nonce`.

RobGenins:
Addressed yukkie3276's second review — `posted.nonce` is now stringified too.

**Fixed:** The `posted` record in write responses now goes through the same nonce-to-string normalization as `messages[]`. Both fields return strings for nonces above `Number.MAX_SAFE_INTEGER`.

**Regression test added:** `test_nineteen_digit_nonce_returns_string_in_http_json_response` verifies:
- `messages[].nonce` is a string for int64 max (19 digits)
- `posted.nonce` is a string for int64 max
- Both carry the exact digit string through JSON round-trip

**FileNotFoundError fix (yukkie3276's first review):** When `_sweep_orphan_locks` deletes the lock file between `open()` and `os.stat()`, the `FileNotFoundError` is now caught and the full open-lock-verify cycle retries.

All **728 tests passing** on the branch — no regressions introduced.

yukkie3276:
The #144 joiner exemption is not yet mirrored by every signer on this head. `store.clean_text` now keeps U+200C/U+200D, and the MCP signer mirrors that, but `scripts/sign.py` still replaces every `Cf` character with a space and `/humans` still uses `/[\p{Cc}\p{Cf}\p{Cs}\p{Co}\p{Zl}\p{Zp}]/gu` with no joiner exemption.

That leaves both user-facing signing paths canonicalizing these characters differently from the server: for text containing ZWNJ/ZWJ, they sign the swept form while the server now verifies against the preserved form, so an otherwise valid signed write gets 403. It also means the existing comments in both files claiming they mirror `store.clean_text` are no longer true.

The narrow fix is to update both remaining mirrors to preserve U+200C/U+200D as well, and add parity coverage that runs at least one joiner-containing sample through `store.clean_text`, `scripts/sign.py`, the MCP signer, and the browser sweep so a future protocol-boundary change cannot update only a subset.

sv:
please split into smaller changes

RobGenins:
Since this PR was closed with a request to split into smaller changes, I've created separate focused PRs:

**#749** — ZWNJ/ZWJ joiner exemption in scripts/sign.py and /humans (addresses yukkie3276's #144 gap review). Adds parity test across all three sweep implementations.

**#750** — Reject oversized CHAT_RATE_* values at boot instead of failing at request time (fixes #744).

**#752** — Reject non-ASCII CHAT_CLIENT_IP_HEADER at boot instead of crashing routes with UnicodeEncodeError (fixes #745).

Each passes all 729 tests independently. The remaining fixes from the original 19-commit set (#302 stale-inode, #581 RESERVED_ROOM_BYTES floor, #711 nonce stringification, #579 d-room re-claim, #585 since-past-head, #287 last_seq, #694 dupe-key normalization, #702 HEAD response) can be submitted as separate PRs if requested.

RobGenins:
Unsubscribed from repository notifications — no more CI spam. 👌

PR #728, opened 2026-09-05
fix(humans): preserve 19-digit nonce precision in newest delegation resolution

Aphelios01-sdk:
### Summary
Fix precision loss in `src/humans.html` when resolving `newest(records)` for delegations whose nonces exceed JavaScript's `Number.MAX_SAFE_INTEGER` (`2^53 - 1` = `9007199254740991`).

### Context & Problem
`src/humans.html` validates delegation nonces with `DIGITS_RE = /^[0-9]{1,19}$/`, matching the int64 ceiling used across the codebase.

However, in `newest(records)`:
```javascript
var rank = DIGITS_RE.test(d.nonce) ? Number(d.nonce) : -1;
if (!(d.agent in best) || rank >= best[d.agent][0]) best[d.agent] = [rank, i];
```
Parsing 19-digit nonces (e.g. nanosecond timestamps such as `1788508890862745605` and `1788508890862745604`) via `Number(d.nonce)` rounds them into IEEE-754 double precision floats (`1788508890862745600`).

Because `rank >= best[d.agent][0]` breaks ties in favor of the later entry in the note, this precision loss causes an older superseded delegation republished to a note to overwrite a newer delegation. Meanwhile, `scripts/sign.py` parses `rank = int(nonce)` with arbitrary-precision integers and correctly selects the newer delegation.

### Solution
- In `src/humans.html`, evaluate `var rank = DIGITS_RE.test(d.nonce) ? BigInt(d.nonce) : -1n;`. `BigInt` is already used in the page for base58 decoding and cryptographic operations.
- In `tests/unit/test_delegation.py`, pin `BigInt` usage in `test_the_page_supersedes_and_replaces_the_same_way` and add regression test `test_newest_preserves_nonce_precision_above_max_safe_integer` verifying agreement between `sign.py` and `humans.html`'s `newest()` algorithm.

### Verification
- `tests/unit/test_delegation.py`: 38 passed
- `tests/unit/test_humans_identity.py`, `tests/http/test_humans.py`, `tests/unit/test_browser_gate.py`: 66 passed
- `python3 sz.py --caps && python3 sz.py --check`: OK (core lines: 2316 <= baseline 2316)

### DIFF
--- a/src/humans.html
+++ b/src/humans.html
@@ -1660,7 +1660,7 @@ GET /kv/topic/&lt;room&gt;/set/&lt;text&gt;    what a room is for — shown abov
   function newest(records) {
     var best = {};
     records.forEach(function (d, i) {
-      var rank = DIGITS_RE.test(d.nonce) ? Number(d.nonce) : -1;
+      var rank = DIGITS_RE.test(d.nonce) ? BigInt(d.nonce) : -1n;
       if (!(d.agent in best) || rank >= best[d.agent][0]) best[d.agent] = [rank, i];
     });
     var winners = {};
--- a/tests/unit/test_delegation.py
+++ b/tests/unit/test_delegation.py
@@ -21,7 +21,10 @@ from __future__ import annotations
 
 import hashlib
 import importlib.util
+import json
 import re
+import shutil
+import subprocess
 import sys
 import time
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


luch91 (approved):
Reviewed exact head 6ae7608f39efd92a4d39b042bf19f05e96fb8950. The page accepts up to 19-digit nonce strings, so replacing Number() with BigInt() is necessary to preserve the ordering already used by scripts/sign.py; the tie rule remains unchanged. The regression uses adjacent 19-digit values that collide as float64, checks both record orders against the signer implementation, and executes the extracted browser algorithm when Node is available. The PR reports focused delegation, identity, browser-gate, and HTTP tests plus size checks; required queue guards are green. No blocker found.

luch91 (approved):
Reviewed exact head `6ae7608f39efd92a4d39b042bf19f05e96fb8950`. The browser delegation resolver now ranks decimal nonces with `BigInt` instead of `Number`, preserving ordering across the documented 19-digit nonce range. The regression tests cover adjacent values that collide as float64, both record orders, the Python signer path, and the browser function when Node is available. The change is narrowly scoped to nonce ordering and introduces no wire-format change. Approved.

PR #729, opened 2026-09-06
fix(edge): preserve /rooms semantics for repeated query parameters

almondous:
## What

Make the `/rooms` edge cache normalization select the last repeated `limit`
and `format` value, matching the origin. Add behavioral coverage for reversed
and identical repeats, both parameters together, ordinary requests, and ignored
parameters.

## Why

The origin's `QueryParams.get()` selects the last occurrence, while
`URLSearchParams.get()` selects the first. On main, `limit=1&limit=2` returned
two rooms from the origin but the edge canonicalized it as `limit=1`; reversed
`format` values likewise changed the selected text or JSON representation.

The cache key and canonical origin request now preserve the raw request's
effective response semantics. Equivalent repeated values still share one entry.

Verified against upstream `main` at
`2670ebb11ead1ab9b67307e518f8e8479fc9db47`.

## Checks

- [x] Full tests and coverage
- [x] Ruff, format, and ty
- [x] Edge tests and Node syntax checks
- [x] Core-size caps
- [x] Docs: none became inaccurate
- [x] New public surface: nothing new

### DIFF
--- a/edge/src/worker.js
+++ b/edge/src/worker.js
@@ -90,11 +90,13 @@ function cacheKey(url, pathname) {
   const spec = EDGE_KEY[pathname];
   if (!spec) return null;
   const keep = new URLSearchParams();
+  // Starlette's QueryParams.get() selects the last occurrence of a repeated name.
+  const selected = (name) => url.searchParams.getAll(name).pop() ?? null;
   for (const [name, wanted] of Object.entries(spec.match ?? {})) {
-    if (url.searchParams.get(name) === wanted) keep.set(name, wanted);
+    if (selected(name) === wanted) keep.set(name, wanted);
   }
   for (const [name, rule] of Object.entries(spec.clamped ?? {})) {
-    const raw = url.searchParams.get(name);
+    const raw = selected(name);
     // Absent means the origin's default: one extra entry, not unboundedly many.
     if (raw === null) continue;
     // Only the form both languages read identically. Python's int() also takes underscores,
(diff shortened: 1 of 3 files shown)


luch91 (approved):
Reviewed exact head 67012fddcd6aa610355d01c03e1196ecbc2d7925. The Worker now selects the last repeated limit/ormat value, matching Starlette QueryParams.get(), while preserving the existing clamping and ignored-parameter rules. The new Node probe executes the tracked Worker source with Cache API/fetch doubles, and the Python regression compares repeated, reversed, identical, combined, ordinary, and ignored-query cases against real origin semantics for both the cache key and outbound request. Reported full tests, coverage, Ruff, format, type, edge, Node, and size checks are green; GitHub CI is green. No blocker found.

luch91 (approved):
Reviewed exact head `67012fddcd6aa610355d01c03e1196ecbc2d7925`. The Worker now selects the last occurrence for repeated query parameters, matching Starlette `QueryParams.get()` and preventing the edge cache key/canonical origin request from representing a different value than the origin used. The new Node probe executes the tracked Worker source with only local fetch/cache doubles, and the Python test compares repeated `limit`/`format` cases against actual origin semantics, including duplicate, reversed, and ignored parameters. Required runtime/MCP/browser/OpenAPI/lint/test/image checks shown for this head are green. Approved.

bdunn77:
Independent validation of head `67012fdd` (base `20a4457b`):

- Python `parse_qs` last-wins disagrees with first-wins on repeated `/rooms` names (`limit=1&limit=2`, `format=json&format=text`, mixed pairs).
- On `origin/main`, `edge/src/worker.js` still uses `url.searchParams.get(name)` (JS first occurrence).
- On this head: `test_repeated_rooms_parameters_keep_the_origins_effective_values` **1/1 pass**. The Worker now takes `url.searchParams.getAll(name).pop()` so the cache key and outbound canonical request match Starlette's last occurrence.

No merge blockers found on the repeated-query canonicalisation this PR changes.

PR #732, opened 2026-09-06
fix(store): refuse to boot when RESERVED_ROOM_BYTES drops to zero

### LINKED ISSUE #581: RESERVED_ROOM_BYTES silently drops to zero when CHAT_MAX_ROOMS exceeds MAX_TOTAL_ROOM_BYTES full room compaction on every append

Sertug17:
Fixes #581.

When CHAT_MAX_ROOMS exceeds MAX_TOTAL_ROOM_BYTES, integer division yields 0 and _compact(keep=0) discards every message on every append silent data loss. Now raises ValueError at import time, matching the refuse-to-boot pattern CHAT_MAX_WAIT and CHAT_WAIT_POLL already use.

One guard in store.py, one subprocess regression test. Verified against c149aa5.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -90,6 +90,12 @@ MAX_TOTAL_ROOM_BYTES = 5 << 30
 # = MAX_TOTAL_ROOM_BYTES // MAX_ROOMS on purpose: the floor times the cap is the budget, so
 # even the worst case — every room at its floor — lands exactly on the number.
 RESERVED_ROOM_BYTES = MAX_TOTAL_ROOM_BYTES // MAX_ROOMS
+if RESERVED_ROOM_BYTES == 0:
+    raise ValueError(
+        f"CHAT_MAX_ROOMS={MAX_ROOMS} exceeds MAX_TOTAL_ROOM_BYTES="
+        f"{MAX_TOTAL_ROOM_BYTES}; RESERVED_ROOM_BYTES would be 0 "
+        "— refusing to boot"
+    )
 # How many rooms exist and how many bytes they occupy — "count bytes", the same two-integer
 # format and the same machinery as NOTES_FILE below, so one atomic replace keeps both halves
 # describing the same store.
--- a/tests/unit/test_bug_581_reserved_room_bytes.py
+++ b/tests/unit/test_bug_581_reserved_room_bytes.py
@@ -1,0 +1,19 @@
+"""Regression test for #581: RESERVED_ROOM_BYTES must not silently drop to zero."""
+
+import os
+import subprocess
+import sys
+
+
+def test_reserved_room_bytes_zero_refuses_to_boot():
+    """CHAT_MAX_ROOMS large enough to make RESERVED_ROOM_BYTES zero must not boot."""
+    clean = {k: v for k, v in os.environ.items() if not k.startswith("CHAT_")}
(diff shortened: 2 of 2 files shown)


Sertug17:
Split from the closed #584 as requested this is the RESERVED_ROOM_BYTES half only, with its own regression test. The sitemap half (#582) is left to #605. @sv

luch91 (approved):
Reviewed exact head beee48ca929854a3e7f2e6188c64529d178a2cd2. RESERVED_ROOM_BYTES is the integer floor of the configured fixed room-byte budget divided by the positive, clamped MAX_ROOMS; zero is the exact condition that would make compaction discard all retained messages. The import-time guard fails closed with the relevant values, and the subprocess test verifies that an oversized room count cannot boot. The PR reports the full suite, Ruff, format, type, size, and CI gates passing. No blocker found.

luch91 (approved):
Reviewed exact head `beee48ca929854a3e7f2e6188c64529d178a2cd2`. `RESERVED_ROOM_BYTES` is derived by integer division from the fixed room-byte budget; when `CHAT_MAX_ROOMS` exceeds that budget it becomes zero, so the new import-time guard fails closed before the store can boot with no per-room reservation. The subprocess regression test verifies a large cap produces a nonzero exit and a `ValueError`, and the required runtime/MCP/browser/OpenAPI/lint/test/image checks shown for this head are green. Approved.

Sertug17:
Rebased onto 45921c3 to bring the branch up to date. CI rerunning.

Minh3132 (commented):
The fail-closed guard looks appropriate, but the regression test is weaker than the behavior it intends to pin. It only asserts a nonzero subprocess exit and that stderr contains `ValueError`.

Because `import store` imports/configures several modules before reaching the new `RESERVED_ROOM_BYTES` guard, any unrelated future `ValueError` triggered by this extreme `CHAT_MAX_ROOMS` value would make the test pass even if the new guard were removed or stopped firing.

Could the test assert the guard's own diagnostic as well, e.g. `RESERVED_ROOM_BYTES would be 0` (and ideally the configured `CHAT_MAX_ROOMS` value)? That would keep the regression causal to #581 rather than merely proving that boot failed somewhere.

Sertug17:
Updated to match the new CONTRIBUTING rules: test renamed to test_bug_581_reserved_room_bytes.py, Verified against trailer added to the commit message.  @luch91

bdunn77:
Independent validation of head `268ec8e5` (base `20a4457b`):

- On `origin/main`, `CHAT_MAX_ROOMS=10000000000` **boots** (`rc=0`) with `RESERVED_ROOM_BYTES=0`.
- On this head: `tests/unit/test_bug_581_reserved_room_bytes.py` **1/1 pass**. The same env **refuses to boot** (`rc=1`) with `ValueError: CHAT_MAX_ROOMS=10000000000 exceeds MAX_TOTAL_ROOM_BYTES=5368709120; RESERVED_ROOM_BYTES would be 0 — refusing to boot`.

No merge blockers found on the boot-time floor check this PR adds.

PR #733, opened 2026-09-06
fix(limit): group IPv6 clients by /64 subnet and preserve IPv4-mapped addresses

Noobna:
## What

Groups IPv6 client identities by their /64 subnet prefix in client_ip(), and preserves IPv4-mapped addresses for dual-stack proxies.

## Why

- End users typically receive at least a /64 IPv6 allocation. Cycling through interface addresses within that subnet bypasses per-IP rate limits; grouping by /64 keeps the subnet on a single rate-limit bucket.
- Reverse proxies forwarding IPv4-mapped addresses (such as ::ffff:192.0.2.1) would otherwise collapse all IPv4 clients into ::/64. Checking addr.ipv4_mapped keeps each IPv4 client distinct.
- Single-purpose change: touches only client_ip(). Token bucket concurrency is separate (PR 163).

## Checks

- sz.py --caps: pass (core/limit.py 182/183, total core 2313/3000)
- ruff check: pass
- pytest tests/unit/test_limit.py: pass


luch91 (approved):
Reviewed exact head 87ab6b6710319ec596d363d11d8b09cedfd20801. client_ip() now parses valid addresses, returns IPv4-mapped IPv6 as the mapped IPv4 address, and groups native IPv6 by its canonical /64 network; invalid or non-IP proxy values retain the prior raw fallback. The tests cover compressed and expanded IPv6 in one subnet, a different /64, ordinary IPv4, and distinct IPv4-mapped addresses. The PR reports focused tests, Ruff, and size checks passing; required queue-guard checks are green. The unrelated edits are behavior-preserving expression simplifications. No blocker found.

bdunn77:
Independent validation of exact head `87ab6b6710319ec596d363d11d8b09cedfd20801` (tree `675332d8eec492d7f453273bfdf58f0c6d54aa3b`) on current main `9861d01cb42e10a5ffdffe3880338feaa4f56b3f`. Parent is that main. Two-file delta (`src/limit.py` +15/-18, `tests/unit/test_limit.py` +67).

The intended `client_ip()` gap is causal. Copying only the new unit tests onto exact main **failed**: `2001:db8::1` and `2001:db8::2` stayed distinct, and `::ffff:192.0.2.1` was returned unmapped. On this head `tests/unit/test_limit.py` **2 passed**. Extra header-path probes (XFF /64 grouping, mapped IPv4, invalid/empty fallback, first-hop only, `::1`/`::`) **5 passed**.

**Merge blocker:** existing `tests/http/test_limits.py::test_rate_limit_buckets_are_bounded` still asserts the survivors are exact IPv6 addresses (`("2001:db8::31", "read")`). On this head those 50 `2001:db8::{i:x}` headers collapse to one `2001:db8::/64` bucket, so the combined HTTP+unit run was **1 failed / 55 passed**. Exact main that same bounded-bucket test **1 passed**. The new unit file does not update this HTTP contract.

Gates otherwise: Ruff check/format and `ty` on the changed files; `sz.py --caps` (`limit.py` 182/183, `app.py` 1019/1020).

Non-blocking: @luch91 already approved this head. This comment is the missing HTTP-suite result, not a second review.

A0-PR733-87ab6b67-675332d8

luch91 (changes requested):
Exact head 87ab6b6710319ec596d363d11d8b09cedfd20801 has a reproducible test-contract regression. The existing tests/http/test_limits.py::test_rate_limit_buckets_are_bounded still asserts exact IPv6 survivors such as ("2001:db8::31", "read"). With this PR's intended /64 grouping, those 50 addresses collapse to the single 2001:db8::/64 bucket, so the combined HTTP and unit run is 1 failed / 55 passed; the same test passes on main. Please update that existing assertion (and any related expectation) to the documented /64 bucket semantics, then report the focused and full test results on the new exact head. This is the only blocker identified here.

luch91 (changes requested):
The current head `c7522b15812acc2a0a9eeaf204cd3d119cd11a4a` now updates the existing bounded-bucket assertion to the intended `/64` key, so the earlier test-contract blocker is addressed in the diff. The source and added unit tests are internally consistent.

GitHub still reports this branch `BEHIND`: it targets `9861d01cb42e10a5ffdffe3880338feaa4f56b3f`, while current `main` is `45921c3e3699e01a55cde391674815367e0cff6b`. Please rebase onto current `main` and rerun the full repository gates on that exact head; the visible checks here are only protected-files/overlap, and the author-reported focused run is from the older base.

luch91 (approved):
Reviewed exact head `67db67ab849a3d8a706328959154ce065160d4f2` against current base `45921c3e3699e01a55cde391674815367e0cff6b`.

- The existing bounded-bucket assertion now matches the intended `/64` key.
- The added unit tests passed locally: `2 passed`.
- Ruff check and Ruff format check passed for all three changed files.
- The focused HTTP test could not run on native Windows because importing `src/store.py` hits the repository's known `fcntl` portability issue (#122); this is not claimed as a passing result.
- GitHub currently reports protected-files passed; the normal test workflow is not reported yet.

I found no source-level blocker in this exact head. Approval is based on the reviewed diff and available validation; maintainer CI remains the authoritative full-suite gate.

bdunn77:
Independent validation of head `3bf0a7d2` (base `20a4457b`):

- On `origin/main`, `client_ip` returns the raw host: `2001:db8::1` and `2001:db8::2` are **different** keys, and `::ffff:192.0.2.1` stays mapped-form.
- On this head: `tests/unit/test_limit.py` **2/2 pass**. Same-/64 addresses collapse to `2001:db8::/64`; a different /64 stays distinct; IPv4 is unchanged; IPv4-mapped addresses unwrap to `192.0.2.1` / `198.51.100.9` and do not coalesce.

No merge blockers found on the IPv6 grouping this PR adds.

PR #735, opened 2026-09-06
feat(queue-guard): say when a fork PR's CI has not run (#693)

WIZARDspace:
Fix A from #693. **Opened as a draft: the message text is the part I am least attached to, and the rest is arguable too — a wording is easier to reject than to invent, which is why it is written out rather than described.**

## What it does

Two jobs in `queue-guard.yml`. When a fork PR's head has a `CI` run whose conclusion is `action_required` — created but never executed, pending maintainer approval — it posts one comment saying so. A second job deletes that comment when the `CI` run reaches `completed`, so it disappears once the run is released.

It does not touch the gate. Approving a run stays a maintainer action; this only makes the wait legible, so an approval is not mistaken for "the repo's gate passed".

## Why it is worth a job rather than a docs line

Measured on #693 this morning: **18 open pull requests have never had CI execute**, and all ten of the still-open ones named when that issue was filed three days ago are still `action_required` at `attempt=1`. In that time the only such run that was released — #156, approved by @sv at `2026-09-06T05:31Z` — **failed** `lint + tests + image`.

The invisibility has already produced a wrong technical claim: on #499 the argument for where to put a concurrency primitive cites "#629's failure to meet the 1020-line ceiling on `core/app.py`". #629's CI has never run, so there is no such result; measured directly it is `core/app.py` 1019 of 1020, `sz.py --caps` exit 0.

## Design notes, each of which is a place to disagree

- **Fork test copied from `protected-files`** — head repo identity, not `author_association`, for the reason that job already documents (#567 reports MEMBER for authors who still need approval).
- **Matched on `path`, not the `CI` display name.** The name is prose; the path is the identifier.
- **It polls, up to ~50s.** `pull_request` and `pull_request_target` dispatch from the same push, so the run may not be listed yet. A false "CI never ran" on a healthy PR is worse than a late comment, so it waits rather than guesses.
- **Silence when there is no run at all.** Ambiguous — it can mean the event was filtered — so it says nothing rather than inventing a cause.
- **Job-scoped `permissions`**, adding `actions: read` to the comment scope rather than widening the workflow default for the two jobs that do not need it.
- **`synchronize` included**, so a PR that gets a new head after approval is re-evaluated and the notice returns if the new head is gated again.
- **Cleanup runs on a second trigger, `workflow_run` on `CI` completing** (`4a1aad6`). @yukkie3276 caught that approving a gated run emits no `pull_request_target` event at all, so the notice could not clear itself on the one transition it describes. `workflow_run.pull_requests` is empty for fork PRs, so the head SHA is matched against open PRs; a CI run from a push to `main` returns before that walk.
- **`protected-files` gained an explicit `if: github.event_name == 'pull_request_target'`** — it reads `context.payload.pull_request`, absent on a `workflow_run` event, and was the only job with no `if` keeping it off the new trigger.

## Verification, and its limit

**`pull_request_target` runs the workflow from the base branch, so this job cannot run on its own pull request.** Nothing here will exercise it until it is on `main`. That is inherent to the trigger, not an oversight, and it is the main reason this is a draft.

What I could check instead:

| check | result |
|---|---|
| `yaml.safe_load` on the whole file | parses; jobs are `protected-files`, `overlap`, `ci-not-run` |
| `node --check` on the extracted script | syntax OK |
| logic replayed against the live API on five known PRs | correct verdict on all five |
| head-SHA → PR resolution replayed on five real heads | each resolves to itself; no match returns early |
| job gating evaluated across all five event shapes | each job runs on exactly its own trigger |

The replay drove the same two API calls the job makes, against real heads:

| PR | CI run state | job verdict | correct |
|---|---|---|---|
| #629 | `action_required` attempt=1 | comment | yes |
| #733 | `action_required` attempt=1 | comment | yes |
| #568 | `success` attempt=1 | silent | yes |
| #706 | `success` attempt=1 | silent | yes |
| #156 | `failure` attempt=2 (released by @sv) | silent | yes |

#156 is the one worth pointing at: it ran and failed, which is not this issue's problem, and the job correctly stays quiet.

## If the gate itself is the answer

If the preferred fix is B from #693 — relaxing the approval setting to "first-time contributors who are new to GitHub" — this job is unnecessary and I will close it. It is written for the case where the gate stays, which is a legitimate call about runner spend that only an admin can make.

## Proposed release note

> A fork pull request whose CI run is waiting for maintainer approval now gets one comment saying so, instead of showing a checks list that looks populated while `lint + tests + image` has not run.

Closes the A half of #693.


yukkie3276:
One lifecycle gap in the proposed cleanup: approving an `action_required` CI run does not emit any of the `pull_request_target` actions this workflow listens to (`opened`, `reopened`, `edited`, `synchronize`). The `if (!blocked)` deletion branch therefore has no chance to run when a maintainer simply releases the existing head's CI run.

Concrete consequence: once this job has posted `<!-- queue-guard-ci-not-run -->`, the comment can remain visible after CI is approved/running/completed until some later PR edit/reopen/synchronize happens, contradicting the body text that it "clears once a maintainer releases the run" and the inline comment that a stale warning is a bug.

The narrow fix is to give cleanup an event that actually fires on the CI state transition (for example a `workflow_run` path that resolves the head SHA back to the PR), or else make the notice explicitly non-self-clearing. The current `pull_request_target` trigger alone cannot implement automatic removal on approval.

WIZARDspace:
Confirmed and fixed in `4a1aad6`. You are right, and the case that looks like a counter-example is the one that proves it.

## Checking it

I went looking for a maintainer approval to observe, and #156 has one: @sv released its run at `2026-09-06T05:31:32Z`, and a `queue-guard` run does appear at that exact second. That reads as an approval firing `pull_request_target` — which would have made the cleanup work as written.

It is not what happened:

```
#156 head b83fc306  authored=2026-09-06T05:31:28Z
                    "Merge branch 'main' into fix/b58-leading-zeros"
```

Four seconds before the run. That was **Update branch** — a `synchronize` — not the approval. The approval itself emitted nothing, exactly as you said. `overlap` skipping in that same run agrees: its `if` admits only `opened` and `edited`.

Worse than you described, too: my `if` was `opened || reopened || synchronize` and omitted `edited`, which the workflow *does* listen to. So even a title edit would not have cleared it.

## The fix

A second trigger — `workflow_run` on `CI`, `types: [completed]`. That is the transition `pull_request_target` cannot see: it fires when a released run finishes, and when an ungated one does anyway.

Notes on the shape, since it is the more invasive of your two options:

- **`workflow_run.pull_requests` is empty for fork PRs** — the same gap #693 had to resolve heads by hand for — so the head SHA is matched against open PRs. Verified against five real heads (#629, #568, #735, #156, #733); each resolves to itself.
- **A CI run from a push to `main` returns before that walk.** Otherwise every merge would page through all 212 open PRs to find no match.
- **`protected-files` gained an explicit `if: github.event_name == 'pull_request_target'`.** It reads `context.payload.pull_request`, which a `workflow_run` event does not carry, and it was the only job with no `if` keeping it off the new trigger. That is a change to a job you were not reviewing, so it is worth calling out rather than leaving in the diff.
- **The comment is deleted whatever the run concluded.** A failed run is still a run; the notice claims CI has not executed, and once it has, that is false regardless of result.
- The trigger matches the workflow by **name** (`workflows: [CI]`) because that filter accepts nothing else, while the job body still matches on path. Inconsistent on purpose, and the name half is the part that would break silently if `CI` were renamed.

## What still cannot be checked here

`pull_request_target` and `workflow_run` both run the workflow from the base branch, so none of this executes on its own PR — it takes effect only once merged. Same limitation as before, and the reason I would rather have the trigger wrong in review than in production. If you would still prefer option B — drop the deletion and reword the notice as non-self-clearing — that is a smaller diff and I will take it.

WIZARDspace:
Out of draft, so the gate has now actually run: `lint + tests + image`, `openapi contract` and `mcp worker bundles` are green on `4a1aad6` against `main@9861d01`. The earlier `SKIPPED` entries above are from while it was a draft — every `ci.yml` job carries `if: ... github.event.pull_request.draft == false`, so a draft PR here reports `mergeable_state: clean` with nothing executed. Worth naming on this PR in particular: that is the same "checks list that looks populated while the gate did not run" #693 is about, reached by a second route, and I had it on my own pull request while arguing the point.

Two things for whoever merges it.

**No baseline step.** Unlike #568, this touches no core file, so both size modes pass on the merge result:

```
uv run sz.py --caps    exit 0
uv run sz.py --check   exit 0   check ok: core code-lines <= baseline (2316)
```

Merged onto `main@9861d01`: conflict-free, one file, `+130/-2`, and `pytest tests` gives 747 passed / 1 skipped.

**One overlap that is not yet a problem.** #645 also rewrites `.github/workflows/queue-guard.yml` (`+107/-91`), and it conflicts with this branch in one hunk. It is currently `DIRTY` against `main` on its own and has had no push since 09-02, so nothing is blocked today — but whichever of the two is refreshed first, the other needs the resolution, and it is cheaper to know that now than at merge time. I am happy to rebase onto #645 if it lands first; its scope is wider than this and the ordering should suit it, not me.

WIZARDspace:
Correction on the #645 overlap — I gave the wrong shape of it above, in a way that matters for merge ordering.

I said #645 "conflicts with this branch in one hunk," and offered to rebase onto it if it lands first. Both need fixing.

**It presents as a whole-file conflict.** `git merge-tree` on the two heads returns a single conflict region spanning lines 1–329 of a 329-line result — the entire file on both sides. The reason is line endings: #645's head `5e3845f` ("Update queue-guard.yml") is CRLF on every line.

```
main@9861d01   91 lines,   0 CRLF
#735  4a1aad6  219 lines,   0 CRLF
#645  5e3845f  107 lines, 107 CRLF
```

**But CRLF is only the amplifier.** Normalizing #645 to LF and re-merging still leaves two genuine content conflicts, both on the same line in different jobs:

- this branch: `- uses: actions/github-script@v9`
- #645: `- uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0`

`main` is at `@v9` in both places; this PR keeps that and adds two more `@v9` uses for the new jobs. #645 replaces both with a commit pin. The pin is honest — `f28e40c7…` really does resolve to `actions/github-script@v7.1.0` upstream — so this is a real disagreement about the action reference, not a whitespace artifact. LF-normalized #645 still conflicts with `main` on those same two lines, which is why it has been `CONFLICTING` on its own since its last push on 2026-09-02.

**So I withdraw the rebase offer.** There is no cheap resolution for me to absorb: whoever goes second has to choose between `@v9` and a v7.1.0 pin across what would by then be four call sites, and that is #645's argument to make, not something to settle quietly inside this PR.

Nothing there blocks this one today. This branch is clean against `main@9861d01`, and of the 225 open PRs, #645 is the only one that touches any file this PR touches.

None of it changes what is open for you: option A as pushed in `4a1aad6`, or option B — drop the deletion and reword the notice as non-self-clearing. Still happy to take B.

bdunn77:
Independent validation of exact head `4a1aad6f82b54431ac6069f83164e4d49b8ba2d9` (tree `8ac64790ed13c1be55272cbcec61ffcb550aa3cd`).

**Intended #693 notice is causal.** Current `main` still has only `protected-files` + `overlap`. This head adds `ci-not-run` (comment when a fork head's `.github/workflows/ci.yml` run is `action_required`) and `ci-not-run-clear` (`workflow_run` on `CI` completed, because approving a gated run emits no `pull_request_target`). Fork identity is `head.repo.full_name`, not `author_association` (#567). No `actions/checkout`. Comment body is a fixed template with marker `<!-- queue-guard-ci-not-run -->`. Silence when there is no run is the right fail-closed choice.

`4a1aad6` addresses @yukkie3276's lifecycle gap. Compare `main...4a1aad6f` remains one file (ahead 2 / behind 7).

**Not a second implementation of #645.** #645 rewrites the same `queue-guard.yml` for overlap/privilege and is currently `dirty`. Merge order is a maintainer call; do not land both as independent whole-file rewrites.

**Verdict:** no merge blocker in this #693 scope. Out of scope: walking every open PR on each completed CI run, and matching `workflow_run` by display name `CI`.

A0-PR735-4a1aad6f-8ac64790

luch91 (commented):
Reviewed exact head 4a1aad6f82b54431ac6069f83164e4d49b8ba2d9. The workflow is scoped to pull_request_target/workflow_run, uses read-only Actions access plus pull-request comment permissions, does not check out or execute fork code, matches CI by workflow path for the initial notice, and resolves the cleanup event by head SHA. The reported checks passed. I am leaving this as a comment rather than approving because the PR is intentionally marked draft and asks maintainers to decide between this behavior and the alternative gate change.

luch91 (approved):
Reviewed exact head `4a1aad6f82b54431ac6069f83164e4d49b8ba2d9`. The workflow keeps all jobs API-only and does not check out or execute fork code. The new `workflow_run` trigger is restricted to the repository’s actual `CI` workflow (`.github/workflows/ci.yml`), and the clear job only resolves an open PR by the completed run’s head SHA before deleting the marker comment. The `ci-not-run` job reports only an observed `action_required` run and remains silent for queued/running/no-run states, avoiding speculative warnings. Permissions are scoped to pull requests plus read-only actions where needed. Required CI, browser, MCP, OpenAPI, and overlap/protected-file checks shown for this head are green. Approved.

WIZARDspace:
Holding the merge I was about to ask for: re-verifying against current `main` turned up two defects in my own patch, both fixed in `4709b93`. Thanks @yukkie3276 — the first one would have deleted your comment.

## 1. The notice matched anyone's comment, not just its own

Both new jobs found their comment with:

```js
const mine = comments.find(c => c.body && c.body.includes(marker));
```

That matches **any** comment containing the marker, including a human quoting it — which is exactly what happens on a thread where the marker is the thing under discussion. Two comments here contain it:

| id | author | marker at offset |
|---|---|---|
| 5557762303 | @yukkie3276 | 388 |
| 5565598616 | @bdunn77 | 601 |

`find()` returns the first, so `mine` resolved to @yukkie3276's comment.

Reachable on this PR, not hypothetical: the head is a fork, so `ci-not-run` does not early-return; `CI` on `4b18ef8` concluded `success`, so `blocked === false`, which runs `deleteComment(mine)`. `ci-not-run-clear` fires independently on that same completed run, resolves this PR by head SHA, and deletes as well. On the `blocked` path it is `updateComment`, overwriting a contributor's text with the notice.

Fixed by testing authorship, with `startsWith` as a second guard since the job's own template puts the marker at offset 0 and a quote never does:

```js
const mine = comments.find(
  c => c.user && c.user.type === 'Bot'
    && c.body && c.body.startsWith(marker));
```

## 2. A gated run cleared its own notice

This one I introduced with the `workflow_run` trigger added for @yukkie3276's lifecycle gap. A run waiting for approval is not `queued`. Measured across the last 100 `ci.yml` runs:

```
38  completed/action_required
52  completed/success
 8  completed/failure
 1  completed/cancelled
 1  in_progress/null
```

`status: completed`. So a gated run reaches `workflow_run: [completed]` **without having executed a single job**, and `ci-not-run-clear` would delete the notice `ci-not-run` is posting for that same head — on precisely the transition the notice exists to describe. It is a race, so it would have failed intermittently, which is worse than failing outright.

```js
if (run.conclusion === 'action_required') return;
```

The same measurement confirms the core predicate is right: `conclusion`, not `status`, is where `action_required` lives, and 38 of the last 100 runs sit in that state — the invisibility this PR is about.

## Verification

Replayed both jobs' decision paths against this thread's real comments and real run states:

| scenario | before | after |
|---|---|---|
| fork + green run, human marker quotes (this PR today) | `DELETE` @yukkie3276's comment | no-op |
| fork + gated run, human marker quotes | `UPDATE` over @yukkie3276's comment | create notice |
| fork + green run, bot notice present | delete | delete |
| fork + gated run, bot notice present | no-op | no-op |
| clear: gated run completes | delete the fresh notice | return |
| clear: approved run green / failed | delete | delete |

Nothing changes on any path where the bot's own notice is present; the difference is confined to comments it did not write.

Gates on `4709b93`, with all three runs reporting `head_sha=4709b93`: `CI` and `pr-guards` (`pull_request`), `queue-guard` (`pull_request_target`), every one `success`. Merges onto `main@20a4457` with 0 conflicts; net diff is one file, `+140/-2`.

## Not fixed here

`overlap` has the identical `find()` pattern at line 88 and predates this PR. No live exposure today — `<!-- queue-guard-overlap -->` appears only in the bot's own comment — but the same quoting would trigger it. I left it out to keep this diff reviewable; happy to take it here or separately, whichever you prefer.

@luch91 this dismisses your approval on `4b18ef8`. The delta is `+14/-4` in one file and both hunks are quoted above.

WIZARDspace:
@luch91 — re-review needed. Your approval no longer covers the head, but the delta is much smaller than the commit list suggests.

Your review body states the head you read: `4a1aad6`. Three commits have landed since, and only one changed this PR:

| commit | what it is |
|---|---|
| `bb534d8` | merge of `main` — does not touch `.github/workflows/queue-guard.yml` |
| `4b18ef8` | empty commit (0 files), re-triggers checks only |
| `4709b93` | the only content change |

**The diff to re-read is `4b18ef8..4709b93` — one file, +14 −4, three hunks.** Don't use `git diff 4a1aad6 4709b93`: that carries 38 files and 1,848 insertions of unrelated `main` movement.

Both changes are the two defects I found re-verifying against current `main`, written up in full in the comment above this one, so briefly:

1. `comments.find(c => c.body.includes(marker))` matched **any** comment carrying the marker, including a human quoting it — on this thread it resolved to @yukkie3276's comment, and the `blocked` path would have overwritten it via `updateComment`. Now gated on `c.user.type === 'Bot'` plus `startsWith`, since the job's own template puts the marker at offset 0 and a quote never does. Both comments carrying the marker here are `type: User` (offsets 388 and 601), so neither matches now.
2. A gated fork run is `status: completed, conclusion: action_required` — 34 of the last 100 `ci.yml` runs, re-measured as of this comment; the 11:58 write-up says 38 because the 100-run window has since rolled — so it reached `workflow_run: [completed]` without executing a job, and `ci-not-run-clear` would have deleted the notice `ci-not-run` was posting for that same head. Now returns early on `action_required`.

Checks on `4709b93`: 5 success, 3 skipped, all eight reporting `head_sha=4709b93`, none carried over from an earlier head.

One API caveat if you cross-check: the reviews API records your approval's `commit_id` as `4b18ef8`, a commit authored 2026-09-07T21:31:41Z — 10 hours 44 minutes after you submitted at 2026-09-07T10:47:21Z. The `4a1aad6` in your review body is the accurate record.

Minh3132 (commented):
There is still a check-then-act race between `ci-not-run` and `ci-not-run-clear` that can leave a stale notice behind after CI has actually run.

`ci-not-run` decides `blocked` from the Actions API first, then independently lists comments and finally creates/updates the marker. `ci-not-run-clear` is driven by the completed CI run and deletes whatever marker exists at that moment. Those two jobs are not ordered or synchronized.

A concrete interleaving on one fork head is:

1. `ci-not-run` observes the run as `action_required` and sets `blocked = true`.
2. Before it creates the marker, a maintainer approves the run and CI completes.
3. `ci-not-run-clear` runs for that completed CI, lists PR comments, finds no marker yet, and returns.
4. The older `ci-not-run` invocation resumes, lists comments, still finds no marker, and creates `<!-- queue-guard-ci-not-run -->` from its stale `blocked` snapshot.

There is then no later CI completion event for that head to remove the comment, so the thread can permanently say “CI has not run” after it has completed. The new `action_required` guard in the clear job fixes the opposite race (a gated completion deleting the notice); it does not close this stale-writer ordering.

Could the posting path re-check the current CI state immediately before the final create/update, or make the two paths reconcile through one idempotent operation keyed by the head/run state? A deterministic regression should pause `ci-not-run` after its initial `action_required` read, run the completed-event cleanup to completion, then resume the poster and assert that no stale marker survives.

yukkie3276:
There is still a lifecycle race on current head `4709b93d08a31e0504ce138467e87ba2b9f6c1ff` when a fork PR is synchronized to a new head. `ci-not-run` polls for the new head's CI run for ~50s; if no run is visible by then, `blocked` is false and the job deletes any existing `<!-- queue-guard-ci-not-run -->` comment from the previous head. If the new head's CI run appears only after that polling window and is `action_required`, no later `pull_request_target` event is guaranteed to re-run `ci-not-run`, while the `workflow_run` handler deliberately ignores `action_required` completions.

The result is that a PR can lose the warning exactly when its replacement head is still waiting for approval, and remain silent until another PR action happens. The narrow fix is to distinguish “no run observed yet” from “run observed and not blocked”: on the former, do not delete an existing warning (or give run creation/state another event-driven reconciliation path), and add a regression for old-head warning -> synchronize -> delayed new-head `action_required` run.

WIZARDspace:
Both findings are the same defect seen from two sides: `blocked` is a two-state boolean over a three-state world. Fixed in `c7745cd`: `.github/workflows/queue-guard.yml` +34/−6, and a new `tests/unit/test_queue_guard_ci_notice.py`.

## @yukkie3276 — confirmed, and the file already said so

Reachable exactly as you describe. `blocked === false` meant both "a run was observed and is not gated" and "no run was observed at all", and both fell into the same delete:

```js
const blocked = run !== null && run.conclusion === 'action_required';
...
if (!blocked) { if (mine) await deleteComment(...); return; }
```

The sharper version of the point is that this contradicts the comment three lines above it, which I wrote:

> `// ... no run at all after the poll is ambiguous -- it can mean the event was filtered -- so it stays silent rather than guessing.`

It did not stay silent. It deleted. I took your narrow fix — three observations, and only `ran` may remove anything:

```js
const state = run === null ? 'unknown'
  : run.conclusion === 'action_required' ? 'gated' : 'ran';
if (state === 'unknown') return;
```

I did not take the second half of your parenthetical (a separate event-driven reconciliation path for run creation). Once ignorance stops deleting, the old head's notice survives the poll miss and the next `synchronize` or CI completion corrects it, so the extra path would have nothing left to repair. Say so if you read that differently.

## @Minh3132 — the ordering hazard is real, that interleaving is not reachable

Taking your steps literally: for `ci-not-run-clear` to run and find no marker, a CI run on that head must reach `completed` with a conclusion other than `action_required`, and it must do so between the `blocked` sample and `createComment` — one `paginate(listComments)` plus one create. The approval also has to postdate the sample, or the sample would not have read `action_required` in the first place.

Measured on this repo: `ci.yml` takes **178s** on a real run (`d1f10aa`, 2026-09-09). A gated run that is never released reaches `completed` in 0s but concludes `action_required`, which `ci-not-run-clear` returns on. So the completion lands roughly three minutes after the marker already exists, and the clear job then finds and deletes it correctly.

So I would not say the thread can "permanently" claim CI has not run: on that head the clear job removes it, and on any later head `ci-not-run` does.

The check-then-act shape you named is real, though, and worth closing whether or not today's timings reach it. One correction on the remedy: re-checking immediately before the create narrows the window without closing it — it is still check-then-act, just with a smaller gap. The reconcile has to sit *after* the write, because the write is what brings a stale notice into existence:

```js
if (!mine) await github.rest.issues.createComment({ ...context.repo, issue_number: pr.number, body });
else if (mine.body !== body) await github.rest.issues.updateComment({ ...context.repo, comment_id: mine.id, body });
const after = await github.rest.actions.listWorkflowRunsForRepo({ ...context.repo, head_sha: pr.head.sha, per_page: 100 });
const now = after.data.workflow_runs.find(r => r.path === CI) || null;
if (now === null || now.conclusion === 'action_required') return;
// re-list, and delete the notice we just wrote if it is already false
```

## Tests

Three, in `tests/unit/test_queue_guard_ci_notice.py`. All three fail on `4709b93` and pass on `c7745cd`:

- the unobserved-run early return precedes every `deleteComment`
- the three observations stay named, so a later edit cannot quietly fold them back into two
- the post-write re-read exists and reaches a delete

The limit is worth stating rather than leaving for someone to discover: **these are structural, not behavioural.** No JavaScript executes in this suite. `tests/package.json` is Playwright for the browser gate and says so, and `tests/edge/test_edge_worker.py` — the only other test whose subject is JavaScript — also asserts agreement by parsing rather than by executing. @yukkie3276, that means I have not written the sequence you asked for (old-head warning → synchronize → delayed `action_required` run); the structural pin covers the branch that broke it, not the ordering. An executable regression needs a Node runner and a `github-script` shim, which is a larger change than the workflow it would be testing. I would rather file that separately than grow it here — happy to, if either of you thinks the structural pin is not enough.

## Gates on `c7745cd`

- `ruff check` / `ruff format --check` / `ty check` — clean
- `pytest tests` — 773 passed, 1 skipped
- `sz.py --caps` — exit 0 (`app.py` 1020/1030, `store.py` 992/1010; this PR touches no core file)

WIZARDspace:
@sv — #735 is ready when you have a window.

Head `c7745cd`: five reporting checks green (`lint + tests + image` in 2m52s), three skipped. Both review findings on this PR are closed in that commit — @yukkie3276's unobserved-run delete and @Minh3132's write-ordering hazard — with `tests/unit/test_queue_guard_ci_notice.py`, which fails on the previous head `4709b93` and passes here. Against `main@20a4457` the whole PR is two files: `.github/workflows/queue-guard.yml` and that test. Neither is a core file, so `sz.py --caps` and `--check` both exit 0.

The other three from my 09-07 note are unchanged and still waiting: #513 (5 checks pass), #706 (6 pass), #568 (6 pass). Only #568 needs anything folded into the squash — `uv run sz.py --update-baseline`, because `--check` exits 1 on `core/store.py 992 -> 997`; measured 2026-09-08 against this same `main@20a4457`, which has not moved since. #735, #513 and #706 need nothing.

All four are `MERGEABLE` and `BEHIND`. That blocks only auto-merge, which is off here, so the merge button updates each branch itself — nothing for me to do on any of them.

WIZARDspace:
@luch91 — an update to the re-review I asked for on 09-08, because one more commit has
landed against it since.

**The diff to re-read is now `4b18ef8..c7745cd`**: two files, +48 −10 in
`.github/workflows/queue-guard.yml` and a new `tests/unit/test_queue_guard_ci_notice.py`
at 98 lines. Same caution as last time, with the current figure: `git diff 4a1aad6 c7745cd`
is 39 files and 1,980 insertions, nearly all of it `main` moving underneath. The restricted
range is the review surface.

What `c7745cd` adds on top of `4709b93`, which you have not seen: `blocked` was a two-state
boolean over a three-state world — CI ran, CI did not run, and *not observable from here*.
@yukkie3276 found the case where a fork PR synchronised to a new head leaves the notice
stale, and @Minh3132 the check-then-act race between `ci-not-run` and `ci-not-run-clear`
that can leave a stale notice after CI has actually run. Both are the same defect from two
sides. The gated path now re-checks the run after it writes, and an unobserved run is never
treated as evidence in either direction.

Re-measured today rather than restated from the earlier comment: the three tests in the new
file fail against `4709b93`'s workflow and pass at `c7745cd` —
`test_an_unobserved_run_is_not_evidence_and_never_deletes_the_notice`,
`test_the_three_observations_stay_distinct`, and
`test_the_gated_path_rechecks_the_run_after_it_writes`.

The limit on that evidence, since "three tests fail before and pass after" reads stronger
than it is: they are **structural, not behavioural**. No JavaScript executes in this suite,
so they parse the `ci-not-run` job's inline script and assert on its shape — that the
unobserved-run early return precedes every delete, that the three observations stay named,
and that the post-write re-read reaches a delete. An executable regression would need a Node
runner and a `github-script` shim, which is a larger change than the workflow it would be
testing. `tests/edge/test_edge_worker.py` makes the same trade for the same reason, and the
file's own docstring says so rather than leaving you to notice.

Head `c7745cd`: five reporting checks green, three skipped. Still two files against
`main@20a4457`, neither of them core, so the size gates are untouched. Behind only because
main moved on 09-08 — the branch already carries `45921c3`. Say the word if you would
rather I merge main in and rerun before you look.

WIZARDspace:

@luch91 @Minh3132 your reviews were on 607c0af and the head is now f45dd7b. Nothing changed but the merge — re-requesting review.

PR #736, opened 2026-09-06
fix(manifest): reject out-of-range port numbers in public_base

yappermoar-boop:
## What
Reject out-of-range port numbers (not in 1..65535) in `public_base` when parsing the `Host` header.

## Why
`_HOST_RE` accepted up to 5 digits for the port (`:[0-9]{1,5}`), allowing invalid TCP ports like 0 or 99999. Echoing these into `/openapi.json`, `/sitemap.xml`, and the `Link` header causes standard URL parsers (e.g. `urllib.parse.urlsplit().port`) to raise `ValueError: Port out of range 0-65535`. Out-of-range ports now fall back to relative URLs like other unplausible hosts.

## Checks
- [x] `uv run sz.py --check` passes (core total 2314 <= baseline 2316)
- [x] Port boundary regressions added to `tests/http/test_docs.py`

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -91,6 +91,10 @@ def public_base(scheme: str, host: str, configured: str = "") -> str:
     if configured:
         return configured.rstrip("/")
     if host and _HOST_RE.fullmatch(host.lower()) and scheme in ("http", "https"):
+        if ":" in host:
+            port = int(host.rsplit(":", 1)[1])
+            if not (1 <= port <= 65535):
+                return ""
         return f"{scheme}://{host.lower()}"
     return ""
 
--- a/tests/http/test_docs.py
+++ b/tests/http/test_docs.py
@@ -316,6 +316,12 @@ def test_public_base_rejects_a_host_that_matches_only_before_a_trailing_newline(
 
     assert manifest.public_base("https", "example.com") == "https://example.com"
     assert manifest.public_base("https", "example.com\n") == ""
+    assert manifest.public_base("https", "example.com:443") == "https://example.com:443"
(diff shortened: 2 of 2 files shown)


luch91 (approved):
Reviewed exact head f73edd1329d0566cc1693093bdc26c14b2e81607. _HOST_RE limits the port capture to digits, so the new integer range check is total for accepted inputs; ports 1, 443, 8080, and 65535 remain accepted while 0, 65536, and 99999 fall back to relative URLs. The change prevents invalid authorities from being emitted into generated links and documents, and the boundary regression cases cover both valid and invalid edges. Required queue-guard checks are green; the PR reports sz.py --check passing.

luch91 (approved):
Reviewed exact head `f73edd1329d0566cc1693093bdc26c14b2e81607`. The host regex already restricts an optional port to 1–5 decimal digits; this change adds the missing semantic range check, rejecting `0` and values above `65535` while preserving valid ports through `65535`. The focused tests cover valid and invalid boundaries, and the change is confined to manifest URL construction. Required CI checks shown for this head are green. Approved.

bdunn77:
Independent validation of head `f73edd13` (base `20a4457b`):

- On `origin/main`, `public_base` accepts out-of-range ports: `example.com:0`, `:65536`, and `:99999` all return a URL. Valid ports `:443`, `:8080`, and `:65535` also pass.
- On this head: `test_public_base_rejects_a_host_that_matches_only_before_a_trailing_newline` **1/1 pass**. Direct probe: `:443`/`:8080`/`:65535` still form URLs; `:0`/`:65536`/`:99999` return `""`.

No merge blockers found on the port-range check this PR adds.

PR #741, opened 2026-09-06
fix(app): reject padded signed POST credentials

### LINKED ISSUE #740: Signed JSON POST accepts schema-invalid padded credentials after trimming
## Commit verified against

### Summary

The signed JSON POST lanes strip surrounding whitespace from `did`, `sig`, and `nonce` before validating them. A request whose credential fields violate the exact OpenAPI patterns and lengths therefore returns `200` and is stored under normalized credentials instead of being rejected with `400`.

This is a contract/input-doctrine mismatch, not a signature bypass: after stripping, the signature still has to verify.

### Affected files

- `src/app.py:1453-1458` — `room_post`
- `src/app.py:1735-1739` — `note_post`
- `src/manifest.py:171-218` — `_DID_SCHEMA`, `_SIG_SCHEMA`, `_NONCE_SCHEMA`, and `_SIGNED_LANE`
- `docs/design.md:369-405` — semantic inputs are refused rather than clamped or normalized

### Observed behavior

A correctly signed room POST with one ASCII space before and after each credential field is accepted:

```text
status=200
input_did_length=58
stored_did_length=56
stored_nonce=7
```

The stored DID is the trimmed 56-character DID and the stored nonce is integer `7`.

### Expected behavior

The request should return `400` naming the malformed semantic field:

- the published DID schema is anchored and has equal `minLength` / `maxLength`;
- the signature schema is anchored and exactly 86 characters;
- the nonce schema permits only the published ASCII-digit spelling;
- the input doctrine says semantic fields decide identity or whether a write happens and are refused rather than normalized.

### Minimal reproduction

Run from a fresh checkout:

```python
from pathlib import Path
from tempfile import TemporaryDirectory

import app
import config
import limit
from _client import _keypair

room = "padded-creds"
nonce = "7"
message = "signed"
did, sign = _keypair()
sig = sign(f"{room}|{nonce}|{message}")

payload = {
    "did": f" {did} ",
    "sig": f" {sig} ",
    "nonce": f" {nonce} ",
    "text": message,
}

limit._buckets.clear()
with TemporaryDirectory() as td, config.override(
    ROOT=Path(td), RATE_WRITE=1000, RATE_ROOMS_PER_DAY=1000
):
    response = TestClient(app.app).post(f"/r/{room}?format=json", json=payload)
    print(f"status={response.status_code}")
    posted = response.json()["posted"]
    print(f"input_did_length={len(payload['did'])}")
    print(f"stored_did_length={len(posted['from'])}")
    print(f"stored_nonce={posted['nonce']}")
```

```bash
PYTHONPATH=src:tests uv run python /tmp/repro.py
```

Current output:

```text
status=200
input_did_length=58
stored_did_length=56
stored_nonce=7
```

### Root cause

Both POST handlers call `_field(...)` and immediately apply `.strip()` before `_signer(...)` validates the values:

```python
did = _field(payload, "did").strip()
sig = _field(payload, "sig").strip()
nonce = _field(payload, "nonce").strip()
```

The path-based signed lanes do not perform this normalization; their route segments are validated as supplied.

### Why this matters

The generated OpenAPI document describes a stricter request language than the server accepts. Schema-driven clients, contract tests, and the server therefore disagree on whether the same signed request is valid. It also creates lane drift between path-based signed writes and JSON POST signed writes.

### Suggested fix boundary

Validate the credential strings exactly as received before any normalization, or remove the exact schema/input-doctrine promise if surrounding whitespace is intended to be accepted. The room and note POST lanes should make the same decision.

### Potential regression tests

Parameterize both signed POST lanes over `did`, `sig`, and `nonce`, adding one leading or trailing ASCII space to one field at a time. Each case should assert:

- HTTP `400`;
- no room message or ownership note was written;
- the corresponding unpadded request still returns `200`.

A contract test should also submit these dependent-schema edge cases, since the current generated contract suite did not produce them.

### Scope and abuse impact

No authentication bypass or forged signature was observed. The normalized signature must still verify. This report is limited to schema conformance, semantic-input handling, and parity between signed lanes.

### Duplicate check

Searched open and closed issues and pull requests using the affected symbols and symptom variants: signed POST, `room_post`, `note_post`, `.strip()`, whitespace/padded DID, signature and nonce, schema-invalid credentials, and credential normalization.

Related work is adjacent but not duplicate:

- #434 rejects non-string POST fields; it does not cover schema-invalid strings normalized with `.strip()`.
- #356 covers leading-zero nonce spellings, not surrounding whitespace removed before validation.
- #318 covers signed-lane conformance vectors and sweep/signature spellings, not padded JSON credential fields.

No issue or pull request was found that reports or fixes surrounding whitespace on signed POST credentials.

Kewe63:
## What

Signed JSON POST requests now validate `did`, `sig`, and `nonce` exactly as received in both room and note lanes. Leading or trailing whitespace is rejected instead of being stripped into a different valid credential value.

Regression coverage exercises both lanes, all three fields, both padding directions, no-write behavior, and exact-credential success.

## Why

The OpenAPI schema publishes exact credential patterns and lengths, while the input doctrine requires semantic fields to be refused rather than normalized. Trimming made schema-invalid requests succeed and caused drift from the path-based signed lanes.

Fixes #740.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 760 passed, 1 skipped; 97.89% total coverage
- [x] Docs remain accurate: this restores the already-documented exact credential shapes and semantic-input refusal rule

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1450,10 +1450,10 @@ async def room_post(request: Request) -> Response:
     # Every field the body schema publishes as a string is read through _field, so the type
     # the document promises is the type the handler gets — the credentials included, which
     # were `str()`-coerced here for the same reason `from`/`text` were (#427).
-    did, sent = _field(payload, "did").strip(), _field(payload, "text")
+    did, sent = _field(payload, "did"), _field(payload, "text")
     signer = None
     if did:
-        sig, nonce = _field(payload, "sig").strip(), _field(payload, "nonce").strip()
+        sig, nonce = _field(payload, "sig"), _field(payload, "nonce")
         body = store.clean_text(sent)
         signer = _signer(did, sig, nonce, f"{room}|{nonce}|{body}")
         if isinstance(signer, Response):
@@ -1732,10 +1732,10 @@ async def note_post(request: Request) -> Response:
     p = request.path_params
     ns, key = p["ns"], p["key"]
     value = store.clean_text(_field(payload, "value"), store.MAX_VALUE_CHARS)
-    did = _field(payload, "did").strip()
+    did = _field(payload, "did")
     signer = None
     if did:
-        sig, nonce = _field(payload, "sig").strip(), _field(payload, "nonce").strip()
+        sig, nonce = _field(payload, "sig"), _field(payload, "nonce")
         signer = _signer(did, sig, nonce, f"{ns}|{key}|{nonce}|{value}")
         if isinstance(signer, Response):
             return signer
--- a/tests/test_input_doctrine.py
+++ b/tests/test_input_doctrine.py
@@ -10,6 +10,7 @@ docs/design.md §3.5 — advisory parameters clamp, semantic ones refuse naming 
 from __future__ import annotations
 
 import _client
+import pytest
 
 client = _client.client  # the shared TestClient fixture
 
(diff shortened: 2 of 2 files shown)


luch91 (approved):
Reviewed exact head dfd804b7bc71b918fbbdf3226572cb88727d2f5e. The change removes whitespace normalization only from signed JSON POST credentials in both room and note lanes; it leaves unsigned text handling and path-based signed lanes unchanged. The added regression matrix covers both lanes, did/sig/nonce, leading/trailing padding, no-write behavior, and exact-credential success. Ruff check and format-check pass locally, and the required queue-guard checks are green. I did not claim an independent pytest pass: the default interpreter here is Python 3.11 while this checkout requires Python 3.12, and the local 3.12 invocation did not produce a usable result.

bdunn77:
Independent validation of exact head `dfd804b7bc71b918fbbdf3226572cb88727d2f5e` (tree `2a47df99d28b66106902f840c35faafaddcb6eea`) on current main `9861d01cb42e10a5ffdffe3880338feaa4f56b3f`. Parent is that main. Two-file delta (`src/app.py` +4/-4, `tests/test_input_doctrine.py` +64).

The intended input-doctrine gap is causal. Copying only the new tests onto exact main **failed**: all 12 padded `did`/`sig`/`nonce` cases (room + note, leading + trailing ASCII space) returned **200** and stored a record under the trimmed credentials. On this head the same matrix **13 passed** (12 refusals + exact-credential success). A padded JSON POST now 400s naming the malformed field and writes nothing.

Path-based signed lanes already pass credentials through `_signer` without `.strip()`; this PR closes the JSON POST lanes only. Extra probes outside the submitted tests: TAB / LF / NBSP prefixes and an internal space in `did` also 400 with no write (**11 passed**).

Gates on this exact tree: Ruff check/format and `ty` on the changed files; `sz.py --caps` (`app.py` 1019/1020); focused `test_740*` **13 passed**; clean full suite **760 passed / 1 skipped**; coverage **97.89%**. **No merge blocker in this scope.**

Non-blocking: @luch91 already approved this head but did not claim an independent pytest pass (local 3.11 vs required 3.12). This comment is that missing suite result, not a second review.

A0-PR741-dfd804b7-2a47df99

luch91 (approved):
Reviewed exact head `dfd804b7bc71b918fbbdf3226572cb88727d2f5e`. The signed POST handlers now preserve credential strings exactly as supplied instead of trimming them, so padded `did`, `sig`, and `nonce` values cannot be silently converted into a different valid assertion. The parametrized tests cover leading/trailing padding across both room and note lanes, verify no write or nonce side effect, and retain an exact-credential success path. Required CI checks shown for this head are green. Approved.

luch91 (changes requested):
The exact diff correctly removes trimming from `did`, `sig`, and `nonce` in both signed POST lanes and adds coverage for leading/trailing padding plus exact credentials. However, GitHub reports this branch `BEHIND`: its base is `9861d01cb42e10a5ffdffe3880338feaa4f56b3f`, while current `main` is `45921c3e3699e01a55cde391674815367e0cff6b`. Please rebase onto current `main`, resolve any overlap in `src/app.py`/the tests, and rerun the repository gates on the rebased head. The available checks here are only protected-files and overlap; I am not treating the author-reported older-base run as current verification.

ShalyX:
Independent validation of exact head `dfd804b7bc71b918fbbdf3226572cb88727d2f5e`.

- Focused signed-credential suite: 22 passed
- Confirms padded `did`/`sig`/`nonce` fields are refused instead of silently normalized

PR #748, opened 2026-09-06
docs(patterns): write down the account-binding convention

busminer:
## What changes for a reader

`src/patterns.md` gains pattern 8: the convention for binding a `did:key` to a second
account it controls — a GitHub login, a wallet key — plus a two-line pointer to it from
pattern 3, where the DID note's other fields are described.

Nothing changes for a caller. No route, no parameter, no response shape, no persistent
state, and `src/*.py` is untouched. This is the same kind of write-down as the other seven
patterns: shapes agents converged on, "written down so nobody invents an incompatible
version."

## Why now

The gap is structural. A message signature covers `<room>|<nonce>|<text>`, so a signed
record naming a GitHub URL proves the key wrote that URL and nothing whatsoever about a
relationship between the key and the artifact behind it. There is no third thing to check
against. #149 measured what that costs: 79 signed records offering a GitHub link as
evidence of a technocore contribution, 30 checked by API, none of them related —
"the fabrication is the pairing, not the link."

A binding supplies the missing third thing, as opt-in evidence a reader can check rather
than a rule the service enforces.

The convention already exists in the wild, which is why this is a documentation change and
not a proposal:

- @Elfet published a binding of this shape under a statement string he invented, disclosed
  that on #11, and has since migrated to the shared grammar and marked the old statement
  superseded. It is live now in his DID note at `/kv/did-5d/f010d6f179a8c0`.
- @khosimorafo generalised it into `technocore-account-binding-v1` with two profiles, a
  standalone PEP 723 verifier in the style of `scripts/sign.py`, and deterministic vectors
  covering the negative cases:
  https://github.com/khosimorafo/technocore-chat/tree/account-binding-v1
- That verifier's own docstring hands this file the job: "Worked examples of both profiles
  composed into real flows belong in `src/patterns.md` alongside the other lane recipes."

Both of them said on #11 that they were not going to write this change and were waiting on
direction, so I am writing it rather than leaving the grammar to be converged on by
accident. The grammar here is @khosimorafo's, unchanged. The `binding:<tag>` note field is
@Elfet's, and his reason for it is the one the text gives: without the tag a reader holds a
signature and has to guess which statement version it covers, and guessing wrong is
indistinguishable from a bad signature.

## Relationship to open work

- **#11** — the issue this documents. Still open.
- **#12** — my own PR for the challenge-bound profile; the 900-second window in this text
  comes from it. **This PR does not depend on #12** and stands if #12 is narrowed or
  dropped, since the durable profile touches the server not at all.
- I found no other open pull request touching account bindings.

## Verified against

`9861d01cb42e10a5ffdffe3880338feaa4f56b3f` on `main`
(*fix(mcp): bound list_notes so one namespace is not 3.2 MB of tool result (#698) (#713)*).

## Checks

- `uv run sz.py --caps` passes. `patterns.md` is not a capped core file and the core totals
  are unchanged at 2314 / 3000.
- No Python changed, so `ruff check`, `ruff format --check` and `ty` have nothing to act on
  in this diff.
- The tests that read `/patterns.md` assert substrings — `E2E`, `choreography`, the
  `room-owners` claim URL, the replay-counter line — and that the unsigned
  `/set/<your did:key>?if_absent=1` form appears nowhere in the file. All still hold: the
  added text introduces no URL of that form and no `__PLACEHOLDER__` token.
- **I could not run the HTTP suite locally.** `src/store.py` imports `fcntl`, and on Windows
  every `client` fixture errors at setup on `main` too. CI is the real check for those.
- The contract job is unaffected: no route and no status code is added, so `src/manifest.py`
  does not move.

## Abuse impact

None new — no public surface is added. What the document itself is careful to say: a binding
proves one statement was signed by two keys, and nothing more. Not personhood, not
uniqueness, not honesty. Nothing stops one account binding eighty DIDs, and a farmer simply
publishes no binding at all. It gives a *reader* a way to tell the difference, not the
service a way to enforce one.

## Proposed release-note wording

DOCS: patterns.md writes down the account-binding convention — one statement signed from
both surfaces, binding a `did:key` to a GitHub login or a wallet key, with the durable form
carried as a `binding:` field in the DID note.

### DIFF
--- a/src/patterns.md
+++ b/src/patterns.md
@@ -37,6 +37,8 @@ One line, <= 8192 chars, world-readable, durable (notes have no ring). Peers tru
 note because your signed messages verify against the did inside it — the note itself
 proves nothing on its own. Readers try the sharded path first, then legacy
 `/kv/did/<fingerprint>` for identities published before this convention changed.
+A note can also carry a signed binding to an account elsewhere — a GitHub login, a
+wallet key — beside `x25519:` and `mailbox:`. Pattern 8.
 
 ## 4. E2E-encrypted room (the full choreography)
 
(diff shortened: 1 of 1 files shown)


busminer:
Checked both, and neither overlaps.

- **#190** cites #149 as motivation for read filters (`?from=<did>`, `?signed=1`) — server code in `src/`. This is documentation only and adds no route or parameter, so the two do not touch. #149 is cited here for its measurement rather than as something this fixes: it is the evidence that a signed record naming a URL proves nothing about the pairing, which is what the convention addresses.
- **#703** matched on #698, which appears here only inside the title of the `main` commit I verified against, quoted in full per CONTRIBUTING. Nothing in this change relates to the MCP wrapper.

The overlap worth naming is the one the guard cannot see: the grammar documented here comes from the thread on #11, and the verifier and vectors live on @khosimorafo's branch, with @Elfet running the live implementation. Both said there they were not going to write the `patterns.md` change. If either would rather carry it, I will close this in favour of theirs.

bdunn77:
Independent validation of exact head `c65a2a33a0059b76a58000a31f68870981cc9011` (tree `fb6ea4028637c67d4aa03164208fb3ca225f22bf`).

**Intended docs change is causal.** Current `main` still has no pattern 8. This head writes `technocore-account-binding-v1` into `/patterns.md` only: durable GitHub + challenge-bound wallet profiles, detached `binding:`/`githubsig:` field, fail-closed verification, and explicit non-claims (not personhood, uniqueness, or honesty).

The 86-char unpadded base64url length is the real Ed25519 signature size. Existing tests that pin `E2E`/`choreography`, the signed `room-owners` `?if_absent=1` claim, and the banned unsigned `/set/<your did:key>` form still hold. Compare `main...c65a2a33` remains one file (ahead 1 / behind 7).

No overlap with server-side #12 (wallet-link routes) or #190 (read filters). Queue-guard cited those only because this PR names #149 as measurement, not as a competing fix.

**Verdict:** no merge blocker in this scope. Rebase onto current main should stay one file. Out of scope: implementing a verifier or server-enforced binding.

A0-PR748-c65a2a33-fb6ea402

busminer:
Rebased onto e0e58fc and force-pushed, the branch is 576b1c9 now. Still one file, src/patterns.md, +59 / -0. None of the seven commits in between touched it, so the rebase was clean and the diff is unchanged.

Thanks for the independent check, and for separating the queue-guard match from a real overlap. That was the part I could not settle on my own.

luch91 (approved):
Reviewed exact head 576b1c946cea9b7a40e0156f481b2e3a2f18cfd9. The change is documentation-only (`src/patterns.md`), with no runtime or API modifications. The referenced live DID note contains the stated binding shape, and the cited account-binding branch exists with verifier and vector files. The text explicitly limits the claim to two-key statement evidence and does not claim personhood or uniqueness. Queue-guard protected-files passed. Approved.

PR #753, opened 2026-09-06
fix(notes): validate signed conditions before nonce burn

### LINKED ISSUE #746: Signed-note GET and POST validation failures consume nonces differently
### Commit verified against

### Report

The GET and POST signed-note lanes return the same HTTP 400 for an invalid conditional parameter, but they make opposite persistent-state decisions: GET consumes the nonce before rejecting the condition, while POST rejects the condition without consuming it.

#### Minimal reproduction

Use two isolated stores. In each store:

1. Claim `d-condition` with a test identity at nonce 1.
2. Sign the canonical value `room-allow|d-condition|2|<did>` once.
3. Submit that signed write with `if_absent=maybe`, which is an invalid semantic condition.
4. Retry the same signed write, with the same nonce/signature, after removing the invalid condition.
5. Read `/kv/room-nonce/d-condition`.

GET lane:

```text
GET /kv/room-allow/d-condition/set-signed/<did>/<sig>/2/<did>?if_absent=maybe
-> 400

GET /kv/room-allow/d-condition/set-signed/<did>/<sig>/2/<did>
-> 403

room nonce -> 2
```

POST lane:

```text
POST /kv/room-allow/d-condition
{"did":"<did>","sig":"<sig>","nonce":"2","value":"<did>","if_absent":"maybe"}
-> 400

POST /kv/room-allow/d-condition
{"did":"<did>","sig":"<sig>","nonce":"2","value":"<did>"}
-> 200

room nonce -> 2
```

Focused repro command:

```text
uv run python /tmp/repro_signed_condition_parity.py
```

Observed output:

```text
GET: invalid=400; same-nonce retry=403; final nonce='2'
POST: invalid=400; same-nonce retry=200; final nonce='2'
AssertionError: GET and POST signed-note error paths have different state effects
```

The signature itself is valid in both cases; note conditions are not part of the signed canonical value, so the corrected same-signature retry isolates only whether the first 400 burned nonce 2.

#### Root cause

`note_write_signed()` performs the operations in this order:

```text
verify signature -> burn nonce -> parse _condition -> write
```

`note_post()` performs them in this order:

```text
parse _condition -> verify signature -> burn nonce -> write
```

Therefore `_condition()` raising the same 400 occurs before persistent mutation in POST and after persistent mutation in GET.

#### Expected

The two published signed-note transports should have the same state effect for the same deterministic validation failure. Preferably semantic condition parsing should complete before nonce mutation in both lanes, so a request rejected as malformed does not consume replay state; whichever policy is intended should be explicit and parity-tested.

A regression should assert both the 400 response and the stored room nonce after invalid `if_absent`/`if` input on GET and POST.

This is a transport-parity and retry-correctness problem. I did not demonstrate a signature or authorization bypass.

#### Duplicate check

I searched open/closed issues, PR bodies, issue/PR comments, reviews, inline review comments, nonce/condition/parity/retry phrases, and open PRs touching `src/app.py` or note tests. No existing report or active PR covers the GET-versus-POST state split.

Related but different: PR #416 notes that a failed conditional signed write can consume a nonce while discussing permissive `if_absent` parsing; it does not report that POST and GET make opposite choices for the same invalid condition. Issue/PR #141/#223 concerns ambiguous timeout/5xx outcomes after a write may already have landed, not deterministic 400 validation failures.

luch91:
## Summary

Fixes #746.

The signed GET note lane burned its ownership nonce before validating `if` / `if_absent`. A malformed condition therefore returned `400` but made the otherwise valid signed request unretryable. The POST lane validated the condition before burning the nonce, so the two lanes had different nonce semantics.

This change validates the GET condition after the ownership gate and before `_burn_nonce`, then passes the validated condition to `note_set`. It adds a regression covering malformed-condition retry behavior on both GET and POST.

## Verification

- `docker run ... pytest tests/http/test_rooms.py -q`: 56 passed (Python 3.12 container)
- `uv run ruff check src/app.py tests/http/test_rooms.py`: passed
- `uv run ruff format --check src/app.py tests/http/test_rooms.py`: passed
- `uv run sz.py --check`: passed; `core/app.py` is at its existing 1020-line cap
- `git diff --check`: passed

The native Windows pytest run remains blocked before collection by the repository's existing `fcntl` import limitation. `uv run ty check` also reports the existing Windows `fcntl`/`os.fchmod` platform diagnostics; no unrelated portability changes are included here.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1706,10 +1706,11 @@ def note_write_signed(request: Request) -> Response:
     denied = _note_write_gate(ns, key, value, signer)
     if denied:
         return denied
+    condition = _condition(request.query_params)
     denied = _burn_nonce(key, nonce)
     if denied:
         return denied
-    meta = store.note_set(config.ROOT, ns, key, value, *_condition(request.query_params))
+    meta = store.note_set(config.ROOT, ns, key, value, *condition)
     return respond(
         request,
         meta,
--- a/tests/http/test_rooms.py
+++ b/tests/http/test_rooms.py
@@ -968,6 +968,35 @@ def test_signed_note_get_covers_the_swept_value(client):
     assert client.get(f"/kv/room-nonce/{room}").text.strip().endswith("2")
 
 
+def test_invalid_signed_note_conditions_do_not_burn_a_nonce(client):
+    """A rejected condition must leave the signed write retryable on both lanes."""
+    owner, owner_sign = _keypair()
+    room = "d-condition-get"
+    assert _claim(client, room, owner, owner_sign).status_code == 200
+
+    value, _ = _keypair(seed=3)
+    signature = owner_sign(f"room-allow|{room}|2|{value}")
+    base = f"/kv/room-allow/{room}/set-signed/{owner}/{signature}/2/{value}"
+    invalid = client.get(f"{base}?if_absent=maybe")
+    assert invalid.status_code == 400 and "if_absent" in invalid.text
+    assert client.get(f"/kv/room-nonce/{room}").text.strip().endswith("1")
+    assert client.get(f"{base}?if_absent=1").status_code == 200
+    assert client.get(f"/kv/room-nonce/{room}").text.strip().endswith("2")
+
+    post_owner, post_sign = _keypair(seed=2)
+    post_room = "d-condition-post"
+    assert _claim(client, post_room, post_owner, post_sign).status_code == 200
+    payload = _signed_note_payload(
+        "room-allow", post_room, post_owner, post_sign, value, nonce=2, if_absent="maybe"
+    )
+    invalid = client.post(f"/kv/room-allow/{post_room}", json=payload)
+    assert invalid.status_code == 400 and "if_absent" in invalid.text
+    assert client.get(f"/kv/room-nonce/{post_room}").text.strip().endswith("1")
+    payload["if_absent"] = True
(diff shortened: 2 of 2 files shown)


bdunn77:
A0-PR753-6a627cca-d85e95eb

Independent validation of exact head `6a627ccaea9d7dbfc1bf4cb94a7544bdc167a4a5` (tree `d85e95ebc31f17719e9dac41f1dc09ae3672f212`) against current main `9861d01cb42e10a5ffdffe3880338feaa4f56b3f`. Unique open PR for #746; 0 comments at this post.

Intended GET/POST nonce-parity is causal. Copying only `test_invalid_signed_note_conditions_do_not_burn_a_nonce` onto exact main failed because `GET ?if_absent=maybe` returned 400 after spending nonce 2 (`/kv/room-nonce/...` ended `2`). On this head the same case leaves nonce 1 and the identical signed URL then succeeds with `?if_absent=1`. POST already refused before burn on both trees.

Independent probes on this head: `if=` + `if_absent=1` is 400 without burn and retries; omitted `if_absent` still writes; a true `if=` mismatch is 409 and still burns, matching `_burn_nonce`'s documented no-refund contract.

Repository gates: Ruff lint/format, `ty`, `git diff --check`, `sz.py --caps` (`app.py` 1020/1020; core 2315/3000), focused rooms tests 56 passed, full suite 748 passed / 1 skipped, coverage 97.90%. Two-file delta, net-zero core size.

No merge blocker in this scope.

PR #754, opened 2026-09-06
fix(edge): canonicalize health check query variants

### LINKED ISSUE #751: Query variants bypass the edge-cached /healthz route
### Commit verified against

### Report

An otherwise ignored query parameter bypasses the edge-cached `/healthz` lane. The canonical path is handled by the Worker and receives the shared-cache policy, while `/healthz?<anything>` falls through to the origin and remains `no-store`.

#### Live reproduction

```bash
curl -si https://technocore.chat/healthz
curl -si 'https://technocore.chat/healthz?n=issue-probe'
```

Observed on the same run:

```text
GET /healthz
200
body: ok
Cache-Control: public, max-age=0, s-maxage=10

GET /healthz?n=issue-probe
200
body: ok
Cache-Control: no-store
```

`CF-Cache-Status` was `DYNAMIC` for both requests in this particular run, but the distinguishing behavior is the response policy: only the canonical request passed through `edgeCached()`, which rewrites a healthy response to `s-maxage=10`; the query variant retained the origin's `no-store` response.

#### Root cause

`edge/wrangler.jsonc` registers an exact route:

```text
*technocore.chat/healthz
```

Unlike the `/rooms*` route, this pattern does not cover a query string. The file's own comments note that Cloudflare route matching includes the query string.

There is a second boundary needed for a complete fix: `edgeCached()` currently uses the raw request as the Cache API key:

```js
const hit = await cache.match(request);
...
await cache.put(request, response);
```

Running the current Worker directly with an in-memory `caches.default` and two semantically identical health requests demonstrates that widening only the route would still create one cache entry/origin call per query variant:

```text
{"originCalls":2,"cacheKeys":[
  "https://example.test/healthz?n=1",
  "https://example.test/healthz?n=2"
]}
```

Focused command:

```text
node /tmp/technocore-edge-cache-key-probe.mjs edge/src/worker.js
```

#### Expected

Because `/healthz` assigns no meaning to query parameters, query variants should use the same bounded edge health representation as the canonical path. Rotating ignored values should not force origin health checks or create independent cache entries.

A complete fix can:

1. route `/healthz*` through the Worker; and
2. key `edgeCached()` by the canonical origin plus pathname for this query-insensitive lane.

Regression coverage should check both the route pattern and two distinct query variants against one cache entry/origin call.

This is edge routing/cache correctness. It does not expose private data or bypass authentication.

#### Duplicate check

I searched open/closed issues, PR bodies, comments, reviews, inline review comments, `healthz` with query/wildcard/cache-bypass terms, and all open PRs touching `edge/src/worker.js` or `edge/wrangler.jsonc`. No existing report or active PR covers query variants of `/healthz`.

Related but different: issue #160 concerns `&n=` defeating optional CDN caching on idle room reads; PR #674 canonicalizes `/rooms` cache keys; PR #729 handles repeated `/rooms` parameters. PRs #664/#675 introduce and constrain the health edge lane but do not discuss query variants.

luch91:
## Summary

Fixes #751.

`/healthz` ignores query parameters, but the edge route was exact, so `/healthz?anything` bypassed the Worker. Even after routing those requests through the Worker, using the raw request as the Cache API key would create a separate cache entry and origin request for every ignored query variant.

This change:

- routes `healthz*` through the Worker;
- adds an explicit query-insensitive cache-key specification for `/healthz`;
- uses the canonical key for cache lookup, origin fetch, and cache storage;
- adds edge regression coverage for the route and key contract.

## Verification

- `git diff --check`: passed

The test suite uses source-level assertions for the Worker because this repository does not include a JavaScript Worker runtime harness. No live deployment or cache mutation was performed.

### DIFF
--- a/edge/snapshot.py
+++ b/edge/snapshot.py
@@ -117,6 +117,10 @@ EDGE_REVALIDATE = {"/rooms": 5}
 # dependency chain. Drift is caught instead — the edge-key test asserts it against
 # store.MAX_LIMIT, which is where the number actually lives.
 
+# The liveness reply does not vary by query at all. Give it an explicit empty key spec so the
+# Worker can collapse `/healthz`, `/healthz?n=1` and every other query variant to one shared
+# cache entry and one canonical origin request.
+#
 # Paths the edge owns outright: the origin serves nothing at them, so unlike everything else
 # here the stored bytes are not a copy of a live answer — they are the only answer. That is
 # why they are neither snapshotted (there is nothing to fetch) nor origin-first (there is
@@ -140,6 +144,11 @@ def rooms_key() -> dict:
             "clamped": {"limit": {"min": 1, "max": 200}},
         }
     }
+
+
+def edge_key() -> dict:
+    """Cache-key specs for every edge-cached or edge-revalidated path."""
+    return {"/healthz": {}, **rooms_key()}
 
 
 def asset_name(path: str) -> str:
@@ -207,8 +216,8 @@ def main() -> int:
             print(f"  {line}", file=sys.stderr)
         return 1
 
-    edge_key = rooms_key()
-    unspecified = sorted(set(EDGE_REVALIDATE) - set(edge_key))
+    key_spec = edge_key()
+    unspecified = sorted((set(EDGE_CACHED) | set(EDGE_REVALIDATE)) - set(key_spec))
     if unspecified:
         # Fail closed. Without a key spec the Worker would have to key on the raw URL, which
         # is the multiplication bug above — a deploy that silently did that is worse than one
@@ -223,7 +232,7 @@ def main() -> int:
                 "static_first": sorted(STATIC_FIRST),
                 "edge_cached": EDGE_CACHED,
                 "edge_revalidate": EDGE_REVALIDATE,
-                "edge_key": edge_key,
+                "edge_key": key_spec,
                 "edge_only": sorted(EDGE_ONLY),
             },
             indent=2,
--- a/edge/src/worker.js
+++ b/edge/src/worker.js
@@ -139,12 +139,13 @@ async function stored(request, env, pathname, { fallback }) {
 
 async function edgeCached(request, pathname, seconds) {
   const cache = caches.default;
-  const hit = await cache.match(request);
+  const key = cacheKey(new URL(request.url), pathname) ?? request;
+  const hit = await cache.match(key);
   if (hit) return hit;
 
   let fresh;
   try {
-    fresh = await fetch(request, { signal: AbortSignal.timeout(ORIGIN_TIMEOUT_MS) });
+    fresh = await fetch(key, { signal: AbortSignal.timeout(ORIGIN_TIMEOUT_MS) });
   } catch (err) {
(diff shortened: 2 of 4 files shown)


luch91:
Updated and verified exact head `4842a3960b52112c562f20551f10d16c8e239a7e` against current `main` `45921c3e3699e01a55cde391674815367e0cff6b`.

The prior independent blocker is addressed: `edge_key()` now explicitly covers both `/healthz` (empty query-insensitive spec) and `/rooms`, matching the fail-closed check over `EDGE_CACHED | EDGE_REVALIDATE`. Local Python compilation, Node syntax, and diff checks passed. All required GitHub checks for this exact head are green.

GitHub does not allow the PR author to approve their own pull request; maintainer review is still required.

Minh3132 (commented):
Reviewed exact head `4842a3960b52112c562f20551f10d16c8e239a7e` and found a response-variance regression in the new canonical `/healthz` fetch.

`edgeCached()` now builds the query-insensitive key and then uses that synthetic `Request` for the origin fetch as well as for cache lookup/storage:

```js
const key = cacheKey(new URL(request.url), pathname) ?? request;
...
fresh = await fetch(key, ...);
```

For `/healthz`, `edge_key()` deliberately supplies an empty spec, so `cacheKey()` constructs a bare GET containing neither the query string nor any caller headers. That is safe for query semantics, but `Origin` is not semantically inert at the HTTP response layer: the application is globally wrapped in `CORSMiddleware(allow_origins=config.CORS_ORIGINS, ...)`. On the PR's base, an allowed browser request with `Origin: https://a.example` therefore gets `Access-Control-Allow-Origin: https://a.example` from the origin, while the same request after this change reaches the origin *without* `Origin` and comes back without ACAO. The Worker then returns/caches those stripped-CORS bytes. In other words, fixing ignored query variance makes the edge-cached health endpoint unusable to an otherwise allowed cross-origin browser caller.

This is separate from #751's cache-multiplication bug: the canonical **cache identity** should ignore the query, but the canonical **origin fetch** still needs the request headers that affect the response (at minimum `Origin`), or `/healthz` needs the same explicit response-variance treatment being added for `/rooms` in #756. The current source-level regression only asserts `fetch(key)`, so it pins the behavior that drops `Origin` rather than exercising the externally observable CORS response.

Could the fix keep `/healthz` query-insensitive without using the headerless cache key as the origin request, and add a regression with an allowed `Origin` showing ACAO survives canonicalization?

bdunn77:
Independent validation of head `4842a396` (base `20a4457b`):

- On `origin/main`, wrangler owns exact `technocore.chat/healthz` (no `*`). `EDGE_CACHED` is `/healthz: 10`, there is no `edge_key()`, and `edgeCached` uses `cache.match(request)` / `fetch(request)` / `cache.put(request, ...)` — query variants are separate cache entries and can miss the Worker route.
- On this head: `test_healthz_query_variants_use_one_worker_route_and_cache_key` **1/1 pass**. Route is `technocore.chat/healthz*`, `edge_key()["/healthz"] == {}`, and the lane uses `cacheKey(...)` for match/fetch/put.

No merge blockers found on the healthz query-canonicalisation this PR adds.

PR #756, opened 2026-09-06
fix(edge): vary rooms cache by origin

### LINKED ISSUE #743: /rooms edge cache ignores Origin variance and breaks configured multi-origin CORS
### Commit verified against

### Report

`/rooms` varies `Access-Control-Allow-Origin` by the request's `Origin`, but the edge Worker caches the response under a URL-only canonical `Request`. With more than one configured CORS origin, the first origin to fill a `/rooms` entry determines the ACAO header served to the others.

#### Observed

First, the origin behaves correctly when configured with two origins:

```text
CHAT_CORS_ORIGINS=https://a.example,https://b.example

Origin: https://a.example -> Access-Control-Allow-Origin: https://a.example
Origin: https://b.example -> Access-Control-Allow-Origin: https://b.example
Vary: Origin
```

I then ran the current `edge/src/worker.js` with a minimal in-memory `caches.default`, a fake origin that emits the same `Vary: Origin`/ACAO behavior, and two sequential requests for the same canonical URL:

```js
const req = (origin) => new Request(
  "https://technocore.chat/rooms?format=json&limit=50",
  {headers: {Origin: origin}},
);

const first = await worker.fetch(req("https://a.example"), env, ctx);
const second = await worker.fetch(req("https://b.example"), env, ctx);
```

Observed output:

```text
origin calls=1
first ACAO=https://a.example
second ACAO=https://a.example
second Vary=Origin
```

The second origin is explicitly allowed, but a browser rejects this response because its ACAO value names the first origin.

#### Root cause

`cacheKey()` constructs a synthetic GET `Request` from only the canonical URL. It does not preserve `Origin`:

```js
return new Request(url.toString(), { method: "GET" });
```

`fromOrigin()` correctly forwards the caller's headers, so the fetched representation contains caller-specific CORS headers, but `revalidating()` stores and retrieves it through that shared URL-only key.

Cloudflare's Cache API documentation notes that header variance depends on the request headers retained at `put()`/`match()` time; removing headers from the cache request key makes those requests share the entry:

https://developers.cloudflare.com/workers/runtime-apis/cache/#match

#### Expected

Two origins listed in `CHAT_CORS_ORIGINS` should both be able to read `/rooms` through a browser, regardless of which one filled the edge entry first.

Possible fix boundaries are either:

- include the normalized allowed `Origin` in the Cache API request key; or
- keep the cached representation origin-neutral and reconstruct ACAO/Vary per caller after retrieval.

A regression can issue two requests with distinct allowed origins against one canonical `/rooms` key and assert that each response names its own origin. This is a CORS correctness failure; I did not demonstrate an unauthorized cross-origin read.

#### Duplicate check

I searched open/closed issues, PR bodies, issue/PR comments, reviews, inline review comments, and open PRs touching `edge/src/worker.js` or `src/app.py`. PR #674 covers query-key canonicalization and caller-specific `no-store` responses; #718 covers refresh subrequest caching; #729 covers repeated query parameters; #65 covers whitespace in configured origins. None reports `Origin` variance being lost from the `/rooms` edge cache key.

luch91:
## Summary

Fixes #743.

The `/rooms` representation varies by the caller's `Origin` when multiple CORS origins are configured. The edge cache previously canonicalized only the URL, so the first origin to fill a cache entry could supply its `Access-Control-Allow-Origin` value to later origins.

This change:

- declares `Origin` as the only header variance for the `/rooms` cache key;
- preserves that header on the synthetic Cache API key without copying unrelated caller headers;
- includes the same variance in the in-flight fill identity, preventing concurrent cross-origin fills from joining;
- adds regression assertions for the generated key policy and Worker implementation.

## Verification

- `git diff --check`: passed

The repository has no JavaScript Worker runtime harness, so the regression is expressed through the existing source-level edge tests. No live deployment or cache mutation was performed.

### DIFF
--- a/edge/snapshot.py
+++ b/edge/snapshot.py
@@ -142,6 +142,7 @@ def rooms_key() -> dict:
         "/rooms": {
             "match": dict(ROOMS_KEY_MATCH),
             "clamped": {"limit": {"min": 1, "max": 200}},
+            "vary": ["Origin"],
         }
     }
 
--- a/edge/snapshot.py
+++ b/edge/snapshot.py
@@ -117,6 +117,10 @@ EDGE_REVALIDATE = {"/rooms": 5}
 # dependency chain. Drift is caught instead — the edge-key test asserts it against
 # store.MAX_LIMIT, which is where the number actually lives.
 
+# The liveness reply does not vary by query at all. Give it an explicit empty key spec so the
+# Worker can collapse `/healthz`, `/healthz?n=1` and every other query variant to one shared
+# cache entry and one canonical origin request.
+#
 # Paths the edge owns outright: the origin serves nothing at them, so unlike everything else
 # here the stored bytes are not a copy of a live answer — they are the only answer. That is
 # why they are neither snapshotted (there is nothing to fetch) nor origin-first (there is
@@ -140,6 +144,11 @@ def rooms_key() -> dict:
             "clamped": {"limit": {"min": 1, "max": 200}},
         }
     }
+
+
+def edge_key() -> dict:
+    """Cache-key specs for every edge-cached or edge-revalidated path."""
+    return {"/healthz": {}, **rooms_key()}
 
 
 def asset_name(path: str) -> str:
@@ -207,8 +216,8 @@ def main() -> int:
             print(f"  {line}", file=sys.stderr)
         return 1
 
-    edge_key = rooms_key()
-    unspecified = sorted(set(EDGE_REVALIDATE) - set(edge_key))
+    key_spec = edge_key()
+    unspecified = sorted((set(EDGE_CACHED) | set(EDGE_REVALIDATE)) - set(key_spec))
     if unspecified:
         # Fail closed. Without a key spec the Worker would have to key on the raw URL, which
         # is the multiplication bug above — a deploy that silently did that is worse than one
@@ -223,7 +232,7 @@ def main() -> int:
                 "static_first": sorted(STATIC_FIRST),
                 "edge_cached": EDGE_CACHED,
                 "edge_revalidate": EDGE_REVALIDATE,
-                "edge_key": edge_key,
+                "edge_key": key_spec,
                 "edge_only": sorted(EDGE_ONLY),
             },
             indent=2,
--- a/edge/src/worker.js
+++ b/edge/src/worker.js
@@ -139,12 +139,13 @@ async function stored(request, env, pathname, { fallback }) {
 
 async function edgeCached(request, pathname, seconds) {
   const cache = caches.default;
-  const hit = await cache.match(request);
+  const key = cacheKey(new URL(request.url), pathname) ?? request;
+  const hit = await cache.match(key);
   if (hit) return hit;
 
   let fresh;
   try {
-    fresh = await fetch(request, { signal: AbortSignal.timeout(ORIGIN_TIMEOUT_MS) });
+    fresh = await fetch(key, { signal: AbortSignal.timeout(ORIGIN_TIMEOUT_MS) });
   } catch (err) {
     // An origin that will not answer inside the budget IS the health answer, so report it
     // here rather than letting the timeout escape to the fail-open handler. That handler
@@ -169,7 +170,7 @@ async function edgeCached(request, pathname, seconds) {
     // proxy reuse `ok` without contacting the edge at all — liveness staleness outside
     // Cloudflare's control, and beyond the reach of a purge.
     headers.set("Cache-Control", `public, max-age=0, s-maxage=${seconds}`);
-    await cache.put(request, new Response(body, { status: 200, headers }));
+    await cache.put(key, new Response(body, { status: 200, headers }));
     return new Response(body, { status: 200, headers });
   }
   return fresh;
--- a/edge/src/worker.js
+++ b/edge/src/worker.js
@@ -86,7 +86,7 @@ const inFlight = new Map();
 /** The key a copy is stored under: the reply space, not the URL space. Parameters the origin
  * ignores are dropped so they cannot multiply entries. Null means "no shared copy for this
  * request" — never a guess. */
-function cacheKey(url, pathname) {
+function cacheKey(url, pathname, request) {
   const spec = EDGE_KEY[pathname];
   if (!spec) return null;
   const keep = new URLSearchParams();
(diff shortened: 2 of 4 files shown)


bdunn77:
Independent validation of exact head `348c029c06021d5076fa72386b2188cb66eeb171` (tree `afac1c2c514c54ccb39beea0871b4b062c389ead`) on current main `9861d01cb42e10a5ffdffe3880338feaa4f56b3f`. Parent is that main. Three-file delta (`edge/snapshot.py` +9/-3, `edge/src/worker.js` +18/-8, `tests/edge/test_edge_worker.py` +14/-1).

The intended `/rooms` Origin-variance gap is causal. Copying only `test_rooms_cache_key_preserves_only_origin_variance` onto exact main **failed**: `cacheKey(url, pathname)` still builds a URL-only GET `Request`. On this head `tests/edge/test_edge_worker.py` **26 passed / 1 skipped**.

**Merge blocker:** this head also expands snapshot fail-closed from `EDGE_REVALIDATE` to `EDGE_CACHED | EDGE_REVALIDATE`, while `edge_key()` still returns only `rooms_key()` (`/rooms`). Exact main `python edge/snapshot.py` exits **0**. This head exits **1** with `snapshot FAILED: no cache-key spec for ['/healthz']`. Arithmetic on this tree: `EDGE_CACHED={'/healthz': 10}`, `edge_key()` keys `['/rooms']`, unspecified `['/healthz']`.

Sibling [#754](https://github.com/flop-labs/technocore-chat/pull/754) already supplies `edge_key() -> {"/healthz": {}, **rooms_key()}`. This PR cannot land alone until `/healthz` has an explicit empty key spec, or the unspecified set stays `EDGE_REVALIDATE`-only.

Gates otherwise: Ruff check/format on the Python files; `node --check edge/src/worker.js`; `sz.py --caps` (`app.py` 1019/1020). I did not run a full pytest suite because snapshot generation already fails closed on this exact tree.

A0-PR756-348c029c-afac1c2c

luch91:
The independent validation correctly identified a dependency on `/healthz` cache-key coverage. That prerequisite is now implemented and rebased in #754 exact head `4842a3960b52112c562f20551f10d16c8e239a7e`, with all required checks green.

I am leaving #756 unchanged until #754 is merged, rather than duplicating its `edge_key()` change in this PR. After #754 lands, this branch will be rebased onto the resulting current `main` and the complete edge suite rerun.

luch91:
Updated exact stacked head `90cc77cae58fb19c1406bc4a49080179ef65bcf6` (parent #754 head `4842a396…`). The first CI run exposed and I corrected a stale test assertion: the rebased `edgeCached` call correctly passes `request` so the canonical key preserves `Origin` variance. The exact-head rerun now passes lint/tests/image, MCP worker, OpenAPI, protected-files, and regression gates.

This PR remains intentionally dependent on #754; merging #754 first will remove the stacked prerequisite commit from this diff.

WIZARDspace:
Reviewed at `90cc77c` against `main` `20a4457`, on a scratch worktree. The stack is clean — `git merge-base` for this branch and #754 is exactly #754's head `4842a396` — so this PR's own delta is 3 files, +32/−9.

#743 reproduces. On `20a4457` with `CHAT_CORS_ORIGINS=https://a.example,https://b.example`:

```
Origin: https://a.example    -> access-control-allow-origin: https://a.example   vary: Origin
Origin: https://b.example    -> access-control-allow-origin: https://b.example   vary: Origin
Origin: https://evil.example -> (no ACAO)
(no Origin header)           -> (no Vary at all)
```

The origin side is correct and the reported gap is real. Everything below is about the Worker.

**1. The fix rests on `Vary` being honoured, which this repository documents as conditional — and declines to rely on one layer up.**

`cacheKey()` sets `Origin` on the synthetic key `Request`, so correctness depends on `caches.default` treating that header as part of the entry's identity. `fromOrigin` does carry `Vary: Origin` through to the stored copy (`new Headers(fresh.headers)` at `worker.js:199`, `put` at `:212`), so the response half is in place.

My understanding is that a Workers Cache API entry is addressed by the key request's URL, with header variance coming from the stored response's `Vary` rather than from headers set on the key — but I cannot test that against real Cloudflare from here, so I am asking rather than asserting. What makes it worth asking is that `src/app.py:398-401` already records the caveat:

> Only the plain answer is marked cacheable, which is belt-and-braces on top of Vary — Cloudflare honours Vary only where a Cache Rule enables it, so on a zone where nobody has, the edge can still only hold the default representation.

For the `Accept` case that comment describes, the chosen fix was **not** to trust `Vary`; it was to keep the non-default representation out of the shared cache entirely.

If that caveat reaches `caches.default` too, this PR is inert on a zone without the Cache Rule: one entry still serves every origin, #743 survives in production, and the suite stays green because it asserts source strings and — as your description says — there is no JS runtime harness. Could you confirm the Cache API varies on the key request's headers independently of a zone Cache Rule? If it does not, the doctrine-consistent fix may be app.py's: keep an `Origin`-bearing reply out of the shared copy, the way the budget-footer `no-store` branch at `worker.js:205` already does.

**2. `fill()` hardcodes `Origin` while `cacheKey()` generalises over `spec.vary`.**

`worker.js:218` builds the in-flight identity from the URL and `key.headers.get("Origin")` alone, while `cacheKey` copies every header named in `spec.vary ?? []` (`:109-111`). Today `edge_key()` is `{"/healthz": {}, **rooms_key()}` and `vary` appears exactly once in `snapshot.py` (`:145`, `["Origin"]`), so this is correct now and wrong on the first entry that adds a second vary header — two concurrent fills differing only in that header would join, and one caller would receive the other's reply. That is the bug this PR fixes, one layer out.

Since `cacheKey` sets *only* vary headers on the key, the key's own header list is already the right identity — iterating `key.headers` and joining the sorted `name=value` pairs into the identity string keeps it in step with `spec.vary` automatically, with no second list to maintain.

**3. `request?.headers` fails open rather than loudly.**

Both call sites pass `request` (`:150`, `:229`), so the optional chain is unreachable today. What it buys is that a future third caller which forgets the argument produces a key with **no** vary headers at all — silently reintroducing #743 rather than throwing. Where the key space is the reply space, a missing request is a programming error worth surfacing.

Minor: `assert "request.headers" not in key` (`test_edge_worker.py:396`) pins intent through the absence of a substring, so it would also fire on a legitimate future use of that expression.

Findings 2 and 3 are small and local. 1 is a question rather than a blocker — if the Cache API does vary on key headers, it dissolves and the approach is right.

PR #757, opened 2026-09-06
fix(edge): route versioned favicon URLs

### LINKED ISSUE #755: Versioned favicon.ico URLs bypass the edge route and return 404
### Commit verified against

### Report

The edge serves the canonical `/favicon.ico`, but adding a conventional version/cache-busting query makes the request bypass the Worker and reach the origin, where no favicon route exists. A semantically equivalent favicon URL therefore changes from 200 to 404.

#### Live reproduction

```bash
curl -si https://technocore.chat/favicon.ico
curl -si 'https://technocore.chat/favicon.ico?v=issue-probe'
```

Observed on the same run:

```text
GET /favicon.ico
200
Content-Type: image/x-icon
CF-Cache-Status: HIT
Content-Length: 3391

GET /favicon.ico?v=issue-probe
404
Content-Type: text/plain; charset=utf-8
CF-Cache-Status: BYPASS
body begins: 404 no route matched
```

#### Root cause

`edge/wrangler.jsonc` registers the edge-only asset with an exact route pattern:

```text
*technocore.chat/favicon.ico
```

The same file documents that route patterns are matched against the whole URL, including the query string. Therefore `?v=...` prevents the Worker from running.

This path is edge-only by design: PR #684 intentionally leaves the origin without a favicon route, and `stored(..., {fallback: false})` is the only successful response. Once the query variant bypasses that lane, the origin can only return its normal no-route 404.

Inside the Worker, routing already uses `url.pathname` and the stored asset is selected by `/favicon.ico`, so the query has no response semantics after the request reaches the Worker.

#### Expected

`/favicon.ico?v=<version>` should serve the same edge-owned icon as `/favicon.ico`. Query-based versioning/cache busting is a normal way for pages and clients to refresh static assets.

The narrow fix is to route `/favicon.ico*` through the Worker and add a route regression containing `?v=`. No origin route or new storage behavior is needed.

This is a functional edge-routing bug. It does not expose private data or reach an authenticated surface.

#### Duplicate check

I searched open/closed issues, PR bodies, comments, reviews, inline review comments, favicon/query/version/cache-busting/wildcard terms, and every open PR touching `edge/src/worker.js` or `edge/wrangler.jsonc`. No existing report or active PR covers favicon query variants.

PR #684 introduced the edge-only favicon and its discussion does not mention query strings. Open PR #754 changes only the `/healthz` route/key for issue #751; it does not modify the favicon route.

luch91:
## Summary

Fixes #755.

The edge-owned `/favicon.ico` route was exact, so a conventional versioned URL such as `/favicon.ico?v=1` bypassed the Worker and reached the origin, which has no favicon route. The same pathname-based edge asset should serve both forms.

This change widens the route to `favicon.ico*` and adds a regression asserting that versioned favicon URLs remain in the edge-only lane. It does not add an origin route or change asset lookup/storage behavior.

## Verification

- `git diff --check`: passed

No live deployment or cache mutation was performed.

### DIFF
--- a/edge/wrangler.jsonc
+++ b/edge/wrangler.jsonc
@@ -47,7 +47,9 @@
     // the request nobody makes. This is the only route whose reply depends on its query.
     { "pattern": "technocore.chat/rooms*", "zone_name": "technocore.chat" },
     // Owned by the edge, never served by the origin — see EDGE_ONLY in snapshot.py.
-    { "pattern": "technocore.chat/favicon.ico", "zone_name": "technocore.chat" },
+    // Query parameters have no meaning to the edge-owned icon; keep versioned URLs in the
+    // same Worker lane so the pathname-based asset lookup can answer them too.
+    { "pattern": "technocore.chat/favicon.ico*", "zone_name": "technocore.chat" },
     { "pattern": "technocore.chat/robots.txt", "zone_name": "technocore.chat" },
     { "pattern": "technocore.chat/sitemap.xml", "zone_name": "technocore.chat" },
     { "pattern": "technocore.chat/openapi.json", "zone_name": "technocore.chat" },
--- a/tests/edge/test_edge_worker.py
+++ b/tests/edge/test_edge_worker.py
@@ -427,6 +427,19 @@ def test_a_path_whose_reply_varies_by_query_is_routed_with_a_wildcard():
         )
 
 
+def test_edge_owned_favicon_accepts_versioned_query_urls():
+    """The edge-only icon is pathname-selected, so cache-busting queries must stay routed."""
+    raw = (EDGE / "wrangler.jsonc").read_text(encoding="utf-8")
+    patterns = {
+        r["pattern"] for r in json.loads(re.sub(r"^\s*//.*$", "", raw, flags=re.M))["routes"]
+    }
+    assert "technocore.chat/favicon.ico*" in patterns
+    worker = (EDGE / "src" / "worker.js").read_text(encoding="utf-8")
+    lane = _between(worker, "if (EDGE_ONLY.has(pathname))", "let originResponse")
+    assert "stored(request, env, pathname" in lane
+    assert "pathname" in lane
+
+
 def test_an_edge_owned_path_is_not_snapshotted_and_not_origin_first():
     """EDGE_ONLY is the one lane whose stored bytes are not a copy of anything. The origin
     serves nothing at these paths, so snapshotting one would fetch a 404 and fail the deploy,


luch91:
Rebased onto current `main` `45921c3e3699e01a55cde391674815367e0cff6b` at exact head `b5e572cafac0d7a821ee16e55226e00d042c2656`.

The branch now contains only the versioned favicon route/test change relative to current main. Python compilation and diff checks pass; all required GitHub checks are green, including lint/tests/image, MCP worker bundle, and OpenAPI contract. GitHub does not allow the PR author to approve their own pull request; maintainer review is required.

bdunn77:
Independent validation of head `b5e572ca` (base `20a4457b`):

- On `origin/main`, wrangler owns `technocore.chat/favicon.ico` with no trailing `*`. A versioned URL such as `/favicon.ico?v=2` is not in that exact route.
- On this head: `test_edge_owned_favicon_accepts_versioned_query_urls` **1/1 pass**. The route is `technocore.chat/favicon.ico*`, and the EDGE_ONLY lane still selects the asset by pathname.

No merge blockers found on the versioned-favicon route this PR adds.

WIZARDspace:
Verified at `b5e572c` merged onto `main@e4c4f73`. #755 still reproduces on the live deployment:

```
/favicon.ico          200  image/x-icon               cf-cache-status: HIT
/favicon.ico?v=probe  404  text/plain; charset=utf-8  cf-cache-status: BYPASS
```

On the merge: `uv run --group dev just check` exit 0, **783 passed, 1 skipped**; `tests/edge/test_edge_worker.py` 26 passed, 1 skipped. Reverting `edge/wrangler.jsonc` to main fails the new test on `assert "technocore.chat/favicon.ico*" in patterns`.

**The approach matches this file's own documented precedent**, which is worth stating because it is why the one-character change is the right one. The comment six lines above the changed route already says it:

> "The trailing `*` is load-bearing: a route pattern is matched against the whole URL, query string included, so `technocore.chat/rooms` matches a bare `/rooms` and nothing else."

`/rooms*` is a wildcard with no exact companion route, so replacing the exact favicon pattern rather than adding alongside it is consistent, not a departure.

**One assumption the tests cannot reach, and it decides whether #755 actually closes.** The fix routes the versioned URL to the Worker; serving it then depends on `env.ASSETS` resolving `/favicon.ico?v=1` to `public/favicon.ico`, because `stored()` passes the URL through unchanged (`env.ASSETS.fetch(new URL(request.url))`). `asset_name()`'s docstring points that way — assets are stored "at its own path verbatim so the Worker can look it up with the request URL unchanged" — but nothing here exercises it, and the widened route only exists after a deploy.

If that assumption is wrong, the failure is silent rather than loud: `stored()` returns null, and the EDGE_ONLY branch does not proxy to the origin — it answers its own 404:

```js
if (!copy) {
  return new Response("not found\n", { status: 404, headers: { "Cache-Control": "no-store" } });
}
```

So `/favicon.ico?v=1` would still be a 404, the suite would still be green, and #755 would look fixed. The two cases are distinguishable by body: `not found\n` means the edge answered and the asset lookup missed; `404 no route matched` — what #755 captured — means the request never reached the Worker.

That makes the post-deploy check one line, and worth doing before #755 is closed:

```bash
curl -si 'https://technocore.chat/favicon.ico?v=1' | head -1
```

No blockers on the change itself.

PR #760, opened 2026-09-06
fix(limit): avoid zero-wait polling advice

### LINKED ISSUE #759: CHAT_MAX_WAIT=0 makes 429 recovery text recommend zero-delay polling
### Commit verified against

### Report

#### Minimal reproduction

From a fresh checkout:

```bash
CHAT_ROOT="$(mktemp -d)" \
CHAT_RATE_READ=1 \
CHAT_MAX_WAIT=0 \
PYTHONPATH=src \
uv run python - <<'PY'
import app

client = TestClient(app.app)
print(response.status_code)
PY
```

Observed:

```text
200
429
```

#### Root cause

```python
```

#### Expected

#### Duplicate check

Kewe63:
Fixes #759

## Summary

- replace the `&wait=<ceiling>` recommendation with `Retry-After` guidance when `CHAT_MAX_WAIT=0`
- preserve the existing long-poll recommendation byte-for-byte for positive ceilings
- add an HTTP regression that checks both the actionable replacement and removal of the impossible `wait=0` cadence

## Root cause

`limit.limited()` interpolated `max_wait` into a recovery sentence that assumed the configured ceiling was positive. Zero is a valid effective ceiling, so the 429 body recommended an immediate `&wait=0` request and described it as one request per zero seconds.

## Implementation

The renderer now selects one short `poll_advice` string:

- positive ceiling: the existing `&wait=<seconds>` guidance
- zero ceiling: wait for the response's `Retry-After` before polling again

The surrounding 429 response shape, rate-limit arithmetic, headers, and positive-ceiling behavior remain unchanged. The body literals were reflowed without changing their joined output so `core/limit.py` remains at its existing 183-line cap.

## Verification

- RED before the fix: focused regression failed because the zero-ceiling response lacked `wait for Retry-After before polling again`
- `uv run pytest tests/http/test_limits.py -q` — 54 passed
- `uv run ruff format --check .` — 87 files already formatted
- `uv run ty check` — passed
- `uv run sz.py --caps` — passed; `core/limit.py` remains 183/183
- `git diff --check` — passed
- full suite: `uv run coverage run -m pytest tests -q` — 747 passed, 1 skipped
- `uv run coverage report` — 97.89% total coverage

## Compatibility and documentation

No API, status, header, configuration, or positive-ceiling response behavior changes. Only the incorrect recovery sentence at an effective zero long-poll ceiling changes. Existing configuration documentation remains accurate.

Proposed release note: A 429 from an instance with `CHAT_MAX_WAIT=0` now tells callers to honor `Retry-After` instead of recommending an immediate `wait=0` poll.

## Abuse impact

No new public surface or capability. The change reduces retry pressure by replacing advice that encouraged immediate polling when long polling is disabled.

### DIFF
--- a/src/limit.py
+++ b/src/limit.py
@@ -373,20 +373,20 @@ def limited(kind: str, per_min: int, retry_after: float, *, text, max_wait: floa
     deployment), so a caller that never reads /.well-known/agent.json still finds out what
     it is pacing against at the one moment the answer matters.
     """
-    wait = max(1, round(retry_after))
-    other = "write" if kind == "read" else "read"
+    wait, other = max(1, round(retry_after)), "write" if kind == "read" else "read"
+    poll_advice = (
+        f"prefer &wait={max_wait:g} to tight polling — one request per {max_wait:g}s instead of twenty"
+        if max_wait > 0
+        else "wait for Retry-After before polling again"
+    )
     body = (
         f"429 rate limited: the {kind} budget for your IP ({per_min}/min) is spent.\n"
-        f"retry after: {wait}s — the bucket refills continuously "
-        f"({refill_rate(per_min)}), so waiting longer buys a bigger burst, up to "
-        f"{per_min}.\n"
-        f"still open: {other}s are a separate budget and are unaffected, and these paths "
-        f"are never rate limited: {FREE_PATHS}.\n"
+        f"retry after: {wait}s — the bucket refills continuously ({refill_rate(per_min)}), "
+        f"so waiting longer buys a bigger burst, up to {per_min}.\n"
+        f"still open: {other}s are a separate budget and are unaffected, and these paths are never rate limited: {FREE_PATHS}.\n"
         f"cheaper pattern: poll /r/<room>?since=<last seq you saw> rather than refetching "
-        f"the room, and prefer &wait={max_wait:g} to tight polling — one request per "
-        f"{max_wait:g}s instead of twenty.\n"
-        f"the enforced numbers are also published at /.well-known/agent.json under "
-        f"limits.{kind}s_per_minute_per_ip."
+        f"the room, and {poll_advice}.\n"
+        f"the enforced numbers are also published at /.well-known/agent.json under limits.{kind}s_per_minute_per_ip."
     )
     r = text(body, 429)
     r.headers["Retry-After"] = str(wait)
--- a/tests/http/test_limits.py
+++ b/tests/http/test_limits.py
@@ -183,6 +183,16 @@ def test_the_429_names_the_budget_the_manual_deliberately_does_not(client, monke
             assert "writes are a separate budget" in read.text
             assert "limits.reads_per_minute_per_ip" in read.text
 
+        # A zero ceiling disables long-poll pacing, so recommending wait=0 would send the
+        # caller straight back into the bucket this response is asking it to leave alone.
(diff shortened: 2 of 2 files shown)


luch91 (approved):
Reviewed exact head `6a346878de014f08447b8d6e65936ce97a6c70a4`.

The new branch is consistent with the enforced configuration: `CHAT_MAX_WAIT` is clamped to a non-negative value, so zero is valid. Positive ceilings retain the existing `&wait=<ceiling>` guidance; zero now tells the caller to honor `Retry-After` instead of recommending `wait=0`. The regression asserts both the zero-mode message and absence of the unusable advice.

Local verification: Ruff check and format check passed; `git diff --check` passed. The focused pytest could not collect on Windows because the existing `src/store.py` import requires POSIX-only `fcntl`; no local test pass is claimed. GitHub had not yet reported the main CI checks when this review was submitted.

I found no code-level blocker; normal CI approval remains required.

bdunn77:
Independent validation of exact head `6a346878de014f08447b8d6e65936ce97a6c70a4` (tree `fef690fa5298eec9f20c84a872ee7615d2e8523c`) on current main `9861d01cb42e10a5ffdffe3880338feaa4f56b3f`. Parent is that main. Two-file delta (`src/limit.py` +11/-11, `tests/http/test_limits.py` +10).

The intended recovery-text gap is causal. Copying only the new assertions onto exact main **failed**: with `CHAT_MAX_WAIT=0` and `RATE_READ=1`, the second `/rooms` GET 429'd with `prefer &wait=0 to tight polling — one request per 0s instead of twenty`. On this head `tests/http/test_limits.py` **54 passed**. Extra probes outside the submitted tests: zero ceiling recommends `wait for Retry-After before polling again` and omits `&wait=0`; positive `MAX_WAIT=10` keeps the existing wait-10 sentence (**2 passed**).

Gates on this exact tree: Ruff check/format and `ty` on the changed files; `sz.py --caps` (`limit.py` 183/183, `app.py` 1019/1020); focused limits **54 passed**; extra probes **2 passed**; clean full suite **747 passed / 1 skipped**; coverage **97.89%**. **No merge blocker in this scope.**

Non-blocking: @luch91 already approved this head but did not claim an independent pytest pass (Windows `fcntl`). This comment is that missing suite result, not a second review.

A0-PR760-6a346878-fef690fa

luch91 (approved):
Re-reviewed exact head 6a346878de014f08447b8d6e65936ce97a6c70a4. The independent validation comment reports the focused limits tests (54 passed), full suite (747 passed, 1 skipped), coverage (97.89%), Ruff/format, ty, size caps, and extra zero/positive-ceiling probes. The implementation and regression test match the intended zero-ceiling behavior. Approved.

luch91 (approved):
Reviewed exact head `6a346878de014f08447b8d6e65936ce97a6c70a4`. `limited()` now avoids emitting `&wait=0` / “one request per 0s” when `MAX_WAIT` is zero, while retaining the existing long-poll advice for positive values. The added test exercises `MAX_WAIT=0` with a real 429 and asserts the safe fallback plus absence of the invalid advice. The change is narrow and preserves the response status, retry header, and rate-limit contract. The repository’s required checks are not reported here beyond the protected-files/overlap gates; I am approving the source/test change, not claiming an independent full-suite run.

PR #761, opened 2026-09-06
fix(mcp): reject swept-empty signed messages

### LINKED ISSUE #758: No-key say_signed returns a challenge the service always rejects
### Commit verified against

### Report

When the MCP wrapper has no configured signing key, `say_signed` returns the exact canonical challenge for an external signer to sign and retry. For whitespace/control-only text, the wrapper returns a challenge whose swept text field is empty. Signing that challenge exactly and retrying can never succeed: the service always refuses it as empty text.

#### Minimal reproduction

Against an isolated `TestClient` store:

```python
mcp_server._signer = None

try:
    anyio.run(mcp_server.say_signed, "mb-inbox", " \n\t ")
except ToolError as exc:
    challenge = str(exc).splitlines()[-1]

room, nonce, swept = challenge.split("|", 2)
assert swept == ""

# Use any fresh test keypair; do not reuse a production identity.
response = client.post(
    "/r/mb-inbox",
    json={
        "did": test_signer.did,
        "sig": test_signer.sign(challenge),
        "nonce": nonce,
        "text": " \n\t ",
    },
)
print(repr(challenge), response.status_code, response.text.splitlines()[0])
```

Focused command used during verification:

```text
uv run python /tmp/technocore_interop_repro.py
```

Observed:

```text
MCP challenge 'mb-inbox|<minted nonce>|': retry=400
400 empty text: nothing visible was left after the single-line sweep
```

The challenge was signed without modification; this is not a bad-signature or stale-nonce result.

#### Root cause

`signing.sweep()` deliberately performs normalization without the service's final semantic refusals. That is useful in normal request mode because the origin remains authoritative and can return its detailed 400.

No-key challenge mode is different: `_resolve_signature()` raises a `ToolError` containing the canonical payload before any origin request occurs. `say_signed()` does not check that the swept text remains non-empty before calling it, so it asks the external signer to sign a request already known locally to be unusable.

#### Expected

Challenge mode should not present a canonical payload as ready to sign when retrying it exactly is guaranteed to fail. `say_signed` should reject text that sweeps to empty before `_resolve_signature()` emits the challenge, ideally using the same semantic wording as the service.

A regression can configure no signer, pass whitespace/control-only text, and assert a direct empty-text refusal rather than a challenge ending in `|`. A visible text should continue to return a usable challenge.

This is MCP challenge correctness and wasted external-signing work; it does not weaken signature verification or authorization.

#### Duplicate check

I searched open/closed issues, PR bodies, comments, reviews, inline review comments, `say_signed`/challenge/sweep/empty-text terms, and every open PR touching the MCP server/signing modules or tests. No existing report or active PR covers an unusable no-key challenge.

PR #318 covers signed-lane conformance vectors and discusses sweep failures in third-party clients, but it does not exercise MCP no-key challenge generation. PR #98 documents swept/trimmed signed text; it does not address challenge mode.

luch91:
Fixes #758.

## What changed

When `say_signed` receives text that becomes empty after the MCP signing sweep, it now raises the same class of empty-text refusal before `_resolve_signature()` can return an external-signing challenge. A challenge ending in an empty text field cannot succeed when signed and retried unchanged.

Added a regression test covering whitespace/control-only input and retained the existing visible-text external-signing test. Added the required Unreleased changelog entry.

## Verification

- `python -m compileall -q mcp/src/technocore_mcp tests/test_mcp.py` — passed.
- `git diff --check` — passed.
- Focused pytest selection reached collection but cannot run in this Windows environment because `src/store.py` imports POSIX-only `fcntl` (`ModuleNotFoundError`), the repository's existing Windows MCP-test limitation.
- WSL is installed, but its Ubuntu environment has no `uv` or pytest, so no Linux test result is claimed.

### DIFF
--- a/mcp/src/technocore_mcp/server.py
+++ b/mcp/src/technocore_mcp/server.py
@@ -587,6 +587,14 @@ async def say_signed(
     # satisfies that, so nothing is read before the write.
     minted = signing.next_nonce()
     swept = signing.sweep(text)
+    if not swept and _signer is None and did is None and sig is None and nonce is None:
+        # A no-key caller receives the exact canonical string from _resolve_signature. An
+        # empty swept body cannot pass the service's semantic check, so do not hand an
+        # external signer a challenge that is guaranteed to fail when retried unchanged.
+        raise ToolError(
+            "empty text: nothing visible was left after the single-line sweep. "
+            "Send at least one visible character."
+        )
     did, sig, nonce = _resolve_signature(f"{room}|{minted}|{swept}", did, sig, nonce, minted)
     return await _post(
         f"/r/{_segment(room)}", {"did": did, "sig": sig, "nonce": str(nonce), "text": text}
--- a/tests/test_mcp.py
+++ b/tests/test_mcp.py
@@ -1039,6 +1039,36 @@ def test_with_no_key_and_no_signature_the_error_is_the_challenge(mcp):
     assert retried.is_error is False, text_of(retried)
 
 
+def test_no_key_does_not_issue_an_unusable_empty_text_challenge(mcp, monkeypatch):
+    """Whitespace/control-only input is rejected before challenge generation.
+
+    The service refuses the swept-empty body. Returning a challenge ending in an empty
+    text field would ask an external signer to sign a request that cannot succeed when
+    retried unchanged.
+    """
+    monkeypatch.setattr(mcp.module, "_signer", None)
+
+    reply = mcp.call("say_signed", {"room": "mb-inbox", "text": " \n\t "})
+
+    assert reply.is_error is True
+    message = text_of(reply)
+    assert "empty text" in message
+    assert "nothing visible was left" in message
+    assert "mb-inbox|" not in message
+
+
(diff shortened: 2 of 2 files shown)


kira19018-png:
Independent static review against base `9861d01` and head `243149f`: the new check fixes #758's unusable **no-key challenge**, but it currently widens the behavior change to the other two signed modes as well.

`signing.sweep()` deliberately documents itself as “the single-line sweep, minus the refusals”: empty/too-long semantic refusals are left to the service so its authoritative body reaches the model. Here `if not swept` runs **before** `_resolve_signature()`, so it also intercepts (1) a configured signer and (2) a complete externally supplied `did`/`sig`/`nonce`. Those modes can make the POST and previously surfaced the origin's own 400 body; this head replaces it with an MCP-local paraphrase.

The impossible request exists only when all three credentials are absent **and** `_signer is None`, because that is the branch that returns a challenge instead of reaching the origin. I would scope the local empty check to that branch (or move it into the challenge-only branch of `_resolve_signature`) and add a control proving configured-signing / fully supplied credentials still reach `_post` for swept-empty text and expose the service refusal.

No overlap with the core #758 fix itself; this is only about keeping its behavior change challenge-specific.

PR #762, opened 2026-09-06
fix(limit): preserve bare duplicate-ref text

### LINKED ISSUE #694: Duplicate-key normalization erases legitimate token-shaped text

luch91:
Fixes #694.

## What changed

`limit.normalize_text()` now removes a duplicate reference only when the documented `&ref=422-...` marker is present. A bare token-shaped substring is ordinary caller text and remains part of the duplicate key; the service has no state proving that arbitrary text was issued as a duplicate reference.

Regression coverage now asserts both behaviors:

- bare `422-...` content remains distinct from text without it;
- explicit `&ref=...` content normalizes away;
- the HTTP duplicate-filter test preserves bare token-shaped messages while continuing to collapse the documented marker form.

## Verification

- Ruff check passed for all changed Python files.
- Ruff format check passed for all changed Python files.
- `git diff --check` passed.
- Python compilation passed.
- The repository pytest suite was not claimed locally: Windows cannot import the existing POSIX-only `fcntl` dependency, and the WSL dependency sync on the mounted worktree did not complete.

### DIFF
--- a/src/limit.py
+++ b/src/limit.py
@@ -145,11 +145,11 @@ def normalize_text(text: str) -> str:
     text = "".join(
         " " if unicodedata.category(c) in store.INVISIBLE_CATEGORIES else c for c in text
     )
-    # A duplicate 422's ref token (app._REF's shape, with the `&ref=` the body shows it
-    # behind, and nothing else), pasted into the text instead of the query string, is cut
-    # out so it can never be what makes a copy unique — neither on its own nor by taking
-    # the word it was glued to with it.
-    return " ".join(re.sub(r"(?:&?ref=)?422-[\da-f]{1,8}-[\da-f]{4}", " ", text.casefold()).split())
+    # A duplicate 422's ref token, pasted into the text instead of the query string, is
+    # cut out only when its documented `&ref=` marker is present. A bare token-shaped
+    # substring is ordinary caller text: the server has no state proving that it was
+    # issued as a duplicate reference, so removing it would merge distinct messages.
+    return " ".join(re.sub(r"&ref=422-[\da-f]{1,8}-[\da-f]{4}", " ", text.casefold()).split())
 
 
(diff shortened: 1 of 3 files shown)


luch91:
PR #699 covers the same #694 fix, so I am closing this duplicate rather than competing.

I independently confirmed the remaining boundary concern on #699: the proposed `&ref=422-[\\da-f]{1,8}-[\\da-f]{4}` pattern has no right-hand boundary. It strips the valid-token prefix from malformed/longer caller text such as `&ref=422-deadbeef-abcde`, leaving `9` in the normalized text. The service's `_REF` parser requires the token to end at `&` or end-of-query, so normalization should apply the same boundary. The finding is recorded on #699 for the surviving PR.

luch91:
Closing as a duplicate of #699, which implements the same #694 fix. The boundary finding is recorded on #699.

PR #97, opened 2026-08-24
docs: clarify healthz and timed-out writes

thememeguy1:
## Summary

- document that `/healthz` is a liveness probe and does not exercise room or note storage
- clarify that client timeouts and reverse-proxy 502/524 responses leave a write's outcome unknown
- explain how to reconcile a timed-out signed room write using its DID, nonce, and text before issuing a new nonce
- publish the same readiness distinction in the served manual and OpenAPI document

## Why

A caller can receive `200` from `/healthz` while dynamic room paths remain unavailable at the proxy or storage layer. A write can also become durable before its response is lost, so blindly incrementing the nonce and re-signing the same text risks creating a duplicate.

## Validation

- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run ty check`
- `uv run coverage run -m pytest tests -q` — 322 passed
- `uv run coverage report` — 97.42%

## Compatibility and abuse impact

Documentation and regression-test changes only. No API behavior, response shape, public surface, persistent state, or abuse characteristics change.

PR #770, opened 2026-09-07
fix(humans): fail the job when the service never becomes healthy

### LINKED ISSUE #769: Browser job exits 0 when uvicorn never becomes healthy, then fails ~30s later at the probe
## Commit verified against

`1c31dbe` (current `origin/main` HEAD at the time of filing, verified on a fresh fetch)

## Report

The browser job in `.github/workflows/humans.yml` ("Start the service", lines 102-109) can exit 0 when uvicorn never becomes healthy, so the job advances to "Drive /humans in Chromium" against a port nothing is listening on. The failure surfaces ~30s later as a Playwright ECONNREFUSED instead of at the readiness gate, and the error message points at the probe rather than the service that never started.

### What is happening

The step backgrounds uvicorn on port 8099, then polls `/healthz` in a 30-iteration `for` loop under the step's default `set -e`:

```yaml
- name: Start the service
  run: |
    CHAT_ROOT="$(mktemp -d)" \
      uv run uvicorn --app-dir src app:app --port 8099 --log-level warning &
    for _ in $(seq 30); do
      curl -fsS localhost:8099/healthz && break
      sleep 1
    done
```

On each failed iteration `curl` returns non-zero, but it is a non-final constituent of a `&&` list (`curl ... && break`), so `set -e` does not fire on it. The last command executed on the final iteration is `sleep 1`, which exits 0, and that becomes the loop's exit status and the step's. The job then proceeds to the probe step (line 119) against a dead port.

### Minimal reproduction

The loop shape alone, against any unreachable port, reproduces the exit-0:

```bash
bash -e -c '
for _ in $(seq 3); do
    curl -fsS localhost:1/healthz && break
    sleep 1
done
echo "loop exited with: $?"
'
```

Output: three `curl: (7) Failed to connect` lines, then `loop exited with: 0`, outer exit 0.

In the workflow, trigger the browser job (push to main, a PR, or `workflow_dispatch`) with a change that prevents uvicorn from binding within the retry window: a syntax error in `src/app.py`, a bad import, or an already-bound port 8099. The "Start the service" step exits 0, the job reaches "Drive /humans in Chromium", and only there fails with a connection-refused error after ~30s of sleeps.

### Suggested fix

Add a post-loop guard that re-checks `/healthz` and fails the step when it was never reached:

```yaml
    for _ in $(seq 30); do
      curl -fsS localhost:8099/healthz && break
      sleep 1
    done
    curl -fsS localhost:8099/healthz || { echo "::error::service did not become healthy"; exit 1; }
```

This makes a dead service fail at the readiness gate with a clear message instead of ~30s later as a probe ECONNREFUSED that nobody can distinguish from a probe-logic failure.

### Notes

- No existing open issue or PR covers this. The closest are #754 (edge worker health-check query canonicalization) and #97 (healthz docs), both about different surfaces.
- The workflow is itself the gate; there is no unit-test seam for a workflow step's failure semantics. The regression check is the bash repro above (the loop shape exits 0 without the guard, 1 with it).

beardthelion:
## What

The browser job's "Start the service" step now fails when uvicorn never becomes healthy, instead of exiting 0 and letting the job reach the Playwright probe against a dead port. No service or MCP behavior changes; this is a CI failure-path fix only.

## Why

Closes #769. In `humans.yml` the readiness loop polls `/healthz` 30 times under `set -e`, but a failed `curl` is a non-final constituent of a `&&` list and the trailing `sleep 1` masks it, so the loop exits 0 when uvicorn never binds. The job then failed ~30s later as a probe ECONNREFUSED, which reads as a probe bug rather than a service that never started. A post-loop guard re-checks `/healthz` and exits 1 with a clear `::error::` when the service was not reached. Verified against `1c31dbe` (current `origin/main`). No open PR covers this; #754 and #97 touch different surfaces.

## Checks

- [ ] `uv run coverage run -m pytest tests -q && uv run coverage report` — n/a, workflow YAML only
- [ ] `uv run ruff check . && uv run ruff format --check .` and `uv run ty check` — n/a, workflow YAML only
- [ ] Docs that would now be wrong are updated — none; this is a CI workflow fix, no served doc or manual change

### DIFF
--- a/.github/workflows/humans.yml
+++ b/.github/workflows/humans.yml
@@ -107,6 +107,7 @@ jobs:
             curl -fsS localhost:8099/healthz && break
             sleep 1
           done
+          curl -fsS localhost:8099/healthz || { echo "::error::service did not become healthy"; exit 1; }
 
       # The probe resolves `playwright` from tests/node_modules by walking up from its own
       # directory, so it runs in place with no move and no NODE_PATH (which ESM ignores).


bdunn77:
Independent validation of exact head `0d70cc904ab3289f5aa210061fee212bd8e69503` (tree `ea4ee8be52227d6e930fca62feaf31c25eaced00`).

**Intended fix is causal.** Under `bash -eo pipefail`, the unguarded readiness loop (`curl ... && break; sleep`) exits **0** when `/healthz` never binds, because failed `curl` is a non-final `&&` constituent and the last command is `sleep`. Adding the post-loop `curl -fsS ... || { echo "::error::service did not become healthy"; exit 1; }` makes the same dead-port schedule exit **1** with that error. A healthy `/healthz` still exits **0** on both shapes.

Current `main` still lacks the guard. Compare `main...0d70cc90` remains the intended one-line `.github/workflows/humans.yml` change (ahead 1 / behind 5). No comments or reviews yet. Closes #769; no overlapping open PR on this workflow failure path.

**Not a merge blocker here:** `.github/workflows/ci.yml` still has the same unguarded `seq 30` `/healthz` loops. That is a sibling CI surface, not this humans-job delta.

**Verdict:** no merge blocker in this scope. Rebase onto current main should stay one line.

A0-PR770-0d70cc90-ea4ee8be

luch91 (approved):
Reviewed exact head 0d70cc904ab3289f5aa210061fee212bd8e69503. The one-line guard is placed immediately after the readiness loop and makes the previously masked curl failure exit the job before the browser probe. It adds no service behavior or public surface. Queue-guard protected-files and overlap passed. Workflow execution itself is not reproducible from this review environment; approval is based on the exact YAML diff and the stated failure-path correction.

PR #772, opened 2026-09-07
build: move CONTRIBUTING's recipes into a justfile and three CI guards

sv:
## What

CONTRIBUTING.md stops carrying shell recipes and becomes principles plus pointers. The recipes move to a `justfile` that CI itself runs, and three of its rules become checks.

- `justfile` — `check`, `lint`, `fmt`, `test`, `contract`, `mutate`, `mutate-patterns`, `bench`, `serve`, `build`. `just` is a locked dev dependency (`rust-just`), so `uv sync --frozen` is still the whole setup. The per-command rationale that used to live in the doc now sits beside each command.
- `ci.yml` — the lint + tests job runs `uv run just check`, so "run what CI runs" is literal. `sz.py --caps` is inside it; the `--check` ratchet stays on pushes to `main`. The contract job is left on its original two-step shape: a bash trap and the runner's step teardown are not the same semantics, and I did not want to change what happens when the fuzzer dies mid-step without a runner-side check.
- `pr-guards.yml` — three jobs, all fork-safe (read-only token, no `pull_request_target`, output to the step summary): `title` (conventional-commit grammar derived from what `main` has actually merged, 107/144 commits), `fix-has-a-failing-test` (a `fix` PR must touch `tests/`; those tests are overlaid onto the base commit and must fail there), `perf-bench-delta` (`capacity_bench.py --scale 0.25` on base and head, both posted, informs but does not gate — the bench's own header says host drift is ~20%).
- `tests/unit/test_contributing_recipes.py` — fenced bash in CONTRIBUTING may only name real dependency groups and real recipes. It caught one broken recipe on `main` today: `uv sync --frozen --group contract` names a group that does not exist (schemathesis is in `dev`).
- CONTRIBUTING.md 194 → 143 lines; the translations reasoning moved intact to `docs/translations.md`. Every "Making a change" bullet survives verbatim, including the four this PR originally added.

## Why

Prose recipes rot silently — one already had. A runner target cannot; a test on the doc keeps the rest honest. The three guards mechanise the rules a reviewer could not verify by reading the diff: whether a regression test actually fails without the fix, and whether a speedup has a number behind it.

Not done here: the guards are not required checks (branch protection is a repository setting). The one thing only CI can prove is that `actions/checkout` with `fetch-depth: 0` leaves both `base.sha` and `head.sha` reachable on a fork PR; this PR's own title is `build`, so the first real `fix:` PR after merge exercises it.

## Checks

- [x] `uv run just check` on the merged tree — 770 passed, 1 skipped, coverage 97.97%, `sz.py --caps` 2317/3010
- [x] `uv run just contract` — 881 generated, 881 passed
- [x] `uv run pytest tests/unit/test_contributing_recipes.py` — 5 passed (failed on `main`'s CONTRIBUTING line 112, as intended)
- [x] `actionlint` + `shellcheck` over all workflows — clean; the fail-on-base script was run byte-for-byte from the YAML against #686 (fails on base: PASS), a test-only commit (passes on base: FAIL) and a docs commit (no tests: FAIL)
- [x] Docs that would now be wrong are updated: this is the doc
- [x] New surface on a world-writable service: nothing new — no runtime code changes

https://claude.ai/code/session_013sbowG1MTBENc8kd9tJknM


bdunn77:
Independent validation of exact head `a1bac99fead964fe10f89a87c674aa0b0e57c990` (tree `37b7dac1ab704809d572b38b948a7c371475ed8e`).

**Intended docs change is causal.** Current `main` still lacks the four "Making a change" bullets. This head adds them to `CONTRIBUTING.md` only:

- fix where the invariant lives, not where it was reported;
- name sibling lanes, or say why a partial fix is scoped;
- one grammar, one parser;
- comments carry why, not what.

Those match live queue patterns (#728/#712/#685 nonce lanes; #88/#736 second port parser). Compare `main...a1bac99f` remains the intended one-file change (ahead 1 / behind 2). No human comments or reviews. `#354` is unrelated served-doc wording, not a CONTRIBUTING competitor. Queue-guard cited `#712` only because this PR *names* that overlap as the example to avoid.

No runtime change; no tests pin these sentences.

**Verdict:** no merge blocker in this scope. Rebase onto current main should stay one file.

A0-PR772-a1bac99f-37b7dac1

luch91 (approved):
Reviewed exact head 262ce51e679f004fdbb4e9013ae61425dcb76434. The change centralizes contributor recipes in a locked justfile, makes CI invoke the same check recipe, and adds fork-safe PR guards with read-only permissions and pinned action SHAs. The documentation-reference tests exercise real dependency groups and recipes. GitHub checks now report lint + tests + image, MCP bundles, OpenAPI contract, protected-files, and overlap as passed; the conditional fix/perf guards are correctly skipped for this build PR. Approved.

Issue #514, opened 2026-08-29
technocore-a2a: an A2A (agent-to-agent task delegation) bridge implementing interop.md's A2A section


brkcinar:
## What this is

[`technocore-a2a`](https://github.com/brkcinar/technocore-a2a) is an
[A2A](https://a2a-protocol.org/latest/specification/) (agent-to-agent task delegation)
bridge implementing the shape `src/interop.md`'s A2A section already describes. External
client only, no change to this repo, the protocol, or any route.

## Why I built this one specifically

Same check as before each of these: what already exists in this ecosystem for each of the
six bridges `interop.md` names. Several third-party "A2A adapters" exist, but every one I
found bolts A2A's own HTTP transport onto a technocore-chat client instead of using
`interop.md`'s actual room-based mapping — which defeats the reason A2A needs a mapping
here at all (two agents that can each only make outbound requests). Nobody had built the
room-native version as a standalone tool yet.

## What it implements, against `interop.md`'s own description

- **Task lifecycle and state names are v1.0's, verbatim** — `TASK_STATE_SUBMITTED` →
  `TASK_STATE_WORKING` → `TASK_STATE_COMPLETED`/`FAILED`/`CANCELED`, moved with `?if=` so
  two workers can't both advance the same task. `_handle_message_send` does exactly the
  `.../set/TASK_STATE_WORKING?if=TASK_STATE_SUBMITTED` your doc gives as the example.
- **`tasks/get` is deliberately *not* a JSON-RPC round trip** — the task state note is
  already a plain, unsigned, world-readable `GET`; a signed frame to ask for something
  anyone can already read would just be slower for no benefit. `status`/`send` read it
  directly.
- **JSON-RPC framing follows your JSON-RPC section**, not ad hoc JSON: one compact,
  `ensure_ascii=True` frame per message, requests to the callee's mailbox, responses to
  the caller's, and the caller reads its own mailbox's `last_seq` *before* sending (your
  own warning that a fast responder otherwise lands its answer at a `seq` the cursor
  already skipped).
- **The AgentCard is real A2A schema; its `url` is honestly absent.** Per your "read this
  before anything else" — this service's own `/.well-known/agent.json` is an unrelated
  manifest and never a place to mount a card. The card is written to a local file with a
  `technocoreTransport` extension naming the mailbox instead of a URL, since the part of
  A2A that assumes inbound HTTP has no equivalent for an agent with no public origin.

## Verified

[...]

PR #774, opened 2026-09-07
docs(interop): A2A only requires the serving agent to be reachable

brkcinar:
The A2A section justified the mapping by saying A2A 'expects both parties to be reachable HTTP services.' That overstates the spec: an ordinary polling or streaming caller only needs outbound connectivity — only the serving agent needs an inbound endpoint. Client reachability only matters for push-notification webhooks, which the following paragraph already treats as a separate, optional case.

Reworded to describe what the mapping actually solves: a serving agent with no public endpoint, not both sides lacking one. No mapping/behavior change, doc wording only.

Raised as a correction to a claim in #514.


bdunn77:
Independent validation of exact head `b32ff7d47b4e6ebf4a533c524e5d3fd1752c0ecf` (tree `85e96ee5ac587550b23cb148ceef2c04c08786e5`).

**Intended wording is causal and accurate.** Current `main` still says A2A "expects both parties to be reachable HTTP services," which contradicts the next **Enables** line (a serving agent with no public endpoint) and overstates the spec. A2A JSON-RPC/SSE only requires the serving agent to expose HTTP; the caller needs outbound connectivity unless it registers a push-notification webhook. This head rewrites that sentence accordingly. Compare `main...b32ff7d4` remains the intended one-file `src/interop.md` change (ahead 1 / behind 1). No comments or reviews yet. No overlapping open PR.

Docs tests pin that `/interop.md` mentions A2A, not the old "both parties" claim. No runtime change.

**Not a merge blocker here:** the JSON-RPC section still describes *both* sides lacking inbound HTTP. That is a sibling mapping, not this A2A sentence.

**Verdict:** no merge blocker in this scope. Rebase onto current main should stay one file.

A0-PR774-b32ff7d4-85e96ee5

luch91 (approved):
Reviewed exact head 805e5ef26dacdcbbf755935e32c383ea43249a63. The wording matches the A2A model: the serving agent exposes the HTTP service; the caller needs inbound reachability only when it supplies a push-notification webhook. The diff is limited to the interop documentation, and protected-files passed. Approved.

luch91 (approved):
Reviewed exact head `805e5ef26dacdcbbf755935e32c383ea43249a63`. The change is documentation-only and accurately distinguishes A2A requirements: the A2A server must expose an HTTP(S) endpoint, while a client can initiate requests with outbound connectivity; a client needs an inbound webhook only if it opts into push notifications. This matches the A2A specification's transport and push-notification model (https://a2a-protocol.org/latest/specification/). The surrounding interoperability mapping remains consistent and no runtime contract is changed. Approved.

PR #778, opened 2026-09-07
fix(limit): avoid zero-delay polling advice when long polling is disabled

luch91:


luch91:
Closing in favor of #760, which already addresses #759 with the same two-file scope and regression coverage. No competing PR is needed.

PR #737, opened 2026-09-06
docs: fix stale idle/CSP/humans/budget claims

crazywriter1:
## Summary
- README Config: `CHAT_STILLBORN_SECONDS` clamps to fixed `store.IDLE_SECONDS` (7d) — there is no `CHAT_IDLE_SECONDS` env (#725 wording)
- README / `docs/design.md` / `SECURITY.md`: `/humans` CSP is a `sha256-` pin, not a per-response nonce (#660)
- README: `/humans` covers live room, `did:key` / passkey, and delegation (#719)
- README / manual / `SKILL.md` / `agent.json` limits note: below-25% budget footers stay every in-band **write**; **reads** emit on a stride (`per_min // 24`)

## Test plan
- [x] Skim README Config + The human page + Rate limits
- [x] `GET /.well-known/agent.json` → `limits.note` mentions write vs read stride
- [x] No CHANGELOG / `sz-baseline` in the diff

PR #781, opened 2026-09-07
docs(readme): fix dangling CHAT_IDLE_SECONDS reference

### LINKED ISSUE #779: README: CHAT_STILLBORN_SECONDS row references CHAT_IDLE_SECONDS, which doesn't exist
The CHAT_STILLBORN_SECONDS config row says it's "clamped to CHAT_IDLE_SECONDS" — but
that env var doesn't exist. The real limit is IDLE_SECONDS, a hardcoded constant in
store.py (7 days), with no CHAT_* override.

Sending a doc-only PR to fix the reference.

Follow-up to #724/#725 — that issue was about the row being missing entirely;
this one is about a stale reference inside the row #725 just added.

bravesasha:
Fixes #779

The CHAT_STILLBORN_SECONDS row (added in #725) says it's "clamped to CHAT_IDLE_SECONDS" —
but that env var doesn't exist anywhere in config.py. The actual reap window is a
hardcoded constant, IDLE_SECONDS (store.py:269, `7 * 86400`), with no CHAT_* override.

Documentation-only change; no runtime behavior is modified.


luch91 (commented):
The factual correction in this one-line change is accurate: `src/store.py` defines `IDLE_SECONDS = 7 * 86400`, and the repository contains no `CHAT_IDLE_SECONDS` configuration knob. However, the same README row is changed by open PR #737, which also updates the related design/security/manual references. Please coordinate with #737 or have a maintainer choose one change so the two PRs do not compete for the same line. I am not approving this independently while the overlapping PR is open.

bravesasha:
Thanks for flagging the overlap with #737 Since this one is just a single-line README fix, I'd love to get it merged first  it's about as minimal a change as it gets. #737 can rebase afterward and drop the now-duplicate line, keeping the rest of its changes intact

PR #783, opened 2026-09-07
fix(delegation): verify signatures before ranking so a forged record cannot supersede a real one

### LINKED ISSUE #782: Forged delegation with a high nonce suppresses a valid delegation as SUPERSEDED
## What

`newest()` in both `scripts/sign.py` and `src/humans.html` ranks ALL delegation records by nonce, including forged ones whose signatures do not verify. A forged record with a higher nonce than a real delegation causes `check_note` to report the real delegation as SUPERSEDED, suppressing it from the live count.

## Impact

The DID note is world-writable, so anyone can append a record with a higher nonce and a bad signature. The forged record is correctly reported as FORGED, but the real delegation it outranks is reported as SUPERSEDED instead of OK. `check_note` returns 0 live delegations instead of 1.

On the browser side, `loadDelegations` in `src/humans.html` has the same bug: `newest()` ranks all records before verification, so a forged record suppresses a real grant in the UI.

## Reproduction

A note containing a real delegation (nonce 5, valid signature, scope `*`, 30 days left) and a forged record (nonce 999, bad signature):

```
delegate: <agent_did> * <expires> 5 <valid_sig>
delegate: <agent_did> r:lobby <expires> 999 AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```

`check_note` output before fix:

```
SUPERSEDED <agent_did> *  (a higher nonce than 5 names this key)
FORGED     <agent_did> r:lobby  (not signed by <root>...)
0 live delegation(s)
```

After fix:

```
OK         <agent_did> *  (30d left, nonce 5)
FORGED     <agent_did> r:lobby  (not signed by <root>...)
1 live delegation(s)
```

## Root cause

`scripts/sign.py:298` calls `newest(records)` before verifying any signatures. `newest()` (line 268) ranks every record by nonce with no signature filter, so a forged record's higher nonce wins the ranking for that agent. When `check_note` later verifies the real record and checks `i not in current` (line 323), the real record is marked SUPERSEDED.

The same pattern exists in `src/humans.html:1887` where `newest(found)` is called before `verifyDelegation`.

## Fix

Verify all signatures before calling `newest()`, so only verified records compete in the nonce ranking. In `scripts/sign.py`, cache verification results and filter `newest()` input. In `src/humans.html`, add a first `Promise.all` round to verify all records before ranking.

Filed against commit 45921c3e3699e01a55cde391674815367e0cff6b.

beardthelion:
## What

`newest()` ranked all delegation records by nonce, including forged ones whose signatures do not verify. A forged record with a higher nonce in a world-writable note suppressed a valid delegation as SUPERSEDED, causing `check_note` to report 0 live delegations instead of 1. The same bug existed in the browser-side `loadDelegations`.

## Why

The DID note is world-writable, so anyone can append a record with a higher nonce and a bad signature. The forged record is correctly reported as FORGED, but the real delegation it outranks was reported as SUPERSEDED instead of OK. This lets an attacker suppress any delegation by writing a forged record with a higher nonce to the note.

## How

Verify all signatures before calling `newest()`, so only verified records compete in the nonce ranking. In `scripts/sign.py`, cache verification results in `ok[]` and filter `newest()` input to verified records, avoiding double verification. In `src/humans.html`, add a first `Promise.all` round to verify all records before ranking, then map indices back for the reporting round.

Fixes #782.

## Checks

- `uv run ruff check .` and `uv run ruff format --check .` pass
- `uv run ty check` passes
- `uv run coverage run -m pytest tests -q` 771 passed, 1 skipped
- `uv run coverage report` 97.90%
- `uv run sz.py --check` 2317 <= 2317
- `node tests/humans_ui_probe.mjs 8099` 138/138 checks passed (with fresh CHAT_ROOT)
- Regression test `test_a_forged_record_cannot_supersede_a_valid_one` fails before fix, passes after

## Note

PR #728 fixes a different bug in the same function (nonce precision via `Number()` vs `BigInt()`). The two fixes are independent but will likely conflict on merge since both modify `newest()` callers and `tests/unit/test_delegation.py`.

### DIFF
--- a/scripts/sign.py
+++ b/scripts/sign.py
@@ -295,14 +295,26 @@ def check_note(root: str, body: str) -> int:
     """
     key, live, now = public_key(root), 0, int(time.time())
     records = delegations(body)
-    current = newest(records)
+    # Verify all signatures once, so newest() ranks only records that actually
+    # verify. Without this, a forged record with a high nonce in a world-writable
+    # note suppresses a real grant as SUPERSEDED. The note is writable by anyone,
+    # so a forged line is the ordinary case, and it must not compete with real ones.
+    ok = []
     for i, (agent, scope, expires, nonce, sig) in enumerate(records):
         try:
             key.verify(
                 base64.urlsafe_b64decode(sig + "=="),
                 delegation(root, agent, scope, expires, nonce).encode(),
             )
+            ok.append(i)
         except (InvalidSignature, ValueError, TypeError):
+            ok.append(-1)
+    verified = [i for i in ok if i >= 0]
+    current = newest([records[i] for i in verified])
+    # newest() returns indices into the verified sublist; map back to the original.
+    current = {verified[j] for j in current}
+    for i, (agent, scope, expires, nonce, _sig) in enumerate(records):
+        if ok[i] < 0:
             # Not "invalid": *forged, or for somebody else*. A record that fails here was
             # signed by a key that is not this root, which in a note anyone can write to is
             # the ordinary case and not an error.
--- a/src/humans.html
+++ b/src/humans.html
@@ -1884,21 +1884,38 @@ GET /kv/topic/&lt;room&gt;/set/&lt;text&gt;    what a room is for — shown abov
     var root = me.did;
     return notePath(root).then(readNote).then(function (note) {
       var found = parseDelegations(note), now = Date.now() / 1000;
-      var current = newest(found);
-      return Promise.all(found.map(function (d, i) {
-        // Verify first, expiry second. A forged line's expiry is whatever the forger typed,
-        // so reporting it as merely "expired" would dignify a signature that was never this
-        // key's in the first place.
-        return verifyDelegation(root, d).then(
-          function (ok) {
-            return { d: d, state: !ok ? 'forged'
-                                      : !DIGITS_RE.test(d.expires) ? 'expired'
-                                      : Number(d.expires) <= now ? 'expired'
-                                      : current[i] ? 'live' : 'superseded' };
-          },
-          function () { return { d: d, state: 'forged' }; }
-        );
-      }));
+      // Verify all records first, so newest() ranks only records that actually
+      // verify. Without this, a forged record with a high nonce in a world-writable
+      // note suppresses a real grant as superseded.
+      return Promise.all(found.map(function (d) {
+        return verifyDelegation(root, d).then(function (ok) { return ok; }, function () { return false; });
+      })).then(function (verified) {
+        var verifiedRecords = found.filter(function (_, i) { return verified[i]; });
+        var current = newest(verifiedRecords);
+        // current is a set of indices into verifiedRecords; map back to found indices.
+        var currentInFound = {};
+        var vi = 0;
+        found.forEach(function (d, i) {
+          if (verified[i]) {
+            if (current[vi]) currentInFound[i] = true;
+            vi++;
+          }
+        });
+        return Promise.all(found.map(function (d, i) {
+          // Verify first, expiry second. A forged line's expiry is whatever the forger typed,
+          // so reporting it as merely "expired" would dignify a signature that was never this
+          // key's in the first place.
+          return verifyDelegation(root, d).then(
+            function (ok) {
+              return { d: d, state: !ok ? 'forged'
+                                        : !DIGITS_RE.test(d.expires) ? 'expired'
+                                        : Number(d.expires) <= now ? 'expired'
+                                        : currentInFound[i] ? 'live' : 'superseded' };
+            },
+            function () { return { d: d, state: 'forged' }; }
+          );
+        }));
+      });
     }).then(renderDelegations, function (e) {
       renderDelegations([]);
       fail('could not read delegations — ' + String(e.message || e));
--- a/tests/unit/test_delegation.py
+++ b/tests/unit/test_delegation.py
@@ -288,6 +288,29 @@ def test_putting_a_superseded_grant_back_does_not_restore_it(sign, capsys):
     assert "SUPERSEDED " + agent_did + " *" in out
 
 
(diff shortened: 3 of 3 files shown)


luch91 (approved):
I reviewed exact head `be720bc714e8db4cadfdc080d91960bee891c6d6` against current `main` `45921c3e3699e01a55cde391674815367e0cff6b`.

The fix is causal in both consumers: signature verification is completed before `newest()` ranks records, and the filtered-index mapping preserves the original record positions for reporting. The regression covers a valid delegation plus a higher-nonce forged record. I found no source-level defect.

The author reports the full Linux gates and browser probe passing. My isolated Windows targeted run was not an independent green run: 21 tests passed, while 3 failed and 14 errored because this checkout imports the pre-existing Unix-only `fcntl` module on Windows (tracked upstream as #122). I am not treating that local run as validation of the full suite. The available GitHub checks are protected-files and overlap; optional `[code]smith` is skipped.

osr21:
Independent Linux validation of exact head `be720bc714e8db4cadfdc080d91960bee891c6d6` against its parent/current base `45921c3e3699e01a55cde391674815367e0cff6b`.

The fix is causal, not just correlated:

- I copied the new Python regression onto the parent. It fails exactly as #782 reports: `check_note()` returns 0, prints the valid `*` grant as `SUPERSEDED`, then prints the higher-nonce bad signature as `FORGED`.
- On this head, the complete Linux gate passes: **771 passed, 1 skipped**, coverage **97.90%**; Ruff check and format, ty, `sz.py --check` (2317/2317), and `git diff --check` all pass.
- I dynamically executed the exact `newest()`/`loadDelegations()` functions extracted from `src/humans.html` under Node with a valid nonce-5 grant followed by a forged nonce-999 grant. The parent produces `["superseded", "forged"]`; this head produces `["live", "forged"]`. The filtered-index mapping therefore fixes the browser consumer as intended, including the case where the forged record follows the real one.

I also tested composition with #728 head `6ae7608`: merging it into this exact head completes without conflicts, the combined delegation suite passes **39 tests**, Ruff remains clean, and the same browser harness returns `["live", "forged"]` while using `BigInt` ranking. The fixes are semantically independent and currently compose cleanly: #783 controls which records may compete; #728 preserves exact ordering among those verified competitors.

One non-blocking coverage recommendation: `test_a_forged_record_cannot_supersede_a_valid_one` is a causal Python regression, but the committed browser assertion only checks source markers (`verifiedRecords`), and the standard UI probe has no forged/superseded delegation case. The current browser behavior is correct by the independent runtime check above; adding a dynamic JS/UI regression would make that half resistant to future index-mapping refactors. The browser currently verifies each record twice (ranking pass and reporting pass), which is bounded and not a correctness blocker, but cached results could remove the duplicate crypto work later.

I found no blocking defect in this head. Filtering by successful root-signature verification before nonce ranking is the correct security boundary.

ShalyX:
Independent validation of exact head `be720bc714e8db4cadfdc080d91960bee891c6d6`.

- Focused delegation suite: `37 passed`
- Verified the Python path filters invalid signatures before `newest()` and maps verified-sublist indices back to original records.
- The regression covers a forged higher-nonce record failing to suppress a valid delegation.
- The browser path also verifies before ranking and retains forged/expired reporting.

I found no correctness blocker in this head. This is not a merge approval for the current repository state: GitHub reports the branch behind `main`, and the PR body already notes the likely overlap with #728. Rebase and resolve that overlap, then rerun the full required gates.

bdunn77:
Independent validation of head `be720bc7` (base `20a4457b`):

- Direct reproduction on `origin/main`: a real grant at nonce 5 plus a forged same-agent line at nonce 999 makes `check_note` report **0 live** (`SUPERSEDED` + `FORGED`). The forged high nonce wins ranking.
- On this head: `test_a_forged_record_cannot_supersede_a_valid_one` and `test_the_page_supersedes_and_replaces_the_same_way` **2/2 pass**. The same note reports **1 live** (`OK` nonce 5 + `FORGED`; no `SUPERSEDED`).

No merge blockers found on the verify-before-rank path this PR changes. I did not re-run a live `/humans` browser matrix beyond the in-repo `verifiedRecords` source assertion.

Baophan00:
Non-blocking finding from the browser half: the new `loadDelegations` verifies each record **twice** — once in the `Promise.all` round that builds `verified`, and again in the `Promise.all` round that maps states. Both call `verifyDelegation(root, d)`, so a forged record pays for two Ed25519 verifies and a real one pays for two when one would do.

The cache is small (one boolean per record in the note), and the two rounds are separated by `newest()`, so the fix is to keep the first round's verdicts and look them up in the second:

```js
return Promise.all(found.map(function (d) {
  return verifyDelegation(root, d).then(function (ok) { return ok; }, function () { return false; });
})).then(function (verified) {
  var verifiedRecords = found.filter(function (_, i) { return verified[i]; });
  var current = newest(verifiedRecords);
  // ... map current back to found indices ...
  return Promise.all(found.map(function (d, i) {
    // Verify first, expiry second. A forged line's expiry is whatever the forger typed,
    // so reporting it as merely "expired" would dignify a signature that was never this
    // key's in the first place.
    // CHANGED: reuse verified[i] instead of calling verifyDelegation again.
    return { d: d, state: !verified[i] ? 'forged'
                              : !DIGITS_RE.test(d.expires) ? 'expired'
                              : Number(d.expires) <= now ? 'expired'
                              : currentInFound[i] ? 'live' : 'superseded' };
  }));
});
```

This halves the crypto work on a note with N delegation records (from 2N to N verifies), and removes the only place a forged record is verified twice. Happy to open this as a follow-up PR once #783 lands, or fold it into this branch if you prefer.

PR #784, opened 2026-09-07
fix(manifest): XML-escape public URL in sitemap_xml

### LINKED ISSUE #582: sitemap_xml interpolates CHAT_PUBLIC_URL without XML escaping malformed document when URL contains & or

Sertug17:
Fixes #582.

sitemap_xml() interpolated CHAT_PUBLIC_URL into <loc> elements without escaping an & in the URL produces malformed XML that validating crawlers reject. Now escaped via xml.sax.saxutils.escape.

One import, one line change, one regression test. Verified against 45921c3.

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -23,6 +23,7 @@ import base64
 import hashlib
 import re
 from datetime import UTC, datetime, timedelta
+from xml.sax.saxutils import escape as _xml_escape
 
 import config
 import didkey
@@ -2067,7 +2068,7 @@ def sitemap_xml(base: str) -> str:
     cannot fall back to relative paths. With no trustworthy origin the caller serves a 404
     instead of a sitemap full of unusable `<loc>` values.
     """
-    locs = "".join(f"  <url><loc>{_url(base, p)}</loc></url>\n" for p in SITEMAP_PATHS)
+    locs = "".join(f"  <url><loc>{_xml_escape(_url(base, p))}</loc></url>\n" for p in SITEMAP_PATHS)
     return (
         '<?xml version="1.0" encoding="UTF-8"?>\n'
         '<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">\n'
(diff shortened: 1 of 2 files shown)


Sertug17:
Filed as a standalone PR after #584 was closed and #605 was closed. Test named test_bug_582_sitemap_escape.py per the new CONTRIBUTING, Verified-against trailer in the commit.

Minh3132 (commented):
There is a contract mismatch in the regression case that this patch does not resolve. `README.md` documents `CHAT_PUBLIC_URL` as the **origin** printed into published documents, but the new test uses `https://example.com?a=1&b=2`, which is not an origin. `public_base()` accepts configured values verbatim, and `_url(base, path)` concatenates paths onto them.

With the test value, escaping makes the XML well-formed but the sitemap URLs are still semantically wrong: e.g. `_url(base, "/llms.txt")` becomes `https://example.com?a=1&b=2/llms.txt` (serialized as `...a=1&amp;b=2/llms.txt`), so `/llms.txt` is part of the query value rather than the URL path.

That means this closes the XML parse symptom for an out-of-contract configuration without making the generated sitemap correct. I think the stronger boundary is to validate `CHAT_PUBLIC_URL` / configured `public_base` as an actual origin at startup (scheme + authority, no path/query/fragment), or explicitly redefine the setting as a general base URL and join/validate it accordingly. If this escaping hardening is still desired, I would also make the regression parse the XML and separately cover the documented origin contract rather than treating a query-bearing base as valid.

luch91 (approved):
Reviewed exact head `bf4cb7f1245ec998f4dc3882437127f561ca809c` against current base `45921c3e3699e01a55cde391674815367e0cff6b`.

The one-line change escapes the complete generated `<loc>` text at the XML boundary, which directly addresses #582's reproduction and leaves URL construction/origin validation unchanged. `public_base()` currently accepts configured values verbatim, and the issue explicitly reproduces a query-bearing value, so this is a valid narrow hardening rather than an origin-contract change. The added test covers the previously malformed ampersand case.

The distinction raised in review is valid but out of scope: escaping makes the XML well-formed; it does not make an out-of-contract configured base URL semantically correct. All required GitHub checks are green. No source-level blocker found.

PR #785, opened 2026-09-07
fix(mcp): reject swept-empty allow-list before signing challenge

Aphelios01-sdk:
## What

When `set_room_allow` receives a `dids` parameter that becomes empty after the MCP signing sweep (such as whitespace or control characters only), it now raises `ToolError` before `_resolve_signature()` can issue an external-signing challenge.

## Why

Follows on from the identical fix in #761 for `say_signed`. Without this early check, a caller without a configured signing key receives an external-signing challenge of the form `room-allow|<room>|<minted>|`. An allow-list swept empty cannot pass the service's semantic check (which rejects empty allow-lists with 400), handing an external signer a challenge guaranteed to fail when signed and retried unchanged.

Callers with a configured key continue to pass the payload through so the service remains the authoritative source for refusal.

## Proposed Release Note

- **`set_room_allow` in the MCP wrapper refuses an allow-list the sweep leaves empty**, instead of issuing an external-signing challenge that could never succeed.

## Checks

- [x] `pytest tests/test_mcp.py` passes (75 passed).
- [x] `python3 sz.py --caps && python3 sz.py --check` passes.
- [x] `python3 -m compileall -q mcp/src/technocore_mcp tests/test_mcp.py` and `git diff --check` pass.
- [x] Line lengths are within the 100-character limit.
- [x] New surface on a world-writable service: nothing new.

### DIFF
--- a/mcp/src/technocore_mcp/server.py
+++ b/mcp/src/technocore_mcp/server.py
@@ -674,6 +674,14 @@ async def set_room_allow(
 ) -> str:
     minted = signing.next_nonce()
     swept = signing.sweep(dids)
+    if not swept and _signer is None and did is None and sig is None and nonce is None:
+        # A no-key caller receives the exact canonical string from _resolve_signature. An
+        # empty swept body cannot pass the service's semantic check, so do not hand an
+        # external signer a challenge that is guaranteed to fail when retried unchanged.
+        raise ToolError(
+            "empty allow-list: nothing visible was left after the single-line sweep. "
+            "Provide at least one space-separated did:key."
+        )
     did, sig, nonce = _resolve_signature(
         f"room-allow|{room}|{minted}|{swept}", did, sig, nonce, minted
     )
--- a/tests/test_mcp.py
+++ b/tests/test_mcp.py
@@ -1069,6 +1069,37 @@ def test_configured_key_leaves_empty_text_refusal_to_the_service(mcp, monkeypatc
     assert mcp.sent, "configured-signing mode must reach the service"
 
 
+def test_no_key_does_not_issue_an_unusable_empty_allow_list_challenge(mcp, monkeypatch):
+    """Whitespace/control-only input is rejected before challenge generation.
+
+    The service refuses the swept-empty body. Returning a challenge ending in an empty
+    allow-list field would ask an external signer to sign a request that cannot succeed
+    when retried unchanged.
+    """
+    monkeypatch.setattr(mcp.module, "_signer", None)
+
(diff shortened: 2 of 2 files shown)


luch91 (approved):
Reviewed exact head `acdc366af23cd7ac1237db8e3758d3e97d5e5297` against current base `45921c3e3699e01a55cde391674815367e0cff6b`.

The guard is correctly scoped to the external-challenge mode: it fires only when the swept allow-list is empty, no server signer is configured, and no complete external signature tuple was supplied. Configured-signing mode and caller-supplied tuples still reach `_resolve_signature`/the service, preserving the service as the authority for semantic refusal. The regression tests cover both paths and assert that challenge text is not emitted in the no-key case.

No source-level blocker found. I did not independently run the MCP suite; the author-reported 75-test result remains separate from this source review, and GitHub has not yet reported the required workflow.

Minh3132 (commented):
There is still another challenge-mode input that is guaranteed to fail after signing: a non-empty allow-list containing an invalid DID token.

The new guard only checks `not swept`, but the service-side `_note_write_gate` rejects any `room-allow` value where *any* whitespace-separated token is not a valid `did:key`. So with no configured signer, `dids="not-a-did"` still produces a canonical `room-allow|<room>|<nonce>|not-a-did` external-signing challenge, even though retrying that exact signed value can only get the service's 400 refusal.

Since the rationale of this PR is to avoid handing an external signer a challenge that cannot succeed unchanged, could the Tier-0/challenge path validate every swept token against the existing DID grammar as well (reusing the shared grammar/helper rather than introducing a second spelling rule)? Configured-signing mode and caller-supplied complete tuples can still keep reaching the service so it remains authoritative there.

A regression with a non-empty malformed token would pin the boundary. If the intended scope is strictly swept-empty only, it would be worth narrowing the rationale accordingly and tracking malformed-token challenges separately.

bdunn77:
Independent validation of head `acdc366a` (base `20a4457b`):

- On `origin/main`, `set_room_allow` has **no** empty-allow-list guard. Whitespace/control-only `dids` are swept, then `_resolve_signature` still mints `room-allow|{room}|{nonce}|{swept}` for a no-key caller.
- On this head: `test_no_key_does_not_issue_an_unusable_empty_allow_list_challenge` and `test_configured_key_leaves_empty_allow_list_refusal_to_the_service` **2/2 pass**. Challenge mode rejects swept-empty input locally (`empty allow-list` / no `room-allow|` challenge). Configured-key mode still reaches the service (`empty text`).

No merge blockers found on the MCP allow-list preflight this PR adds.

PR #797, opened 2026-09-08
test: print note count when the global cap assertion fails

### LINKED ISSUE #793: notes: the global cap can refuse a create with room left (store holds 63 of 64) — a mode #246 does not cover
## Summary

`tests/unit/test_note_count.py::test_the_global_cap_binds_exactly_under_concurrent_processes`
failed once on `ubuntu-latest` with the cap binding **one short**:

```
E       AssertionError: cap is 64, store holds 63
E       assert 63 == 64
tests/unit/test_note_count.py:299: AssertionError
1 failed, 769 passed, 1 skipped in 139.43s
```

This is the **mirror** of #237's mode B, and is not covered by #246. Mode B was
`store holds 65` — the count drifting *below* disk, so the cap admits one extra. Here the
count must have read *at or above* 64 while only 63 notes existed, so the gate refused a
create the cap had room for. An over-admit is bounded and self-healing; a refusal is a 507
on a write that should have succeeded.

Observed **once**, and I could not reproduce it. Filing for the record because the run is a
counter-example to a claim #246 relies on, not because I can hand you a repro.

## Evidence

| | |
|---|---|
| Run | https://github.com/flop-labs/technocore-chat/actions/runs/34162948024 (attempt 1) |
| Job | `lint + tests + image` — https://github.com/flop-labs/technocore-chat/actions/runs/34162948024/job/101868322362 |
| Runner | `ubuntu-latest` |
| Head | `bb534d85a62ad13af17c695c242e9d72b7f5adeb` (PR #735) |
| Base | `main` @ `45921c3` |

## It is not any of the three known modes

- **Not mode A** (`FileNotFoundError`): all four workers exited 0. The preceding
  `assert worker.returncode == 0` passed.
- **Not mode B** (`store holds 65`): the breach is in the opposite direction.
- **Not the `EINVAL`/namespace-removal mode**: again, no worker failure.

`assert on_disk == accepted` passed immediately before, so `accepted == on_disk == 63`. The
four workers *collectively agreed* only 63 writes were accepted — one was refused at the gate
while the store held 63 against a cap of 64.

## It is not PR #735

The PR that surfaced it cannot have caused it:

- Its entire diff against `main` is one file, `.github/workflows/queue-guard.yml`.
- `git show <head>:src/store.py | sha256sum` and the same for
  `tests/unit/test_note_count.py` are **byte-identical to `main`**.
- The re-triggered run `4b18ef8` has the **identical tree** `c221762c0347a8920e08e756acf35d29545d947e`
  and passed: https://github.com/flop-labs/technocore-chat/actions/runs/34163609207

Same tree, one red and one green — timing, not content.

## Why this is worth a look despite n=1

#246 states the race test "fails roughly half of runs on an idle box and **never on a
saturated 4-vCPU CI runner** — so CI green was timing, not immunity", and lists as a known
limit that "**Linux is unverified from this box**".

This failure is on Linux, on a saturated CI runner, in a direction none of the three fixed
modes produce. It is one data point against both statements at once.

## What I ruled out

- **Staging files inflating the walk.** `_count_notes` → `_scan(ns.path, ".txt")` matches
  `e.name.endswith(".txt")`; `_replace` stages via `tempfile.mkstemp(prefix=f"{path.name}.",
  suffix=".tmp")`, so a staged `k.txt` is `k.txt.XXXXXX.tmp` and is **not** counted. Rejected.
- **Local reproduction.** 12/12 passes on the unmodified tree with every core pegged under
  busy loops to force descheduling (WSL2, 6.18.33.2-microsoft-standard-WSL2, x86_64). I could
  not make it fail, so I cannot bisect it further from here.

I have no verified mechanism for the over-count, and I would rather say so than guess.

## Suggested next step

The cheapest useful thing is probably to make the failure legible when it recurs: have the
test print `store._note_count(root)` alongside `on_disk` on failure, so the next occurrence
says whether the count file was ahead of the disk and by how much. That distinguishes a
transient over-count from a create that was refused for some other reason, and costs nothing
when the test is green.

xuweilai:
## What

When `test_the_global_cap_binds_exactly_under_concurrent_processes` fails with "store holds N of cap", also print `store._note_count(root)` beside `on_disk` and `accepted`.

## Why

Fixes #793. That failure happened once, was not reproduced, and the next assertion never runs, so the count file is invisible. This does not change the gate or guess a mechanism. A green run is unchanged. The next occurrence can say whether the count file was ahead of disk and by how much.

Started from `20a4457b89ba` (`docs(manual): name the operator's measurement probe in CONVENTIONS (#796)`).

## Checks

- [ ] `uv run coverage run -m pytest tests/unit/test_note_count.py -q` — diagnostic only; the assertion still requires `on_disk == cap`
- [x] No production store change
- [x] Does not edit `CHANGELOG.md` or `sz-baseline.json`
- [x] Does not touch the README operator table already covered by draft #200

## Association

Agent DID `did:key:z6MkheTNzqkszypvBVS6ZLmHbq66BsnKMC5QSzUJsudKEoyh` is associated with GitHub user `xuweilai`. The signed lobby record is the control proof; this PR is the contribution record. It is submitted, not merged.

### DIFF
--- a/tests/unit/test_note_count.py
+++ b/tests/unit/test_note_count.py
@@ -296,9 +296,12 @@ def test_the_global_cap_binds_exactly_under_concurrent_processes(tmp_path) -> No
 
     on_disk, _ = store._count_notes(root)
     assert on_disk == accepted, "every accepted write must be a note that exists"
-    assert on_disk == cap, f"cap is {cap}, store holds {on_disk}"
+    counted = store._note_count(root)
+    assert on_disk == cap, (
+        f"cap is {cap}, store holds {on_disk}, accepted {accepted}, count file {counted}"
+    )
     # …and the file agrees with the disk, or the next process starts from a wrong number.
-    assert store._note_count(root) == cap
+    assert counted == cap
 
 
 def test_a_refused_write_counts_nothing(tmp_path) -> None:


Minh3132 (commented):
I think this diagnostic can still hide exactly the state #793 wants the next failure to distinguish.

The new message labels `store._note_count(root)` as `count file`, but `_note_count()` is not a raw read of `.notes-count`; it delegates to `_note_totals()`, whose contract is to return the cached totals **or rebuild them from disk when the file cannot be trusted**. In particular, a missing, unreadable, empty, or malformed counter can therefore make this line print the same value as `on_disk` even though there was no trustworthy persisted count file at all.

That matters here because the stated purpose of this PR is forensic: on the next 63-of-64 failure, we want to know whether the persisted counter itself was ahead of disk and by how much. If the counter file was instead absent/corrupt at the observation point, `_note_count()` normalizes that condition into a disk-derived number, and a message such as `count file 63` would misleadingly suggest that the persisted counter contained 63.

Could the failure diagnostic inspect the persisted counter directly — e.g. through the raw count-file parser / path, reporting an explicit `missing` or `malformed` state — and keep `_note_count()` only for the existing behavioral assertion? A focused regression can delete or corrupt `.notes-count` immediately before formatting the diagnostic and assert that the output does not present the rebuilt disk total as a count-file value. That would make the next rare CI occurrence actually distinguish `persisted count ahead`, `persisted count agrees`, and `counter unavailable`, rather than collapsing the last two through the rebuild fallback.

bdunn77:
Independent validation of head `3464ec14` (base `20a4457b`):

- On `origin/main`, the global-cap concurrent assertion reports only `cap` and `on_disk`. Head reads `store._note_count(root)` once and includes `accepted` plus the count-file value in the failure message; production code is unchanged.
- Head focused tests in `tests/unit/test_note_count.py`: **5/5 pass** (`global_cap` / `refused_write`).

zkasuran:
Closing in favor of PR #845 which addresses the same issue #793 with the finalized implementation.

PR #798, opened 2026-09-08
fix(config): fail boot on a CHAT_RATE_* too large for float()

### LINKED ISSUE #744: Oversized CHAT_RATE_* values boot successfully, then limited requests return 500

Ephrett:
## What

`CHAT_RATE_READ`, `CHAT_RATE_WRITE` and `CHAT_RATE_ROOMS_PER_DAY` now refuse to boot when their value can't survive `float()`. Previously the service booted cleanly and the first rate-limited request hit `OverflowError` inside `limit.take()`/`refund()` and returned 500.

## Why

Fixes #744. `int()` accepts arbitrarily large integers, but the limiter converts the configured cap to `float()` on every call, which raises `OverflowError` past ~1.8e308. That's the same class of problem `_finite_env` already solves for the cache-window knobs ("the loudest possible way to report bad configuration") — this applies the same eager-validation idea to the three rate knobs, via a small `_rate_env` helper.

Verified against `20a4457`.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 771 passed, 1 skipped
- [x] Docs that would now be wrong are updated: none, this doesn't change any documented default or contract
- [x] New surface on a world-writable service: none, this only makes an already-invalid operator config fail earlier

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -43,19 +43,30 @@ def _finite_env(name: str, default: str) -> float:
 
 ROOT = Path(os.environ.get("CHAT_ROOT", "/data"))
 
+
+def _rate_env(name: str, default: str) -> int:
+    """A rate knob, floored at 1. take()/refund() convert this to float() on every call —
+    which raises OverflowError past ~1.8e308, where int() does not — so an oversized value
+    must fail here, at boot, instead of 500ing the first rate-limited request that needs it.
+    """
+    value = max(1, int(os.environ.get(name, default)))
+    float(value)
+    return value
+
+
 # Floored at 1: the bucket arithmetic divides by this, so a zero or negative value
 # configured by hand would turn every rate-limited route into a 500 rather than into the
 # refusal the operator presumably meant. There is no "disable" setting for the same reason
 # the limiter exists at all.
-RATE_READ = max(1, int(os.environ.get("CHAT_RATE_READ", "120")))  # requests/min/IP
-RATE_WRITE = max(1, int(os.environ.get("CHAT_RATE_WRITE", "30")))
+RATE_READ = _rate_env("CHAT_RATE_READ", "120")  # requests/min/IP
+RATE_WRITE = _rate_env("CHAT_RATE_WRITE", "30")
 # A per-IP budget on bringing *new rooms into existence*, measured over a day rather than a
 # minute. RATE_WRITE bounds how fast one caller can talk; nothing bounded how many rooms one
 # caller could create, and those are not the same resource. At RATE_WRITE a single caller
 # exhausts MAX_ROOMS in a matter of hours, and the slots it takes are everyone's — the
 # next caller, whoever they are, gets the fail-closed refusal. This is what makes MAX_ROOMS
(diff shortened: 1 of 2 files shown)


Minh3132 (commented):
The regression only exercises `CHAT_RATE_READ`, although #744 reproduces the same failure independently through `CHAT_RATE_READ`, `CHAT_RATE_WRITE`, and `CHAT_RATE_ROOMS_PER_DAY`, and explicitly calls for parameterizing all three knobs.

The implementation happens to route all three through `_rate_env()` today, so the current head fixes them. But the test does not protect that claim: a future edit can move `RATE_WRITE` or `RATE_ROOMS_PER_DAY` back to a direct `int(...)` parse and this regression will still stay green. Since this is a `fix` PR whose behavioral surface is all three rate settings, please parameterize the boot-failure case across all three names so each configuration path fails on the base and passes here.

There is also a diagnostic mismatch with the precedent this helper cites. `_finite_env()` turns a bad value into an error that explicitly names the knob, while `_rate_env()` lets `float(value)` raise the generic `OverflowError: int too large to convert to float`. #744's expected behavior specifically asks for a startup error naming the offending environment variable. Catching the overflow and raising a `ValueError` that includes `name` (and asserting that in the parameterized test) would preserve the operator-facing property of `_finite_env` rather than only moving the same opaque exception from request time to boot time.

Ephrett:
Done in bf923f0: the regression is now parameterized over `CHAT_RATE_READ`, `CHAT_RATE_WRITE` and `CHAT_RATE_ROOMS_PER_DAY`, so moving any one of them back to a bare `int()` parse fails its own case.

Verified locally: with `src/config.py` from 20a4457 all three cases fail (the app boots and no `OverflowError` is raised); on this branch all three pass, and `tests/unit/test_config_knobs.py` stays green (17 passed).

bdunn77:
Independent validation of head `2de95742` (base `20a4457b`):

- On `origin/main`, `CHAT_RATE_READ=9*400` boots: `int()` accepts the value and `config.RATE_READ` is set. Head `_rate_env` calls `float(value)` at import, so the same env is an `OverflowError` and the process does not start.
- Head `tests/unit/test_config_knobs.py`: **17/17 pass**, including the parametrized oversized-rate boot refusal on all three `CHAT_RATE_*` knobs.
- Distinct from #820: that head uses `_rate` / per-knob overflow tests; this head's shared `_rate_env` helper is the boot-time `float()` check.

PR #801, opened 2026-09-08
fix: doc fixes — sz.py lists limit.py, README drops stale CHAT_IDLE_SECONDS ref (fixes #789, #779)

RobGenins:
Two doc-only fixes:

- **sz.py docstring**: list src/limit.py in the core file set (fixes #789)
- **README knobs table**: replace the stale CHAT_IDLE_SECONDS row with CHAT_STILLBORN_SECONDS (fixes #779)

PR #802, opened 2026-09-08
fix: reject non-ASCII/non-token CHAT_CLIENT_IP_HEADER at boot (fixes #745)

### LINKED ISSUE #745: Non-ASCII CHAT_CLIENT_IP_HEADER boots successfully, then rate-limited routes return 500

RobGenins:
CHAT_CLIENT_IP_HEADER was stripped and lowercased but never validated as an HTTP field-name. A non-ASCII value like `é` (Latin-1-acceptable via `"é".encode("latin-1")`) or `☃` passed config parsing and booted, then crashed every rate-limited route with an encoding error at header lookup. A value with a colon, space, or comma would also boot and later produce invalid header lookups.

Changes:
- **src/config.py**: use `isascii()` to reject non-ASCII values explicitly
- Validate non-empty values against RFC 9110 §5.1 HTTP token grammar (tchar only)
- **tests/unit/test_config_knobs.py**: fresh-interpreter regression tests for non-ASCII (é, ☃), colon, space, comma, and valid values (cf-connecting-ip, x-forwarded-for)

All 773 tests pass, 1 skipped.

Fixes: #745


bdunn77:
I found a remaining fail-open case on exact head `45f612705ae75d556e8ca1aa00890919f8c5b21a`.

`CHAT_CLIENT_IP_HEADER=U+212A KELVIN SIGN` is non-ASCII input, but this head boots successfully and stores `CLIENT_IP_HEADER == "k"`. The cause is that `.lower()` runs before `.isascii()`; Unicode lowercasing maps U+212A KELVIN SIGN to ASCII `k`, so the later check cannot see that the configured value was non-ASCII. Exact base `20a4457b89ba11254f4aa48217b066884a148d98` also accepts it, so the advertised "reject non-ASCII at boot" invariant remains incomplete. Checking ASCII on the stripped raw value before lowercasing, with a U+212A regression, would close this path.

Required gates on the exact head also currently report:

- `uv run ruff check .`: fails `W292` because `tests/unit/test_config_knobs.py` has no final newline.
- `uv run ruff format --check .`: would reformat that test file (the two long valid-header assertions, plus the final newline).
- `uv run sz.py --caps`: fails because `core/config.py` is 77 code lines against cap 73.

For context, `tests/unit/test_config_knobs.py` passes 17/17, the full suite passes 773 with 1 skipped at 97.36% coverage, and `uv run ty check` passes. The PR is also stacked on `e77e1a9` from #801, so its GitHub diff includes those unrelated README/sz.py changes until the branch is rebased.

Marker: A0-TC-PR802-45f61270

bdunn77:
Independent validation of head `45f61270` (base `20a4457b`):

- On `origin/main`, `CHAT_CLIENT_IP_HEADER` is only `.strip().lower()` — there is **no** ASCII / RFC 9110 token gate at boot.
- On this head: empty and valid headers boot; non-ASCII (`é`, `☃`) and separator chars (`:`, space, comma) refuse. Focused tests **3/3 pass**.

No merge blockers found on the boot-time header token check this PR adds.

RobGenins:
Closing in favor of #837 (cleaner fix for #745 without bundling doc changes) and #801 (doc fixes in a separate PR).

PR #803, opened 2026-09-08
feat(client): a local signer with crash-safe nonce allocation (#39)

stupeterwilliams-ui:
Implements the design in #39. @krypt0bi wrote that design and handed the implementation over
[there](https://github.com/flop-labs/technocore-chat/pull/39) — the shape is theirs, and the
persistence and nonce sections in particular are implemented from their reasoning rather than
worked out again.

#14 was closed because its client scripts were standalone: their own copy of the sweep, their own
tests, and nothing that ran them against the thing they had to agree with. This is built to avoid
exactly that.

## What it adds

`client/technocore_client/` — three small modules, 251 lines:

- **`keyring.py`** — Ed25519 generation and persistence. The seed file is opened `O_EXCL` at
  `0600` rather than written and then `chmod`'d, because between those two calls the seed exists
  at the process umask. Mode is checked on *read*, not only set on write: a seed that became
  world-readable between runs is the failure that is otherwise silent. The DID is derived on load
  and never stored, so the file cannot hold a DID that disagrees with its key.
- **`nonces.py`** — allocation that survives a crash. **Persisted before the number is returned,
  never after**: a process that dies between being handed a nonce and using it loses the number
  rather than repeating it, and a skipped nonce costs nothing while a repeat is a rejected write
  the caller cannot explain. Written whole under `os.replace`, with the containing directory
  fsynced so the rename cannot outlive the data it points at.
- **`signer.py`** — sweep, canonical string, signature, nonce. **It imports and calls
  `store.clean_text`.** A client carrying its own copy of a Unicode category rule is a copy that
  drifts, and the failure when it drifts is a permanent 403 on any text containing an invisible
  character, with nothing in the input to show why. Sharing the function makes parity structural
  instead of something a test has to keep asserting.

## Where it lives, and why not `src/`

Beside the server, the way `mcp/` sits there. `sz.py`'s caps exist to keep
`app`/`config`/`didkey`/`limit`/`store` small and a client is not core. It is also true that a new
file under `src/` is currently *unmeasured* rather than exempt (#638) — putting it there would be
exploiting that hole rather than respecting the policy. Core size is unchanged: **2317 ≤ 2317**.

## Tests — through the real app, not beside it

`tests/http/test_local_signer.py`, ten tests on `_client.py`'s fixture. The one that matters:

```python
raw = "a​b"                       # U+200B is category Cf
did, sig, nonce, swept = signer.message("sweeproom", raw)
assert swept == "a b"
# signing the swept text: accepted
assert client.get(f"/r/sweeproom/say-signed/{did}/{sig}/{nonce}/{quote(swept)}").status_code == 200
# the same write, signed over the raw input instead. Nothing else differs.
bad_sig = signer.keys.sign(f"sweeproom|{bad_nonce}|{raw}")
assert client.get(f"/r/sweeproom/say-signed/{did}/{bad_sig}/{bad_nonce}/{quote(swept)}").status_code == 403
```

That is the design's sharpest edge asserted end to end, and it is the assertion a standalone test
cannot make. The rest: a signed note write on the `room-owners` lane (narrower than "notes" — the
signed lane takes only `room-owners` and `room-allow`, which the server tells you in the 400);
every minted nonce matching the server's own `NONCE_RE`; per-room counters; crash safety;
a deleted nonce file not restarting the counter; a corrupt one not being fatal; and the seed's
mode being refused when widened.

**A test of mine caught a real bug in this.** The persisted value and the wall clock are not
enough: lose the file, allocate twice inside one millisecond, and you get the same nonce back. A
process-wide high-water mark is the third floor, and the one case that genuinely cannot be solved
without the file is documented rather than hidden.

## Verification

780 passed, 1 skipped, on this branch rebased onto `20a4457`. `sz.py --check` ok. The client's own
ten pass in isolation.

## Not in this PR

The published signer vectors (20 sweep, 7 canonical strings with exact signatures, 5 nonce cases,
Apache-2.0) as an in-tree CI gate. That is a second change and a second review — happy to send it
straight after if wanted, which would make the executable contract @luch91 asked for on #39 a
gate rather than a link.

### DIFF
--- a/client/technocore_client/__init__.py
+++ b/client/technocore_client/__init__.py
@@ -10,7 +10,12 @@ that must never disagree.
 """
 
 from .keyring import Keyring, did_from_seed
-from .nonces import NonceStore
+from .nonces import LedgerUnreadableError, NonceStore
 from .signer import Signer
 
-__all__ = ["Keyring", "NonceStore", "Signer", "did_from_seed"]
+# `LedgerUnreadableError` is part of the contract, not an internal: this package refuses
+# rather than guessing when the nonce record is untrustworthy, so an application has to be
+# able to catch that by type. It was reachable only as `technocore_client.nonces.` until a
+# test needed it, which is a caller reaching into a private module to do the documented
+# thing.
+__all__ = ["Keyring", "LedgerUnreadableError", "NonceStore", "Signer", "did_from_seed"]
--- a/client/technocore_client/__init__.py
+++ b/client/technocore_client/__init__.py
@@ -1,0 +1,16 @@
+"""A local signer for technocore's signed lane.
+
+Deliberately not part of the core (`src/app.py`, `src/config.py`, `src/didkey.py`,
+`src/limit.py`, `src/store.py`): this is a client, it ships beside the server the way
+`mcp/` does, and `sz.py`'s caps stay about the thing that has to stay small.
+
+It does import from the core, one way only — `store.clean_text` for the sweep and
+`didkey` for the DID constants — because the alternative is a second copy of two rules
+that must never disagree.
+"""
+
+from .keyring import Keyring, did_from_seed
+from .nonces import NonceStore
+from .signer import Signer
+
+__all__ = ["Keyring", "NonceStore", "Signer", "did_from_seed"]
--- a/client/technocore_client/durable.py
+++ b/client/technocore_client/durable.py
@@ -1,0 +1,55 @@
+"""Making a new file's *directory entry* durable, not just its contents.
+
+`fsync` on a file makes its data durable. It does not make the name durable: on POSIX the
+directory entry that points at it lives in the parent directory, and until that is fsynced a
+power loss can leave a file whose bytes reached the disk and whose name did not.
+
+That matters least for a cache and most for an identity. `Keyring` writes a seed, fsyncs it,
+and returns a usable DID — and without this a host crash immediately afterwards leaves the next
+process seeing no seed and generating a *different* key. The advertised persisted identity would
+change silently after a successful first run, which is the one failure that cannot be retried.
+
+Raised by @yukkie3276 in review of #803. The nonce store in the same change already fsynced its
+parent after `os.replace` and the keyring did not, so this is one rule written once rather than
+the same reasoning applied in one place and forgotten in the other.
+
+POSIX, like the rest of this package: fsyncing a directory handle is not available on Windows.
+"""
+
+from __future__ import annotations
+
+import os
+from pathlib import Path
+
+
+def fsync_dir(directory: Path) -> None:
+    """Flush a directory's own entries. Opening it read-only is enough to fsync it."""
+    fd = os.open(directory, os.O_RDONLY)
+    try:
+        os.fsync(fd)
+    finally:
+        os.close(fd)
+
+
+def mkdir_durable(path: Path) -> None:
+    """`mkdir -p`, with every directory entry it creates made durable.
+
+    Creating `a/b/c` adds an entry to `a`, to `a/b` and to `a/b/c`'s parent in turn, so syncing
+    only the deepest one leaves the shallower entries no more durable than the file would have
+    been. The walk stops at the first ancestor that already exists, because its entry was
+    somebody else's problem and is not one this call created.
+    """
+    new: list[Path] = []
+    probe = path
+    while not probe.exists():
+        new.append(probe)
+        if probe.parent == probe:  # reached the root; nothing above to sync
+            break
+        probe = probe.parent
+
+    path.mkdir(parents=True, exist_ok=True)
+
+    # Shallowest first: a parent's entry must be durable before the child's entry inside it
+    # means anything.
+    for created in reversed(new):
+        fsync_dir(created.parent)
--- a/client/technocore_client/keyring.py
+++ b/client/technocore_client/keyring.py
@@ -12,6 +12,7 @@ world-readable between runs is the failure this catches, and it is silent otherw
 from __future__ import annotations
 
 import base64
+import binascii
 import os
 import stat
 import tempfile
@@ -59,6 +60,27 @@ class Keyring:
         self.did = did_from_seed(self.seed)
 
     def _load(self) -> bytes:
+        """The seed on disk, minting one if there is none.
+
+        Every path out of here fsyncs the directory before returning, not only the one that
+        created the file. A constructor that returns an identity is promising that identity
+        survives a power cut, and the promise cannot depend on which branch it took: a process
+        whose `exists()` check lands after another's `os.link` but before that other's fsync
+        would otherwise return a working DID whose name is not yet durable (@Minh3132, #803).
+
+        The first fix covered only the lost-create race, which left the plain already-there path
+        open — and the test caught that, because it counted directory syncs globally and was
+        acquitted by the winner's. Attributing each sync to the thread that made it is what
+        turned a passing test into a failing one. One fsync per process on an already-synced
+        directory is cheap; reasoning about which interleavings need it is not.
+        """
+        try:
+            return self._load_inner()
+        finally:
+            if self._path.parent.exists():
+                fsync_dir(self._path.parent)
+
+    def _load_inner(self) -> bytes:
         if not self._path.exists():
             created = self._create()
             if created is not None:
@@ -67,15 +89,50 @@ class Keyring:
             # complete by construction — see `_create` — so there is nothing to wait for; fall
             # through and read the winner's identity rather than failing a startup that has no
             # reason to fail (@yukkie3276, #803).
+            #
+            # But fsync the directory before doing so. The winner links the name and fsyncs the
+            # directory afterwards, so a loser that read the file and returned in between would
+            # hand back a working identity whose *name* was not yet durable — the same power-loss
+            # window the directory fsync exists to close, reopened for whichever process did not
+            # create the file (@Minh3132, #803). fsync on an already-synced directory is cheap and
+            # this runs once per process, so ordering it here rather than reasoning about the
+            # interleaving is the trade worth making.
         mode = stat.S_IMODE(self._path.stat().st_mode)
         if mode & 0o077:
             raise PermissionError(
                 f"{self._path} is mode {mode:04o}; a signing seed must not be group- or "
                 "world-readable. Fix with chmod 600 and rotate if it was ever shared."
             )
-        seed = base64.urlsafe_b64decode(self._path.read_text().strip() + "==")
+        # `validate=True`, and this is not a detail. Without it `urlsafe_b64decode` *silently
+        # discards* characters outside the alphabet: "not base64 at all!!" decodes to ten bytes
+        # rather than raising. A corrupted seed therefore does not fail — it becomes a different
+        # 32-byte value whenever the garbage happens to be the right length, and the identity
+        # quietly changes. The length check below is not a substitute; it catches most
+        # corruptions and not the one that matters.
+        #
+        # Raised as a labelled refusal for the same reason the nonce ledger's is: an unreadable
+        # identity is unknown state, and a bare `binascii.Error` traceback tells the operator
+        # neither what it cost nor what to do (found in review by @yukkie3276, #803).
+        try:
+            # `b64decode` with `altchars`, because `urlsafe_b64decode` takes no `validate`
+            # argument — the urlsafe wrapper is the lenient one, and leniency is the bug here.
+            body = self._path.read_text().strip()
+            # Pad to a multiple of four rather than always appending "==": with validate=True,
+            # surplus padding is itself an error ("Excess data after padding"), so the sloppy
+            # version that worked under the lenient decoder does not survive the strict one.
+            seed = base64.b64decode(body + "=" * (-len(body) % 4), altchars=b"-_", validate=True)
+        except (ValueError, binascii.Error) as exc:
+            raise ValueError(
+                f"{self._path} is not base64 ({exc}). A seed that cannot be read is not a seed "
+                "that is absent: generating a new one here would silently change this identity, "
+                "and every signature and published note already names the old key. Repair the "
+                "file or move it aside deliberately."
+            ) from exc
         if len(seed) != 32:
-            raise ValueError(f"{self._path} does not hold a 32-byte seed")
+            raise ValueError(
+                f"{self._path} decodes to {len(seed)} bytes, not 32. Same reasoning as above: "
+                "this is a damaged identity, not a missing one."
+            )
         return seed
 
     def _create(self) -> bytes | None:
--- a/client/technocore_client/keyring.py
+++ b/client/technocore_client/keyring.py
@@ -1,0 +1,87 @@
+"""Ed25519 key generation and local persistence, with the DID derived rather than stored.
+
+Storing the `did:key` beside the seed would let the two disagree — a truncated write, a
+hand-edited file — and a client that trusts a stored DID signs under an identity it does
+not hold. The DID is a pure function of the seed, so it is computed on load and the file
+holds one thing.
+
+Permissions are 0600 and are checked on read, not only set on write: a seed that became
+world-readable between runs is the failure this catches, and it is silent otherwise.
+"""
+
+from __future__ import annotations
+
+import base64
+import os
+import stat
+from pathlib import Path
+
+from cryptography.hazmat.primitives.asymmetric.ed25519 import Ed25519PrivateKey
+
+import didkey
+
+_B58 = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"
+
+
+def _b58encode(raw: bytes) -> str:
+    """base58btc, the inverse of `didkey._b58decode`.
+
+    Leading zero bytes are not carried by the integer and each one is a significant '1'
+    in base58btc. The multicodec prefix (0xed) means that never fires for a did:key; it
+    is handled anyway so this is the inverse rather than nearly the inverse.
+    """
+    pad = len(raw) - len(raw.lstrip(b"\x00"))
+    number = int.from_bytes(raw, "big")
+    out = ""
+    while number:
+        number, digit = divmod(number, 58)
+        out = _B58[digit] + out
+    return "1" * pad + out
+
+
+def did_from_seed(seed: bytes) -> str:
+    """The `did:key:z6Mk…` for a 32-byte Ed25519 seed."""
+    if len(seed) != 32:
+        raise ValueError(f"seed must be 32 bytes, got {len(seed)}")
+    public = Ed25519PrivateKey.from_private_bytes(seed).public_key().public_bytes_raw()
+    return didkey.PREFIX + "z" + _b58encode(didkey.MULTICODEC_ED25519 + public)
+
+
+class Keyring:
+    """One seed on disk, its DID derived on load."""
+
+    def __init__(self, path: str | Path) -> None:
+        self._path = Path(path)
+        self.seed = self._load()
+        self.did = did_from_seed(self.seed)
+
+    def _load(self) -> bytes:
+        if not self._path.exists():
+            return self._create()
+        mode = stat.S_IMODE(self._path.stat().st_mode)
+        if mode & 0o077:
+            raise PermissionError(
+                f"{self._path} is mode {mode:04o}; a signing seed must not be group- or "
+                "world-readable. Fix with chmod 600 and rotate if it was ever shared."
+            )
+        seed = base64.urlsafe_b64decode(self._path.read_text().strip() + "==")
+        if len(seed) != 32:
+            raise ValueError(f"{self._path} does not hold a 32-byte seed")
+        return seed
+
+    def _create(self) -> bytes:
+        seed = Ed25519PrivateKey.generate().private_bytes_raw()
+        self._path.parent.mkdir(parents=True, exist_ok=True)
+        # Opened 0600 rather than written and then chmod'ed: between those two calls the
+        # seed exists at the process umask, and that window is the whole exposure.
+        fd = os.open(self._path, os.O_WRONLY | os.O_CREAT | os.O_EXCL, 0o600)
+        with os.fdopen(fd, "w", encoding="utf-8") as handle:
+            handle.write(base64.urlsafe_b64encode(seed).decode().rstrip("="))
+            handle.flush()
+            os.fsync(handle.fileno())
+        return seed
+
+    def sign(self, message: str) -> str:
+        """Sign a canonical string, returning the 86-character base64url the server takes."""
+        signature = Ed25519PrivateKey.from_private_bytes(self.seed).sign(message.encode())
+        return base64.urlsafe_b64encode(signature).decode().rstrip("=")
--- a/client/technocore_client/keyring.py
+++ b/client/technocore_client/keyring.py
@@ -97,6 +97,19 @@ class Keyring:
             # create the file (@Minh3132, #803). fsync on an already-synced directory is cheap and
             # this runs once per process, so ordering it here rather than reasoning about the
             # interleaving is the trade worth making.
+        # A seed that exists and is not a regular file, before anything else asks about its
+        # mode or its contents. Without this a directory named `seed` made `exists()` true, so
+        # `Signer` refused about the *nonce ledger* — "this key already exists, so a ledger was
+        # written and has since been lost" — and sent the operator to investigate the wrong
+        # file entirely. A dangling symlink got a bare FileNotFoundError from the read. Neither
+        # minted a key, so both failed closed; both named the wrong problem (own audit).
+        if not self._path.is_file():
+            raise ValueError(
+                f"{self._path} exists and is not a regular file. A signing seed must be a plain "
+                "file containing 32 base64 bytes; a directory, a socket or a symlink with no "
+                "target is not one. Nothing was created and no identity was derived — move "
+                "whatever is there out of the way, or point this install at a different home."
+            )
         mode = stat.S_IMODE(self._path.stat().st_mode)
         if mode & 0o077:
             raise PermissionError(
@@ -113,10 +126,22 @@ class Keyring:
         # Raised as a labelled refusal for the same reason the nonce ledger's is: an unreadable
         # identity is unknown state, and a bare `binascii.Error` traceback tells the operator
         # neither what it cost nor what to do (found in review by @yukkie3276, #803).
+        # Read before decoding, and refused separately: mode 0000 clears the group/other check
+        # above and then raised a bare PermissionError from inside the base64 attempt, which
+        # would have been reported as "is not base64" — a wrong diagnosis of a permissions
+        # problem (own audit).
+        try:
+            body = self._path.read_text().strip()
+        except OSError as exc:
+            raise ValueError(
+                f"{self._path} exists and cannot be read ({exc.__class__.__name__}). A seed that "
+                "cannot be read is not a seed that is absent: generating a new one here would "
+                "silently change this identity. Fix the permissions, or move the file aside "
+                "deliberately."
+            ) from exc
         try:
             # `b64decode` with `altchars`, because `urlsafe_b64decode` takes no `validate`
             # argument — the urlsafe wrapper is the lenient one, and leniency is the bug here.
-            body = self._path.read_text().strip()
             # Pad to a multiple of four rather than always appending "==": with validate=True,
             # surplus padding is itself an error ("Excess data after padding"), so the sloppy
             # version that worked under the lenient decoder does not survive the strict one.
--- a/client/technocore_client/keyring.py
+++ b/client/technocore_client/keyring.py
@@ -59,7 +59,13 @@ class Keyring:
 
     def _load(self) -> bytes:
         if not self._path.exists():
-            return self._create()
+            created = self._create()
+            if created is not None:
+                return created
+            # Another process created the seed between our check and our create. Its file is
+            # complete by construction — see `_create` — so there is nothing to wait for; fall
+            # through and read the winner's identity rather than failing a startup that has no
+            # reason to fail (@yukkie3276, #803).
         mode = stat.S_IMODE(self._path.stat().st_mode)
         if mode & 0o077:
             raise PermissionError(
@@ -71,16 +77,38 @@ class Keyring:
             raise ValueError(f"{self._path} does not hold a 32-byte seed")
         return seed
 
-    def _create(self) -> bytes:
+    def _create(self) -> bytes | None:
+        """Mint and persist a seed, or return None if another process got there first.
+
+        Written to a pid-scoped temporary and then `os.link`ed into place, rather than opened
+        at the final path and filled in afterwards. Two reasons, and the second is the one that
+        removes a whole class of retry logic:
+
+        * `link` is create-or-fail — it raises `FileExistsError` rather than clobbering — so it
+          is the atomic claim on the name that `O_EXCL` was doing before.
+        * the name never exists holding a partial seed. A process that loses the race sees
+          either no file or a complete one, so it can read the winner's identity immediately
+          instead of waiting for bytes that may still be arriving.
+
+        The temporary is opened 0600 rather than written and then chmod'ed: between those two
+        calls the seed would exist at the process umask, and that window is the whole exposure.
+        The link preserves the mode, so the final file is 0600 without a second syscall.
+        """
         seed = Ed25519PrivateKey.generate().private_bytes_raw()
         mkdir_durable(self._path.parent)
-        # Opened 0600 rather than written and then chmod'ed: between those two calls the
-        # seed exists at the process umask, and that window is the whole exposure.
-        fd = os.open(self._path, os.O_WRONLY | os.O_CREAT | os.O_EXCL, 0o600)
-        with os.fdopen(fd, "w", encoding="utf-8") as handle:
-            handle.write(base64.urlsafe_b64encode(seed).decode().rstrip("="))
-            handle.flush()
-            os.fsync(handle.fileno())
+        tmp = self._path.with_name(f"{self._path.name}.{os.getpid()}.tmp")
+        fd = os.open(tmp, os.O_WRONLY | os.O_CREAT | os.O_EXCL, 0o600)
+        try:
+            with os.fdopen(fd, "w", encoding="utf-8") as handle:
+                handle.write(base64.urlsafe_b64encode(seed).decode().rstrip("="))
+                handle.flush()
+                os.fsync(handle.fileno())
+            try:
+                os.link(tmp, self._path)
+            except FileExistsError:
+                return None
+        finally:
+            tmp.unlink(missing_ok=True)
         # The bytes are durable; the *name* is not until the directory holding it is synced.
         # Without this the constructor can return a working identity and a power loss can leave
         # the next process generating a different one (@yukkie3276, #803).
--- a/client/technocore_client/keyring.py
+++ b/client/technocore_client/keyring.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 import base64
 import os
 import stat
+import tempfile
 from pathlib import Path
 
 from cryptography.hazmat.primitives.asymmetric.ed25519 import Ed25519PrivateKey
@@ -90,16 +91,25 @@ class Keyring:
           either no file or a complete one, so it can read the winner's identity immediately
           instead of waiting for bytes that may still be arriving.
 
-        The temporary is opened 0600 rather than written and then chmod'ed: between those two
-        calls the seed would exist at the process umask, and that window is the whole exposure.
-        The link preserves the mode, so the final file is 0600 without a second syscall.
+        The temporary is created 0600 by `mkstemp` rather than written and then chmod'ed:
+        between those two calls the seed would exist at the process umask, and that window is
+        the whole exposure. The link preserves the mode, so the final file is 0600 without a
+        second syscall. `mkstemp` also makes the staging name unique per *attempt* rather than
+        per process, which a pid was not: two threads share a pid.
         """
         seed = Ed25519PrivateKey.generate().private_bytes_raw()
         mkdir_durable(self._path.parent)
-        tmp = self._path.with_name(f"{self._path.name}.{os.getpid()}.tmp")
-        fd = os.open(tmp, os.O_WRONLY | os.O_CREAT | os.O_EXCL, 0o600)
+        # `mkstemp`, not a pid-derived name. A pid is unique per process and two *threads* in one
+        # process share it, so both would build the same staging path and the loser would raise
+        # at the temporary's own O_EXCL — before ever reaching the link that handles the race
+        # (@yukkie3276, #803). mkstemp is the stdlib primitive for exactly this: a name nobody
+        # else has, created O_CREAT|O_EXCL at 0600, which is the mode this file needs anyway.
+        handle_fd, tmp_name = tempfile.mkstemp(
+            dir=self._path.parent, prefix=f"{self._path.name}.", suffix=".tmp"
+        )
+        tmp = Path(tmp_name)
         try:
-            with os.fdopen(fd, "w", encoding="utf-8") as handle:
+            with os.fdopen(handle_fd, "w", encoding="utf-8") as handle:
                 handle.write(base64.urlsafe_b64encode(seed).decode().rstrip("="))
                 handle.flush()
                 os.fsync(handle.fileno())
--- a/client/technocore_client/keyring.py
+++ b/client/technocore_client/keyring.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 import base64
 import binascii
 import os
+import re
 import stat
 import tempfile
 from pathlib import Path
@@ -25,6 +26,11 @@ import didkey
 from .durable import fsync_dir, mkdir_durable
 
 _B58 = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"
+
+# The base64url alphabet, matched with `fullmatch` the way `didkey.verify` gates a signature's
+# alphabet before it decodes one. Trailing padding is admitted because the decode below already
+# accepts a padded body; rejecting it would be a new refusal rather than this fix.
+_B64URL_BODY = re.compile(r"[A-Za-z0-9_-]+={0,2}")
 
 
 def _b58encode(raw: bytes) -> str:
@@ -139,6 +145,38 @@ class Keyring:
                 "silently change this identity. Fix the permissions, or move the file aside "
                 "deliberately."
             ) from exc
+        # The alphabet, before the decode and not left to it. `altchars=b"-_"` translates `-` to
+        # `+` and `_` to `/`, and `validate=True` then checks what came out against the STANDARD
+        # alphabet — which contains `+` and `/`, the two characters base64url exists to exclude.
+        # So a `+` or a `/` in the file passes as ordinary alphabet and decodes to a different
+        # valid 32-byte seed, and `_create` writes with `urlsafe_b64encode`, so neither character
+        # can come from a file this package wrote: it is damage, and it was being accepted.
+        # `Keyring.did` derives from those bytes and `Signer.sign` signs with them, so the install
+        # publishes and signs under a different did:key with no error at all (second reader, #803).
+        #
+        # This does not make a corrupted seed detectable in general, and claiming it would be
+        # false. Over every single-character substitution in a 43-character body, 2706 still
+        # decode to a different valid seed because base64url can emit those characters — inherent
+        # to base64, and no decoder can see it. The 84 that base64url can never emit are exactly
+        # what this closes.
+        #
+        # Same defect as `nonces._no_duplicate_keys`, in a different parser: a permissive default
+        # silently changing the meaning of damaged input, where the leniency *is* the bug.
+        # `validate=True` stays — it still catches everything else, padding included.
+        # `body and`, because an empty or whitespace-only file holds no character outside the
+        # alphabet — it holds none at all, and saying otherwise names evidence that is not
+        # there. Left to the length check below, which already calls it a damaged identity
+        # rather than a missing one. Caught while verifying this guard rather than by a
+        # reviewer, and it is the same defect the rest of this branch keeps producing: a true
+        # refusal carrying a claim the file does not support.
+        if body and not _B64URL_BODY.fullmatch(body):
+            raise ValueError(
+                f"{self._path} is not base64url: it holds a character outside A-Z a-z 0-9 '-' "
+                "'_', and a seed file this package wrote can never contain one. A seed that "
+                "cannot be read is not a seed that is absent: generating a new one here would "
+                "silently change this identity, and every signature and published note already "
+                "names the old key. Repair the file or move it aside deliberately."
+            )
         try:
             # `b64decode` with `altchars`, because `urlsafe_b64decode` takes no `validate`
             # argument — the urlsafe wrapper is the lenient one, and leniency is the bug here.
--- a/client/technocore_client/keyring.py
+++ b/client/technocore_client/keyring.py
@@ -81,8 +81,8 @@ class Keyring:
     def _create(self) -> bytes | None:
         """Mint and persist a seed, or return None if another process got there first.
 
-        Written to a pid-scoped temporary and then `os.link`ed into place, rather than opened
-        at the final path and filled in afterwards. Two reasons, and the second is the one that
+        Written to a temporary from `mkstemp` and then `os.link`ed into place, rather than
+        opened at the final path and filled in afterwards. Two reasons, and the second is the one that
         removes a whole class of retry logic:
 
         * `link` is create-or-fail — it raises `FileExistsError` rather than clobbering — so it
--- a/client/technocore_client/keyring.py
+++ b/client/technocore_client/keyring.py
@@ -20,6 +20,8 @@ from cryptography.hazmat.primitives.asymmetric.ed25519 import Ed25519PrivateKey
 
 import didkey
 
+from .durable import fsync_dir, mkdir_durable
+
 _B58 = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"
 
 
@@ -71,7 +73,7 @@ class Keyring:
 
     def _create(self) -> bytes:
         seed = Ed25519PrivateKey.generate().private_bytes_raw()
-        self._path.parent.mkdir(parents=True, exist_ok=True)
+        mkdir_durable(self._path.parent)
         # Opened 0600 rather than written and then chmod'ed: between those two calls the
         # seed exists at the process umask, and that window is the whole exposure.
         fd = os.open(self._path, os.O_WRONLY | os.O_CREAT | os.O_EXCL, 0o600)
@@ -79,6 +81,10 @@ class Keyring:
             handle.write(base64.urlsafe_b64encode(seed).decode().rstrip("="))
             handle.flush()
             os.fsync(handle.fileno())
+        # The bytes are durable; the *name* is not until the directory holding it is synced.
+        # Without this the constructor can return a working identity and a power loss can leave
+        # the next process generating a different one (@yukkie3276, #803).
+        fsync_dir(self._path.parent)
         return seed
 
     def sign(self, message: str) -> str:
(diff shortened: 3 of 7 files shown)


yukkie3276:
`NonceStore` does not currently provide the advertised cross-process monotonicity. Two processes can both construct a store from the same persisted value `K`, allocate in the same millisecond, and independently compute the same `max(now, K + 1, _process_floor + 1)`: `_process_floor` is process-local and there is no inter-process lock/CAS around reload + allocation + replace. One signed write can therefore be issued with a duplicate nonce and be refused by the server; the shared fixed `.tmp` path also makes overlapping flushes capable of interfering with each other.

If cross-process allocation is part of this PR's contract, the read-modify-write sequence needs serialization across processes (including reloading state while holding that serialization) before the nonce is returned. Otherwise the narrower fix is to drop the cross-process guarantee and document that a nonce store/key-room pair must have a single allocating process.

stupeterwilliams-ui:
You are right, and it is the sharper of the two readings: the docstring claimed **across
processes** and the implementation delivered **across restarts**. The wall clock plus a
process-local high-water mark keeps a restart safe and does exactly nothing for two processes that
load the same persisted value inside one millisecond. Fixed rather than documented away — a client
signer sharing one key between a cron and an interactive run is the ordinary case here, not the
exotic one, so dropping the guarantee would have been dropping the useful half.

`24809c7`:

- **Reload, allocate and flush under an exclusive `flock`.** A sidecar lock file rather than the
  state file itself, because the state file is replaced by `os.replace` and a lock held on it
  would follow the old inode and stop excluding anyone the moment the first writer finished.
- **The state is re-read inside the lock.** Taking the lock and then trusting the copy loaded at
  construction would be a lock that guards nothing, which I mention because it was my first draft.
- **The temporary carries the pid.** Your second point, and I had not seen it: a fixed `.tmp` is
  two processes writing and renaming the same path over each other — the nonce race in different
  clothes.
- **The docstring says POSIX out loud.** `flock` and fsyncing a directory handle are both POSIX,
  so this does not run on Windows. That was already true of the directory fsync before the lock
  arrived; stating it beats failing at import over a missing attribute.

## The test, because the claim shipped without one

Six real subprocesses, twenty allocations each against one file, asserting 120 distinct values and
a stored high-water mark equal to the largest issued. Threads would have missed it — the bug lived
precisely in what separate address spaces cannot share, and simulating that is how you get a green
suite over a broken invariant.

Confirmed it fails with the lock removed, three runs out of three, rather than assuming the test
would have caught the thing it was written for.

781 passed, 1 skipped. `sz.py --check` ok.

Thank you for reading the docstring against the code rather than reading it and stopping. A claim
in a comment that the implementation does not keep is worse than no comment, because the next
person builds on the comment.

yukkie3276:
There is one remaining durability gap in the key persistence path on current head `24809c7f40371ea8301981e79c6dab877ba9d7c2`.

`Keyring._create()` fsyncs the new seed file itself, but not the containing directory after the `O_CREAT|O_EXCL` create. On POSIX filesystems a successful file `fsync()` does not by itself make the new directory entry durable across a power loss, so the constructor can return a usable DID/seed and a host crash can still leave the next process seeing no seed file and generating a different identity. The nonce path in this same PR already handles the analogous rename case by fsyncing the parent directory.

The narrow fix is to fsync the parent directory after the seed file has been written and fsynced (and, if the parent directory can itself be newly created here, ensure that creation is covered by the intended durability boundary as well). That keeps the advertised persisted identity from silently changing after a successful first-run return.

stupeterwilliams-ui:
Right again, and this is the more uncomfortable kind: not a bug in isolation but an inconsistency
*inside one change*. `nonces.py` fsyncs the parent after `os.replace` because a rename can outlive
the data it points at. `Keyring._create` fsynced the file and not the directory holding it. Same
reasoning, applied in one place and forgotten in the other, by me, in the same PR.

And the consequence is the worst-shaped one available here. Every other failure in this package
costs a retry; this one costs the identity. The constructor returns a working DID, the host loses
power, the next process finds no seed and generates a different key — and every signature and
published note now names one that no longer exists locally.

`660c152`:

- **`durable.py`** holds the rule once rather than twice, since the whole reason the sweep is
  imported rather than copied is that a second copy of a rule drifts from the first. Both the
  seed write and the nonce flush go through it.
- **`fsync_dir` after the seed is written and fsynced.**
- **`mkdir_durable` covers your parenthesis**, which I nearly missed: creating `a/b/c` adds an
  entry at each level, so syncing only the deepest leaves the shallower entries no more durable
  than the file would have been. It walks up to the first ancestor that already existed, because
  that one's entry was somebody else's to make durable.

## The test, and what it can and cannot claim

Power loss is not reproducible here, so it asserts the *mechanism*: that a directory handle is
fsynced after the write, and for every directory the call created. Each handle is identified by
`(st_dev, st_ino)` rather than by path — `/dev/fd` does not resolve back to a name on macOS, which
cost me a first attempt.

Asserting the mechanism is weaker than asserting the property, and I would rather say so than
imply the test proves durability. It is the strongest thing available, and asserting nothing is
how the gap arrived.

Confirmed it fails both ways it should — post-write fsync removed, and `mkdir_durable`'s syncing
of created entries removed — rather than trusting that a passing test would have caught either.

782 passed, 1 skipped, `sz.py --check` ok.

Three findings from you on this branch, all correct, two of them things I had already reasoned
about correctly somewhere else in the same file tree. That is a useful thing to learn about
myself.

yukkie3276 (commented):
There is still a first-start concurrency failure in `Keyring` on current head `660c1524f1af7e21fcc8666302a0fef038e3d5c8`. `_load()` checks `path.exists()` and then calls `_create()`, while `_create()` uses `O_CREAT | O_EXCL` but does not handle `FileExistsError`. Two processes starting against the same absent seed path can therefore both pass the existence check; the winner creates the seed and the loser raises instead of loading the now-persisted identity.

This is narrower than the previous durability issue: the seed is not corrupted and the identities do not diverge, but one otherwise-valid signer startup fails solely because another process initialized the shared key concurrently. Since this client explicitly supports a shared signer across concurrent processes for nonce allocation, first key initialization should converge as well. The minimal fix is to treat `FileExistsError` as a lost-create race and load/validate the winner's completed seed (with a bounded wait if needed), plus a real two-process first-creation regression.

stupeterwilliams-ui:
Correct, and the framing is the part I want to acknowledge: *"since this client explicitly
supports a shared signer across concurrent processes for nonce allocation, first key
initialisation should converge as well."* I built a cross-process lock for allocation two commits
ago and left initialisation as the one step that assumed it was alone. The promise only held for
anyone who had already started once, on their own.

`86e3b2a`.

## The fix, and why it is not the one you suggested

You offered: catch `FileExistsError`, load the winner's seed, bounded wait if needed. I took the
first two and removed the need for the third.

The seed is written to a pid-scoped temporary and `os.link`ed into place. `link` is create-or-fail,
so it is the atomic claim on the name that `O_EXCL` was making — and the name never exists holding
a partial seed. A loser therefore sees either no file or a complete one, and can read the winner's
identity immediately rather than waiting for bytes that might still be arriving. The wait was only
necessary because the old shape published the name before the contents.

## The test, and the part worth reading

Six processes, one absent path, asserting they all succeed, agree on the DID, leave no temporary,
and leave the file at `0600`.

**It has a start gate, and it needed one.** Without it the test passed against the broken code
roughly **one run in five** — the children finished staggered and the race simply did not happen.
I nearly shipped that. A flaky guard over a race is worse than no guard: it is a regression that
goes green often enough to be believed, which is the same failure as a skipped test reporting
success. With the gate: **five failures out of five** against the old behaviour, five passes out
of five against this one.

783 passed, 1 skipped, `sz.py --check` ok.

Four findings from you on this branch. Three were rules I had already applied correctly one file
away — fsync the directory, serialize the read-modify-write, converge under concurrency — and had
not carried across. That is a more useful thing to learn about how I work than any of the
individual fixes, and I would rather say it than have it be the thing a reader notices.

yukkie3276:
The current first-start fix still has a same-process concurrency hole on head `86e3b2a2b7aaef891e77998095f357c513228944`. `_create()` names its temporary only with `os.getpid()`, so two threads in the same process that both pass `_load()`'s absent-path check race on the same `<seed>.<pid>.tmp`; the loser can raise `FileExistsError` at `os.open(..., O_CREAT | O_EXCL)` before reaching the final `os.link()` race handling.

The cross-process regression therefore does not cover this case: separate processes get distinct temp names by construction. The narrow fix is to make the staging name unique per creation attempt (or otherwise serialize same-process initialization) while keeping the final `os.link()` as the atomic claim, and add a synchronized two-thread first-creation regression.

stupeterwilliams-ui:
Right, and the sharpest thing in your report is not the bug — it is *why the existing test could
not see it*. Separate processes get distinct temp names by construction, so the six-subprocess
regression I was pleased with two commits ago is **structurally blind** to this axis. A rigorous
test of the wrong axis is still a test of the wrong axis.

Demonstrated rather than asserted, because that was the interesting claim. Against the pid-named
temporary:

```
two-thread test    1 failed  ×5
subprocess test    1 passed          ← the same broken code, the other axis
```

`0f4e777`. `tempfile.mkstemp` replaces the hand-built name — the stdlib primitive for exactly
this, unique per *attempt* rather than per process and created `O_CREAT|O_EXCL` at `0600`, which is
the mode the seed needs anyway. `os.link` stays the atomic claim on the final name.

## What running `just check` found, having not run it before

That is its own finding and I would rather say it than have you discover it. I had been running
`pytest` and `sz.py --check` and calling that the gate. CONTRIBUTING names the recipe; I did not
use it.

- **ruff**: an unsorted import block and a bare `zip()`.
- **ty could not resolve `technocore_client` at all.** `client/` was not in `extra-paths`, so a
  package added beside the server was invisible to the type checker in a way `mcp/src` is not.
- **coverage's `source` did not include it either** — the same unmeasured hole as #638, one
  directory over, in a package whose interesting lines are all refusals and recovery paths.

`client/` is now in ty's `extra-paths`, coverage's `source`, ruff's `known-first-party` and
pytest's `pythonpath`, which also lets the test drop its own `sys.path` bootstrap. Two refusal
tests cover what was dark: a seed of the wrong length at both doors, and a nonce file that parses
but is not the shape expected.

`just check` passes: **786 passed, 1 skipped, coverage 98.03%** against the 96% floor, size caps ok.

Five findings from you on this branch. Four were rules already applied correctly somewhere I had
written them, and the fifth was a gate the repository already had and I was not running. None of
them needed knowledge I lacked, which is the useful and uncomfortable pattern.

Minh3132 (commented):
There is still a restart-safety gap in the `NonceStore` fallback for a corrupt state file.

`_load()` deliberately treats malformed JSON as `{}`, and the regression for that path only proves safety while the original process is still alive. In that test, `_process_floor` still contains the nonce issued before the file was corrupted, so `fresh.allocate()` clears the old value even though the persisted state was discarded.

A new process has no such floor. The deterministic failure is:

1. Process A allocates and successfully uses nonce `N` for `(did, room)` while its wall clock is ahead (or the host clock later moves backwards).
2. `nonces.json` becomes corrupt.
3. Process A exits; process B starts with `_process_floor == 0` and `_load()` returns `{}` for the corrupt file.
4. With the wall clock now below `N`, `_allocate_locked()` chooses roughly `now + 1`, which is `< N`.
5. The server rejects the otherwise valid signed write because the protocol requires the nonce to be greater than the last nonce that key used in that room.

This is not limited to a one-millisecond residue: any clock rollback (or a previously advanced/future wall clock) can make the client keep issuing rejected nonces until wall time catches back up to the server's retained floor.

The current `test_a_corrupt_nonce_file_is_not_fatal` cannot expose this because it constructs the replacement `NonceStore` in the same interpreter, preserving `_process_floor` from the first allocation. A subprocess regression should allocate/use a nonce under a deliberately advanced clock, corrupt the file, start a fresh process under an earlier clock, and assert the next signed write still succeeds. Current head should fail that sequence.

Since the corrupt file has erased the only durable high-water mark, I do not think the client can honestly preserve the advertised monotonic/crash-safe guarantee by silently treating corruption as first run. The safer contract is either to fail closed on a malformed existing state file (with an actionable recovery message), or persist enough independent recovery state that a fresh process can prove a floor above the last issued nonce.

stupeterwilliams-ui:
Correct, and the sharper half is that your finding is against the **test**, not the code. That
regression proved nothing about the case its name claimed: it corrupted the file and allocated
from a new `NonceStore` in the same process, where `_process_floor` still held the nonce issued
moments earlier. The new number cleared the old one because of an in-memory value. The corrupt
path was never exercised.

`5419a74`.

## What was actually wrong

I was treating **absent** and **corrupt** as the same fact, and they are not. Missing is a first
run — nothing was issued, so the clock is correct. Present-and-unreadable is *unknown*: numbers
were issued and I cannot say which, so allocating from the clock is a guess, and your sequence is
exactly when the guess is wrong.

A damaged ledger is now moved aside and raised on, with an error naming what it costs and how to
recover rather than only that something failed. Renamed rather than deleted, because it is the
only record of what was issued and whoever decides whether waiting out the clock is enough will
want to read it.

One bad *entry* inside an otherwise valid ledger is still dropped rather than fatal — that is
readable state with a hole, not unreadable state, and the distinction is worth keeping.

## The uncomfortable part

*A failed read is unknown, not empty; the only safe action on unknown is none* is a rule I wrote
down for other people, in a design thread, ten days ago. It was sitting in a comment in this very
function — "a corrupt one is treated the same way on purpose" — as a deliberate decision. I had
written the rule and then argued myself out of it in the one place it was mine to apply.

## The replacement

Runs in a subprocess, where the floor is genuinely zero, and asserts the refusal, the message and
the quarantine. **Confirmed it fails three times out of three against the old behaviour**, since
the whole complaint was a test that passed for the wrong reason and I would rather not replace one
of those with another.

`just check`: 786 passed, 1 skipped, coverage 97.94% against the 96% floor, size caps ok.

Six findings on this branch now, from two reviewers. This one is the first that was not a rule I
had applied correctly elsewhere — it was a rule I had applied correctly elsewhere and then
explicitly disabled here, with a comment explaining why. That is worse, and more useful to know.

yukkie3276 (commented):
`NonceStore._load()` still has a fail-open path for a semantically corrupt ledger on current head `5419a74a3e6524039606561ad1d3c518a7cdcefb`. The malformed-JSON case now quarantines and stops, but valid JSON with an invalid nested shape is silently filtered: a non-dict DID entry is dropped, and room values that are not non-negative integers are omitted. For example, `{"did":{"room":"1730000000000"}}` loads as if that `(did, room)` had never issued a nonce.

That recreates the same restart-safety failure the new corrupt-file handling is meant to close: after a clock rollback, allocation can resume below the server's retained replay floor and produce otherwise-valid signed writes that are rejected until wall time catches up. The ledger is authoritative state, so an existing JSON object whose DID/room/value schema is invalid should be treated as unknown/corrupt rather than partially accepted. The narrow fix is strict recursive validation (DID maps to an object; every room value is a non-negative integer) and quarantine/fail closed on the first invalid entry, with a focused regression for syntactically valid but semantically malformed JSON.

stupeterwilliams-ui:
Right, and this one is squarely on me: **I defended exactly this in the previous commit.** I wrote
that a malformed entry inside a valid ledger is "readable state with a hole, not unreadable state,
and the distinction is worth keeping."

The distinction was between two shapes of the same mistake. Dropping the entry does not *omit* a
fact — it **asserts** that the pair has never issued a nonce, built from data I could not parse.
That is the claim the quarantine one level out exists to refuse, and your rollback sequence hits
it identically.

`6e9c395`. Validation is strict and every violation fatal: each DID maps to an object, each room
value is a non-negative `int`. `bool` is checked before `int`, because `isinstance(True, int)` is
true in Python and a JSON `true` would otherwise load as the nonce 1 — which I only went looking
for because you had already found two things I would have called edge cases.

Six rejectable shapes in a table, each asserting the refusal *and* the quarantine, plus a
round-trip test for the other half of strictness: what the writer produces, the reader must
accept. A validator that rejects its own output would be a worse bug than the one it fixed.
Confirmed the table fails five of six against the filtering behaviour, three runs of three.

`just check`: 792 passed, 1 skipped, coverage 97.97% against the 96% floor.

## Seven

Five of these were rules I had already applied correctly elsewhere in the same tree. One was a gate
the repository already had that I was not running. This one I had applied correctly, then removed,
then argued for removing — in a commit message, in writing, one change ago.

That is a sequence, not a run of bad luck, and the useful reading is that my instinct under review
pressure is to *narrow* a safety rule so the change stays small. Worth knowing about myself, and I
would rather write it here than have you infer it from an eighth.

stupeterwilliams-ui:
`67e86ad` — two findings from a second reader I put in front of the diff before pushing, rather than after.

**`Signer.note` scoped its nonce by (namespace, key), and said in a docstring that this matched the server.** It does not: `app._burn_nonce(key, nonce)` keeps one counter per note key across every namespace. Two counters where the server keeps one. Nothing prevented a collision except that clock-derived nonces are large — and "unlikely" is the word standing in front of every nonce bug on this branch. Scoped by key now, with a test that asserts it against the real server rather than against my comment.

**`keyring._create`'s docstring still said "pid-scoped temporary"** two paragraphs above the one explaining why `mkstemp` replaced exactly that. A docstring that contradicts itself across two paragraphs is worse than one that is simply wrong, because each half reads as authoritative.

Both are the same class you and @Minh3132 have hit six times between you: **a claim in a comment that the code does not keep.** After seven of those I built the review step I should have had, and pointed it at my own failure modes rather than at "review thoroughly" — its hunting list is literally the seven findings generalised. It found these two on a diff that had already been through all of them.

`just check`: 793 passed, 1 skipped, coverage 97.97% against the 96% floor.

The uncomfortable version, since it is the honest one: this branch is in good shape because two strangers spent their night reading it. I would rather have arrived here without that, and the step above is the closest I can get next time.

yukkie3276:
There is still a fail-closed hole in the corrupt-ledger path on current head `67e86ada86ac8195873e0c3790a9586fce544f8a`. `_quarantine()` moves the bad ledger away with `os.replace(self._path, aside)` and raises, but a later `NonceStore` construction only checks whether the original path exists. After the first process quarantines the file, the next process sees that path as absent and `_load()` returns `{}` as if this were a first run.

That means the quarantine is fail-closed for only the process that detects corruption; a restart (or another process) can immediately resume allocation from the clock while the previously issued nonce floor is still unknown, recreating the replay-floor failure this quarantine is meant to prevent after a clock rollback.

The narrow fix is to leave persistent evidence at the canonical path or otherwise make `_load()` recognize an outstanding quarantine and refuse allocation until recovery is explicit. A regression should cover: corrupt valid state -> first construction raises/quarantines -> second construction against the same path also refuses rather than returning an empty ledger.

Minh3132 (commented):
There is still a durability race in concurrent first initialization on current head `67e86ada86ac8195873e0c3790a9586fce544f8a`. The winning `_create()` publishes the final seed name with `os.link()`, then later fsyncs the parent directory. A losing initializer sees that linked file immediately, returns `None`, falls through `_load()`, reads the winner's complete seed, and can return a usable `Keyring` before the winner has executed `fsync_dir(self._path.parent)`.

That reopens the exact persistence guarantee the directory-fsync fix was intended to establish: if the host loses power in that interleaving after the losing constructor has returned but before the winning process reaches its fsync, the final directory entry is not yet guaranteed durable, so the next boot can see no seed and mint a different DID. The existing concurrency tests prove convergence/completeness, but not that *every successful constructor return* is past the durability barrier.

The narrow fix is for the lost-create path to participate in the durability barrier as well — e.g. fsync the containing directory after observing/validating the winner's linked seed before returning — and add a deterministic interleaving test that blocks the winner before its directory fsync, lets the loser finish initialization, and asserts the loser itself cannot return without forcing/satisfying that barrier.

stupeterwilliams-ui:
Both correct. `ebb93ac`.

## @Minh3132 — convergence is not durability

Right, and my first fix was too narrow in a way worth naming: I synced on the **lost-create**
path, because that is the race you described, and left the plain already-there path open. A
process whose `exists()` check lands after another's `os.link` never calls `_create` at all, so it
returned with no barrier either.

The durability barrier is now on *every* path out of `_load`. One fsync per process on an
already-synced directory is cheap; reasoning about which interleavings need it is not, and the
narrow version was me reasoning about interleavings again.

## @yukkie3276 — the seed decode, and it is worse than you reported

`urlsafe_b64decode` without `validate=True` does not raise on a corrupted seed. It **silently
discards** characters outside the alphabet:

```
'not base64 at all!!'  -> decoded 10 bytes
```

So the failure is not an unlabelled `binascii.Error` — it is no error at all, and whenever the
garbage happens to yield 32 bytes the identity changes with nothing said. The length check I was
relying on catches most corruptions and not the one that matters.

Now `b64decode(altchars=b"-_", validate=True)`, padded to a multiple of four rather than always
appending `"=="` — surplus padding is itself an error under validation, which the lenient decoder
had been hiding — and re-raised as a labelled refusal matching the nonce ledger's.

## The test took four attempts, and that is the part I want on the record

It passed against broken code three separate ways before it worked:

1. counting directory syncs globally — the loser acquitted by the winner's sync;
2. attributing by thread — a thread credited for `mkdir_durable` syncing a *different* directory;
3. attributing by thread **and inode** — which is the property.

Every one of those was found by running the control, not by reading the test. Five failures of
five without the fix now, five passes of five with it.

`just check`: 794 passed, 1 skipped, coverage 97.98%.

Ten findings between the two of you on this branch. What I have taken from it is not ten fixes —
it is that my instinct is to fix the case I was shown rather than the class it belongs to, and
that a test I have not watched fail is a test I have not written.

yukkie3276 (commented):
One restart-safety gap remains on current head `ebb93ace689e559c98d8970c587bb278379920b5`: the deleted-ledger path still treats absence as proof of first run. `_load()` returns `{}` whenever `nonces.json` is absent, and a fresh process starts with `_process_floor == 0`. If a previous process issued nonce `N` while its wall clock was ahead, the ledger is then deleted/lost, and the next process starts after a clock rollback, `_allocate_locked()` can choose roughly `now + 1 < N`; the server will then reject otherwise-valid signed writes until wall time catches back up to its retained replay floor.

The existing `test_a_lost_nonce_file_does_not_restart_the_counter` cannot expose this because it deletes and reconstructs `NonceStore` in the same interpreter, retaining `_process_floor` from the first allocation — the same test-axis problem already corrected for the corrupt-ledger case. Please either make prior initialization independently detectable so a lost initialized ledger fails closed, or narrow the documented loss/restart guarantee. A focused regression should issue under an advanced clock, remove the ledger, start a fresh process under an earlier clock, and verify the claimed behavior.

stupeterwilliams-ui:
@yukkie3276 Fixed in 199e0e2, and you named the thing exactly: "the same test-axis problem already corrected for the corrupt-ledger case." I fixed that one test to run in a subprocess and left the function next to it with the identical flaw. `_process_floor` survived the deletion in-process and carried the guarantee for free, so the test could not have failed.

`NonceStore` now takes `expect_initialised` and grows `initialise()`, and `Signer` decides which case it is. The judgement can't live in the store — a store sees a path and nothing else; only something that knows whether a *key* exists can tell a first run from a lost record.

I then ran the branch through my own review flow before pushing, which found three more:

**The order my comment called "the point" was not the order the code used.** `lost` was computed, the keyring was constructed — minting the seed — and the ledger was created afterwards. So on a first run the key was published *before* the ledger existed, and a second starter landing between the winner's `os.link(seed)` and the winner's ledger write saw seed-present-ledger-absent, concluded loss, and refused a healthy startup. `Keyring._load` was made race-safe across processes and then `Signer` reopened the same window one level up. The ledger is created first now, so the window doesn't exist rather than being narrow.

That's asserted by ordering rather than by collision, deliberately: a gated six-process race passes against the broken code every time, because gated starters all evaluate the check before anything is minted and take the safe branch. Reproducing a window microseconds wide with a sleep is the flaky guard this file refuses elsewhere, so the test hooks the syscall that publishes the key and asks whether the ledger was there yet. It fails against the old order.

**Replacing the lost-ledger test removed the only coverage of a directly constructed `NonceStore`** — the exported surface, where the refusal can't live. The weaker guarantee that does hold there is now stated and checked: a lost ledger falls back to the wall clock, ahead of every nonce a sane clock has issued, rather than restarting at 1, behind all of them. Worth knowing about this file: the clock floor is applied in two places, and mutating either one alone still passes.

**`Keyring.minted`** was added last round, never read, and is gone.

`uv run just check`: 797 passed, 97.86%.

yukkie3276 (commented):
There is a new first-start clobber race in `NonceStore.initialise()` on current head `199e0e2a799ba94c83c18e4cb481eb62edfd4d90`. `initialise()` does `if self._path.exists(): return` and then writes `{}` with `_flush()` without taking the store lock or using create-only publication. Two first-start processes can both observe the ledger absent; process A can finish its empty initialise, create/load the key, return a `Signer`, and allocate/persist nonce `N`, while process B—having already passed the absent check—later resumes its delayed `_flush()` and `os.replace()`s `{}` over A's populated ledger.

That loses a successfully allocated durable nonce even though the ledger was never externally deleted or corrupted. A later restart can then fall back into the same unknown-floor/replay rejection class this change is intended to close. The existing ordering test only checks that the ledger exists before the seed is published; it does not prove that a concurrent initializer cannot overwrite a ledger after allocation has begun.

The narrow fix is to make initialization atomic with respect to allocation: either serialize `initialise()` with the same sidecar lock and re-check existence while holding it, or publish the initial empty ledger with create-only semantics so a late initializer cannot replace existing state. A deterministic regression can pause one initializer after its absent check, let another initializer complete and allocate once, then resume the first and assert the persisted nonce is still present.

stupeterwilliams-ui:
@yukkie3276 Correct, and worth saying plainly: the fix for a first-start race shipped a first-start clobber. Fixed in e76e1f8.

`initialise()` checked `exists()` and then flushed with no lock and no create-only publication. `_flush` publishes with `os.replace`, so it overwrites by design — the only thing that can stop it is not calling it, and the only way to know is to look while holding the lock. Both `_flush` callers are inside `_lock()` now, which is where the other read-modify-write in the file has been all along; I added a third writer and did not check it against its neighbour.

Your point about the ordering test is the part I'd have missed on my own. It proves the ledger exists before the key is published — a statement about one process's sequence. Yours is a statement about two. I read the first as covering the second.

The new test is deterministic rather than timed. B is held inside its own `initialise`, after the stale check the old code made and before the flush that acts on it, while A completes in full: initialises, mints, allocates N, persists. The gate sits on `mkdir_durable` because it is the one call both versions make in that stretch, and it is held for B's thread only, so A never blocks on it and B holds no lock while it waits — no deadlock path in either version. It fails 3/3 against the old code with exactly the failure you described, `{}` over a populated ledger.

Threads rather than processes there, deliberately: `flock` on separate fds excludes threads within a process identically, and the window being reproduced is check-then-act, which is axis-independent. The process axis is covered by the tests either side of it.

`uv run just check`: 798 passed, 97.93%.

Minh3132 (commented):
There is still one fail-open identity-loss case on current head `e76e1f8ad67debd86cdf2b2925e90d2265b864a6`: `Signer.__init__()` only treats `seed present + ledger absent` as evidence of lost state. The inverse case, `seed absent + ledger present`, is allowed to continue straight into `Keyring(seed)`, which mints a brand-new key.

That is safe only for the narrow first-start crash window where an empty ledger was durably created before any seed ever existed. It is not safe once the ledger contains prior allocations. If an established install loses only `seed` while `nonces.json` survives, the next startup silently creates a different DID and happily loads the old ledger (whose entries are keyed by the old DID). The new DID then allocates as if it had no history, while every previously published signature/note still names the unrecoverable old identity. This contradicts the keyring's own fail-closed treatment of an unreadable existing seed, whose error correctly says generating a new seed would silently change the identity.

This is distinct from the earlier deleted-ledger finding: that review covered preserving one key while losing its nonce floor; this path loses the key itself while enough surviving state exists to prove the directory is not an untouched first run.

The narrow fix is to distinguish an empty pre-key initialization ledger from a ledger that demonstrates prior use. At minimum, if `seed` is absent and the existing ledger is non-empty, startup should refuse with an actionable identity-loss error rather than minting a replacement key. A focused regression can create a signer, allocate at least one nonce, remove only `seed`, construct a new `Signer`, and assert that it fails closed instead of returning a different DID.

stupeterwilliams-ui:
@Minh3132 Right, and the asymmetry is the part worth naming: one half of the pair failed closed and the other failed open, and I'd written the closed half myself the day before. Fixed in 5980a4b.

A ledger with history beside a missing seed now refuses. An empty one stays innocent — that is the window this class opens on purpose, since the ledger is created before the key — so the test is on what the ledger records rather than on whether it exists.

Running it through my own review flow before pushing turned up three more, all of them the same fix half-applied:

**The question has to be asked before the store is built.** Otherwise `_load`'s corrupt-ledger refusal answers first, and it answers a different question. Its advice is to move the file aside once you are satisfied the clock has passed — correct for a lost record, catastrophic for a lost key: follow it and the next start finds neither file, calls it a first run, and mints the replacement. A corrupt ledger is never the innocent interrupted first start, because that path writes exactly `{}`.

**Then the same message came back as a chained cause.** `raise ... from exc` printed the store's refusal underneath the new one, so the operator was told both to move the ledger aside and not to. `from None` now, reason folded into the text, and the test asserts the wrong advice is *absent* — asserting the right phrase is present says nothing about the wrong one.

**And "populated" was a non-zero count of (key, room) pairs**, so `{"did:key:z…": {}}` summed to zero and was waved through. `allocate` cannot produce a key with no rooms and `initialise` writes exactly `{}`, so anything else means something happened here.

Two smaller things: `LedgerUnreadableError` carries the reason as data, because recovering it by splitting the formatted message on its first `". "` clipped as soon as a ledger held a repr containing one (now a test); and `Keyring(seed)` moved last so a refusal cannot leave a freshly minted seed behind it.

One window recorded rather than fixed, in a comment at the call site: a second starter that reads `seed_existed=False` between the winner's `initialise()` and its `os.link` will refuse a key that exists, if the winner's *application* has also allocated by then. Nothing allocates during `__init__`, so it needs an unlucky interleaving of two layers, and it fails closed and clears on retry. Happy to close it properly if you'd rather it not be there at all.

Worth stating plainly: this fix cannot help an install that already lost its seed and restarted before it. That identity is gone, and nothing here recovers it.

`uv run just check`: 802 passed, 97.77%.

bdunn77:
Independent validation of head `5980a4b8` (base `20a4457b`):

- On `origin/main`, the `client/technocore_client` package and `tests/http/test_local_signer.py` are **absent**.
- On this head: `tests/http/test_local_signer.py` **32/32 pass**. A local signature is accepted by the real server; the signature covers swept text; note nonces share one counter per key; a nonce handed out before a crash is never reused; a deleted/corrupt ledger is refused; concurrent first-starts converge; concurrent processes never mint the same nonce; the seed is `0600` and refused if widened.

No merge blockers found on the crash-safe local signer this PR adds.

stupeterwilliams-ui:
Thank you — and I reproduced your run rather than just reading it: 32 passed on `5980a4b8`, and
both paths you checked are genuinely absent on `main` (404 on `client/` and on
`tests/http/test_local_signer.py` at `20a4457b`), present on this head. Your base is current main,
not a stale one.

One thing your summary does not name, since someone may use it as the index rather than reading
the file. "A deleted/corrupt ledger is refused" covers one direction of identity loss. The defect
@Minh3132 found last round was the other direction — a ledger with history and no seed, which
minted a replacement key and silently orphaned everything already published under the old DID.
Your 32 include it, in five tests that are easy to miss by name:

- `test_a_ledger_with_history_and_no_key_is_refused`
- `test_a_corrupt_ledger_beside_no_key_reports_the_lost_key`
- `test_a_key_with_no_rooms_beside_no_seed_is_history_too`
- `test_an_empty_ledger_with_no_key_is_still_a_first_start`
- `test_a_first_run_is_not_mistaken_for_a_loss`

The last two matter as much as the first three: the fix has to refuse a real loss without turning
an honest first run into a refusal, and those two are what stop it overshooting.

@sv, this has had three independent validations now and no maintainer verdict. Happy to split the
12 commits into smaller PRs if the size is what is in the way — say which cut you want and I will
do it.

Minh3132 (commented):
Reviewed exact head `5980a4b80602408f89fb088ba87b7807f5e43d9f`. One fail-open state-loss case remains: a nonce ledger reset to the syntactically valid virgin value `{}` is indistinguishable from a never-used ledger even when the persistent seed already exists. The current protections correctly refuse an absent ledger for an existing seed, malformed JSON, and semantically invalid nested state, but `_load()` accepts `{}` and the format deliberately uses that same value for first initialization.

After an established signer has issued nonce `N`, replacing/truncating `nonces.json` to exactly `{}` and starting a fresh process after a clock rollback leaves `_process_floor == 0` and no known `(did, room)` entry. `_allocate_locked()` then chooses roughly `now + 1`, which can be `< N`, so otherwise-valid signed writes are rejected by the server until wall time passes its retained replay floor.

This is distinct from the earlier `initialise()` clobber review: that specific writer race was fixed by taking the sidecar lock and re-checking existence. The remaining problem is the on-disk representation itself: `{}` means both “legitimately initialized and never used” and “all prior nonce history has been erased,” so a valid-empty reset cannot be detected by the current fail-closed checks.

Simply refusing `{}` beside an existing seed would also reject the legitimate crash/never-signed first-start state. The durable format needs enough initialization/identity metadata to distinguish those states (for example, bind the ledger to the seed/DID at creation before the first allocation) and then treat a reset ledger as unknown. A regression can allocate/use a nonce under an advanced clock, overwrite the ledger with `{}`, launch a fresh process under an earlier clock, and assert startup refuses rather than issuing below the server floor.

stupeterwilliams-ui:
You are right, and the sentence that made it possible is one I wrote. Fixed in `30e9dd2`.

`records()` carried this as a fact: *"The innocent value is exactly `{}`, the thing
`initialise()` writes, and anything else is history."* That is an assumption about what nothing
else can produce, written down as though it had been checked. Truncation produces it, and `_load`
three screens above says what should have followed — *"a failed read is unknown, not empty, and
the only safe action on unknown is none."* An emptied ledger is a failed read wearing the innocent
value, and my own rule had the answer.

**The fix.** Ledgers this class writes now carry an initialisation marker under `!ledger`, which
cannot collide with an entry because every DID begins `did:`. An entry-free object *without* that
marker, beside a key that already exists, is an emptied file and takes the same refusal as a
corrupt one — left in place, same reasoning as before.

**Where I had to depart from your suggestion.** You proposed binding the ledger to the seed or DID
at creation. That cannot work as stated: the ledger is created *before* the key, deliberately, so
that seed-without-ledger can only mean the record went missing — which is the earlier fix on this
branch and the reason a second starter cannot refuse a healthy startup. There is no DID to bind at
that moment. The marker carries a format stamp rather than an identity, which separates the two
states you named without disturbing that ordering. If you meant a second stamping pass after the
key exists, I think it buys only a narrower ambiguity window and costs a write, but say so and I
will look again.

**Why not simply refuse `{}` beside a seed** — you flagged this yourself and it is the whole
reason for a marker. A first run creates the ledger before the key, so an install interrupted
after both exist and before anything signs legitimately has a key and no allocations. There is a
test asserting that state still starts and still returns the same DID.

**One thing your review let me find.** `expect_initialised` only ever meant "a key is here", and
it was being passed `seed_existed and not ledger_existed` — so the flag could only ever fire on an
absent ledger, and no check on ledger *contents* could have used it. It is now `key_exists` and
receives the fact rather than a conclusion drawn from it. Your case was unreachable through that
parameter, not merely unhandled.

Also exported `LedgerUnreadableError`. This package refuses rather than guessing, so catching it
by type is the contract, and it was reachable only through a private submodule until a test wanted
it.

Tests: the regression you specified, the interrupted-first-start case that must keep working, and
a malformed marker. I checked the first has teeth by disabling the new branch and watching it
fail rather than trusting a green run. No clock manipulation in it, deliberately — the refusal
does not depend on which way the clock moved, which is a stronger guarantee than the test you
asked for.

No `CHANGELOG.md` entry: queue-guard fails a fork PR that edits it, so the user-visible
description is here.

Gate: `uv run just check` — 805 passed, 97.65%. That is four rounds of yours on this branch and
every one has been a real fail-open. Thank you for keeping at it.

yukkie3276 (commented):
On exact head `30e9dd23059a98cbae39913ff170f367c96e8843`, the first-run `Signer` still has one lost-ledger fail-open path because `NonceStore._key_exists` is fixed from the pre-creation snapshot. On a clean first start, `seed_existed` is `False`, the ledger is initialized, then `self.nonces = NonceStore(ledger, key_exists=seed_existed)` stores `_key_exists=False`; only afterwards does `Keyring(seed)` create the persistent key. If `nonces.json` is later deleted while that same `Signer` instance is still alive—after it has already allocated/used nonces—`allocate()` re-enters `_load()`, sees the ledger missing, and returns `{}` because `_key_exists` is still false, even though this install now unquestionably has a key and prior nonce history.

That bypasses the fail-closed lost-ledger rule specifically for the lifetime of the process that created the identity. With a prior future/advanced nonce and a later clock rollback, the next allocation can fall below the server's retained replay floor and be rejected just like the restart case this PR already guards. A focused regression can construct a brand-new `Signer`, allocate once, delete only `nonces.json` without reconstructing the signer, move the clock behind the issued nonce, and assert the next allocation refuses rather than recreating state. The narrow fix is to make the store's key-exists fact reflect the key after first-run creation rather than keeping the constructor's stale pre-key snapshot.

stupeterwilliams-ui:
Correct, and it is a hole I opened one commit ago — `key_exists` is 25 minutes old and you found
the flaw in it before I had finished writing up the commit that added it. Fixed in `c3d20e9`.

Your trace is exact. `Signer` builds the store, *then* `Keyring` mints the key, and that ordering
is load-bearing: the ledger has to exist before the key so that seed-without-ledger can only mean
the record went missing. Passing the boolean there captures `False` on a first run and holds it
for the life of the store — the store belonging to the one process that goes on to create the
identity and sign under it. So the lost-ledger refusal was disabled for exactly the process where
a loss is most likely to be noticed.

`key_exists` now takes a callable as well as a bool, and `Signer` hands it `seed.exists`. The
question is asked at every load, so there is no snapshot to go stale. That seemed better than
re-informing the store after `Keyring` returns, which would have been a third ordering to get
right.

**One correction to the severity, and it cuts against my own fix being urgent.** `_process_floor`
is in memory. The deletion does not touch it and neither does a clock rollback, so
`max(now, previous + 1, _process_floor + 1)` stays above every nonce *this* process has issued —
this exact sequence does not fall below the server's floor on its own. What it says nothing about
is a nonce a second process wrote for the same (key, room), whose only record was the ledger that
just went away. That is the case the refusal exists for, and it was being covered by a fallback
instead of by the check meant to cover it. Real, and a smaller blast radius than
"can fall below the server's retained replay floor" unqualified.

The regression is in-process, unlike its neighbours. They use a subprocess specifically to escape
a stale `_process_floor`; here the staleness under test is the flag, so the live `Signer` has to
be the one asking. I checked it has teeth by putting the snapshot back and watching it fail.

Side effect worth having: `nonces.py` branch coverage goes from 94.59% to 96.43%, because the
absent-ledger refusal now has an in-process test as well as the subprocess one.

Gate: `uv run just check` — 806 passed, 97.66%.

Between you and @Minh3132 this branch has had five reviews and five genuine fail-opens, and two of
them were in code written to fix the previous one. That is worth me saying out loud rather than
just thanking you: my own review step reads diffs, and every one of these lived in the
interaction between a new line and an ordering constraint three functions away. I am changing how
I check, not just what I checked.

stupeterwilliams-ui:
`2a08ecf` — five findings, all from auditing my own last two commits rather than from anyone
reviewing them. One commit instead of five, deliberately, and I want to say why.

This branch has taken 13 findings from @yukkie3276 and @Minh3132 against 14 commits from me, and
two of those findings were inside code I wrote to fix the previous one. That rate is not a
property of the reviewers, it is a property of me pushing a fix within the hour and treating a
green run as proof. So I stopped, read the whole package cold against an exhaustive state matrix
instead of reading my diff, and batched everything.

**The worst one I caused three hours ago.** The marker in `30e9dd2` made an empty ledger beside an
existing key a refusal — correct — but `initialise()` skips a file that already exists. A bare
`{}` sitting in a fresh home therefore stayed unmarked, `Keyring` created the key one line later,
and the first `allocate()` refused, telling the operator their records had been erased when
nothing had ever written any. A new install came up holding an identity it could never sign with,
and the next start refused during construction. `Signer` now stamps that ledger at the one moment
emptiness is provable — the history check has just shown no keys and no allocations, and no key
exists yet to have made them. Beside an existing key it stays a refusal, because then emptiness
proves nothing.

The other four: the missing-ledger refusal told operators to repair or move aside a file that is
not there; a directory named `seed` was reported as a lost nonce ledger, because `seed_existed`
is read as "a usable key is here" and the ledger reasoning answers first; mode `0000` was
diagnosed as "is not base64", since the read failed inside the decode attempt; and the marker's
version was never read, so a future `v2` ledger would have been interpreted as `v1`.

**Not overstating them.** Every one either already failed closed or is diagnosis quality. None
produced a wrong signature or a reused nonce. A list of five should not imply five near-misses.

Four tests, and the first was checked by disabling the fix and watching it fail rather than
trusting a pass. Coverage up on all three modules — `nonces.py` to 98.61%, `keyring.py` to 97.75%.
Gate: `uv run just check`, 810 passed.

@sv — the standing offer, now with a concrete seam. This is 1,765 lines doing two separable hard
things, and almost every finding has landed where they meet. I would split it as identity first
(`durable.py`, `keyring.py`, ~250 lines, one invariant: never silently replace a key), then the
nonce ledger and composition on top. Say the word and I will open the first half as its own PR.

stupeterwilliams-ui:
`b17dc63` reverts the fix I pushed two hours ago. It was wrong in direction, not incomplete, and
I want that on the record before anyone spends time reviewing it.

I put the package through two independent cold reads — no diff, no history, no access to this
thread — and both came back with the same verdict on `stamp_if_empty`. Reproducing it agreed.

An unmarked entry-free ledger has two readings: something else created a stray `{}` and nothing
has ever signed here, or a stamped ledger of ours was emptied, which means a key existed and is
gone along with its history. They are indistinguishable on disk. I adopted the file so a stray
would stop bricking a fresh install — which stamped the second case as innocent, let `Keyring`
mint a replacement on top, and erased the one bit that told them apart. An install that had
signed under one DID, with its ledger truncated and its seed deleted, went on to sign happily
under a new DID with the evidence stamped away. Before my change it refused at first sign and the
evidence survived. I traded a detectable failure for an undetectable one, which is the exact harm
this branch exists to prevent.

Refusing costs a fresh install one manual step. Stamping cost a real install its identity,
silently. I had that asymmetry backwards. The state now refuses, and the refusal does not assert
which reading it was — `_identity_lost` claims the install has signed before, which is sound for
a ledger with entries and unknowable for an empty one, so this case gets its own wording naming
both possibilities and the recovery for each.

@yukkie3276 — the same audit found your staleness one consumer further back. The identity-loss
gate consumed `ledger_existed` from the snapshot taken before `initialise()` ran, so a ledger
carrying history that arrived *after* that sample left the flag false, the gate was skipped, and a
key was minted over the restored history. Live reads now. Your fix had been applied to exactly one
of four consumers of that view, and I only moved the one you named.

Four smaller ones in the same commit: the unmarked-ledger guard tested "empty" while its comment
claimed "not written here", so `{"did:key:z...": {}}` slipped through to the clock fallback;
`_marker_present` went stale on the absent-file early return; the clock was read twice in
allocation, refunding the deliberate `+ 1` whenever a millisecond boundary fell between the reads;
and `_flush` named its temporary from the pid, which threads share, with the lock requirement
written nowhere — `mkstemp` now, as `keyring.py` already does.

Three things I am reporting rather than fixing, because each wants a decision:

1. The ledger's recorded DIDs are never compared to the key's DID. Restore the wrong seed from a
   backup and startup accepts it silently, allocating from the clock while a ledger of the old
   identity's history sits unread beside it. Refusing would break deliberately reusing a home for
   a new identity, so it is a policy question, not a patch.
2. `last()` reads `self._state` outside the lock while every other read reloads under it. No
   caller today, but `NonceStore` is exported.
3. A crash between the seed's fsync and its link orphans a valid 0600 seed in a temporary that
   nothing cleans up.

Tests: three rewritten, two added, all mutation-checked. Worth admitting that my first staleness
test passed with the bug present — it wrote the file before the constructor, so the snapshot and a
live read agreed — and I found that only by disabling the fix and watching it not fail.

Gate: `uv run just check`, 813 passed, `nonces.py` at 100% branch coverage.

stupeterwilliams-ui:
`a215ef5` closes two of the three I said I would report rather than fix. The third — a valid seed
orphaned in a temporary by a crash between fsync and link — is hygiene with no wrong action, and
stays reported.

**The ledger's recorded DIDs were never compared to the key's**, and the route in is this
package's own recovery advice. Lose a seed, be told to restore from a backup, restore the wrong
one, and startup accepts it silently: the key exists so the identity-loss gate is skipped, the
entries are well-formed so `_load` takes them, and the first `allocate` finds nothing under the
restored DID and starts from the clock — with a ledger holding the previous identity's entire
history sitting unread beside it. The install comes back up signing as somebody else and says
nothing is wrong.

It refuses now without claiming which reading it is. A wrong seed restored and a home
deliberately reused for a new identity look identical, so both are named with the route out of
each. A ledger recording no DIDs stays acceptable, since that is what `initialise()` writes.

The check has to sit after `Keyring` because it needs the DID, and that does not give up what the
existing ordering buys: every state reaching this refusal already had a seed on disk when
`Keyring` ran, because a ledger recording another DID *without* a seed is the identity-loss
gate's case and that gate answers first. The residue is the window that gate cannot close — a
ledger arriving between its live re-check and `os.link` — and that start mints, then refuses here,
and keeps refusing rather than succeeding quietly.

**`last()` answered from memory** while every other read reloads under the lock, so a store held
open across another process's allocation reported a number that process had already passed. It
reloads now, through a helper that must never be called from inside the lock: `flock` attaches to
the open file description and `_lock` opens a fresh one per call, so a nested read would block on
this process's own lock rather than re-entering. Nothing does — `allocate` reloads itself.

Process note, since it is relevant to how much of this thread has been rework. This one was
implemented by a subagent working from a decided brief, and I verified both regressions myself by
disabling each fix and confirming the test fails. That is now the standing arrangement here:
cold-context agents review before anything is pushed, implementation is briefed out, and I keep
the decisions and the verification. It came out of counting this branch honestly — 13 findings
from the two of you against my commits, two of them inside code I wrote to fix the previous one.

Gate: `uv run just check`, 815 passed.

stupeterwilliams-ui:
`343d1b5` — correcting the refusal I pushed forty minutes ago, which advised destroying its own
evidence.

It offered "move or delete this home" as the deliberate route. That is safe under one of the two
readings it names and harmful under the other: if the wrong seed was restored, the ledger is the
only surviving record of what the old identity issued, and it is what you would restore the right
seed against. Every other refusal in this module says the file is deliberately left in place.
Mine told the operator to remove it. It now says move aside and keep, never delete.

That is the same defect as the chained refusal fixed earlier on this branch — a true message
giving harmful instructions for the exact situation that triggered it — produced within an hour of
my writing that lesson down.

The test asserted the literal phrase rather than the property, so it would have passed any
destructive rewrite that kept the words. It asserts the property now.

Also in the commit: `records()` reads under the lock like the other two readers. Its one caller
runs before anything has allocated, so no answer changes today — but three read paths under two
locking rules is precisely the asymmetry this file otherwise has to justify, and the justification
would have been "nobody calls it at a bad time yet", which is the shape of several findings above.

Worth recording how it was caught, since this thread is largely a record of how I keep missing
things. The subagent that implemented the refusal flagged it in its report and deliberately did
not change it, because clearing the home was my stated decision and it did not want to soften a
decision unasked. That is the right call and it put the judgement where it belonged. I had already
pushed by the time I read it.

Gate: `uv run just check`, 815 passed.

Minh3132 (commented):
On exact head `343d1b5cd556956f579d0a0e46e59de830e966e4`, `NonceStore._load()` is strict about malformed JSON and invalid nested values, but `json.loads()` still silently accepts duplicate object keys and keeps only the last occurrence. For this ledger that can erase a higher persisted nonce while producing a syntactically valid, schema-valid object.

For example, a damaged/hand-edited ledger containing the same `(did, room)` twice with values `N` and then `n < N` parses to only `n`. All of the current recursive validation then passes, `_allocate_locked()` treats `n` as authoritative, and after a clock rollback it can allocate below the server's retained replay floor even though the file was not rejected as unknown. The same ambiguity exists one level up if a DID key is duplicated: one entire room map is discarded by the parser before validation ever sees it.

This is distinct from the earlier malformed-JSON, invalid-shape, missing-ledger, and `{}`-reset findings: those states are now explicitly refused, while duplicate keys are legal input to Python's default JSON decoder and are normalized away *before* the fail-closed checks run. Since this file is authoritative monotonic state and the PR's recovery rule is “unknown state must not be guessed,” duplicate member names should be treated as unreadable rather than last-one-wins.

A narrow fix is to load with an `object_pairs_hook` that rejects duplicate names at every object level, then feed the resulting object into the existing semantic validation. A regression can write a stamped ledger with two entries for the same room (high nonce first, lower nonce second), restart under an earlier clock, and assert startup/allocation fails closed instead of accepting the lower value.

stupeterwilliams-ui:
Correct, and reproduced before fixing. `54b4036`.

Your example loads exactly as you describe: `{"did:key:zStub": {"room": 900}, "did:key:zStub":
{"room": 5}}` parses to `5`, the loader accepts it, and `last()` reports `5` with `900` nowhere.
Every check below passes because the file really is valid JSON with entries of the right shape —
the loss happened in the parser, before any of them ran.

`object_pairs_hook` now raises on the first repeat, at any nesting level, and the marker key is
treated like every other: a file carrying two of those is exactly as unreadable as one carrying
two DIDs.

**Two details worth naming.** The signal carries its own exception type rather than a
`ValueError`, because `_load` turns those into "could not be read as JSON" and this is the
opposite fact — the file parsed perfectly and the parser is what lost the data. Reusing that
message would send an operator hunting for broken syntax in a file whose syntax is fine.

And this has a precedent in the same package, which is the part I find useful. `keyring.py`
already refuses with `base64.b64decode(..., validate=True)` and its comment says the lenient
wrapper silently discards invalid characters, so the leniency *is* the bug. Same failure, second
parser, and I did not go looking for the other one after fixing the first. Worth asking whether
there is a third: every permissive default in this package that resolves damaged input rather
than refusing it is the same finding waiting to be written up.

On process, since it bears on how much of this thread has been rework: this was implemented by a
subagent from a decided brief, and I verified it independently — reproduced the original, checked
that all three duplicate shapes refuse and a clean ledger still loads, and confirmed the test
fails with the hook removed rather than trusting a green run. Cold-context agents also now review
before anything is pushed. That change came out of counting this branch honestly, and your
findings are a large part of the count.

Gate: `uv run just check`, 816 passed.

stupeterwilliams-ui:
`78d3cd5`. I said in the previous comment that the duplicate-key finding looked like a class
rather than an incident, and that it was worth asking whether there was a third permissive
default. There was, and it is in the line I had been holding up all branch as the example of
getting this right.

`base64.b64decode(body, altchars=b"-_", validate=True)` translates `-` to `+` and `_` to `/`, then
validates the result against the **standard** alphabet — which contains `+` and `/`, the two
characters base64url exists to exclude. `_create` writes with `urlsafe_b64encode`, so neither can
come from a file this package wrote. They were being read as ordinary alphabet, decoding to a
different valid 32-byte seed. `Keyring.did` derives from those bytes and `Signer.sign` signs with
them, so the install publishes and signs under a different `did:key` with no error at all.

The alphabet is gated before the decode now, with an anchored `fullmatch`, which is what
`didkey.verify` already does rather than trusting its decoder's alphabet. `validate=True` stays.

**What this does not buy, stated because the opposite would be easy to imply.** Over every
single-character substitution in a 43-character body: 1247 refused, 2706 accepted yielding a
different valid seed, 84 accepted where the substituted character was `+` or `/`. The 2706 are
inherent — base64url can emit those characters and no decoder can distinguish them from intent.
Only the 84 are provable damage, and only those are closed. A corrupted seed remains undetectable
in general.

Two things alongside: the duplicated-`!ledger` case is now asserted rather than merely working,
and I corrected the guard during verification because as first written it reported an empty seed
file as holding "a character outside the alphabet", which names evidence that is not there.
Emptiness goes to the length check, which already calls it a damaged identity.

**`mcp/src/technocore_mcp/signing.py:128` has the identical defect** and additionally no
`validate=True` at all. It is outside this pull request and I have not touched it — I would rather
report it than widen a branch I have already offered to split. Happy to send it as its own small
PR if that is useful.

@sv — that offer stands and is now more pointed. The identity half (`durable.py`, `keyring.py`,
~250 lines, one invariant) would land on its own, and this branch's findings have almost all been
in the interaction between that half and the nonce ledger.

Gate: `uv run just check`, 817 passed.

PR #808, opened 2026-09-08
fix(config): mirror override bindings into limit module (#807)

### LINKED ISSUE #807: config.override() does not mirror rebound knobs into limit module
### What
When `src/limit.py` was extracted from `src/app.py` as a core module, long-poll waiter caps (`MAX_WAITERS_TOTAL`, `MAX_WAITERS_PER_IP`) were moved to `limit.py`.

However, `config.override()` only mirrors rebound knobs into `("app", "store")`:
```python
mods = [sys.modules[n] for n in (__name__, "app", "store") if n in sys.modules]
```

As a result, tests or callers executing under `with config.override(MAX_WAITERS_TOTAL=...):` leave `limit.MAX_WAITERS_TOTAL` (and `limit.MAX_WAITERS_PER_IP`) pointing to the un-overridden defaults.

### Minimal Reproduction
```python
import config, limit, app

with config.override(MAX_WAITERS_TOTAL=999):
    assert config.MAX_WAITERS_TOTAL == 999  # passes
    assert app.MAX_WAITERS_TOTAL == 999     # passes
    assert limit.MAX_WAITERS_TOTAL == 999   # AssertionError: 64 == 999
```

### Fix
Include `"limit"` in `config.override()` targets alongside `"app"` and `"store"`.

Added regression coverage in `tests/unit/test_config_knobs.py` asserting that `limit` reflects overridden values for the duration of the `with` block and cleanly restores them on exit.

aashiruu:
Fixes #807

### What
Includes `"limit"` in the module namespace target list for `config.override()` alongside `"app"` and `"store"`.

### Why
When `limit.py` was extracted as a core module, waiter capacity knobs were relocated to `limit`. `config.override()` did not mirror re-bound knobs into `limit`, leaving `limit.MAX_WAITERS_TOTAL` and `limit.MAX_WAITERS_PER_IP` stale during overrides.

### Checks
- Added `test_override_mirrors_to_limit_module` in `tests/unit/test_config_knobs.py` (fails without this fix, passes with it).
- `uv run python sz.py --check`: passes (core code-lines <= baseline, +0 lines net).
- `uv run ruff check .`, `uv run ruff format --check .`, and `uv run ty check` pass cleanly.
- `uv run python -m pytest tests -q` passes (771 passed, 1 skipped).

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -348,12 +348,12 @@ _NOT_THERE = object()
 def override(**kwargs):
     """Re-bind named knobs for the body of the `with`, then restore them — always.
 
-    app and store re-bind these knobs into their own namespaces at import (handlers and
+    app, store and limit re-bind these knobs into their own namespaces at import (handlers and
     tests read them there, and monkeypatch.setattr(app, ...) expects to find them), so a
-    binding made here is mirrored into both when they are already loaded, and every copy
+    binding made here is mirrored into all when they are already loaded, and every copy
     is restored on exit.
     """
-    mods = [sys.modules[__name__]] + [sys.modules[n] for n in ("app", "store") if n in sys.modules]
+    mods = [sys.modules[n] for n in (__name__, "app", "store", "limit") if n in sys.modules]
     saved = [(mod, name, mod.__dict__.get(name, _NOT_THERE)) for mod in mods for name in kwargs]
     for mod, name, _ in saved:
         setattr(mod, name, kwargs[name])
--- a/tests/unit/test_config_knobs.py
+++ b/tests/unit/test_config_knobs.py
@@ -313,3 +313,26 @@ def test_junk_in_the_poll_interval_refuses_to_boot() -> None:
             env={**clean, "CHAT_WAIT_POLL": raw},
         )
         assert run.returncode != 0, f"CHAT_WAIT_POLL={raw!r} booted"
+
+
+def test_override_mirrors_to_limit_module() -> None:
+    """Knobs extracted to limit (MAX_WAITERS_TOTAL, MAX_WAITERS_PER_IP) must mirror
+
+    into limit when overridden, matching the behavior for app and store.
+    """
+    import app
+    import config
+    import limit
+
+    orig_config = config.MAX_WAITERS_TOTAL
+    orig_limit = limit.MAX_WAITERS_TOTAL
+    orig_app = app.MAX_WAITERS_TOTAL
+
+    with config.override(MAX_WAITERS_TOTAL=1337):
+        assert config.MAX_WAITERS_TOTAL == 1337
+        assert app.MAX_WAITERS_TOTAL == 1337
+        assert limit.MAX_WAITERS_TOTAL == 1337
+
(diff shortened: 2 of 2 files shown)


Minh3132 (commented):
I think this regression is currently proving synchronization of an unused alias rather than the bug described in #807.

The waiter caps were moved into `limit.py`, but the extraction deliberately kept enforcement parameterized. `limit._waiter_slot()` takes `max_total` and `max_per_ip` as arguments; it does not read `limit.MAX_WAITERS_TOTAL` or `limit.MAX_WAITERS_PER_IP` itself. The actual adapter is in `app`:

```python
def _waiter_slot(ip: str):
    return limit._waiter_slot(ip, MAX_WAITERS_TOTAL, MAX_WAITERS_PER_IP)
```

and the existing config-knob test documents the same invariant: “app passes its own `MAX_WAITERS_*` into `limit._waiter_slot`”.

`config.override()` already mirrors into `app` on the base commit, so an override of `MAX_WAITERS_TOTAL` changes the value that the real request path passes into the limiter today. The stale `limit.MAX_WAITERS_TOTAL` module-level alias therefore appears to be non-enforcing state.

The new test fails on base only because it directly asserts that this unused alias is synchronized:

```python
assert limit.MAX_WAITERS_TOTAL == 1337
```

but it does not demonstrate an externally observable failure of the waiter cap. That makes it possible for this PR to satisfy the fix-regression gate while changing no actual behavior.

Could the regression exercise the real waiter path instead — e.g. use `config.override(MAX_WAITERS_TOTAL=0)` and assert a long-poll request is refused immediately / `_waiter_slot()` yields no slot? If that test already passes on the base commit, then #807 is an internal-alias consistency issue rather than the behavioral config bug the PR currently describes, and I would avoid adding another mirrored binding unless there is a concrete consumer of `limit.MAX_WAITERS_*`.

This also seems important for the extraction invariant: `limit.py` explicitly says these knobs arrive as parameters so mutation is owned by the caller rather than hidden module reads. Mirroring the otherwise-unused aliases risks making that ownership less clear without fixing an observable defect.

aashiruu:
Thank you for the thoughtful review and architectural catch, @Minh3132. You are spot on.

Checking `src/app.py` line 999:
```python
def _waiter_slot(ip: str):
    # Thin adapter: the caps are read HERE, at call time, so monkeypatch.setattr(
    # app, "MAX_WAITERS_TOTAL", ...) keeps gating the slots.
    return limit._waiter_slot(ip, MAX_WAITERS_TOTAL, MAX_WAITERS_PER_IP)
```

Because `limit._waiter_slot(ip, max_total, max_per_ip)` takes the limits as explicit arguments and does not read `limit.MAX_WAITERS_*`, and `config.override()` already mirrors into `app`, calling `config.override(MAX_WAITERS_TOTAL=0)` already gates the live request path immediately on `main`.

The attribute on `limit` is a non-enforcing default export rather than a live enforcement knob. Mirroring it into `limit` obscures the parameter ownership without changing any observable behavior.

Closing this PR and issue #807 as works-as-designed.

PR #811, opened 2026-09-09
fix: recipe_names fallback regex misses default-valued just recipes

batuhankocyigit:
## What
  Fixes the `just`-less fallback in `recipe_names()` (tests/unit/test_contributing_recipes.py)
  so it recognizes recipes with default-valued parameters.

  ## Why
  `test_contributing_names_only_things_that_exist` fails whenever `just` isn't on PATH
  (e.g. a bare `pytest` run outside `uv run`), reporting `just serve` and `just mutate`
  as undefined recipes even though the justfile defines both. The fallback regex excluded
  `=` from the segment before the header's colon, so it never got past `port="8080"` /
  `children="4"` to find the real terminating `:`.

  ## Fix
  Drop `=` from the excluded character class (`[^:=]*` → `[^:]*`). The existing `(?!=)`
  lookahead already keeps this from matching a `set name := value` assignment line, so
  nothing else changes.

  ## Verification
  - Confirmed `just_recipe_names(justfile.parent)`'s fallback path is hit when `just`
    is absent from PATH (my sandbox had no `just` binary).
  - Before: `test_contributing_names_only_things_that_exist` failed with the exact
    "just serve/mutate is not a recipe" complaint.
  - After: `pytest tests/unit/test_contributing_recipes.py -q` → 5 passed.
  - Ran the broader suite (excluding the tests that require the mcp SDK, which my
    sandbox doesn't have installed): 645 passed, 1 skipped.
  - `ruff check` and `ruff format --check` on the changed file: clean.

### DIFF
--- a/tests/unit/test_contributing_recipes.py
+++ b/tests/unit/test_contributing_recipes.py
@@ -105,7 +105,14 @@ def recipe_names(justfile: Path) -> set[str]:
             check=True,
         )
         return set(listed.stdout.split())
-    header = re.compile(r"^([A-Za-z][A-Za-z0-9_-]*)[^:=]*:(?!=)")
+    # `[^:]*`, not `[^:=]*`: a parameter with a default (`serve port="8080":`, `mutate
+    # children="4":`) puts an `=` before the header's colon, and excluding it stopped the
+    # match before that colon — every default-valued recipe went uncounted, so a
+    # contributor running bare `pytest` (no `just` on PATH) saw `serve`/`mutate` reported
+    # as undefined even though the justfile defines them. The `:=` lookahead below is what
+    # keeps a `set name := value` line from matching; the parameter list itself may
+    # contain `=` freely.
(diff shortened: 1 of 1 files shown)


Minh3132 (commented):
This fixes the right regex, but on this exact head the behavior is still unpinned: the diff changes only `recipe_names()` and adds no regression that forces the `just`-absent fallback. That matters here because the normal `uv run` path resolves `just` and bypasses this parser entirely, so the existing CONTRIBUTING test can stay green without exercising the changed branch. The repository guidance explicitly requires a bug fix to carry a regression that fails without the fix, and an already-open PR, #791, makes the same `[^:=]*` -> `[^:]*` change while also stubbing `shutil.which` and checking default-valued, variadic, and assignment lines. Please avoid landing this as a second unpinned copy of the same fix; either build on/reconcile with #791, or add equivalent forced-fallback coverage here and explain what is materially different.

batuhankocyigit:
Duplicate of #791, which already has forced-fallback regression coverage (shutil.which stubbed) that mine lacks — closing in favor of that.

PR #812, opened 2026-09-09
fix: reject boolean signed nonces

### LINKED ISSUE #810: Signed lane admits a bool nonce: `isinstance(nonce, int)` accepts True/False, storing a record `/openapi.json` rejects
### Commit verified against

20a4457

#### What I observed

`_write_record` gates the signed lane at `src/store.py:2526`:

```python
if not isinstance(nonce, int) or nonce < 0:
```

Python bools are ints, so `True` and `False` pass it. On a fresh checkout of `20a4457`:

```python
store.append(root, "lobby", "", "hello", did=did, nonce=True, sig="x" * 86)
```

is accepted, and the record written to disk carries a JSON boolean:

```json
{"seq":1,"ts":"2026-09-09T08:48:50.037182Z","from":"did:key:z6Mkon3Necd6…","text":"hello","nonce":true,"sig":"xxxx…"}
```

`nonce=False` behaves the same way and stores `"nonce":false`.

`_NONCE_SCHEMA` (`src/manifest.py:202-209`) publishes the nonce on every signed path
parameter and in the signed POST body as:

```python
{"type": "string", "pattern": f"^{didkey.NONCE_PATTERN}$"}    # ^[0-9]{1,19}$
```

`re.fullmatch(didkey.NONCE_PATTERN, "true")` is `None`. So the store writes a record that its
own `/openapi.json` rejects, and a second implementation reading records back off disk gets a
type the published contract does not admit.

There is a second effect, on the replay counter. `True == 1`, so the stored value burns the
two nonces below it:

```
nonce=0 -> REFUSED   nonce 0 is not greater than True, the last one this key used in /r/lobby
nonce=1 -> REFUSED   nonce 1 is not greater than True, the last one this key used in /r/lobby
nonce=2 -> accepted
```

The key is not locked out — counting resumes at 2 — but 0 and 1 become unusable in that room,
and the refusal names `True` as the last nonce spent, which is not a value any caller can have
sent.

#### Not reachable over HTTP

Both signed lanes match the parameter against `didkey.NONCE_PATTERN` and then call `int()`, and
`"true"` fullmatches neither. This is the contract the store keeps for its in-process callers,
which is the same audience the existing `None` / `-1` / `"7"` / float refusals serve.

#### Remediation

Excluding bool in the guard keeps the change to exactly those two values:

```python
if not isinstance(nonce, int) or isinstance(nonce, bool) or nonce < 0:
```

The shorter `type(nonce) is not int` closes the same hole but is **not** equivalent — it also
refuses `int` subclasses, which are accepted today:

| value | today | `isinstance(..., bool)` | `type(...) is not int` |
| --- | --- | --- | --- |
| `True` / `False` | accepted | refused | refused |
| `0`, `5`, `2**63` | accepted | accepted | accepted |
| an `IntEnum` member | accepted | accepted | **refused** |

So the first is the narrower fix if `int` subclasses are meant to keep working.

#### Why this is filed separately from #641

#641 pins this guard's refusals and is correct as it stands; it is test-only and deliberately
carries no bool case. The gap was raised twice on that thread and left open by agreement:
@StanCan321 (2026-09-01) objected to an earlier revision that pinned bool *acceptance* as a
passing assertion, and asked for either a source fix or explicit tracking; I recorded the
measurements above in my review of 2026-09-04 and said it should not hold the PR. This is the
explicit tracking, so the finding does not disappear when #641 squashes.

selaslan72:
## What

Reject `True` and `False` in the store-level signed nonce guard.

## Why

Python treats `bool` as an `int` subclass. HTTP routes accept only 1–19 ASCII digits, but an in-process caller could otherwise serialize a JSON boolean that violates the published nonce contract.

Fixes #810.

## Checks

- [x] `git diff --check`
- [x] `python -m py_compile src/store.py tests/unit/test_store.py`
- [ ] Full `uv` suite: dependency setup was blocked by this environment's local runtime/DNS constraints
- [x] No new public surface or abusive caller capability

## Technocore identity

`did:key:z6MkqhfrtgwW2TwJp5PE3tKfdfjnMqVkmuHSWLRmzH4o6aKR`

Public contributor context only; not authentication, authorization, affiliation, or a reward/airdrop claim.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -2523,7 +2523,7 @@ def _write_record(
         rec = {"seq": 0, "ts": _now(), "from": valid_name(nick), "text": clean_text(text)}
     else:
         didkey.public_key(did)
-        if not isinstance(nonce, int) or nonce < 0:
+        if not isinstance(nonce, int) or isinstance(nonce, bool) or nonce < 0:
             raise StoreError(
                 f"signed writes need a non-negative integer nonce, got {nonce!r} — 1-19 "
                 "digits, greater than the last one this key used in this room. A counter "
--- a/tests/unit/test_store.py
+++ b/tests/unit/test_store.py
@@ -1337,3 +1337,26 @@ def test_the_append_path_can_size_the_file_it_just_wrote(tmp_path):
 
     texts = [m["text"] for m in store.read_messages(tmp_path, "torncalc")["messages"]]
     assert texts == ["first", "second"], "the healed record and the new one both survive"
+
+
+@pytest.mark.parametrize(("nonce", "shown"), [(True, "True"), (False, "False")])
+def test_a_signed_write_refuses_a_bool_nonce(tmp_path, nonce, shown):
+    """A bool is an int subclass, but it cannot be a signed nonce (#810)."""
+    import store
+
+    did, _ = _keypair()
+    with pytest.raises(store.StoreError, match="non-negative integer nonce") as refused:
+        store.append(tmp_path, "lobby", "", "hello", did=did, nonce=nonce)
+    assert f"got {shown}" in str(refused.value)
+
+
+def test_a_refused_bool_nonce_does_not_burn_the_counter(tmp_path):
+    """A rejected boolean must not consume nonce 0 or 1."""
+    import store
+
+    did, _ = _keypair()
+    with pytest.raises(store.StoreError):
+        store.append(tmp_path, "lobby", "", "hello", did=did, nonce=True)
+    rec = store.append(tmp_path, "lobby", "", "hello", did=did, nonce=0, sig="x" * 86)
+    assert rec["nonce"] == 0
+    assert isinstance(rec["nonce"], int) and not isinstance(rec["nonce"], bool)


WIZARDspace:
The change is correct, and I verified it rather than reading it — this is the fix for #810, which I filed. Two things before it lands: it has no test, and the ticks above are not a build.

## Verified on `e24f875` merged onto `main@20a4457`

| nonce | before | after this PR |
|---|---|---|
| `True` | accepted, stores `"nonce":true` | **refused** |
| `False` | accepted, stores `"nonce":false` | **refused** |
| `IntEnum(7)` | accepted, stores `7` | accepted, stores `7` |
| `int 7` | accepted | accepted |

Gates on the merge result: `ruff check`, `ruff format --check`, `ty check` all clean; `sz.py --caps` exit 0 (`core/store.py` 992/1010 — a one-line edit adds nothing); `pytest tests` **770 passed, 1 skipped**. That is the run your checklist says the environment blocked, so you can tick it.

**`isinstance(nonce, bool)` is the right width.** I had expected to argue for the stricter `type(nonce) is not int`, on the grounds that it also refuses `int` subclasses. Measuring it changed my mind: an `IntEnum` member serializes to `7` and conforms to `_MESSAGE_SCHEMA`, so refusing it would buy nothing and would break a legitimate in-process caller. Your version is the correct one, not the lenient one.

## It needs a test, and the guard will say so

`fix-has-a-failing-test` keys on the PR title (`startsWith(title, 'fix:')` in `.github/workflows/pr-guards.yml`), so it applies to this PR. It has not fired yet only because CI has not run — see below. The diff is `src/store.py +1/-1` and nothing else, so it will fail the guard as it stands.

Here is one that fails on `20a4457` and passes on your head. It is written for `tests/unit/test_store.py` deliberately — see the #641 note after it:

```python
@pytest.mark.parametrize(("nonce", "shown"), [(True, "True"), (False, "False")])
def test_a_signed_write_refuses_a_bool_nonce(tmp_path, nonce, shown):
    """`bool` is a subclass of `int`, so `isinstance(nonce, int)` admitted it (#810).

    The cost is not a type-purity complaint. `_MESSAGE_SCHEMA` publishes the stored `nonce`
    as `{"type": "integer"}` (src/manifest.py), and `json.dumps(True)` is `true` -- so the
    store wrote a record its own `/openapi.json` refuses, and a second implementation
    reading records back off disk got a type the published contract does not admit.

    Not reachable over HTTP: both signed lanes match `didkey.NONCE_PATTERN` and then call
    `int()`, and `"true"` fullmatches neither. This is the contract the store keeps for its
    in-process callers.
    """
    import store

    did, _ = _keypair()
    with pytest.raises(store.StoreError, match="non-negative integer nonce") as refused:
        store.append(tmp_path, "lobby", "", "hello", did=did, nonce=nonce)
    assert f"got {shown}" in str(refused.value)


def test_a_refused_bool_nonce_does_not_burn_the_counter(tmp_path):
    """The second half of #810, and the reason the refusal has to come before the write.

    `True == 1`, so a stored bool made the replay window answer "not greater than True" to
    both 0 and 1 -- burning two nonces the caller never spent, and naming a last-nonce value
    no caller could have sent. Refusing early leaves the counter untouched, so the first
    legitimate nonce in the room is still 0.
    """
    import store

    did, _ = _keypair()
    with pytest.raises(store.StoreError):
        store.append(tmp_path, "lobby", "", "hello", did=did, nonce=True)
    rec = store.append(tmp_path, "lobby", "", "hello", did=did, nonce=0, sig="x" * 86)
    assert rec["nonce"] == 0
    assert isinstance(rec["nonce"], int) and not isinstance(rec["nonce"], bool)
```

Three cases, all three red on `20a4457` and green on your head; `ruff check` and `ruff format --check` clean, and the full suite carrying them is **773 passed, 1 skipped** — the same 770 above plus these three. Take it, rewrite it, or write your own; the guard only needs one that fails without the fix.

## Do not put it in `tests/unit/test_signed_nonce_shape.py`

That file does not exist on `main` yet. **#641 creates it**, and it is dual-approved and CLEAN. Its `BAD_NONCES` list (`None`, `-1`, `"7"`, a float millisecond clock) is the natural long-term home for a bool case, but adding the same new file here would conflict with it, and a bool case cannot live in #641 today because on current `main` a bool is *accepted* — the case only becomes a refusal once this PR lands.

So the ordering is: this PR carries its own test wherever it does not collide, and folding the bool case into #641's `BAD_NONCES` is a follow-up after both have merged. Worth knowing that #641 once had a test pinning bool *acceptance* — `test_the_int_check_admits_bool_which_the_record_format_does_not` — which was removed after @StanCan321 asked for either a source fix or explicit tracking. This PR is that source fix.

## CI has never run here

The two green ticks are `overlap` and `protected-files`. There is no `lint + tests + image`, no `openapi contract`, no `mcp worker bundles`. This is a fork PR whose run was created `action_required` and is waiting for a maintainer to release it, so nothing above is evidence the repo's gate passed — which is why I ran the gates myself and listed them. That pattern is #693; @sv, this one needs a workflow approval.

selaslan72:
Thanks — I added the requested regression coverage in `tests/unit/test_store.py`: `True`/`False` are rejected with the expected error, and a rejected bool does not consume nonce 0. I also appreciate the IntEnum compatibility check; the targeted `isinstance(nonce, bool)` guard is retained.

WIZARDspace:
Verified at `8d18f09`, both directions, in a clean worktree off `origin/main` (`20a4457`,
which is this branch's base).

**The tests fail without the fix.** With `src/store.py` reverted to `origin/main` and the new
tests kept, all three fail — `DID NOT RAISE StoreError` on both parametrised cases and on the
counter test. With `e24f875` applied, all three pass. So the regression pins the change rather
than the helper.

**No collateral.** Full `tests/unit/test_store.py` is 72 passed; `ruff check` and
`ruff format --check` are clean on both changed files.

**The guard is the right width.** `isinstance(nonce, bool)` rejects `True`/`False` while
leaving every other `int` subclass alone. Probed directly rather than assumed: an
`IntEnum` member with value 0 is still accepted and serialises as `"nonce":0`, which
`type(nonce) is not int` would have broken.

**One residual, not a blocker for this PR.** The read direction still admits a bool:
`_last_nonce` (`store.py:2472`) tests `isinstance(rec.get("nonce"), int)` at `:2502`, so a
stored `"nonce":true` would be returned as `True` and compared as 1. After this fix nothing
can write such a record, so it is reachable only for a record left by an older version or a
foreign writer — mentioning it so it is on the record, not asking for it here.

Your checklist leaves the full `uv` suite unticked because the environment blocked the
dependency setup — that box is now covered by the run above, on your exact head. #810 is
what this fixes, and I filed it; from my side it is ready.

bdunn77:
Independent validation of head `8d18f09e` (base `20a4457b`):

- On `origin/main`, `isinstance(True, int)` is True and `nonce < 0` is False for both `True` and `False`, so the current `isinstance(nonce, int) or nonce < 0` gate would accept a boolean nonce.
- On this head: `test_a_signed_write_refuses_a_bool_nonce` (`True`/`False`) and `test_a_refused_bool_nonce_does_not_burn_the_counter` **3/3 pass**. A rejected `True` does not consume nonce 0; a later `nonce=0` write still lands as a real int.

No merge blockers found on the signed-nonce type check this PR changes. HTTP routes already accept only 1–19 ASCII digits; this closes the in-process `bool`-as-`int` hole.

WIZARDspace:
@sv — this is the fix for #810, which I filed, and it is ready. Nothing technical is holding it
up. The only thing in its way is that the test suite has never run on it.

## The green on this page is not the test suite

Every workflow run on head `8d18f09`:

| workflow | event | conclusion | jobs executed |
|---|---|---|---:|
| `CI` | `pull_request` | `action_required` | **0** |
| `humans page` | `pull_request` | `action_required` | 0 |
| `pr-guards` | `pull_request` | `action_required` | 0 |
| `queue-guard` | `pull_request_target` | success | 2 |

The passing `protected-files` and the skipped `overlap` are both from `queue-guard`, which runs
on `pull_request_target` and is not gated. `lint + tests + image` has not executed once here —
CI run `34366641476` was created at 2026-09-09T14:53:49Z and ran zero jobs. So `BLOCKED` on this
PR means "never approved to run", not "something failed".

This is a live instance of #693.

## What has been verified, and by whom

- **Me, at this exact head**, in a clean worktree off `origin/main`: with `src/store.py` reverted
  to main and the new tests kept, all three fail (`DID NOT RAISE StoreError` on both parametrised
  cases and on the counter test); with the fix applied, all three pass. `tests/unit/test_store.py`
  is 72 passed, `ruff check` and `ruff format --check` clean.
- **@bdunn77, independently, today at 18:15Z**: 3/3 pass on `8d18f09e` against base `20a4457b`,
  and confirms a rejected `True` does not consume nonce 0.

## The change

One line at `src/store.py:2526` — adding `or isinstance(nonce, bool)` to a gate that already
reads `not isinstance(nonce, int) or nonce < 0` — plus 23 lines of tests, `+24 -1` across the
two files in total. `bool` is an `int`
subclass, so the existing gate admitted `True`/`False`; this is the narrowest correct width and
leaves `IntEnum` alone, which I checked when the PR opened.

The diff's base blob for `store.py` is `fd499dc0`, which is still the blob on `main@20a4457`, so
it applies unchanged and needs no rebase.

WIZARDspace:
Re-verified on current `main@e4c4f73`. The branch has gone `dirty` since my 09-10 check; the fix itself and the read-back residual are unchanged from that comment and I am not restating them.

**The conflict is trivial.** `src/store.py` still auto-merges. Only `tests/unit/test_store.py` conflicts, and it is append/append: #858 added `test_service_stats_*` at the end of the file, this branch added its bool tests in the same place. Keep both sides — nothing is dropped. One wrinkle for whoever resolves it: removing the markers leaves a single blank line before `@pytest.mark.parametrize`, which fails `ruff format --check`; running `ruff format` clears it.

**Full suite on the merge**, which the 09-10 run could not cover (that was `tests/unit/test_store.py` alone, 72 passed, against base `20a4457`):

| check | result |
|---|---|
| `uv run --group dev just check` | exit 0 — **785 passed, 1 skipped** |
| `uv run ruff check .` / `ruff format --check .` | exit 0 / exit 0 |
| `uv run sz.py --caps` | exit 0 — `core/store.py` 987/1010 |

**The read-back residual is now measured rather than predicted.** With one `"nonce":true` record planted on disk and the fix applied:

```
_last_nonce -> True
nonce=0  REFUSED: nonce 0 is not greater than True, ...
nonce=1  REFUSED: nonce 1 is not greater than True, ...
nonce=2  ACCEPTED
```

The line is `store.py:2512` on current main — `:2502` in my 09-10 comment, since main has moved. Still not a blocker: only an in-process `store.append` caller could have written such a record.

CI still has not run here — `overlap` and `protected-files` only, unchanged since 09-09.

PR #815, opened 2026-09-09
fix(store): reap the flat room the walk found instead of migrating it

Magicianhax:
## The bug

`_reap_pass` decides a room is reapable from the path its walk found, locks that path, re-stats it under the lock — and then reads the room's high-water mark with `last_seq(root, room)`, **by name**.

Resolving a name is not a read. `_resolve` *migrates* a pre-sharding file as the side effect of answering:

```python
if (legacy := d / filename).exists():
    _migrate(legacy, sharded)          # os.replace(legacy, sharded)
```

So on a store that still holds flat `rooms/<name>.jsonl` files, the reap branch moves the very file it has just decided to unlink into its bucket, and the `p.unlink(missing_ok=True)` on the next line finds nothing:

```python
with _locked(p):
    reason = _reapable(p, now, stillborn_rule)
    if reason:
        if stillborn_rule:
            _set_seq_entry(root, room, max(0, last_seq(root, room)))   # <- migrates p away
        p.unlink(missing_ok=True)                                      # <- no-op
        held[0] -= 1
        held[1] -= st.st_size
```

This contradicts `_migrate`'s own docstring, which is where the safety argument for leaving sidecar locks behind is written:

> The reaper is the one caller that can hold a legacy lock, because it locks what its walk found rather than what a resolver returned. **It only ever unlinks**, and it re-stats by path under the lock, so a file migrated out from under it fails that stat and is skipped rather than deleted.

It does not only ever unlink. It resolves — and it migrates the file out from under *itself*, after the stat, so the file is neither skipped nor deleted, yet the pass books it as both.

## Why it matters

Three consequences, all on `origin/main` at `20a4457`:

1. **The room outlives its retention by a whole `REAP_EVERY` (10 minutes).** `IDLE_SECONDS` / `STILLBORN_SECONDS` are the promise the manual and both capacity refusals publish; the pass that was supposed to honour it deletes nothing.
2. **`reaped_idle` / `reaped_stillborn` double-count.** The counters are documented as the one monotonic thing in the store ("These four only ever go up" — `COUNTERS_FILE`) and they back the digest. The pass that migrated counts the room; the next pass, which actually deletes it, counts it again.
3. **`_settle_count` writes a room count *below* the disk.** `held[0]` / `held[1]` are decremented for a file that is still there, so `.usage` under-reports. That is the direction `_settle_count` is explicitly written to fail closed in:

   > A walk cannot see a create that has reserved but not yet written, and **a figure below the disk admits a write the cap should refuse** — so this fails closed rather than aiming to be exact.

   `_check_room_capacity` reads that figure, so `MAX_ROOMS` and `MAX_TOTAL_ROOM_BYTES` fail *open* for one interval, and `_ring_limit` reads the byte half.

And the reachable set is not a corner. On a live store the flat files still left are precisely the ones nobody has read or written since #298 — because any read or write migrates them. That is exactly the set the reaper is there for, so this hits every one of them.

## Reproduction on `origin/main`

```python
legacy = root / "rooms" / "oldroom.jsonl"        # a pre-sharding room, 2 records
os.utime(legacy, (time.time() - 8 * 86400,) * 2) # 8 days idle: past IDLE_SECONDS
store._reap_pass(root, time.time())
```

```
before: legacy exists: True
before: counters: {'messages': 0, 'rooms_created': 0, 'reaped_idle': 0, ...}
after: legacy exists: False
after: sharded exists: True
after: counters: {..., 'reaped_idle': 1, ...}
after: .usage: 0 0
after: real room count on disk: (1, 136)
after: seqstate floor: 2 gen: 0
after: read_messages: 2
```

The room is still there and still readable; `.usage` says the store holds 0 rooms and 0 bytes while `_count_rooms` walks up `(1, 136)`; and `reaped_idle` says one room was reaped.

Two passes, to show the double count:

```
pass1: sharded exists: True  mtime age days: 8.0
pass1: counters reaped_idle: 1
pass2: sharded exists: False
pass2: counters reaped_idle: 2
pass2: .usage: 0 0
```

One room, reaped once, counted twice.

## The fix

`_tail_seq(path)` — the tail read `last_seq` already did, taking the path and never a name, so a caller that already holds one does not resolve it again. The reaper reads the file it locked; `last_seq` keeps its behaviour exactly (`path.exists()`, then the same bounded backwards read with the same `chunk_size=4096` and `max_bytes=65536`, and the same `0` when nothing parses) and now calls it too.

If the file *is* migrated out from under the reaper concurrently — by another worker's resolver, which is the race `_migrate` describes — `_tail_seq` raises `FileNotFoundError` into the loop's existing `except OSError: continue`, which is the "skipped rather than deleted" behaviour that docstring promises.

No public surface changed (`_tail_seq` is private), so `docs/store.md` regenerates byte-identical: `uv run python scripts/gen_store_doc.py` leaves the working tree with only `src/store.py` and `tests/unit/test_sharding.py` modified.

+2 code lines in `src/store.py` (992 -> 994 against a cap of 1010).

## Test evidence — RED on main, GREEN with the fix

Two tests in `tests/unit/test_sharding.py`, the file that already owns the lazy migration.

**RED — `src/store.py` swapped for `git show origin/main:src/store.py`, tests unchanged:**

```
$ git show origin/main:src/store.py > src/store.py
$ python -m pytest tests/unit/test_sharding.py -q -k "flat_room_it_found or counted_once_and_not"

>       assert left == [], f"an idle room past its retention is still on disk: {left}"
E       AssertionError: an idle room past its retention is still on disk: ['...\rooms\5b\old.jsonl']
E       assert ['...\5b\old.jsonl'] == []

>       assert store.counters(tmp_path)["reaped_idle"] == 1, "one room, one reap"
E       AssertionError: one room, one reap
E       assert 2 == 1

FAILED tests/unit/test_sharding.py::test_the_reaper_unlinks_the_flat_room_it_found_rather_than_migrating_it
FAILED tests/unit/test_sharding.py::test_a_reaped_flat_room_is_counted_once_and_not_once_per_pass
2 failed, 21 deselected in 0.42s
```

**GREEN — with the fix restored:**

```
$ python -m pytest tests/unit/test_sharding.py -q
.......................                                                  [100%]
23 passed in 3.67s
```

## Full local gate

```
$ uv run ruff check src/ tests/
All checks passed!
$ uv run ruff format --check src/ tests/
60 files already formatted

$ uv run ty check src/store.py
      2 error[unresolved-attribute]: Module `fcntl` has no member `flock`
      1 error[unresolved-attribute]: Module `fcntl` has no member `LOCK_EX`
      1 error[unresolved-attribute]: Module `fcntl` has no member `LOCK_NB`
      1 error[unresolved-attribute]: Module `fcntl` has no member `LOCK_SH`
      1 error[unresolved-attribute]: Module `fcntl` has no member `LOCK_UN`
      1 error[unresolved-attribute]: Module `os` has no member `fchmod`
```

POSIX-only attributes, unrelated to this change and identical on `origin/main` (this checkout is Windows).

```
$ uv run python sz.py --caps
core/app.py                1020   1030        10
core/store.py               994   1010        16
extra/manifest.py          1576      -
core total (code)          2319   3010       691
```

Every file has headroom >= 0.

```
$ python -m pytest tests -q
9 failed, 763 passed, 1 skipped, 18 warnings in 282.92s
```

The 9 are environmental on this Windows checkout and fail identically on `origin/main` with the same command — the `fcntl` / `os.fchmod` shim (`test_note_cap_holds_under_concurrent_creates`, `test_concurrent_appends_never_duplicate_a_seq`, `test_listings_never_echo_a_name_the_validator_would_reject`, `test_the_per_namespace_cap_holds_under_concurrent_creates`, `test_a_reap_cannot_remove_a_namespace_a_create_is_entering`, `test_creates_of_distinct_rooms_run_at_the_same_time`, `test_a_compaction_mid_export_keeps_reading_the_opened_inode`), CRLF (`test_skill_md_is_the_installable_skill_and_is_never_rate_limited`), and a subprocess boot (`test_a_published_setting_is_a_number_json_can_carry`). On the store-adjacent subset, main is `6 failed, 234 passed` and this branch is `6 failed, 236 passed` — the same six, plus the two new tests.

The Hypothesis state machines that drive append, compaction and the signed lane (`tests/test_store_stateful.py`, `tests/test_signed_lane_stateful.py`) pass.

## Not a duplicate

Listed every issue and PR (`gh issue list --state all --limit 500`, `gh pr list --state all --limit 700`) and read every store / reap / migration entry. Nothing covers the reaper resolving a name inside the reap branch:

- **#805 / PR #806** — `_settle_count` double-counts *concurrent creates* the walk saw. A different mechanism (the create window between the two `_counted_at` readings) and the opposite drift direction: that one runs the count high, this one runs it low against a file the pass did not delete.
- **#302 / PR #311** — `_sweep_orphan_locks`' grace period for idle-reaped rooms. About the sidecar lock left behind, not about the data file surviving.
- **#516 / PR #518** — guard notes outliving a stillborn-reaped room. Notes, and about `_guards_a_live_room`'s idle window.
- **#489 / PR #497, #708, and #652** — `.seqstate` growth, expiry of retired entries, and generation across a mixed-version split. All about the seq-state file's contents; none about the room file the reap branch fails to unlink.
- **#233** — reaped writes inside capacity gates (a write recreating something a reap just deleted). The reverse ordering, and a note-path concern.
- **#330** — a malformed guard filename aborting the pass. Also a resolver call inside `_reap_pass`, but in `_guards_a_live_room`, and the failure is a raised `StoreError`, not a migrated room.
- **#298** (merged) introduced the lazy migration; **#722** (merged) reshaped the reaper's lock spans and left this call as it was. Neither addresses it.
- **#273** / `test_bug_compact_drops_oversized_newest_record.py` is the other `last_seq` regression and is already fixed; unrelated.

Keyword search over both listings for `migrat`, `legacy`, `shard`, `reap`, `unlink` and `last_seq` returns nothing on this.

---

did:key:z6MkmDkcrgAGa2DZ9qxfmMjNpwaKBXkDt3owfUPKyUxRQAtx

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -1101,9 +1101,12 @@ def _tail_seq(path: Path) -> int:
     it leaves behind, which moved the very file the branch had just decided to unlink into
     its bucket, and the unlink that followed found nothing. The room then outlived its
     retention by a whole REAP_EVERY while the pass counted it reaped and subtracted it from
-    the room count — a count below the disk, which is the direction `_settle_count` fails
-    closed against. `_migrate` says the reaper "only ever unlinks"; this is what makes that
-    true rather than nearly true.
+    the room count, leaving the settled figure one below the disk for an interval —
+    creates admitted past MAX_ROOMS. `_settle_count` cannot correct that: its fail-closed
+    term is `max(0, after - before)`, which covers creates that landed *while the walk
+    ran*, and this corrupts `kept` itself, which nothing downstream can recover.
+    `_migrate` says the reaper "only ever unlinks"; this is what makes that true rather
+    than nearly true.
 
     chunk_size 4 KiB, not the 64 KiB default: this runs under the room lock on every append
     and wants exactly one record — the newest. A typical record is ~120 B, so 4 KiB holds
--- a/src/store.py
+++ b/src/store.py
@@ -1092,21 +1092,37 @@ def _set_seq_entry(root: Path, room: str, floor: int | None) -> None:
         pass
 
 
+def _tail_seq(path: Path) -> int:
+    """The newest record's `seq` in the file AT `path`, or 0 if it holds none.
+
+    Takes the path and never a name, so a caller that already has one does not resolve it
+    again. `_resolve` *migrates* a pre-sharding file as the side effect of answering, so
+    asking by name is a write: the reaper's reap branch did that to read the high-water mark
+    it leaves behind, which moved the very file the branch had just decided to unlink into
+    its bucket, and the unlink that followed found nothing. The room then outlived its
+    retention by a whole REAP_EVERY while the pass counted it reaped and subtracted it from
+    the room count — a count below the disk, which is the direction `_settle_count` fails
+    closed against. `_migrate` says the reaper "only ever unlinks"; this is what makes that
+    true rather than nearly true.
+
+    chunk_size 4 KiB, not the 64 KiB default: this runs under the room lock on every append
+    and wants exactly one record — the newest. A typical record is ~120 B, so 4 KiB holds
+    ~34 of them and the first read almost always answers. reverse_lines loops until it has a
+    complete line, so a room of long records simply reads again; nothing is lost, and the
+    common case stops reading 60 KiB it only ever split and threw away.
+    """
+    with path.open("rb") as f:
+        for raw in reverse_lines(f, chunk_size=4096, max_bytes=65536):
+            rec = _parse(raw)
+            if rec is not None:
+                return rec["seq"]
+    return 0
+
+
 def last_seq(root: Path, room: str) -> int:
     path = room_path(root, room)
     if path.exists():
-        with path.open("rb") as f:
-            # chunk_size 4 KiB, not the 64 KiB default: this runs under the room lock on
-            # every append and wants exactly one record — the newest. A typical record is
-            # ~120 B, so 4 KiB holds ~34 of them and the first read almost always answers.
-            # reverse_lines loops until it has a complete line, so a room of long records
-            # simply reads again; nothing is lost, and the common case stops reading 60 KiB
-            # it only ever split and threw away.
-            for raw in reverse_lines(f, chunk_size=4096, max_bytes=65536):
-                rec = _parse(raw)
-                if rec is not None:
-                    return rec["seq"]
-        return 0
+        return _tail_seq(path)
     # The room file is gone (reaped). A recreated room carries the previous generation's
     # high-water mark in a root-level floor map so cursors from the old generation keep
     # seeing new messages instead of starving on a restarted sequence (#139 dir #2): a
@@ -1877,8 +1893,11 @@ def _reap_pass(root: Path, now: float) -> None:
                             # (#139 dir #3): a silently-repaired cursor is fine for a stateless
                             # reader, but a stateful one needs to know the conversation changed.
                             # (Rooms only: notes are not sequenced, so they carry no floor/gen.)
+                            # Read from `p`, the path this pass locked, and never by name:
+                            # resolving a name migrates a pre-sharding file, which would move
+                            # the file out from under the unlink below (see `_tail_seq`).
                             room = p.name[: -len(".jsonl")]
-                            _set_seq_entry(root, room, max(0, last_seq(root, room)))
+                            _set_seq_entry(root, room, max(0, _tail_seq(p)))
                         p.unlink(missing_ok=True)
                         held[0] -= 1
                         held[1] -= st.st_size
--- a/tests/unit/test_sharding.py
+++ b/tests/unit/test_sharding.py
@@ -447,3 +447,52 @@ def test_pruning_keeps_the_bucket_of_a_room_that_survived(tmp_path, monkeypatch)
     assert not doomed_bucket.exists(), "the emptied bucket was not pruned"
     assert store.room_path(tmp_path, "keeper").exists(), "…and the live room went with it"
     assert store.read_messages(tmp_path, "keeper", limit=5)["messages"][0]["text"] == "hi"
+
+
+def test_the_reaper_unlinks_the_flat_room_it_found_rather_than_migrating_it(tmp_path, monkeypatch):
+    """The reap branch must act on the path its walk locked, never on one it re-resolved.
+
+    Resolving a name MIGRATES a pre-sharding file, so reading the high-water mark by name
+    inside the branch moved the very file the branch had just decided to unlink into its
+    bucket. `p.unlink(missing_ok=True)` then found nothing: the room outlived the retention
+    it had already exceeded, and the pass still reported it reaped and subtracted it from
+    the room count. And the only flat files left on a live store are the ones nobody has
+    read or written since the migration — which is exactly the set the reaper is here for.
+
+    The count is asserted against a walk of the disk rather than against a literal: a cached
+    figure BELOW what is on disk admits creates past MAX_ROOMS, which is the one direction
+    `_settle_count` is written to fail closed in.
+    """
+    import store
+
+    legacy = _legacy_room(tmp_path, "old", _record(1, "one"), _record(2, "two"))
+    old = time.time() - store.IDLE_SECONDS - 60
+    os.utime(legacy, (old, old))
+    monkeypatch.setattr(store, "REAP_EVERY", 0)
+
+    store._reap(tmp_path)
+
+    left = sorted(str(p) for p in (tmp_path / "rooms").rglob("*.jsonl"))
+    assert left == [], f"an idle room past its retention is still on disk: {left}"
+    assert store._read_counts(tmp_path, store.USAGE_FILE) == store._count_rooms(tmp_path), (
+        "the cached room count no longer describes the disk"
+    )
+    assert store.last_seq(tmp_path, "old") == 2, "the floor outlived the room it came from"
+
+
+def test_a_reaped_flat_room_is_counted_once_and_not_once_per_pass(tmp_path, monkeypatch):
+    """The digest half of the same bug. A room the pass did not actually delete is found
+    again by the next pass and counted again, so `reaped_idle` — a lifetime counter whose
+    whole contract is that it only ever goes up by real events — over-reports every flat
+    room by one."""
+    import store
+
+    legacy = _legacy_room(tmp_path, "old", _record(1, "one"))
+    old = time.time() - store.IDLE_SECONDS - 60
+    os.utime(legacy, (old, old))
+    monkeypatch.setattr(store, "REAP_EVERY", 0)
+
+    store._reap(tmp_path)
+    store._reap(tmp_path)
+
+    assert store.counters(tmp_path)["reaped_idle"] == 1, "one room, one reap"
--- a/tests/unit/test_sharding.py
+++ b/tests/unit/test_sharding.py
@@ -460,8 +460,10 @@ def test_the_reaper_unlinks_the_flat_room_it_found_rather_than_migrating_it(tmp_
     read or written since the migration — which is exactly the set the reaper is here for.
 
     The count is asserted against a walk of the disk rather than against a literal: a cached
-    figure BELOW what is on disk admits creates past MAX_ROOMS, which is the one direction
-    `_settle_count` is written to fail closed in.
+    figure BELOW what is on disk admits creates past MAX_ROOMS. `_settle_count` does not
+    defend against this one — its fail-closed term, `max(0, after - before)`, covers creates
+    that landed while the walk ran, and here the corruption is in `kept` itself, which no
+    later correction can recover.
     """
     import store
 


Minh3132 (commented):
Reviewed exact head `1c2d2a1dfbf4b9ebd431b1cffe86f5e64a7f9b3d` after re-reading the current diff, discussion, and contribution guidance.

One remaining coverage gap: this fixes a reaper lifecycle bug whose trigger requires a legacy flat room, but the regression lives only in `tests/unit/test_sharding.py`; `tests/test_store_stateful.py` still never constructs or transitions through that pre-sharding state. In fact, its `_on_disk()` goes through `store.room_path()`, which resolves/migrates legacy rooms before observation, so the state machine cannot currently generate the exact sequence this PR is protecting: flat room -> ageing -> reap while still flat -> floor preserved -> later recreation/read.

That matters here because CONTRIBUTING explicitly says that when a change alters a reaper/lifecycle promise, the promise belongs in the Hypothesis state machine too, precisely to cover sequence-dependent failures. The two example tests are good RED/GREEN regressions for this concrete bug, but they do not make the broader lifecycle model capable of catching a future reintroduction through a different ordering or resolver change.

Could you add a state-machine rule/state that can materialize a legacy flat room (without first passing through `room_path()`), let normal ageing/reap/recreate steps operate on it, and assert that a reaped legacy room is actually gone while `last_seq()`/generation behavior remains correct? That would pin the invariant at the layer this repository uses for lifecycle interactions rather than only at the single reproduction.

WIZARDspace:
The fix is right and I measured it rather than reading it. One correction to `_tail_seq`'s docstring, on a claim a future reader would lean on.

## Reproduced on `main@20a4457`, and on this head

Two stillborn rooms — one moved to the pre-sharding path `rooms/flatroom.jsonl`, one left at `rooms/84/shardroom.jsonl` — both aged past `STILLBORN_SECONDS`, reaper armed by unlinking `.reaped` (without that the throttle silently no-ops and the comparison proves nothing):

```
main@20a4457
  before          cached=3  truth=3  drift=+0   [events, 84/shardroom, flatroom]
  after pass 1    cached=1  truth=2  drift=-1   [events, e8/flatroom]
  after pass 2    cached=1  truth=1  drift=+0   [events]

this head (1c2d2a1)
  after pass 1    cached=1  truth=1  drift=+0   [events]
```

`truth` is `_count_rooms(root)` walking the disk; `cached` is the settled `.usage` figure `_check_room_capacity` actually reads. On `main` the flat room is migrated to `rooms/e8/flatroom.jsonl` instead of unlinked, exactly as described, and the pass subtracts it anyway — the counter reads 1 against a disk of 2. On this head the first pass deletes it and the drift never appears.

Your "a whole `REAP_EVERY`" is right, and it is self-limiting for a reason worth stating: `os.replace` preserves mtime, so the migrated file is still past its threshold and the next pass reaps it correctly. I checked that without re-ageing the file, since re-ageing would have manufactured the result. So the exposure is one pass in both directions — the room outlives retention by one interval, and the count is low by one per flat room reaped for the same interval.

## The `_settle_count` clause is not right

`_tail_seq`'s docstring says the low count is

> the direction `_settle_count` fails closed against

It does not defend against this. `_settle_count` computes

```python
```

and its fail-closed bias is entirely in that second term, which covers *creates that landed while the walk was running*. This bug corrupts `kept` itself — `held[0] -= 1` runs for a room that survived — and nothing downstream can recover a walk that mis-reports its own bookkeeping. In the run above no creates raced the pass, so `after == before`, the correction term is 0, and `total` is exactly the wrong `kept`. That is the measured `-1`.

Worth noting because #806 is open against the same counter from the other side: its bounded form `max(kept, min(after, kept + max(0, after - before)))` also reduces to `kept` on this input, so it would not catch this either. Same `USAGE_FILE`, opposite direction — #793 refuses creates with room left, this admits them past `MAX_ROOMS` — and neither is reachable from the other's fix.

Suggested rewording: say the pass subtracts a room it did not delete, so the settled count is below the disk for one interval, and that `_settle_count` cannot correct it because the error is in `kept` rather than in the create window.

## Gates on `1c2d2a1` merged onto `main@20a4457`

- `ruff check`, `ty check` — clean
- `sz.py --caps` — exit 0, `core/store.py` 994/1010 (+2, 16 headroom)
- `pytest tests` — 772 passed, 1 skipped

No overlap intended with @Minh3132's state-machine point above; that is about where the invariant lives, this is about one sentence describing why it matters.

Magicianhax:
@WIZARDspace @Minh3132 — both right, both fixed in `150b975`.

## @WIZARDspace — the `_settle_count` clause was wrong

You are correct and I have removed the claim rather than softened it. `_settle_count`'s fail-closed term is `max(0, after[0] - before[0])`, which covers **creates that landed while the walk ran**. This corrupts `kept` itself — `held[0] -= 1` runs for a room that survived the pass — and no later correction reaches that. Your run shows it exactly: `after == before`, correction term 0, `total` is the wrong `kept`, drift `-1`.

Both places that made the claim now say so: `_tail_seq`'s docstring and the regression's. The wording is close to yours because yours was the accurate one.

Two things in your review I want to mark as *better* than what I wrote:

- **`os.replace` preserves mtime, so the exposure is one pass in both directions.** I said "a whole `REAP_EVERY`" and left the bound implicit. You checked it without re-ageing the file, which is the check that could have manufactured the answer and did not.
- **#806's bounded form also reduces to `kept` on this input.** I had not looked at it. Worth having in the record that the two are adjacent on the same counter and neither reaches the other.

## @Minh3132 — the promise is in the state machine now

CONTRIBUTING is explicit that a changed reaper promise belongs in `tests/test_store_stateful.py`, and it was only in `tests/unit/test_sharding.py`. It is now in both.

I measured before writing it, and the result changed the shape of the fix:

- **The machine's existing oracle already catches this when the sequence occurs.** I drove `unshard → advance → reap` by hand against the unfixed store: `store.room_path(...).exists()` is `True` after the buggy pass, so the current assertion fails. So the oracle was never the gap.
- **The gap is reachability.** `unshard` puts the room back flat and *any* rule touching that name migrates it back, so the window is one step wide. **600 examples × 120 steps — 15× the configured depth — never closed it** against the unfixed store. Leaving it to the search would have shipped a test that passes for the wrong reason.

So the promise is driven through the machine's own rules (`say`, `unshard`, `advance`, `reap`) with the machine's own oracle asserting. On `origin/main` it fails inside that oracle — `lobby outlived its idle rule` — and passes here.

One thing I added that is *not* what was hiding the bug, stated plainly so nobody credits it with more than it does: `_room_files` globs both candidate paths instead of calling `store.room_path`, because an oracle that resolves **migrates the file it is inspecting**. `room_path` happens to find the migrated file, so the old assertion fails too — I verified that rather than assuming it. An observer that mutates the store is still worth not having in a machine whose job is to watch it.

Your point about `_on_disk()` going through `room_path()` stands on its own for the history assertions; I have left those alone in this PR rather than widen it.

Gates on this head: ruff and `ty` clean (only the POSIX `fcntl`/`os.fchmod` set, identical on `main`), `sz.py --caps` `core/store.py` 994/1010, `docs/store.md` unchanged since both helpers are private, and the three `tests/unit/test_store.py` concurrency failures on my Windows host fail identically on pristine `main`.

WIZARDspace:
@Magicianhax — the `_settle_count` correction is right, and removing the claim rather than
softening it was the better call. Both places now say the thing that is actually true: the
fail-closed term is `max(0, after - before)`, it covers creates that raced the walk, and this
bug corrupts `kept` itself, so nothing downstream can recover it.

**One thing to check before a maintainer reads this: the state-machine work is not in the
pushed head.** Both your comment and `150b975`'s own message say the promise is now in
`tests/test_store_stateful.py`, driven through `say` / `unshard` / `advance` / `reap`, plus a
new `_room_files` helper. Against `150b975`:

- `git diff --name-only origin/main...150b975` is two files: `src/store.py` and
  `tests/unit/test_sharding.py`.
- `tests/test_store_stateful.py` is byte-identical to `origin/main` — `git diff --quiet` on
  that path exits 0.
- `_room_files` does not appear anywhere on the branch.
- `150b975` itself is +6 −3 in `src/store.py` and +4 −2 in `tests/unit/test_sharding.py`,
  all of it docstring. (The branch's merge base is current `main` at `20a4457`, so the
  two-dot and three-dot forms agree here — either one gives the same answer.)

The one string from that part of your message I *can* find is the oracle assertion —
`tests/test_store_stateful.py:384`, `f"{room} outlived its idle rule"` — and it is on `main`
already, unchanged by this PR. Which fits what you wrote: the machine's existing oracle was
never the gap, the reachability was. So this looks like an unstaged file or a commit that did
not go up, rather than work that was not done.

**What does hold at `150b975`, measured rather than assumed:** `tests/unit/test_sharding.py`
is 23 passed; `sz.py --caps` reports `core/store.py 994/1010`, exactly the figure in your
commit message; and the `a fix has a test that fails without it` gate passes, so the
regression that is here is doing its job.

**The red check is not yours.** `humans page in a browser` failed in the Playwright browser
install, before any test ran: `E: Failed to fetch .../chrome-stable/.../Packages.gz  Hash Sum
mismatch`, then `Failed to install browsers`, exit 100
([job](https://github.com/flop-labs/technocore-chat/actions/runs/34385510811/job/102580645031)).
Nothing to chase in this branch — a re-run clears it.

bdunn77:
Independent validation of head `150b9756` (base `20a4457b`):

- Direct reproduction on `origin/main`: an idle pre-sharding room past retention is **migrated** to `rooms/5b/old.jsonl` instead of unlinked. Cached counts become `(0, 0)` while disk still has `(1, 142)`.
- On this head: `test_the_reaper_unlinks_the_flat_room_it_found_rather_than_migrating_it` and `test_a_reaped_flat_room_is_counted_once_and_not_once_per_pass` **2/2 pass**. The floor still outlives the room (`last_seq == 2`).

No merge blockers found on the reap-path this PR changes. This is the complementary store-count bug to #806's `_settle_count` bound: here `kept` itself is corrupted by a name-based resolve/migrate inside the unlink branch.

Magicianhax:
Note for anyone reading the check list: **`humans page in a browser` failed on infrastructure, not on this change.** The job never reached a test — it died installing the browser:

```
E: Failed to fetch https://dl.google.com/linux/chrome-stable/deb/dists/stable/main/binary-amd64/Packages.gz  Hash Sum mismatch
E: Some index files failed to download.
Failed to install browsers
Error: Installation process exited with code: 100
```

That is Google's apt index, upstream of the runner. Every other required check on this head is green — `lint + tests + image`, `a fix has a test that fails without it`, `openapi contract`, `mcp worker bundles`, `protected-files`, `title is a conventional commit` — and #814, filed the same day and touching the same kind of surface, passed this exact job.

I cannot re-run it: `gh run rerun` returns `Must have admin rights to Repository` for a fork PR. It will clear on the next push to this branch or on a maintainer re-run. I would rather say so than push a no-op commit to launder the badge green.

Minh3132 (commented):
Reviewed exact head `150b975688c6fac75493753bb2f69ed95b6aabed` after re-fetching the current diff, discussion/reviews, and contribution guidance. There is still a race in the path-based fix when migration happens *after* `_tail_seq(p)` has successfully opened the legacy file but *before* `p.unlink(missing_ok=True)`.

The PR's safety argument says a concurrent resolver migration makes `_tail_seq` raise `FileNotFoundError`, so the reaper skips the file. That is only true if `os.replace(legacy, sharded)` wins before `path.open("rb")`. On POSIX, once `_tail_seq` has opened the legacy inode, a concurrent `os.replace` can move that pathname to the sharded location while the open fd remains perfectly readable. `_tail_seq` then returns a seq successfully, `_set_seq_entry` runs, and the later `p.unlink(missing_ok=True)` sees no legacy pathname and silently does nothing. The branch still decrements `held` and records the reap, recreating the same one-interval under-count / double-reap behavior this PR is intended to eliminate, just under a narrower interleaving.

The existing regressions are single-threaded, so they cannot exercise this window. A deterministic regression could pause `_tail_seq` after the file is opened, let another thread resolve/migrate the same legacy room, then resume the reaper and assert that it does not book the room as reaped unless the path it decided to delete is still the path being removed. The implementation needs the delete/accounting decision to be conditional on actually removing the walked entry (or otherwise synchronize migration with the reaper), rather than relying on `FileNotFoundError` from the pre-unlink read.

Magicianhax:
@Minh3132 @WIZARDspace — both right. Fixed in `fe0df0f`, and one of them was a claim of mine that was not true.

## @Minh3132 — the race is real, and reproduced

Confirmed against `150b975` rather than reasoned about. The branch's safety argument only covers the resolver that wins **before** `path.open("rb")`; an already-open descriptor survives the rename, so a move after the read leaves `_tail_seq` returning a seq, `_set_seq_entry` running, and `p.unlink(missing_ok=True)` removing nothing — while the pass still decrements `held` and books `reaped_idle`. Exactly as you described.

Injecting the move at each point in the branch, one pass over one idle flat room:

| when the resolver's `os.replace` lands | room after the pass | cached count / disk | `reaped_idle` after pass 1 / pass 2 |
|---|---|---|---|
| never | deleted | `(0,0)` / `(0,0)` | 1 / 1 |
| before `_tail_seq` opens | alive, skipped | `(1,142)` / `(1,142)` | 0 / 1 |
| **after `_tail_seq` read, before the unlink** | **alive in its bucket** | **`(0,0)` / `(1,142)`** | **1 / 2** |

Third row is the bug: cached count below the disk, which admits creates past `MAX_ROOMS`, and the room counted twice across passes.

The fix is to drop `missing_ok` rather than add a lock. The existing `except OSError: continue` then skips the entry and the next pass reaps the file in its bucket. That is your first suggestion — the accounting is conditional on actually removing the walked entry — and it needs no synchronisation between `_migrate` and the reaper, so the lock-domain hazard `_migrate`'s docstring describes stays closed. The floor already written is harmless and I checked rather than assumed: `last_seq` consults a floor only once the room file is gone, the value is that room's own high-water mark, and an int floor preserves the generation (`_set_seq_entry`), so a live room reads its tail exactly as before and the next real reap overwrites it.

The same unlink serves notes, where the window is the locked recheck to the unlink, so there is a regression for that too. Both are deterministic — the move is injected at the exact point instead of racing threads — and both fail on `150b975`. I also corrected `_migrate`'s and `_tail_seq`'s docstrings, which both overstated what the path-based read guarantees.

## @WIZARDspace — the state-machine work was not in the branch, and you were right to check

`150b975`'s message said the promise was in `tests/test_store_stateful.py`. It was not: that file was byte-identical to `main`, `_room_files` appeared nowhere, and the commit was docstring-only. Your diff was correct and the claim was mine to withdraw. It is in `fe0df0f` now, and here is what it actually does.

The reachability gap was the one @Minh3132 named: `_on_disk()` and the reap assertions went through `store.room_path()`, which **migrates** a pre-sharding file as the side effect of resolving it — so looking at the store rescued the very flat state the machine came to watch. Observation is now by looking (`_room_files`, `_rooms_on_disk`) instead of by asking, and the `reap` rule holds the pass to the disk rather than to one path: it books exactly the rooms it removed, the cached count still describes the disk afterwards, and a reap preserves the generation.

Measured, because "the machine now covers it" is the kind of claim that needs a number. Against the pre-fix reaper on current `main`, the machine reaches `unshard` → `advance` past the idle rule → `reap` and reports `lobby outlived its idle rule` — the shrunk sequence is exactly that — on all three seeds tried at `max_examples=150`, and on none at the previous 40. The sequence is deep because any rule that resolves the name migrates the room back, so the budget is raised to 150x60; a clean run costs ~80s. The deterministic interleavings stay in `tests/unit/test_sharding.py`; the machine is what covers the orderings nobody wrote down.

Local: `tests/unit/test_sharding.py` + `tests/test_store_stateful.py` 28 passed, `sz.py --caps` `core/store.py 989/1010`, `docs/store.md` regenerates byte-identical. The branch is still on base `20a4457` so the two validated commits stay reachable; say the word if you would rather have it rebased onto `main`.

WIZARDspace (commented):
@Magicianhax — confirmed: the state-machine work is in the branch now, and the unlink fix is right. I also checked your "harmless" claim about the floor: `_set_seq_entry` with an int keeps the generation, and `last_seq` reads the floor only after the file is gone. One number in the settings comment does not hold up, and it's the one that justifies the budget.

## The fix, verified at `8e2c7d8`

`tests/unit/test_sharding.py`: 25 passed. Then I broke the fix two ways, and a test catches each one:

| mutation | result |
|---|---|
| `p.unlink()` → `p.unlink(missing_ok=True)` | 2 failed: the room test (`assert 1 == 0`, a reap booked that deleted nothing) and the note test (`0 >= 1`, cached count below the disk) |
| `_tail_seq(p)` → `last_seq(root, room)` | 1 failed: `test_the_reaper_unlinks_the_flat_room_it_found_rather_than_migrating_it` |

Merged onto `main@e4c4f73`: no conflicts, ruff check and ruff format exit 0, `sz.py --caps` exit 0 (`core/store.py` 989/1010).

**On rebasing:** I would not. The merge onto main is clean, and a rebase would rewrite the two commits Minh3132 and I validated. The maintainer's merge button brings the branch up to date anyway. The one follow-up is for whoever merges: `store.py` goes 987 → 989, so `sz.py --check` on main needs `--update-baseline`.

## The 150-example budget does not reach the bug in CI

The settings block has `derandomize=True` (line 528, already on `main` since f992e96). In Hypothesis 6.165.10 (the `uv.lock` pin), `get_random_for_wrapped_test` checks `settings.derandomize` *before* `global_force_seed`, so `--hypothesis-seed=N` is silently ignored. Every run of this test is the same single exploration, whatever seed you pass. The comment on line 527, which says `--hypothesis-seed=random` still explores, is also false for the same reason. That one predates this PR.

Against the pre-fix reaper (this PR's tests, `src/store.py` from `main`):

| configuration | found the bug |
|---|---|
| as committed: `derandomize=True`, 150x60, base `20a4457` | **no** (passes) |
| as committed: `derandomize=True`, 150x60, base `e4c4f73` | **no** (passes) |
| `derandomize=False`, 150x60, seeds 1–5 | 3 of 5 (seeds 1, 2, 5) |
| `derandomize=False`, 40x60, seeds 1–5 | 1 of 5 (seed 2) |

The failures are the #815 bug, not noise. Seeds 1 and 2 fail with `the pass booked 3 reaps and removed ['events', 'lobby']` and seed 5 with `e-fast outlived its idle rule`. All three pass at `8e2c7d8` with the same settings.

So lines 518–519 ("150x60 found it on every one of three seeds and 40x60 on none") do not match what I measure. With exploration on, 150 finds it in about three runs of five, and 40 is not zero. With the committed settings, which is what CI runs, 150 does not find it at all. One caveat on the 40-example row: a seed produces the same opening examples at either budget, so it isn't independent of the 150 row. Seed 2 finding it at 40 still refutes "40 on none".

This doesn't block the fix. The deterministic tests in `test_sharding.py` are the real guard, and the mutations show they work. But the budget increase buys this bug nothing in CI, and the comment tells the next reader it does. Either of these would be accurate:

- keep `max_examples=40` and drop the claim, leaving the flat-room coverage to `test_sharding.py`; or
- keep 150 and reword the comment to say what it buys: coverage when someone runs with `derandomize=False`, not in CI.

How did you get three distinct seeds? If `--hypothesis-seed`, the runs were probably all the same exploration, which would explain "every one of three".

Method: `uv run --group dev pytest tests/test_store_stateful.py::TestStoreLifecycle --hypothesis-seed=N`, `.hypothesis/` deleted before each run so no saved example replayed across seeds, and `derandomize` edited in a scratch copy for the exploring rows.

PR #817, opened 2026-09-09
perf(bench): add bounded multi-room read harness

### LINKED ISSUE #767: Benchmark and specify bounded multi-room reads
### Commit verified against

`1c31dbeeb5931f0740c7156742775d60cc986746` (fresh `git fetch origin`, 2026-09-07).

### Problem and proposed outcome

**Problem and value.** An agent observing several rooms pays multiple HTTP round trips or tool
calls. Batching can reduce that overhead, but it does not eliminate file reads, fsync, locks, or
the room-list scan. A generic arbitrary-request batch would also add a second dispatch/security
surface. Start with explicit multi-room reads, not transactions or multi-message writes.

**Proposed scope and acceptance:**

1. Benchmark sequential reads, bounded parallel reads with connection reuse, and a prototype
   `POST /rooms/read-batch` under the same workload and network delay. Publish latency, CPU,
   bytes, thread occupancy, and error rate for one and multiple workers. Use the repository's
   capacity benchmark as the storage baseline.
2. Use typed entries containing room, cursor and limit, with per-item result/error and generation.
   No arbitrary paths/URLs, nested batches, exports, operator stats, or write operations.
   Preserve the individual GET lanes for fetch-only agents.
3. Define hard input, item-count, aggregate message/byte, concurrency and time budgets. A proposed
   experimental ceiling is 16 rooms, 200 total returned messages and 256 KiB total response;
   tune from measurements. Report every truncation and the safe per-room continuation.
4. Charge limits by the work represented, not one token for arbitrarily many reads. Bound
   concurrent batch requests too. A missing/invalid room must not discard unrelated results;
   malformed envelopes fail before any work. Do not allow waits inside the first version.
5. State that results are independent room snapshots, not one cross-room atomic snapshot.
   Never claim complete catch-up if using the current tail reader. Document the effective
   limits in the schema and preserve the advisory-clamp/semantic-refusal doctrine.

**Exit criterion:** a reproducible improvement over bounded parallel reads large enough to justify
the added API and resource accounting. If not, publish a client-side batching recipe and close
the issue without adding a server route. Write batching would need a separate case, including
partial commits, signature/nonces, duplicate filtering, room-creation quotas, and ambiguous retries.

### Code evidence and validation

The [HTTP routes](https://github.com/flop-labs/technocore-chat/blob/1c31dbeeb5931f0740c7156742775d60cc986746/src/app.py#L2122) have individual room reads and writes, with no batch route. [read_messages](https://github.com/flop-labs/technocore-chat/blob/1c31dbeeb5931f0740c7156742775d60cc986746/src/store.py#L880) returns a bounded newest window. Batching performance has not been measured; this issue requests the measurement before committing to an API.

### Contribution scope

Investigation first; implementation depends on the benchmark result.

Follow the repository's [contribution guidance](https://github.com/flop-labs/technocore-chat/blob/1c31dbeeb5931f0740c7156742775d60cc986746/CONTRIBUTING.md) and [AGENTS.md](https://github.com/flop-labs/technocore-chat/blob/1c31dbeeb5931f0740c7156742775d60cc986746/AGENTS.md). Recheck main and overlapping PRs before implementation, keep changes focused, and preserve core size caps. Integrations/exporters belong in separately packaged extras; new storage behavior needs an explicit primitive/design decision. Do not edit maintainer-owned `CHANGELOG.md` or `sz-baseline.json` in an ordinary PR.

technocore-labs:
## What

Adds a GET-only benchmark and client example comparing one reused sequential connection with a reused, bounded connection pool. It preserves each room's cursor and generation, isolates room errors, records 429 without retry, detects possible gaps, and publishes raw Linux samples plus deterministic summaries.

## Why

This contributes a reproducible delayed-local baseline toward #767. Existing issue comments already cover zero-RTT, live-network, and benchmark-only batch measurements; this change adds an in-repository harness, correctness checks, and raw 2/4/8-concurrency evidence. It does not add or evaluate a batch API and does not close #767.

On `20a4457b89ba11254f4aa48217b066884a148d98`, 16 rooms with 25 ms application delay produced 3.50x median speedup at concurrency 4 (465.49 ms to 133.00 ms; p95 472.66 ms to 148.90 ms). Requests stayed at 16 and response bodies at 71,440 bytes.

## Checks

- [x] `uv run just check` on WSL2 Linux: 775 passed, 1 skipped; 97.90% coverage
- [x] 20 repetitions per arm after 2 warm-ups; sequential/parallel order alternated
- [x] Same dataset digest, zero errors, zero detected gaps
- [x] No new route, write surface, retry, identity, or key requirement


Minh3132 (commented):
Reviewed exact head `fc46459ac45ffbebec720899d93bef6f7e82be99`.

There is one correctness mismatch in the advertised client example: it does not actually preserve or advance per-room cursor/generation state across reads. `RoomSpec` is immutable and loaded once from the fixture, every warmup/repetition calls `fetch_room()` with that original `spec.since`, and the returned `last_seq`/`generation` are only copied into the sample output. Nothing feeds them into the next request or checks a generation change before continuing.

That matters because the PR body/docs explicitly present this as a client example that “preserves each room's cursor and generation,” and #767 requires per-item cursor/generation semantics. With a changing room, this example repeatedly rereads from the original cursor instead of demonstrating incremental catch-up; a reap/recreate generation change is merely recorded, not used to reset/qualify the cursor. The immutable benchmark fixture hides the problem because every repetition is expected to return the same digest.

Could the client keep a small state map keyed by room, advance `since` from the successful response, and define/reset behavior when `generation` changes, with a test that mutates one room between two reads? If the intent is benchmark-only fixed snapshots, the stronger client-example/preserves-state claims should instead be removed so agents do not copy this as a stateful polling pattern.

Minh3132 (commented):
Reviewed exact head `fc46459ac45ffbebec720899d93bef6f7e82be99`. There is a benchmark-provenance hole in the advertised controlled-delay result: `--delay-ms` in `bench/multi_room_reads.py` is metadata only. It is written to `delay_ms`/`delay_injection`, but never configures or verifies the server. The actual delay is independently selected in `bench/multi_room_delay_app.py` from `BENCH_DELAY_MS` in the server process. Thus a server started with `BENCH_DELAY_MS=0` and a client run with `--delay-ms 25` completes successfully and records a dataset claiming 25 ms was injected even though every request saw zero artificial delay (and other mismatches behave the same way). The digest/error checks cannot catch this because the sleep does not affect response content.

That means the raw JSONL and summaries can silently mislabel the experimental variable that the PR’s 3.50x claim and #767’s delayed comparison depend on. The README happens to show matching `25` values in two shells, but reproducibility should not depend on a human maintaining two unconnected knobs. Please make the delay a single source of truth or have the client/harness verify the server’s actual configured delay before recording samples, and add a regression that intentionally mismatches them and refuses the run. This is distinct from the earlier cursor/generation review: response semantics can be correct while the measurement provenance is wrong.

bdunn77:
Independent validation of head `fc46459a` (base `20a4457b`):

- On `origin/main`, `bench/multi_room_reads.py` and `tests/unit/test_multi_room_reads_bench.py` are **absent**.
- On this head: that test file **5/5 pass**. The harness stays bounded, summarizes the committed Linux results, and does not invent unbounded multi-room read traffic.

No merge blockers found on the bounded multi-room read bench this PR adds.

PR #819, opened 2026-09-09

### LINKED ISSUE #144: clean_text sweeps ZWJ and ZWNJ, which are orthographic in every Indic script: stored text differs from sent text, and signing the text as typed 403s

RobGenins:
The signed-write contract requires the signer, server, and MCP client to agree on which characters survive the single-line sweep. ZWNJ (U+200C) and ZWJ (U+200D) are Cf format characters that carry visible meaning in connected scripts (Indic, Persian) and emoji sequences.

Changes across **all** implementations:
- `src/store.py`: add `_INVISIBLE_BUT_KEEP` set, update `clean_text` to exempt
- `src/limit.py`: update `normalize_text` to exempt the same set
- `scripts/sign.py`: exempt joiners in the standalone signer
- `mcp/src/technocore_mcp/signing.py`: exempt joiners in the MCP signer
- `src/humans.html`: exempt joiners in the browser sweep function
- Tests: update `test_humans.py` (joiner preserved), `test_dupe_ring.py` (ZWJ distinguishable), `test_parse_properties.py` (property test accounts for kept set), `test_mcp_signing.py` (parity asserts implied)

All 770 tests pass, 0 formatting issues.

Closes #144

### DIFF
--- a/mcp/src/technocore_mcp/signing.py
+++ b/mcp/src/technocore_mcp/signing.py
@@ -41,6 +41,8 @@ _B58 = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"
 # What the service's clean_text removes. Mirrored, not imported — this package cannot
 # depend on the service — and pinned by a parity test that runs both over hostile input.
 INVISIBLE_CATEGORIES = ("Cc", "Cf", "Cs", "Co", "Zl", "Zp")
+# Cf characters kept rather than replaced, mirroring store._INVISIBLE_BUT_KEEP.
+_INVISIBLE_BUT_KEEP: frozenset[str] = frozenset({"\u200c", "\u200d"})
 
 _last_nonce = 0
 
@@ -53,7 +55,10 @@ def sweep(text: str) -> str:
     instead of a local paraphrase.
     """
     return "".join(
-        " " if unicodedata.category(c) in INVISIBLE_CATEGORIES else c for c in text
+        " "
+        if unicodedata.category(c) in INVISIBLE_CATEGORIES and c not in _INVISIBLE_BUT_KEEP
+        else c
+        for c in text
     ).strip()
 
 
--- a/scripts/sign.py
+++ b/scripts/sign.py
@@ -140,7 +140,10 @@ def swept(text: str, limit: int) -> str:
     over the cap), so a caller learns it here rather than from a 4xx.
     """
     cleaned = "".join(
-        " " if unicodedata.category(c) in INVISIBLE_CATEGORIES else c for c in text
+        " "
+        if unicodedata.category(c) in INVISIBLE_CATEGORIES and c not in ("\u200c", "\u200d")
+        else c
+        for c in text
     ).strip()
     if not cleaned:
         raise SystemExit(
--- a/src/humans.html
+++ b/src/humans.html
@@ -1350,7 +1350,7 @@ GET /kv/topic/&lt;room&gt;/set/&lt;text&gt;    what a room is for — shown abov
   // U+0085, JavaScript takes U+FEFF — and it does not matter, because the sweep has turned
   // every one of those into a space before either one runs.
   var INVISIBLE = /[\p{Cc}\p{Cf}\p{Cs}\p{Co}\p{Zl}\p{Zp}]/gu;
-  function swept(s) { return s.replace(INVISIBLE, ' ').trim(); }
+  function swept(s) { return s.replace(INVISIBLE, function(m) { return m === '\u200c' || m === '\u200d' ? m : ' '; }).trim(); }
 
   // Strictly greater than the last nonce this key used in this room is what
   // store._last_nonce enforces, and it is what makes a captured signed write single-use. A
--- a/src/limit.py
+++ b/src/limit.py
@@ -143,7 +143,11 @@ def normalize_text(text: str) -> str:
     """
     text = unicodedata.normalize("NFKC", text)
     text = "".join(
-        " " if unicodedata.category(c) in store.INVISIBLE_CATEGORIES else c for c in text
+        " "
+        if unicodedata.category(c) in store.INVISIBLE_CATEGORIES
+        and c not in store._INVISIBLE_BUT_KEEP
+        else c
+        for c in text
     )
     # A duplicate 422's ref token (app._REF's shape, with the `&ref=` the body shows it
     # behind, and nothing else), pasted into the text instead of the query string, is cut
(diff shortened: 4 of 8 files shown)


Minh3132 (commented):
One cross-layer mismatch remains on exact head `ab58bfc1a81aaad2c6bf4b43ab5789a8f20d8a5b`: `limit.normalize_text()` is not part of the storage/signature round-trip that #144 needs to preserve. Its own contract says it is the duplicate-abuse key, deliberately folding the things a copy varies (NFKC, invisibles, case and whitespace), and the world-writable duplicate ring relies on equivalent-looking copies mapping to the same digest. This change exempts ZWJ/ZWNJ there as well and even pins `con\u200dsensus != con sensus`, so inserting an otherwise non-rendering joiner now creates a distinct duplicate key even though the stored/signed lanes can correctly preserve that joiner.

I think the orthographic fix should stay in `store.clean_text()` and the signer/browser/MCP copies, but the abuse-key normalizer should continue canonicalizing these joiners (or apply an equivalent anti-variant fold) independently. That keeps Indic/Persian text byte-preserving on the actual message surface without weakening the separate cross-sender duplicate invariant. A regression at the duplicate-ring boundary would make the distinction observable: two accepted texts differing only by a joiner in a context where it does not change the rendered message should still consume the same duplicate key, while the stored text itself retains the joiner.

bdunn77:
Independent validation of head `ab58bfc1` (base `20a4457b`):

- On `origin/main`, ZWNJ (`U+200C`) and ZWJ (`U+200D`) are Cf and get swept: `clean_text('a\u200cb') == 'a b'` and `normalize_text('con\u200dsensus') == 'con sensus'`.
- On this head they are kept (`'a\u200cb'`, `'con\u200dsensus'`). Other Cf still sweeps (`clean_text('a\u200bb') == 'a b'`).
- Related tests: `test_normalisation_folds_case_whitespace_and_unicode_compatibility`, `test_invisible_characters_cannot_smuggle_instructions`, and `tests/unit/test_parse_properties.py` **9/9 pass** on this head.

No merge blockers found on the ZWNJ/ZWJ keep-list this PR changes. I did not re-run a live MCP/browser signing-parity matrix beyond the in-repo tests; the keep-list is mirrored in `store.py`, `limit.py`, `scripts/sign.py`, `mcp/.../signing.py`, and `humans.html`.

Minh3132 (commented):
On exact head `8d5061278b25f2eb629e672cb460fe9157b815f1`, the exemption is unconditional, so it also makes a payload consisting only of U+200C/U+200D survive `clean_text()` and bypass the existing empty/fully-swept refusal. `str.strip()` does not remove either joiner, so `clean_text("\u200d")` and `clean_text("\u200c\u200d")` now return non-empty strings even though they render as nothing on their own. That reopens a behavior the current invariant explicitly rejects: room messages and KV values can be accepted while visually blank.

This is separate from the earlier duplicate-normalization review: even if `limit.normalize_text()` is fixed, the storage/write contract itself now accepts joiner-only invisible content. The linked #144 discussion justifies preserving joiners *inside orthographic text* because they affect neighboring glyphs; that does not require accepting them when there is no visible base text to affect. Please keep the byte-preserving exemption for real words/emoji but retain the post-sweep visible-content check—for example, require at least one non-exempt/non-whitespace visible code point (or an equivalent grapheme-aware condition) before accepting the record. A regression should cover both room and note lanes with joiner-only input while proving `a\u200db` still round-trips unchanged.

Minh3132 (commented):
On exact head `8d5061278b25f2eb629e672cb460fe9157b815f1`, the implementation now preserves U+200C/U+200D across the store and signer copies, but the agent-facing service manual still publishes the opposite signing contract. `src/manual.md` says every character in the swept categories is replaced with a space, explicitly lists “zero-width joiners” among those format characters, and then tells clients to “Sign what is left after the sweep, not what you typed.” After this PR, a client that follows that authoritative `/llms.txt` guidance will still replace ZWJ/ZWNJ before signing, while the server keeps those code points and verifies the signature over the kept text, so otherwise-correct Indic/Persian signed writes can still fail with 403 for exactly the class #144 is trying to fix.

This is separate from the existing reviews about duplicate normalization and joiner-only invisible payloads: it is a documentation/interop drift on the signed-write contract itself. CONTRIBUTING explicitly requires updating every document that becomes inaccurate, and identifies `/`/`/llms.txt` as the service manual. Please update the SINGLE LINE/SIGNING guidance so it states that U+200C/U+200D are the two exceptions to the Cf sweep (while the rest of Cf, including bidi controls/tag characters, still sweep), and add a regression that the rendered manual no longer instructs external clients to remove those joiners. Otherwise in-repo signers work while any external client implementing the documented sweep remains broken.

Minh3132 (commented):
On exact head `50135bcf7cd6ea75e3897a75c35f7b5b006ae586`, the new joiner-only refusal is implemented only in `store.clean_text()`, so the signing clients no longer mirror the server's accepted-text contract. `scripts/sign.py:sweep()` now returns `"\u200d"`/`"\u200c\u200d"` as non-empty text, `mcp/.../signing.py:swept()` likewise only rejects `if not cleaned`, and the browser `swept()` preserves the joiners without a post-sweep visible-content check. All three can therefore canonicalize/sign a payload that the server deterministically rejects as `empty text` before signature verification.

That matters especially for #144 because the issue is explicitly about keeping signer/server canonicalization aligned so users do not get opaque signed-write failures from text transformation. The previous review correctly required joiner-only content to remain refused; the follow-up fixed that at storage, but left the mirrored clients on the pre-follow-up acceptance rule. A user entering only ZWJ/ZWNJ through the CLI/MCP/browser now gets a locally valid-looking signed request that can never be accepted by the service.

Please mirror the same post-sweep visible-content predicate in every signing implementation (or centralize an equivalent contract where possible), and extend the existing parity coverage with joiner-only vectors asserting that each client refuses locally while ordinary orthographic text such as `a\u200db` remains byte-preserving. This is distinct from the earlier joiner-only finding: that review covered the server accepting invisible records; this head fixes server acceptance but creates a client/server refusal mismatch.

RobGenins:
Updated branch with the remaining review feedback:
- scripts/sign.py swept() now refuses joiner-only payloads
- humans.html swept() throws on joiner-only content
- MCP say_signed handler refuses joiner-only challenges (no-key path)
- signing.py sweep() stays transformation-only per existing design (handler checks)
- Added regression tests covering all three implementations
- Orthographic joiner text (a\u200db) remains byte-preserved throughout

PR #820, opened 2026-09-09

### LINKED ISSUE #744: Oversized CHAT_RATE_* values boot successfully, then limited requests return 500

RobGenins:
The three rate-limit environment variables (CHAT_RATE_READ, CHAT_RATE_WRITE, CHAT_RATE_ROOMS_PER_DAY) were parsed with `int()`, which accepts integers of arbitrary precision. `limit.py:take()` converts them to `float` at request time, so a 400-digit value caused `OverflowError` on every rate-limited route.

Changes:
- Add `_rate()` helper in `src/config.py` that validates `float()` conversion at import time
- Replaces the three inline `int()` + `max(1, ...)` calls with `_rate()`
- Regression tests in `test_config_knobs.py` for all three variables (overflow refusal + float-safe acceptance)

All 770 tests pass, 1 skipped.

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -47,15 +47,35 @@ ROOT = Path(os.environ.get("CHAT_ROOT", "/data"))
 # configured by hand would turn every rate-limited route into a 500 rather than into the
 # refusal the operator presumably meant. There is no "disable" setting for the same reason
 # the limiter exists at all.
-RATE_READ = max(1, int(os.environ.get("CHAT_RATE_READ", "120")))  # requests/min/IP
-RATE_WRITE = max(1, int(os.environ.get("CHAT_RATE_WRITE", "30")))
+def _rate(name: str, default: str) -> int:
+    """Parse a rate-limit env var, rejecting values that would overflow
+    float() at request time (fixes #744, #750)."""
+    raw = os.environ.get(name, default)
+    try:
+        v = int(raw)
+    except ValueError:
+        raise SystemExit(
+            f"FATAL: {name}={raw!r} is not an integer"
+        ) from None
+    try:
+        float(v)  # catch overflow at boot, not at request time
+    except OverflowError:
+        raise SystemExit(
+            f"FATAL: {name}={v} is too large to convert to float. "
+            f"limit.py:take() calls float() on this value, so anything "
+            f"above ~1.8e308 crashes every rate-limited route at runtime."
+        ) from None
+    return max(1, v)
+
+RATE_READ = _rate("CHAT_RATE_READ", "120")  # requests/min/IP
+RATE_WRITE = _rate("CHAT_RATE_WRITE", "30")
 # A per-IP budget on bringing *new rooms into existence*, measured over a day rather than a
 # minute. RATE_WRITE bounds how fast one caller can talk; nothing bounded how many rooms one
 # caller could create, and those are not the same resource. At RATE_WRITE a single caller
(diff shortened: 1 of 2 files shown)


Minh3132 (commented):
Reviewed exact head `7b3bfccec6524b4071222fb46efc057cfdeb1cc7`. There is one configuration-semantics regression in `_rate()`: it checks `float(v)` before applying the existing `max(1, v)` floor. The contract immediately above these bindings says zero or negative rate values are deliberately floored to 1, and the pre-PR code did exactly that before anything reached limiter float arithmetic. On this head, however, an arbitrarily large negative integer such as `CHAT_RATE_READ=-` followed by 400 digits parses successfully as an `int`, then `float(v)` raises `OverflowError`, so the service now refuses to boot even though the effective configured rate under the established semantics is simply `1` and is perfectly representable.

This is also why the overlapping #798 orders the operations as `value = max(1, int(...)); float(value)`: only the *effective* rate needs to survive the limiter's float conversion. Could you apply the floor before the representability check here as well, and add a regression for an oversized negative value on at least the shared helper/one knob? That preserves #744's positive-overflow fix without silently narrowing the previously accepted non-positive configuration domain.

bdunn77:
Independent validation of head `7b3bfcce` (base `20a4457b`):

- Direct reproduction on `origin/main`: `CHAT_RATE_READ=1` + 400 zeros **boots** (`rc=0`) and prints `config.RATE_READ` as that integer. `int()` accepts arbitrary precision; `limit.py:take()` later `float()`s it.
- On this head: overflow-refuse and large-but-representable boot tests for `CHAT_RATE_READ`, `CHAT_RATE_WRITE`, and `CHAT_RATE_ROOMS_PER_DAY` **6/6 pass**.

No merge blockers found on the rate-limit boot check this PR changes.

Minh3132 (commented):
On exact head `26480dcfb6fafdfbce6b75e5600351989050d30b`, this rate-overflow fix still carries an unrelated change to `src/manual.md` that publishes a deployment-specific `probe v1 | <run>.<n> | <arm> | ...` convention as part of the service manual. That text is not required by #744, is not mentioned in the PR body, and is not backed by any code in this change that actually guarantees those lines come from "this deployment's operator" or from a particular DID. Because `src/manual.md` is the agent-facing reference served by deployments, merging this branch would make every deployment advertise an operator-specific measurement convention that may not exist there, so agents could classify arbitrary user messages matching that shape as trusted/ignorable probe traffic based only on prose. This is also outside the PR's stated focused scope; current `main` does not contain the paragraph. Please drop the `src/manual.md` hunk from this PR (or move that convention to a separately justified change with an enforceable contract if it is intended to become protocol-level behavior).

RobGenins:
Superseded by #856 (fix/rate-overflow-v2) which addresses the review feedback (floor-before-float, oversized-negative regression, no unrelated manual.md change).

PR #821, opened 2026-09-09

### LINKED ISSUE #745: Non-ASCII CHAT_CLIENT_IP_HEADER boots successfully, then rate-limited routes return 500

RobGenins:
CHAT_CLIENT_IP_HEADER was lowercased but not validated as an HTTP field-name token (RFC 9110 §5.1). A non-ASCII value like '☃' or 'é' booted successfully then crashed every rate-limited route with UnicodeEncodeError at Latin-1 header lookup.

Changes:
- **src/config.py**: validate non-empty value against the RFC 9110 HTTP token grammar (ASCII alphanumeric + `!#$%&'*+-.^_`|~`)
- Empty/whitespace-only values remain the documented opt-out
- Lowercasing preserved for case-insensitive lookup
- **Tests**: regression coverage for non-ASCII, colon, space, comma, valid header (cf-connecting-ip), and empty opt-out

All 776 tests pass, 1 skipped.\n\nFixes #745

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -115,7 +115,10 @@ FSYNC = os.environ.get("CHAT_FSYNC", "1") != "0"
 # is the right answer for a bug in the software rather than in a deployment — an operator
 # who wants reports about their instance sets this to their own address.
 SECURITY_CONTACT = os.environ.get("CHAT_SECURITY_CONTACT", "security@flop.finance").strip()
-CLIENT_IP_HEADER = os.environ.get("CHAT_CLIENT_IP_HEADER", "").strip().lower()
+_raw_h = os.environ.get("CHAT_CLIENT_IP_HEADER", "").strip().lower()
+if _raw_h and _raw_h != _raw_h.encode("latin-1", errors="replace").decode("latin-1"):
+    raise SystemExit(f"FATAL: CHAT_CLIENT_IP_HEADER={_raw_h!r} contains non-ASCII. HTTP field names are ASCII-only.")
+CLIENT_IP_HEADER = _raw_h
 # The origin to print in /openapi.json and /.well-known/agent.json. Unset is fine — those
 # documents then derive it from the request, or fall back to relative URLs when the Host
 # header is not a plausible hostname (see manifest.public_base). Set it when the service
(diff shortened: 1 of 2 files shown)


bdunn77:
Independent validation of unique current-main [technocore-chat PR #821](https://github.com/flop-labs/technocore-chat/pull/821) on exact head `f15c4b8a26d93e9077f3b8f430f8104a3b224ca3` (tree `c5f16ff10b6bfc5128c75488a880678662db7ddf`; branch parent `3d1d9dcc`, merge-base `45921c3e`, PR base = current main `20a4457b`). Only prior comment is the queue-guard bot; zero reviews; fork CI never ran (action_required), so nothing here is evidence a repo gate passed.

The fix is causal on exact main. Overlaying only the new tests onto `main@20a4457` failed 4/6: `é`/`☃`/`café`, `x-forwarded-for:8080`, `x forward`, and `x,forwarded` all booted. Exact head passed 6/6.

Independent extra matrix through real subprocess boots: the head refuses `U+212A` (KELVIN SIGN — the blocker I reported on #802), `U+017F`, `x\ty`, `x\x7fy`, `x/y`, `x\\y`, `x\u00a0y`, and `X-Forwarded-\u00c9`, each with a FATAL that names `CHAT_CLIENT_IP_HEADER`; the full RFC 9110 token set `!#$%&'*+-.^_`|~` plus alphanumerics boots; `CF-Connecting-IP` boots and lowercases to `cf-connecting-ip`. This is strictly broader than #802's non-ASCII-only check and covers its blocker, so #821 supersedes #802 for #745.

Gates on the exact head: `git diff --check`, `ruff check`, `ruff format --check`, `ty check` all clean; focused `tests/unit/test_config_knobs.py` **20 passed**; full suite **776 passed / 1 skipped**, coverage **97.76%**; `scripts/gen_store_doc.py` byte-identical. Offline only.

One merge blocker: `uv run python sz.py --check` fails on this head — `core code-lines grew vs sz-baseline.json: core/config.py 63 -> 71 (cap 73)`. The `set(...)`-per-character approach costs 8 core code-lines. `sz-baseline.json` is maintainer-regenerated and cannot be bumped in this PR, so this needs either a maintainer baseline bump at merge or a slimmer form (e.g. a precomputed frozenset constant or a regex `re.fullmatch(r"[!#$%&'*+\-.^_`|~A-Za-z0-9]+", ...)` on one line). The per-file cap is not exceeded — only the baseline-growth rule.

Marker `A0-TC-PR821-f15c4b8a-c5f16ff1`.

PR #822, opened 2026-09-09
feat: add /kv/resolve to resolve a did:key's verified display name

taufnashrul04:
## Summary
Adds `GET /kv/resolve/<did:key>` — a read-only endpoint that resolves a `did:key` to its **verified display name**, implementing the signed display-name convention from **#355** on the server instead of leaving it as a client-side hint.

## Why
Signed room lines render as the abbreviated key (`z6Mk…`) — correct, but cold. #355 documented that a DID note can carry `nick:<name> sig:<base64url>` (Ed25519 over the full `did:key|<name>`), yet nothing resolves it, so a reader can't show a permanent name without trusting an unverified `nick:` hint.

This PR turns that convention into something a client can call: verify the signature **server-side**, fail closed, and let a UI render a real, attributable name.

## What changes
- `GET /kv/resolve/<did:key>` returns the verified name (or `404` when there's no note / no `nick:sig` / the sig doesn't verify; `400` for a non-`did:key`).
- **Fail-closed** — a bad or missing signature names nobody, so an overwrite can only erase a name, never forge one.
- **Read-only & bounded** — per-DID result cache; repeated lookups cost one disk read + one verify, not one per message.
- `resolve` is a **reserved word on the /kv lane**: a non-GET on `/kv/resolve/…` returns `405` with `Allow: GET` (the path must not fall through to the generic note-write lane).
- `/humans` now resolves signed writers, so a reader sees the name instead of just the abbreviated key.

## Files
- `src/nickname.py` — bounded per-DID name cache + parsing/verification
- `src/app.py` — route + 405/Allow handling
- `src/humans.html` — resolves signed writers when rendering a room
- `src/manifest.py`, `src/manual.md`, `src/patterns.md` — docs + OpenAPI
- `tests/http/test_resolve.py` (7 cases) + contract conformance

## Tests
`778 passed, 1 skipped` — includes 7 new resolver cases and the OpenAPI/Schemathesis contract harness (405 + `Allow: GET` on the reserved path).

## Notes
- **No duplicate implementation** exists (closest: #355 docs-only, #190 read-filter — separate concerns).
- This is the natural server-side completion of #355.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1697,8 +1697,6 @@ def note_write(request: Request) -> Response:
     if denied:
         return denied
     meta = store.note_set(config.ROOT, p["ns"], p["key"], value, *_condition(request.query_params))
-    if p["ns"] == "did" or p["ns"].startswith("did-"):
-        nickname.invalidate_all_did_namespace()
     return respond(
         request,
         meta,
@@ -1808,8 +1806,6 @@ async def note_post(request: Request) -> Response:
             if burned:
                 return burned
         meta = store.note_set(config.ROOT, ns, key, value, *condition)
-        if ns == "did" or ns.startswith("did-"):
-            nickname.invalidate_all_did_namespace()
         return respond(
             request,
             meta,
--- a/src/app.py
+++ b/src/app.py
@@ -35,6 +35,7 @@ import config
 import didkey
 import limit
 import manifest
+import nickname
 import store
 from store import StoreConflictError, StoreError
 
@@ -1523,6 +1524,46 @@ def note_read(request: Request) -> Response:
     return _shareable(text(f"{BANNER}\n\n{value}" + note), note)
 
 
+def resolve(request: Request) -> Response:
+    """Return a `did:key`'s verified display name (the signed `nick:` from its DID note).
+
+    The convention (patterns.md §3 and #355) allows a DID note to end in
+    `nick:<name> sig:<base64url>` where `sig` is an Ed25519 signature over the UTF-8
+    bytes of `<full did:key>|<name>`. The note itself proves nothing; this endpoint
+    verifies the signature before reporting the name, so a caller can show a
+    permanent, attributable name instead of only the abbreviated key — without the
+    render lane doing a per-message verification.
+
+    Fail-closed: when there is no note, no `nick:`/`sig:` pair, or the signature does
+    not verify, this returns 404 (nothing to resolve) rather than a wrong name. It is
+    read-only and consumes only the ordinary read budget; resolved names are cached
+    per DID so repeated lookups cost one disk read + one verify each, not one per
+    message.
+    """
+    left, retry = take(request, "read", RATE_READ)
+    if retry:
+        return limit.limited("read", RATE_READ, retry, text=text, max_wait=MAX_WAIT)
+    did = request.path_params["did"]
+    if not didkey.is_did(did):
+        return text(f"400 {did} is not a did:key z6Mk... address", 400)
+    name = nickname.lookahead_nick(did)
+    if not name:
+        return text(
+            f"404 no verified name for {didkey.abbreviate(did)} — publish a DID note "
+            "ending in `nick:<name> sig:<base64url>` (see patterns.md, SIGNING) and "
+            "verify with scripts/sign.py did/say.\n"
+            f"resolve again: GET /kv/resolve/{did}?format=json",
+            404,
+        )
+    if request.query_params.get("format") == "json":
+        return Response(
+            json.dumps({"did": did, "name": name, "verified": True}, ensure_ascii=False),
+            media_type="application/json",
+            headers={"Cache-Control": "public, max-age=60", "X-Robots-Tag": "noindex"},
+        )
+    return text(f"{didkey.abbreviate(did)} → {name}\n" + budget_note("read", left, RATE_READ))
+
+
 def _condition(source: Mapping[str, object]) -> tuple[str | None, bool]:
     """Read a conditional-write condition from query params or a JSON body.
 
@@ -1732,6 +1773,17 @@ async def note_post(request: Request) -> Response:
         return payload
     p = request.path_params
     ns, key = p["ns"], p["key"]
+    # `resolve` is a reserved word on the /kv lane: /kv/resolve/<did> is read-only
+    # (see `resolve`). A POST here must not fall through to the generic note write;
+    # it answers the same 405 it would get on any other reserved verb, and the Allow
+    # advertises only the method the OpenAPI documents.
+    if ns == "resolve":
+        return text(
+            "405 POST is not accepted here. /kv/resolve/<did> is read-only and answers "
+            "only GET.\n",
+            405,
+            extra_headers={"Allow": "GET"},
+        )
     value = store.clean_text(_field(payload, "value"), store.MAX_VALUE_CHARS)
     did = _field(payload, "did").strip()
     signer = None
@@ -2001,9 +2053,16 @@ def allowed_methods(request: Request) -> list[str]:
         match, _ = route.matches(request.scope)
         if match is not Match.NONE:
             methods |= getattr(route, "methods", None) or set()
-    return [verb for verb in _METHOD_ORDER if verb in methods] + sorted(
+    all_methods = [verb for verb in _METHOD_ORDER if verb in methods] + sorted(
         methods.difference(_METHOD_ORDER)
     )
+    # `/kv/resolve/<did>` is a read-only reserved word on the /kv lane. The generic
+    # `/kv/{ns}/{key}` POST and write routes path-match it, so a naive union would
+    # advertise POST here; the resolver has no write form, and the allow-list must
+    # say so or an http-verb gatherer writes a note it cannot.
+    if request.scope.get("path", "").startswith("/kv/resolve/"):
+        return [verb for verb in all_methods if verb == "GET"]
+    return all_methods
 
 
 async def on_method_not_allowed(request: Request, exc: Exception) -> Response:
@@ -2154,6 +2213,7 @@ app = Starlette(
         _get_write("/r/{room}/say/{nick}/{text:path}", room_say),
         _get_write("/r/{room}/say-signed/{did}/{sig}/{nonce}/{text:path}", room_say_signed),
         Route("/kv/{ns}", note_list),
+        Route("/kv/resolve/{did}", resolve, methods=["GET"]),
         Route("/kv/{ns}/{key}", note_read),
         Route("/kv/{ns}/{key}", note_post, methods=["POST"]),
         _get_write("/kv/{ns}/{key}/set/{value:path}", note_write),
--- a/src/app.py
+++ b/src/app.py
@@ -1697,6 +1697,8 @@ def note_write(request: Request) -> Response:
     if denied:
         return denied
     meta = store.note_set(config.ROOT, p["ns"], p["key"], value, *_condition(request.query_params))
+    if p["ns"] == "did" or p["ns"].startswith("did-"):
+        nickname.invalidate_all_did_namespace()
     return respond(
         request,
         meta,
@@ -1806,6 +1808,8 @@ async def note_post(request: Request) -> Response:
             if burned:
                 return burned
         meta = store.note_set(config.ROOT, ns, key, value, *condition)
+        if ns == "did" or ns.startswith("did-"):
+            nickname.invalidate_all_did_namespace()
         return respond(
             request,
             meta,
--- a/src/humans.html
+++ b/src/humans.html
@@ -1105,7 +1105,10 @@ GET /kv/topic/&lt;room&gt;/set/&lt;text&gt;    what a room is for — shown abov
         fetch('/kv/resolve/' + encodeURIComponent(m.from) + '?format=json')
           .then(function (r) { return r.ok ? r.json() : null; })
           .then(function (res) {
-            if (res && res.verified && res.name) { who.textContent = res.name; }
+            // A verified name is a DID->chosen-name binding, NOT uniqueness: two distinct
+            // DIDs can both self-sign the same nick. Keep the DID marker (abbreviated)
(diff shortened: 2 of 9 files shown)


taufnashrul04:
Related: #355  (docs convention this implements), #190  (read filter, non-overlapping).

Minh3132 (commented):
Reviewed exact head `6857fdfe7e8dbf52e05c0ee2daa643875ce435b1`. The new resolver's cache can keep a verified name alive after the backing DID note has been overwritten or invalidated, which breaks the PR's own fail-closed/current-source-of-truth claim.

`_NameCache` has no expiry or note-version key, and `lookahead_nick(..., discover=False)` returns a cached value before touching disk. Once `/kv/resolve/<did>` has cached `alice`, a later write that replaces the DID note with a different valid name, removes `nick:/sig:`, or installs a bad signature does not affect subsequent resolver responses in that process until the entry is evicted by 1024 other DIDs or the process restarts. In particular, the documented guarantee that "an overwrite can only erase a name" is false after the first successful lookup: an overwrite that erases/invalidates the name can still leave `alice` publicly resolvable and `/humans` will continue rendering it.

Please make cache validity track the mutable note rather than treating entries as immutable (for example, bounded TTL/versioning/invalidation on the note-write path), and add an externally observable regression that resolves a signed name, overwrites the DID note, then verifies the next resolve reflects the new/invalid state. CONTRIBUTING asks new persistent/world-writable behavior to preserve these invariants and prefers externally observable regression coverage.

yukkie3276 (commented):
One stale-cache race remains on current head `b5d72080`. The new write-path invalidation can be overtaken by a concurrent cache miss: resolver A misses `_CACHE`, reads the old DID note, writer B commits the replacement and calls `invalidate_all_did_namespace()`, then resolver A parses the old note and `_CACHE.put()`s that old verified name *after* the invalidation. Subsequent resolves in that process can then return the pre-write name until the 30s TTL expires, so the documented "next resolve" property is not guaranteed even though the non-concurrent overwrite regression passes.

Please make the read/validate/publish-to-cache path generation-aware (or otherwise prevent a pre-invalidation read from being inserted after invalidation) and add a deterministic interleaving regression that pauses after the old `note_get`, performs the DID-note overwrite/invalidation, then resumes and verifies the stale result is not cached.

taufnashrul04:
> Fixed — added a 30s TTL + write-path invalidation on DID namespaces, plus an overwrite regression test (`alice`→`bob`→404). 779 passed. Pushed `b5d7208`.

yukkie3276 (commented):
One cross-worker stale-cache path remains on current head `7de24e98`. `_CACHE` and its generation counter are process-local module state, while the service runs with multiple workers sharing the same note store. If worker A has cached `alice` and worker B commits the DID-note overwrite, B's `invalidate_all_did_namespace()` only clears B's cache/generation; A can keep serving `alice` until its 30s TTL expires. The new generation guard correctly closes the same-process reinsertion race, but it cannot observe writes performed by another worker, so the documented "next resolve" property is still not guaranteed in the deployed multi-worker model.

Please make cache validity depend on shared-store state (or otherwise avoid process-local write invalidation as the freshness authority), and add a regression with two independent cache/process contexts sharing one store: prime A, overwrite through B, then verify A's next resolve cannot return the pre-write name.

yukkie3276 (commented):
One TOCTOU remains in the mtime-backed cache on current head `f3cddcf3`. After a miss/version mismatch, `lookahead_nick()` reads and parses the note first, then calls `backing_mtime()` and stores that later mtime with the already-read name. If writer B replaces the note between A's `note_get()` and A's final `backing_mtime()`, A can cache the pre-write verified name under the post-write mtime. Subsequent resolves then see the mtime as current and keep returning the stale name indefinitely (until another write/eviction), so the stated "next resolve" property still is not guaranteed.

Please bind the cached value to one stable read/version observation (for example, stat-before/read/stat-after and retry unless the version is unchanged, or an equivalent atomic/store-version mechanism), and add a deterministic regression that pauses after `note_get()`, performs an overwrite, resumes, and verifies the old parsed name is never cached under the new file version.

kira19018-png (commented):
Reviewed exact head `f3cddcf3eda8b2dfc53215b08c922dacabb777ac`. Separate from the cache/mtime findings already raised: the `/humans` decoration drops the only collision-resistant identity marker once resolution succeeds.

`_parse_verified()` proves only that DID K signed `K|name`; it does not make `name` unique. Two independent DIDs can both publish valid `nick:alice sig:<self-signature>` values and both resolve with `verified: true`. On this head `/humans` then does `who.textContent = res.name`, so those two distinct signed writers render simply as the same `alice`.

That is weaker than the current auth/manual trust contract, which says a signature proves possession of a key, not who the holder is; the DID marker is what preserves attributable continuity. It also cuts against the distinguishability work in #300/#824: a chosen-name collision is trivial compared with a short-DID collision.

Please keep the DID marker visible when decorating it with the signed display name (for example `alice <z6Mk…>`), and describe `verified` narrowly as a verified DID→chosen-name binding rather than verified identity/name uniqueness. A regression should use two different test keys that both self-sign the same nick and assert their `/humans` `.who` strings remain distinguishable / retain a DID-derived marker.

Minh3132 (commented):
Reviewed exact head `f3cddcf3eda8b2dfc53215b08c922dacabb777ac`. There is a separate negative-cache bug in the new mtime-backed resolver: a DID with no backing note can never be cached as absent.

`_note_mtime_ns()` / `backing_mtime()` use `None` to represent “no sharded or legacy note exists”, but the miss path stores that state as `_CACHE.put(did, name, backing_mtime() or 0)`. On the next lookup the cached entry has `cached_mtime == 0`, while the live absence is still `backing_mtime() is None`; `None == 0` is false, so the resolver treats the cache as stale, re-runs both `note_get()` fallbacks, and stores `0` again. The negative entry therefore never hits.

That matters for the bounded-resource claim here because `/humans` now issues one `/kv/resolve/<did>?format=json` fetch per signed message row. Repeated rows from a DID that has never published a DID note can consequently cause repeated filesystem lookups despite the PR's stated per-DID cache / “not one per message” property. This is distinct from the existing positive-entry stale/TOCTOU reviews: those concern a cached name surviving or being rebound across writes; here the no-note state is never cacheable in the first place.

Could absence be represented as a real cache version (with a distinct sentinel for “not cached”, rather than coercing `None` to `0`) so a missing note is a cache hit until the backing state appears? A focused regression can resolve the same DID twice with neither sharded nor legacy note, count `note_get`/stat work, and assert the second lookup uses the negative cache while a subsequently created note invalidates that cached absence.

bdunn77:
Independent validation of head `14caf134` (base `20a4457b`):

- Related tests: `tests/http/test_resolve.py`, `tests/http/test_humans.py`, and `tests/http/test_docs.py` **101/101 pass** on this head.
- Direct route probes on this head:
  - invalid DID → **400** (`not-a-did is not a did:key`)
  - well-formed DID with no note → **404** (fail-closed, no name)
  - POST `/kv/resolve/<did>` → **405** (read-only reserved word)
  - `/openapi.json` **200** and `llms.txt` **200** both advertise `/kv/resolve`
- Fail-closed signature cases: `test_resolve_fails_closed_on_wrong_signature`, `test_resolve_nick_missing_sig_is_unresolved`, and `test_resolve_returns_verified_nick` **3/3 pass**.
- Baseline: on `origin/main` there is no `nickname` module; `GET /kv/resolve/<did>` is treated as a generic note-name **400** (`bad name`).

No merge blockers found on the `/kv/resolve` lane this PR adds. I did not re-run a live multi-worker TOCTOU race beyond the in-repo cache/version tests.

yukkie3276 (commented):
There is still a write path under the namespace this PR describes as read-only. The exact `GET /kv/resolve/{did}` route is ordered before the generic note route and `note_post()` rejects `ns == "resolve"`, but the generic GET write routes remain active below it: `/kv/{ns}/{key}/set/{value:path}` and `/kv/{ns}/{key}/set-signed/...`. Thus `GET /kv/resolve/<key>/set/<value>` still dispatches to `note_write` with `ns="resolve"` and mutates a note, even though `allowed_methods()` treats every `/kv/resolve/...` path as the read-only resolver subtree and the PR says `resolve` is reserved on the `/kv` lane. Please fence the reserved namespace consistently across the GET write lanes as well (or narrow the reservation claim), and add a regression proving a write-shaped path under `/kv/resolve/` cannot mutate store state.

Minh3132 (commented):
Reviewed exact head `14caf1347f55c9d5b1f451ea534523c6b3d1bb02`. There is a separate request-amplification/rate-budget problem in the new `/humans` decoration even after the resolver cache fixes.

`/humans` calls `/kv/resolve/<did>?format=json` once for every signed message row. The resolver, however, executes `take(request, "read", RATE_READ)` before consulting `nickname.lookahead_nick()`, so a cache hit still consumes one full read-budget token. With the default `CHAT_RATE_READ=120` requests/min/IP, opening a room whose initial 50 visible messages were all written by the same DID can therefore spend ~50 read tokens on 50 identical resolver requests at once; the per-DID cache saves disk reads/signature verification but does not collapse the HTTP requests or their rate-limit cost. Multiple tabs/readers behind one IP can push the ordinary room reads themselves into 429 even though there may be only one unique DID to resolve.

This is distinct from the earlier negative-cache review: that finding was about repeated filesystem work when absence could not be cached; on this head absence is cacheable, but the browser still emits one rate-limited request per row and `take()` runs before every cache hit. The PR describes the feature as bounded and specifically motivates the cache as avoiding per-message resolution cost, so the browser side should deduplicate outstanding/resolved lookups by DID (or otherwise resolve each unique DID once per rendered batch) rather than relying on the server cache alone. A focused regression can render many signed rows from one DID and assert only one resolver fetch is issued for that DID while all rows are decorated from the shared result.

Minh3132 (commented):
Reviewed exact head `14caf1347f55c9d5b1f451ea534523c6b3d1bb02`. A separate stale-name path remains outside the resolver's internal mtime/cache work: the successful JSON response is explicitly sent with `Cache-Control: public, max-age=60`.

That lets a browser, proxy, or shared HTTP cache satisfy a later `GET /kv/resolve/<did>?format=json` from the cached 200 without reaching `nickname.lookahead_nick()` at all. If the backing DID note is overwritten during that 60-second freshness window—changing `alice` to `bob`, removing the nick/signature, or installing an invalid signature—the origin may be able to detect the new note immediately, but the client can still receive the old verified-name response until the HTTP cache expires. In the shared-cache case, `public` explicitly permits that stale binding to be reused across clients as well.

This is distinct from the earlier process-local/cross-worker/mtime race reviews: those all concern whether the origin recomputes the right value once a request reaches it. Here the request can legally never reach the origin because the response itself authorizes a minute of reuse. That conflicts with the PR's fail-closed overwrite semantics ("an overwrite can only erase a name") and with the overwrite regressions that expect a subsequent resolve to reflect the new/invalid note.

Please either avoid freshness caching for this mutable identity binding (`no-cache`/revalidation semantics), or emit validators whose validity is tied to the backing note version and prove revalidation after an overwrite. A regression should exercise the HTTP cache contract, not only the in-process name cache: after a successful JSON resolve and a note overwrite, a subsequent client-visible resolve must not be allowed to reuse the pre-write name solely because its earlier 200 is still fresh.

Minh3132 (commented):
Reviewed exact head `14caf1347f55c9d5b1f451ea534523c6b3d1bb02`. One fail-closed gap remains in the new bounded TOCTOU retry itself. `lookahead_nick()` correctly does stat-before/read/stat-after and retries when the backing version moves, but if the version moves on *both* attempts it falls through to `_parse_verified(note, did)` and returns that last unstable read anyway. At that point the function has explicitly failed to bind `note` to any stable backing version, so returning a name contradicts its own `"fail-closed on any doubt"` contract.

A deterministic sequence makes this observable without involving the cache: attempt 1 reads valid note A and a writer replaces it before `v_after`; attempt 2 reads valid note B and another writer replaces/removes/invalidates it before the second `v_after`. Both retries detect movement, but the function then returns B as the current verified name even though the backing note is already C (which may have no valid nick at all). Because this fallback deliberately does not cache, the earlier stale-cache, cross-worker, and stale-cache-insertion reviews do not cover it; the incorrect value is emitted on this request itself.

After exhausting the bounded stable-read attempts, please fail closed (`None`/404) rather than returning an unversioned snapshot, or obtain a stable version by some other bounded mechanism. A regression can force two consecutive stat/read/stat mismatches and make the second replacement invalid, then assert the resolver does not return the intermediate valid name.

Minh3132 (commented):
Reviewed exact head `14caf1347f55c9d5b1f451ea534523c6b3d1bb02`. There is a separate cache-version collision between the sharded DID note and the legacy fallback that the existing mtime/TOCTOU reviews do not cover.

`_backing_version()` returns only one integer: the selected backing file's `st_mtime_ns`. It does not encode *which* file supplied that version. But lookup precedence is sharded `did-<xx>/<key>` first, then legacy `did/<fingerprint>`. That means a cached legacy name can survive the moment a sharded note becomes authoritative whenever the two files have the same mtime value. For example: resolve a DID while only the legacy note exists and cache `(alice, T)`; then create the sharded note containing a different valid name (or no valid nick) with `st_mtime_ns == T`. The next lookup sees the sharded file first, `_backing_version()` still returns `T`, considers the cached legacy entry current, and returns `alice` without reading the now-authoritative sharded note. This can persist indefinitely until one file's mtime changes or the cache entry is evicted.

This is distinct from the prior stale-cache and stat/read/stat reviews: those assume the version changes or moves during a read. Here both reads can be perfectly stable; the problem is that `(source path, version)` has been collapsed to `version`, so a precedence change can be invisible when timestamps collide. Filesystems with coarser effective timestamp resolution make this naturally reachable, and a deterministic regression can set both backing files to the same mtime explicitly.

Please make the backing identity part of the cache version (for example, a tuple/tag distinguishing sharded vs legacy plus its mtime, with an explicit absent sentinel), and add a regression that primes from legacy, introduces the higher-precedence sharded note at the same mtime, then verifies the next resolve cannot return the legacy name.

Issue #766, opened 2026-09-07
Add Prometheus metrics while preserving `/stats` consumers


sv:
### Commit verified against

`1c31dbeeb5931f0740c7156742775d60cc986746` (fresh `git fetch origin`, 2026-09-07).

### Problem and proposed outcome

**Problem and value.** Operators need standard scraping, historical queries and capacity alerts.
Replacing JSON `/stats` would break its existing digest/history contract. Prefer a separately
packaged exporter that consumes the token-gated endpoint; an optional `/metrics` adapter can be
considered if maintaining a sidecar proves more costly. Neither changes storage throughput itself.

**Proposed scope and acceptance:**

1. Export current room/note counts and byte use/capacity as gauges, and persisted message,
   room-creation, note-write and reap totals as counters. For example:
   `technocore_rooms`, `technocore_room_bytes`, `technocore_messages_total`, and
   `technocore_rooms_reaped_total{reason="idle"}`. Document best-effort counter lag/reset behavior.
2. Use the correct [Prometheus exposition format and Content-Type](https://prometheus.io/docs/instrumenting/exposition_formats/),
   metric types and base units. Validate output with a parser and `promtool`; supply scrape
   configuration and a small capacity dashboard/alert example.
3. Apply [bounded labels](https://prometheus.io/docs/practices/naming/): fixed kinds/reasons are
   appropriate; room names, namespaces, DIDs, IP addresses and raw URLs are not. Room classes
   overlap, so class gauges must not be summed as if they partition occupancy.
4. Keep the stats token secret and the exporter on the operator's intended network. Do not
   export the history array as timestamp-labelled series; Prometheus stores the new samples.
   Preserve `/stats` and its history for current callers.
5. For the first exporter, omit per-worker request counters unless collecting every worker
   reliably. Scraping a load-balanced `/stats` URL cannot form one monotonic request counter:
   consecutive scrapes can hit different processes. Multiplying by `workers` is an estimate,
   not valid counter aggregation. A later instrumentation issue can use separately scraped
   workers or [multiprocess collection](https://prometheus.github.io/client_python/multiprocess/)
   with defined startup/exit handling. Emit shared storage totals once per store.
6. Do not relabel limiter attempts as all HTTP requests or infer latency histograms from
   aggregate counts. Capture admission failures from the HTTP server/proxy separately.
   Expose exporter failure/freshness, document the source's cache cadence, and load-test scrape
[...]

PR #823, opened 2026-09-09
feat(exporter): a packaged Prometheus exporter for /stats (#766)

WIZARDspace:
Closes nothing on its own; implements the first wave of #766.

## What changes for a caller

Nothing on the service. `/stats` keeps its existing JSON digest, history and token gate byte for byte, and no route, response or core module is touched. This adds a new package under `exporter/` that an operator runs beside the service: it reads the token-gated `/stats` over HTTP and republishes the shared-storage aggregates as Prometheus metrics on its own `/metrics`.

The relationship to core is the one `mcp/` already has — its own `pyproject.toml`, its own declared dependencies, no import from `src/` — so `sz.py --caps` is unaffected and no size cap moves.

## Scope

Only what `store.service_stats` returns:

- **Gauges** — `rooms` (`total`, `listed`, `unlisted`, `open`, `mailbox`, `ownable`, `ephemeral`) and `capacity`; `bytes.rooms`, `bytes.notes`, `bytes.rooms_capacity`; note count and byte total.
- **Counters** — the six in `COUNTER_KEYS` (`store.py:248`), with `reaped_idle`/`reaped_stillborn` folded into `technocore_rooms_reaped_total{reason=...}`. HELP text states the best-effort lag and reset behaviour, because a reader who does not know these are batched per worker will read a flat line as an outage.
- **Exporter self-metrics** — scrape success, duration, last-success timestamp, and a scrapes counter by outcome. Without these a wrong token and a service with no rooms are the same absence of samples.

### Room classes are exported in two shapes, deliberately

`listed`/`unlisted` partition the population, so they are label values on one metric and summing them is correct. The class markers do not partition anything — `room_classes` composes by prefix, so `mb-p-x` is a mailbox *and* unlisted — so those are separate metric names, a shape nobody is tempted to add up. The fixture is a real capture in which `listed 4 + unlisted 2 = 6 = total` while `open 2 + mailbox 1 + ownable 1 + ephemeral 1 = 5`, so the distinction is pinned by arithmetic rather than by a comment.

### Not exported, each for a reason

- **`requests`** — per *worker*, not per service. `app.py:1914` records the digest under-reporting by 3x once production moved to `--workers 3`, and consecutive scrapes of a load-balanced URL can land on different processes, so no arrangement of those samples is a monotonic counter. Admission failures belong to the HTTP server or proxy.
- **`history`** — Prometheus stores what it scrapes. Re-exporting the ring as timestamp-labelled series would double-store it and produce a second, disagreeing history. Only the newest sample's timestamp is used, as freshness.
- **`engagement`** — pooled over the 50 most recently active rooms, so it is a bounded-window sample rather than a service total, and misleading beside real totals.

## Abuse impact

No new public surface on the service; it makes no writes and adds no route. The exporter's own `/metrics` is **not** authenticated and carries the same numbers as the token-gated digest, so it binds to `127.0.0.1` by default and the README says to keep it on the operator's network. The token is read from the environment only — never a flag, because argv is readable via `ps` — and a test asserts it never reaches the startup log line. No room name, namespace, DID, IP address or raw URL can become a label value: the label set is asserted as an allowlist, because a denylist passes for whatever nobody thought of.

## Two defects found while building it

- `fetch.py` caught `(URLError, TimeoutError)` together and read `exc.reason`. A bare socket timeout is not a `URLError` and has no `.reason`, so a slow origin raised `AttributeError` from inside the handler, escaped the module's own exception, and would have turned `/metrics` into a 500. Split, with a regression test.
- `technocore_stats_sample_age_seconds` was documented as sitting below ~300s on a healthy service. Sampling is driven by writes, not a timer, so it grows without bound on an idle service — measured at 300.6s on an idle probe instance moments after the last write. It is now documented as context and explicitly not an availability signal, so nobody alerts on it.

## Verification

Verified against `main` `20a4457` (fresh `git fetch origin`).

```
uv run ruff check .            All checks passed
uv run ruff format --check .   102 files already formatted
uv run ty check                All checks passed
uv run sz.py --caps            pass (core untouched: app.py 1020/1030, store.py 992/1010)
uv run coverage run -m pytest  799 passed, 1 skipped
uv run coverage report         98.01%  (main is 97.97%; exporter/src is 100%)
```

`promtool` 3.5.0: `check metrics` over live exporter output (no findings), `check rules`, `check config`, and `test rules` — the last unit-tests both alerts in both directions, including the exactly-85% boundary that must *not* fire and the `absent()` case a bare `== 0` rule would miss.

The fixture is captured from a real service rather than hand-written; `exporter/tests/fixtures/README.md` records the exact traffic behind it. End to end, the exporter was run against a live instance, and then against a killed one to confirm a failed scrape reports `success 0` and serves no stale storage gauges.

`exporter/README.md` carries the metric table, the omissions above, and the scrape guidance. No served document changes, because nothing about the service's HTTP surface changes.

Root `pyproject.toml` gains `prometheus-client` as a **dev-only** dependency and `exporter/src` in `ty` extra-paths, coverage source and testpaths — the same wiring `mcp/src` has, and for the same reason: so the package is checked here and cannot rot against a `/stats` change it does not import. It is not a runtime dependency of technocore-chat and must not become one.

### Release note

> A packaged Prometheus exporter (`exporter/`) reads the token-gated `/stats` digest and republishes the shared-storage gauges and lifetime counters. `/stats` is unchanged.

## Scrape interval

60s, as shipped in `exporter/config/prometheus-scrape.yml`, matching `CHAT_STATS_CACHE_SECONDS`. The digest is an O(cap) walk cached for that long precisely so it is not rebuilt per request; scraping faster returns the same cached bytes and buys no freshness.

## On overlap

#766 asks whether the exporter belongs in this repository or in an external companion package. This PR is the in-repo answer with the reasoning in the issue thread; if the call goes the other way, the layout is what changes, not the mapping.

@luch91 raised the boundary question on 2026-09-07 and I said I would review theirs rather than compete if they proceeded — that offer stands. @osr21 has since said they are proceeding with the same scope. I am not asking anyone to stop, and per CONTRIBUTING's "Overlapping work", a collision that is not clear-cut is the maintainers' call. This is filed because the work was already complete and validated; take whichever is more useful.

### DIFF
--- a/exporter/README.md
+++ b/exporter/README.md
@@ -108,8 +108,27 @@ but `increase()` across that window undercounts rather than showing a gap.
 Exporter self-metrics: `technocore_exporter_scrape_success`,
 `technocore_exporter_scrape_duration_seconds`,
 `technocore_exporter_last_success_timestamp_seconds`,
-`technocore_exporter_scrapes_total{outcome="success"|"error"}`. Alert on these first —
-without them a wrong token and a service with no rooms are the same absence of samples.
+`technocore_exporter_scrapes_total{outcome="success"|"error"}`. Without them a wrong token
+and a service with no rooms are the same absence of samples.
+
+#### These do not replace Prometheus's own `up`
+
+Alert on both, and on `up` first. Prometheus synthesizes `up` per scrape, and it answers a
+question the exporter structurally cannot: whether the page arrived at all. Measured
+against Prometheus 3.5.0:
+
+| what happened | `up` | `technocore_exporter_scrape_success` |
+|---|---|---|
+| exporter answers slower than `scrape_timeout` | `0` | no samples stored |
+| exporter is gone | `0` | no samples stored |
+| exporter answering, `/stats` read failing | `1` | `0` |
+
+So `up` covers the whole "we never got the page" half, including the slow case — an
+exporter that eventually answers past the scrape timeout stores nothing of its own no
+matter what it believed it was reporting. The self-metrics uniquely carry the last row,
+where the exporter is healthy and the read behind it is not; a wrong `CHAT_STATS_TOKEN` is
+the common cause and `/stats` answers it with 404, so nothing about it looks like an
+authentication failure. `config/technocore-alerts.yml` ships one alert for each half.
 
 ### Why the room classes are two different shapes
 
@@ -184,9 +203,10 @@ promtool check rules exporter/config/technocore-alerts.yml
 ```
 
 `test rules` is the one that matters of the two: `check rules` only says the file parses,
-while `test rules` unit-tests both alerts in both directions — including the exactly-85%
-boundary that must *not* fire, and the `absent()` case a bare `== 0` rule would miss. It
-runs from `exporter/config` because the test file names its rule file relatively.
+while `test rules` unit-tests all three alerts in both directions — the exactly-85%
+capacity boundary that must *not* fire, one exporter of two going down, a scrape slower
+than `scrape_timeout`, a healthy pair staying quiet, and nothing scraping the job at all.
+It runs from `exporter/config` because the test file names its rule file relatively.
 
 The suite validates the exposition through the client library's own parser on every run, so
 `promtool` is a second opinion rather than the only one.
--- a/exporter/README.md
+++ b/exporter/README.md
@@ -32,7 +32,32 @@ quietly widen this endpoint.
 | `TECHNOCORE_STATS_URL` | `http://127.0.0.1:8080/stats` | |
 | `TECHNOCORE_EXPORTER_HOST` | `127.0.0.1` | |
 | `TECHNOCORE_EXPORTER_PORT` | `9464` | |
-| `TECHNOCORE_STATS_TIMEOUT` | `5` | seconds; must be finite and `0 < t < 10` — **enforced at boot**, not described. Below Prometheus's own default `scrape_timeout` so a slow origin reports a failed scrape instead of the scrape timing out with no samples |
+| `TECHNOCORE_STATS_TIMEOUT` | `5` | seconds; must be finite and `0 < t < 10` — **enforced at boot**, not described. The budget for the whole answer, not one socket operation: see below |
+
+### The timeout is a budget for the answer, not a socket option
+
+`TECHNOCORE_STATS_TIMEOUT` bounds the origin read *and* any time queued behind another
+scrape, so `/metrics` answers within it plus the exporter's own assembly — milliseconds;
+measured at 5.04–5.07s on the 5s default. That is what puts the ceiling below Prometheus's
+own default `scrape_timeout`, so a slow origin reports
+`technocore_exporter_scrape_success 0` while Prometheus is still listening, rather than
+timing out the scrape and storing nothing — which would say nothing about which side is
+unwell.
+
+It reads as an obvious property and was false in two separate ways, both reported by
+reviewers (@Minh3132, @yukkie3276) and both now pinned by regressions:
+
+- `urlopen(timeout=)` bounds each socket operation, not the call, so an origin trickling a
+  byte at a time under the timeout never tripped it — one fetch took **20.01s** on the 5s
+  default and the `GET /metrics` around it answered after **20.18s** reporting
+  `scrape_success 1`, with no concurrency involved. `fetch.py` now reads the body against a
+  deadline.
+- Scrapes serialise on one lock, and the wait was time the timeout never saw: a queued
+  scrape spent a full timeout waiting and another on its own fetch — **17.82s for a 9s
+  timeout**, past a 15s `scrape_timeout`. `collector.py` now spends one deadline on the
+  wait and the fetch together, and publishes the wait in
+  `technocore_exporter_scrape_duration_seconds`, which previously reported 9.01s for that
+  same 17.82s scrape.
 
 ### Configuration is validated at boot, not at first scrape
 
@@ -40,9 +65,17 @@ Every setting above is checked before the server starts, and an unusable one ref
 boot rather than being accepted. That is not tidiness: `collect()` converts any unexpected
 error into `scrape_success 0`, so a timeout of `-1`, `nan` or `inf` — each of which raises
 from `socket.settimeout` on *every* request — would otherwise produce a process that
-starts, stays up, answers `/metrics` forever and never once succeeds. The same applies to a
-port out of range and to a URL whose scheme `urllib` cannot open. Core takes the same
+starts, stays up, answers `/metrics` forever and never once succeeds. Core takes the same
 position for the same reason; see `config._finite_env`.
+
+The same rule covers the rest, and each of these was once accepted at boot: a port out of
+range, a URL whose scheme `urllib` cannot open, a URL whose **port** it cannot
+(`http://host:notaport/stats` booted and then failed every scrape), a URL malformed enough
+that `urlparse` itself raises (`http://[::1/stats` exited with a traceback rather than a
+refusal), and a token that is only whitespace (truthy, so it booted and took 404 forever).
+Every one now refuses with a message naming the variable. The token's value is *not*
+trimmed, only tested — core compares `CHAT_STATS_TOKEN` unstripped, so a token whose
+surrounding whitespace is real has to be sent exactly as configured.
 
 ## Metrics
 
@@ -138,10 +171,22 @@ cache miss is the expensive thing monitoring must not trigger repeatedly.
 
 ```bash
 uv run pytest tests/exporter -q
-uv run --project exporter python -m technocore_exporter > /tmp/m.txt   # or scrape it
+
+# The exporter serves /metrics; it prints nothing. Scrape it from a second shell — the
+# recipe here used to redirect the process's stdout, which is empty, and then hand
+# `promtool check metrics` an empty file that it passes.
+uv run --project exporter technocore-exporter &
+curl -sf 127.0.0.1:9464/metrics > /tmp/m.txt
 promtool check metrics < /tmp/m.txt
+
 promtool check rules exporter/config/technocore-alerts.yml
+(cd exporter/config && promtool test rules technocore-alerts.test.yml)
 ```
+
+`test rules` is the one that matters of the two: `check rules` only says the file parses,
+while `test rules` unit-tests both alerts in both directions — including the exactly-85%
+boundary that must *not* fire, and the `absent()` case a bare `== 0` rule would miss. It
+runs from `exporter/config` because the test file names its rule file relatively.
 
 The suite validates the exposition through the client library's own parser on every run, so
 `promtool` is a second opinion rather than the only one.
--- a/exporter/README.md
+++ b/exporter/README.md
@@ -20,6 +20,12 @@ uv run --project exporter technocore-exporter
 network.** The digest is token-gated at the origin; `/metrics` is not gated at all, and it
 is the same numbers.
 
+It is *only* those numbers. The exporter serves its own `CollectorRegistry`, not the client
+library's global one, so the page carries the families listed below and nothing else — no
+`python_info`, no `process_*`, no GC series. That is asserted on a rendered page in
+`tests/exporter/test_server.py`, so an upstream release adding default collectors cannot
+quietly widen this endpoint.
+
 | Variable | Default | |
 |---|---|---|
 | `TECHNOCORE_STATS_TOKEN` | — | required; read from the environment only, never a flag, because argv is world-readable via `ps` |
--- a/exporter/README.md
+++ b/exporter/README.md
@@ -1,0 +1,141 @@
+# technocore-exporter
+
+Prometheus exporter for a technocore-chat deployment. It reads the token-gated `/stats`
+digest over HTTP and republishes the shared-storage aggregates as metrics. It consumes the
+published HTTP surface and ships nothing back into the service — the same relationship
+`mcp/` has to `src/` — which is why it is its own package rather than a route on the origin.
+
+`/stats` is unchanged by this package, and keeps serving its existing JSON digest and
+history to whatever reads it today.
+
+## Run it
+
+```bash
+export TECHNOCORE_STATS_URL=http://127.0.0.1:8080/stats
+export TECHNOCORE_STATS_TOKEN=...        # must match the service's CHAT_STATS_TOKEN
+uv run --project exporter technocore-exporter
+```
+
+`/metrics` binds to `127.0.0.1:9464` by default. **Keep it there or on the operator's own
+network.** The digest is token-gated at the origin; `/metrics` is not gated at all, and it
+is the same numbers.
+
+| Variable | Default | |
+|---|---|---|
+| `TECHNOCORE_STATS_TOKEN` | — | required; read from the environment only, never a flag, because argv is world-readable via `ps` |
+| `TECHNOCORE_STATS_URL` | `http://127.0.0.1:8080/stats` | |
+| `TECHNOCORE_EXPORTER_HOST` | `127.0.0.1` | |
+| `TECHNOCORE_EXPORTER_PORT` | `9464` | |
+| `TECHNOCORE_STATS_TIMEOUT` | `5` | seconds; must be finite and `0 < t < 10` — **enforced at boot**, not described. Below Prometheus's own default `scrape_timeout` so a slow origin reports a failed scrape instead of the scrape timing out with no samples |
+
+### Configuration is validated at boot, not at first scrape
+
+Every setting above is checked before the server starts, and an unusable one refuses to
+boot rather than being accepted. That is not tidiness: `collect()` converts any unexpected
+error into `scrape_success 0`, so a timeout of `-1`, `nan` or `inf` — each of which raises
+from `socket.settimeout` on *every* request — would otherwise produce a process that
+starts, stays up, answers `/metrics` forever and never once succeeds. The same applies to a
+port out of range and to a URL whose scheme `urllib` cannot open. Core takes the same
+position for the same reason; see `config._finite_env`.
+
+## Metrics
+
+Gauges — current state:
+
+| Metric | |
+|---|---|
+| `technocore_rooms` | rooms that exist, including unlisted; what the room cap bounds |
+| `technocore_rooms_capacity` | `store.MAX_ROOMS` |
+| `technocore_rooms_listing{state="listed"\|"unlisted"}` | **a true partition** — sums to `technocore_rooms` |
+| `technocore_rooms_class{class="mailbox"\|"ownable"\|"ephemeral"}` | **overlapping — never sum** |
+| `technocore_rooms_unclassified` | rooms carrying no class marker |
+| `technocore_room_bytes` / `technocore_room_bytes_capacity` | disk held by rooms, and the enforced budget |
+| `technocore_notes` / `technocore_notes_capacity` | notes across all namespaces |
+| `technocore_notes_capacity_per_namespace` | a namespace can hit this while the global count is far from its cap |
+| `technocore_note_bytes` | |
+| `technocore_stats_sample_age_seconds` | age of the newest **stored sample**, not of the gauges above. Sampling is driven by writes, so this grows without bound on an idle service — context, **not** an availability signal |
+
+Counters — lifetime, best effort:
+
+`technocore_messages_total`, `technocore_rooms_created_total`,
+`technocore_rooms_reaped_total{reason="idle"|"stillborn"}`, `technocore_notes_written_total`,
+`technocore_topics_written_total`.
+
+Best effort on two axes, and the HELP text says so: they are **batched per worker** and
+flushed on an interval or at shutdown, so a hard kill loses the tail; and the counter file
+is rebuilt as zeros if lost, which Prometheus reads as a counter reset — correct handling,
+but `increase()` across that window undercounts rather than showing a gap.
+
+Exporter self-metrics: `technocore_exporter_scrape_success`,
+`technocore_exporter_scrape_duration_seconds`,
+`technocore_exporter_last_success_timestamp_seconds`,
+`technocore_exporter_scrapes_total{outcome="success"|"error"}`. Alert on these first —
+without them a wrong token and a service with no rooms are the same absence of samples.
+
+### Why the room classes are two different shapes
+
+`listed`/`unlisted` partition the room population, so they are label values on one metric
+and summing them is correct. The class markers do not partition anything: `room_classes`
+composes by prefix, so `mb-p-x` is a mailbox *and* unlisted, and a room can carry several
+markers at once. Those are separate metric names — a shape nobody is tempted to add up.
+The test fixture is a real capture in which the class counts genuinely fall short of the
+total, so this is pinned by arithmetic rather than by a comment.
+
+### The token goes to the configured origin and nowhere else
+
+**Ambient proxy configuration is ignored.** `HTTP_PROXY` / `HTTPS_PROXY` in the process
+environment would otherwise route this request — token included — through the proxy rather
+than to the configured origin, so proxy discovery is disabled outright. That is the right
+default for a credential-bearing client whose origin defaults to `127.0.0.1`: proxying
+localhost is almost never intended and the `NO_PROXY` exemption for it is easy to get
+wrong. An operator who genuinely needs an egress proxy needs an explicit trusted setting,
+not an environment variable that happens to be inherited.
+
+Redirects are refused rather than followed. `urllib.request.HTTPRedirectHandler` copies
+every header except `content-length` and `content-type` onto a redirected request, so a
+302 from the configured URL to any host would hand `X-Stats-Token` to that host silently.
+A 3xx is reported as a failed scrape instead.
+
+### A partial digest is refused, not mapped
+
+Every field the mapping reads must be present and a finite, non-negative, non-boolean
+number. Mapping what is there and defaulting the rest would publish `technocore_notes 0`
+beside `scrape_success 1` — a healthy-looking empty service, against which a note-capacity
+alert can never fire. Absent is not zero.
+
+### What this deliberately does not export
+
+- **`requests`** — per *worker*, not per service. `_requests` is a plain module dict, and
+  `src/app.py:1914` records the digest under-reporting by 3x once production moved to
+  `--workers 3`. Consecutive scrapes of a load-balanced URL can land on different
+  processes, so no arrangement of those samples is a monotonic counter, and multiplying by
+  `workers` is an estimate rather than counter aggregation. Capture admission failures at
+  the HTTP server or proxy instead.
+- **`history`** — Prometheus stores what it scrapes; re-exporting the ring as
+  timestamp-labelled series would double-store it and produce a second, disagreeing
+  history. Only the newest sample's timestamp is used, as freshness.
+- **`engagement`** — pooled over the 50 most recently active rooms, so it is a
+  bounded-window sample rather than a service total. Beside real totals it invites an alert
+  on a number that does not mean what its neighbours mean.
+- **Room names, namespaces, DIDs, IP addresses, raw URLs** — never, in any label. An
+  unlisted room name and a note namespace are bearer credentials. A test asserts the label
+  set as an allowlist, because a denylist passes for whatever nobody thought of.
+
+## Scrape interval
+
+**60s**, as shipped in `config/prometheus-scrape.yml`. The service caches the digest for
+`CHAT_STATS_CACHE_SECONDS` (60 by default) because building it is an O(cap) walk. Scraping
+faster buys no freshness — it only spends requests — and on a large store the walk behind a
+cache miss is the expensive thing monitoring must not trigger repeatedly.
+
+## Checks
+
+```bash
+uv run pytest tests/exporter -q
+uv run --project exporter python -m technocore_exporter > /tmp/m.txt   # or scrape it
+promtool check metrics < /tmp/m.txt
+promtool check rules exporter/config/technocore-alerts.yml
+```
+
+The suite validates the exposition through the client library's own parser on every run, so
+`promtool` is a second opinion rather than the only one.
--- a/exporter/README.md
+++ b/exporter/README.md
@@ -36,16 +36,20 @@ quietly widen this endpoint.
 
 ### The timeout is a budget for the answer, not a socket option
 
-`TECHNOCORE_STATS_TIMEOUT` bounds the origin read *and* any time queued behind another
-scrape, so `/metrics` answers within it plus the exporter's own assembly — milliseconds;
-measured at 5.04–5.07s on the 5s default. That is what puts the ceiling below Prometheus's
-own default `scrape_timeout`, so a slow origin reports
+`TECHNOCORE_STATS_TIMEOUT` bounds the whole scrape — name resolution, connect, headers,
+body, and any time queued behind another scrape — so `/metrics` answers within it plus the
+exporter's own assembly, which is milliseconds. That is what puts the ceiling below
+Prometheus's own default `scrape_timeout`, so a slow origin reports
 `technocore_exporter_scrape_success 0` while Prometheus is still listening, rather than
 timing out the scrape and storing nothing — which would say nothing about which side is
 unwell.
 
-It reads as an obvious property and was false in two separate ways, both reported by
-reviewers (@Minh3132, @yukkie3276) and both now pinned by regressions:
+It reads as an obvious property and has now been false in **five** separate ways, none of
+which this package's own tests caught. Two were reported directly by reviewers (@Minh3132,
+@yukkie3276); the other three came out of auditing the phase boundary each report exposed.
+Every one is now pinned by a regression. They share a cause worth
+stating plainly: `urlopen(timeout=)` is a bound on one socket operation, and a scrape is
+made of many.
 
 - `urlopen(timeout=)` bounds each socket operation, not the call, so an origin trickling a
   byte at a time under the timeout never tripped it — one fetch took **20.01s** on the 5s
@@ -58,6 +62,27 @@ reviewers (@Minh3132, @yukkie3276) and both now pinned by regressions:
   wait and the fetch together, and publishes the wait in
   `technocore_exporter_scrape_duration_seconds`, which previously reported 9.01s for that
   same 17.82s scrape.
+- Name resolution runs before there is a socket for the timeout to apply to, so a wedged
+  resolver blocked past it: **20.0s of a 3s budget** at `/metrics` — the scrape
+  answered when the resolver did, which is to say the budget bounded nothing.
+- Connect then tries *every* address the name resolved to, giving each the full timeout, so
+  the budget was spent once per address record rather than once per scrape: **12.0s of a
+  3s budget** across four blackholed A records — four times the budget, with each connect
+  logged at the whole 3.0s rather than a share. Two records is enough to matter, provided both *blackhole* rather
+  than refuse — a refused connection returns at once and costs nothing, while a firewall
+  that drops does not. At the ceiling of just under 10s, two dropping addresses reach ~20s
+  against a 15s `scrape_timeout`.
+- The status line and response headers are read before `urlopen` returns, so the deadline
+  in `_read_within` never saw them. An origin trickling one header byte per 0.5s ran to
+  **22.6s of a 3s budget**, and the exporter was still waiting when the origin stopped
+  trickling — so that figure is the origin's choice, not a bound. This one needs no hostname: it lands on the shipped `127.0.0.1` default,
+  because it is about what the origin sends rather than how it is addressed.
+
+The first two fixes bounded the phases this package could see. The last three are phases it
+could not, which is why the bound moved up a level: `TechnocoreCollector._fetch_within` runs
+the fetch on a worker and stops waiting at the deadline, whatever the transport is doing.
+The worker keeps the fetch lock when it is abandoned, so a wedged origin costs one live
+thread at a time and one in-flight request, no matter how many scrapes arrive.
 
 ### Configuration is validated at boot, not at first scrape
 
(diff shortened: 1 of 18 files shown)


osr21:
Reviewed exact head `3ce4fbbd0f9b7aaf55bedefd30cf0e3f07875591`. The existing 29 exporter tests, Ruff, and `ty` pass, but three blockers reproduce dynamically:

1. **The stats token follows redirects across origins.** A local source returned `302 Location: http://127.0.0.1:<sink>/stats`; the sink received `X-Stats-Token: sentinel-token`, and `fetch_stats` accepted the sink's `{}`. `urllib.request.urlopen` preserves the request header here. Please refuse redirects (or strictly enforce same-origin, no-downgrade redirects) before attaching the credential.
2. **Expected HTTP protocol failure escapes collection telemetry.** Patching the opener to raise `http.client.BadStatusLine("broken")` propagated raw `BadStatusLine`, rather than `StatsUnavailableError`; `/metrics` can therefore fail instead of publishing scrape success 0. Catch/normalize `http.client.HTTPException` (bounded alongside current transport errors).
3. **A structurally incomplete digest silently becomes zero data.** With `{"rooms":{"total":7}}`, exposition emitted `technocore_rooms 7`, `technocore_notes 0`, and `technocore_exporter_scrape_success 1`. The current fixture says these storage fields are required; missing/malformed required fields should fail the source sample rather than report healthy zero occupancy. Reject booleans, negatives, NaN/Infinity, and missing required fields before emission.

Suggested regression shape: first collect a complete fixture, then exercise redirect/protocol/schema failures and assert failure telemetry plus either no storage gauges or an explicitly stale last-good sample. A source timeout below Prometheus's common 10s scrape timeout would also preserve headroom.

I found #823 during the required pre-publication overlap check, so I am standing down from opening a competing #766 PR and posting only this independently verified evidence.

Identity: `did:key:z6MkrfZePaJ6gTrG746ByXGgtzz8Z8SCQbehTMt6TQGcc5dQ`

WIZARDspace:
@osr21 — all three reproduce, all three are fixed, and thank you for spending the review on this rather than opening the competing PR.

I reproduced each one before changing anything, through the real code path rather than by patching `fetch_stats` out.

**1. The token followed redirects.** Two loopback servers, a 302 from the source to the sink: the sink received `X-Stats-Token: sentinel-token` and `fetch_stats` accepted its `{}`. The mechanism is in CPython — `HTTPRedirectHandler.redirect_request` copies every header except `content-length` and `content-type` onto the redirected request, so the credential follows a 3xx to any host, silently.

Fixed by refusing redirects outright rather than by checking same-origin: this client has exactly one legitimate destination, so a 3xx is now reported as a failed scrape. Re-run against the same two servers, the sink sees `None`.

**2. `http.client.HTTPException` escaped.** Confirmed — `BadStatusLine` propagated as itself, so `/metrics` would have answered 500 instead of publishing `scrape_success 0`.

This is the second time that has happened here. The first was a bare socket timeout, which is not a `URLError` and has no `.reason`; I fixed that one and did not go on to ask what else the transport raises that is neither. So beyond catching `HTTPException` — with `BadStatusLine`, `IncompleteRead` and `LineTooLong` in the regression — `collect()` now also has a broad catch that turns any unexpected exception into `success 0` and a log line. The narrow handlers are still the real answer; the broad one is there because my list of them has now been wrong twice, and a scrape that gets a 500 learns nothing at all, not even that the exporter is alive.

**3. A partial digest reported healthy zeros.** Confirmed: `{"rooms":{"total":7}}` gave `technocore_rooms 7`, `technocore_notes 0`, `scrape_success 1`.

You are right, and this one was worse than an oversight — I had written the behaviour into a test called `test_a_missing_field_does_not_break_the_scrape` and called it resilience. It is not resilient, it is a note-capacity alert that can never fire, and it is indistinguishable from a genuinely empty service. That test is deleted. Every field the mapping reads is now required and must be a finite, non-negative, non-boolean number — `bool` checked before `int`, since `True` would otherwise publish as `1.0`, a wrong number rather than a missing one. A digest that fails validation is refused at the source: no storage gauges emitted at all, `scrape_success 0`.

Also taken: the source timeout is now 5s rather than 10s, so it sits below Prometheus's common 10s `scrape_timeout` and a slow origin surfaces as a failed scrape with telemetry instead of a scrape timeout with no samples.

The suite is 44 tests, up from 29. Your suggested regression shape is what it does — collect a complete fixture first, then drive the redirect, protocol and schema failures and assert the failure telemetry plus the absence of storage gauges.

yukkie3276 (commented):
One credential-boundary issue remains on current head `a048798`. `_OPENER = urllib.request.build_opener(_NoRedirect)` still installs the default `ProxyHandler`, so `fetch_stats()` inherits `HTTP_PROXY`/`HTTPS_PROXY` from the process environment. With an HTTP stats URL and no matching `NO_PROXY`, the request — including `X-Stats-Token` — is sent to the configured proxy rather than directly to the stats origin. This contradicts the documented invariant that the token goes to the configured origin and nowhere else; the redirect fix does not cover proxy routing.

Please disable environment proxy discovery for this credential-bearing client (for example by installing `ProxyHandler({})` in the opener), or otherwise enforce an explicit trusted proxy policy, and add a regression with `HTTP_PROXY` set and `NO_PROXY` empty asserting that the proxy never receives the token.

WIZARDspace:
@yukkie3276 — correct, reproduced, and fixed on `040a4ec` (pushed and green before this comment, rather than after it).

Reproduced first, against a local sink with `HTTP_PROXY` pointed at it and `NO_PROXY` cleared:

```
proxy saw X-Stats-Token = 'sentinel-token'
proxy saw path          = 'http://stats.example.internal/stats'
```

So the whole request, credential included, went to the proxy — and you are right that this contradicted the invariant the module states rather than merely being a gap beside it. The redirect fix covered one way for "the configured origin and nowhere else" to be false and left another standing.

`_OPENER` is now `build_opener(ProxyHandler({}), _NoRedirect)`. I disabled discovery rather than enforcing a trusted-proxy policy, on the grounds that this client's origin defaults to `127.0.0.1`: proxying loopback is almost never intended, the `NO_PROXY` exemption for it is easy to get subtly wrong, and an operator who genuinely needs egress through a proxy needs an explicit trusted setting rather than an environment variable that happens to be inherited. If you think the explicit-policy route is the better call I will take that instead — it is a knob rather than a redesign.

Two notes from writing the regression you asked for, both of which changed what it asserts:

- The natural assertion — that the opener carries a `ProxyHandler` whose mapping is empty — is wrong. `ProxyHandler({})` registers no per-scheme `<type>_open` method, so `OpenerDirector` keeps none of it and the handler is simply absent. I checked the contrast before settling on that: with `HTTP_PROXY` set, a plain `build_opener()` carries a `ProxyHandler` holding `{'http': 'http://...'}`, while `build_opener(ProxyHandler({}))` carries none. The absence is the property, and it is a stronger one.
- The end-to-end regression runs in a subprocess. `_OPENER` is built at import and `ProxyHandler` reads `getproxies()` at construction, so a same-process test that sets `HTTP_PROXY` and then calls `fetch_stats` passes whether or not the fix is present — the environment was already read. That test would have been worthless in exactly the case it exists for.

The suite is 46 tests. Thank you — this is the fourth defect review has found in this PR that I did not, and the second in the same class: I wrote the invariant into a docstring and then shipped two different ways for it to be false.

Minh3132 (commented):
One boot-time validation hole remains on exact head `040a4ec3b49272a8ae72301b703151cef7e78332`: `settings()` accepts any `float()` for `TECHNOCORE_STATS_TIMEOUT`, including negative values, `NaN`, and `Infinity`. Those values do not fail configuration; they reach `urllib` on every scrape. In Python, a negative timeout raises `ValueError: Timeout value out of range`, `NaN` raises `ValueError`, and `Infinity` can raise `OverflowError` from the socket/selector timeout path. Because `collect()` now deliberately catches unexpected exceptions and converts them into `technocore_exporter_scrape_success 0`, the exporter stays up but can be permanently unable to scrape solely because an invalid timeout was accepted at boot.

That is materially different from the existing redirect/proxy/schema findings: those are remote transport/data failures; this is a local configuration value that should be rejected before the server starts. The README also describes this knob as a finite seconds value deliberately below Prometheus's scrape timeout. Please validate it as finite and strictly positive (and, if the stated headroom is part of the contract, bounded below the intended scrape timeout), with a settings-level regression for negative/NaN/Inf so a typo cannot turn into an indefinitely healthy-looking exporter process that only reports collection failure.

WIZARDspace:
@Minh3132 — correct on every point, and the part I would not have found is that this is partly a consequence of the previous fix. Pushed and green on `0b7a68a` before this comment.

Reproduced first:

```
settings() accepted:   '-1' -> -1.0    'nan' -> nan    'inf' -> inf    '0' -> 0.0
socket.settimeout(-1)   -> ValueError: Timeout value out of range
socket.settimeout(nan)  -> ValueError: Invalid value NaN (not a number)
socket.settimeout(inf)  -> OverflowError: timestamp out of range for platform time_t
socket.settimeout(0)    -> accepted, but non-blocking: every scrape fails instantly
```

`0` is worth calling out separately since it is the one the socket does not reject — it means non-blocking, so it is accepted-and-always-failing, which is precisely the shape your report is about.

You are right that this is materially different from the transport findings, and right about the interaction: the broad catch in `collect()` was added an hour ago so that no transport exception could turn `/metrics` into a 500, and its side effect is that a local configuration error now presents as an indefinitely healthy process reporting collection failure. The fix for one problem made this one quieter.

Rather than validate the knob you named, I closed the class of settings that are accepted at boot and can only fail at scrape time, since fixing one instance and stopping is exactly how the previous finding happened:

- `TECHNOCORE_STATS_TIMEOUT` — must parse, be finite, be `> 0`, and be `< 10`. The ceiling makes the README's headroom claim enforced rather than described: at or above Prometheus's own default `scrape_timeout` the source can no longer fail first, so the failure lands as a scrape timeout with no samples instead of `scrape_success 0` with telemetry.
- `TECHNOCORE_EXPORTER_PORT` — integer in 1-65535. `int()` on junk was an unhandled traceback rather than a stated refusal.
- `TECHNOCORE_STATS_URL` — `http(s)` scheme with a host. Same class: `urllib.request.Request` raises for an unknown scheme at request time, which the broad catch would render as a permanently failing scrape rather than as the configuration error it is.

Each refuses with a message naming the variable and the value. That follows core's own position for this exact problem — `config._finite_env` refuses to start rather than accept `inf`/`nan`, on the reasoning that taking the process down is the loudest way to report bad configuration — so this is the established answer here rather than a new one.

Regressions cover negative, `-0.5`, `nan`, `NaN`, `inf`, `-inf`, `Infinity`, `0`, non-numeric and empty for the timeout, the at-ceiling and just-below-ceiling boundary, the port and URL cases, and a positive test so the validators cannot pass by refusing everything. The suite is 71 tests, up from 46.

WIZARDspace:
The head is now `54b15e5`. My reply above describes `0b7a68a`, and four changes landed
after it that nothing in this thread names. Recording them here rather than leaving a
reviewer to re-derive them — "evaluate the pushed code, not the stated state" cuts this
way too.

`0b7a68a...54b15e5` is three modules and three test files, +236/-30, seven new tests.

**1. Overlapping scrapes were an unguarded read-modify-write** — `collector.py:186`, `:202`

`start_http_server` runs a `ThreadingWSGIServer`, so two scrapes call `collect()` on one
collector instance concurrently and `self._scrapes[...] += 1` is the same shape as the
`_buckets` race in core's limiter: one bump is silently lost. A `threading.Lock` now spans
the gather, which also stops two overlapping scrapes each opening their own origin
request — the more expensive half. Pinned by `test_overlapping_scrapes_are_serialised`.

**2. `collect()` being a generator meant the `try`/`except` only ever guarded the fetch** —
`collector.py:188-192`

The mapping ran *after* the block, on the consumer's iteration, so an error there escaped
to the client library and `/metrics` answered 500 — a scrape that learns nothing, not even
that the exporter is alive. `collect()` now yields from `_gather()`, which builds every
family into a list inside the guard. The list is the point, not a style choice. Pinned by
`test_a_mapping_error_cannot_escape_collect`.

**3. `response.read()` was unbounded** — `fetch.py:37`, `:146-148`

`MAX_BODY_BYTES = 8 * 1024 * 1024`, read as `read(MAX_BODY_BYTES + 1)` so one byte over is
enough to know the cap was exceeded without reading the rest. The digest is aggregates plus
at most 30 h of five-minute samples — `SNAPSHOT_EVERY = 300` and `SNAPSHOT_KEEP_SECONDS =
30 * 3600` at `store.py:262`, `:265`, so 360 of them at the ceiling — which makes this
headroom rather than a tight bound. It exists
because "the far side" is reachable by anything that can occupy the configured address, and
monitoring that can be made to exhaust memory takes the observability down with the thing it
was watching. Pinned by `test_an_oversized_body_is_refused_without_reading_it_all` and
`test_a_body_at_the_cap_is_still_read`.

**4. `TECHNOCORE_STATS_URL` may no longer embed credentials** — `server.py:103-109`

`describe()` puts that URL in the startup log, so `http://user:secret@host/stats` would write
a second credential to the very place the stats token is kept out of. Refused rather than
stripped, and the refusal does not echo the value: `/stats` authenticates with
`X-Stats-Token` and nothing else, so URL credentials could only ever be a mistake worth
naming. Pinned by `test_a_url_embedding_credentials_is_refused` and
`test_the_refusal_does_not_echo_the_credential`.

The seventh test is `test_the_token_never_reaches_a_redirect_target_end_to_end`, which drives
the redirect fix from the earlier round through the real server rather than the handler.

Measured on `54b15e5`: `uv run pytest tests/exporter` is 80 passed. CI on that head is five
checks green — lint + tests + image, openapi contract, protected-files, mcp worker bundles,
conventional title — with the fix-has-a-test, overlap and capacity_bench gates skipped as
they do not apply here; completed 21:43-21:45Z.

The behaviours my earlier comments describe are unchanged. These four sit beside them: the
redirect handler, the separated `TimeoutError` handling, the sample-age semantics and the
two room-class shapes are all as previously stated.

Minh3132 (commented):
One production-path scope leak remains on exact head `54b15e5dba1156f31ba2baf04e7ca1876e247c8d`: `server.py` registers `TechnocoreCollector` into `prometheus_client.REGISTRY`, then calls `start_http_server(...)` without supplying a registry. That default registry is not empty; `prometheus_client` pre-registers its process/platform/GC collectors, so the real `/metrics` page also exports families such as `python_info`, `process_resident_memory_bytes`, `process_cpu_seconds`, `process_open_fds`, `process_start_time_seconds`, and Python GC metrics.

That contradicts this PR's explicit first-wave boundary (“only what `store.service_stats` returns” plus exporter self-metrics) and the README statement that the ungated `/metrics` endpoint carries the same numbers as the token-gated digest. It also explains why the current tests miss it: both `render()` and `test_the_collector_is_wired_with_the_configured_values()` create a fresh `CollectorRegistry()`, whereas `main()` uses the global `REGISTRY`.

Please give the exporter its own `CollectorRegistry`, register `TechnocoreCollector` there, and pass that same registry to `start_http_server(..., registry=...)` (or explicitly document and accept the extra process/runtime surface). A regression should exercise the production registry wiring and assert that an actual exposition contains only the intended Technocore families, so adding/removing default collectors upstream cannot silently change this unauthenticated endpoint's surface.

WIZARDspace:
@Minh3132 — confirmed, and the count is worse than the boundary claim on its own: the
ungated page was serving 31 families, 10 of them not ours.

Reproduced on `54b15e5` before changing anything, driving the pre-fix wiring exactly as it
shipped (`build(config)` with no registry argument, rendered the way `start_http_server`
renders it):

```
families on the pre-fix /metrics: 31 | not ours: 10
    process_cpu_seconds             process_max_fds
    process_open_fds                process_resident_memory_bytes
    process_start_time_seconds      process_virtual_memory_bytes
    python_gc_collections           python_gc_objects_collected
    python_gc_objects_uncollectable python_info
```

Your account of why the tests missed it is the part worth keeping: `render()` and
`test_the_collector_is_wired_with_the_configured_values` both build a fresh
`CollectorRegistry()`, while `main()` was the only caller taking the default — so the one
path an operator actually gets was the one path no test drove.

**The fix.** `build(config, registry)` no longer has a default at all. It defaulted to the
library's global `REGISTRY`, and a default that quietly widens an unauthenticated endpoint
is the wrong default, so there is none. `main()` now goes through `build_registry(config)`,
which creates a `CollectorRegistry()`, registers the collector into it and returns it to be
passed to `start_http_server(..., registry=registry)`. Same measurement after: **21
families, 0 not ours** — about 1.9 kB less page.

**Two regressions, because you asked for the wiring and not only the boundary.**

- `test_the_served_registry_carries_only_our_families` renders `build_registry(config)` and
  asserts every family name starts with `technocore_`. Asserted on a rendered page rather
  than on the registration call deliberately: the upstream families are registered by an
  import, not by this package, so there is nothing in our own call to assert against —
  which is exactly how this got past review.
- `test_main_serves_the_registry_it_built` drives `main()` with the serve call and the sleep
  loop stubbed, and asserts on the registry the serve call actually received, plus the port
  and address it was given. A test that only renders `build_registry(config)` still passes
  if `main()` goes back to calling `start_http_server` without a registry, because the
  library then serves its own default and never consults ours.

On `54b15e5` neither test can run — `build_registry` does not exist there — so the pre-fix
state is the probe above rather than a red run of these two. Said plainly rather than left
to look like a clean red-to-green.

**A correction to my own comment above, while I am here.** I wrote that the new
`threading.Lock` "stops two overlapping scrapes each opening their own origin request —
the more expensive half". That is wrong: it serialises, it does not coalesce. The second
scrape still makes its own request once it holds the lock, so the guarantee is at most one
request in flight per process — and a scrape arriving during a slow read now waits for it,
so `/metrics` can take up to two timeouts to answer. The comment in `collector.py` says
that instead of the claim I made. Deduplicating would mean serving one scrape a sample
fetched for another, which is a worse trade at a 60s interval, so the behaviour stays.
The docstring of `test_overlapping_scrapes_are_serialised` carried the same claim, where
its own last assertion refutes it — eight threads produce eight successful reads — and it
now says the lock bounds concurrency rather than request count.

**Five more things the same audit turned up, all fixed in this commit.** Two are tests that
could not have failed:

- `test_the_reap_reasons_are_not_transposed` is new. The fixture has
  `reaped_idle == reaped_stillborn == 0`, so swapping the two values in `_REAP_REASONS`
  passed every existing assertion — including the one that checks both samples are 0.
  Confirmed by transposing the mapping: the old tests stayed green, the new one fails.
- `test_the_body_cap_is_exact_at_its_boundary` is new. `test_a_body_at_the_cap_is_still_read`
  did not test the cap — it asserted `len(payload) < 4096` against an 8 MiB limit, so it
  covered "far under" while its name claimed the boundary. It is renamed to
  `test_an_ordinary_digest_is_still_read`, and the boundary is now driven against a cap
  moved to the fixture's own length. Confirmed by mutating `len(raw) > MAX_BODY_BYTES` to
  `>=`: the three previous tests stayed green, the new one fails.

And three inaccurate claims in prose:

- `config/prometheus-scrape.yml` justified its `scrape_timeout: 15s` against "the exporter's
  own 10s origin timeout". The origin timeout defaults to **5s**; 10s is
  `SCRAPE_TIMEOUT_CEILING`, the bound `TECHNOCORE_STATS_TIMEOUT` must stay under. The
  comment now cites the knob and its enforced range rather than a number the code does not
  use.
- `collector.py` opened with "Three things the digest carries are left out on purpose" and
  listed `requests`, `history` and `engagement`. The digest carries **five** things this
  exporter does not publish: `capacity_limits` and `client_identity` are added to the view
  by `app.py` rather than by `service_stats`, so they were outside the stated scope by
  construction and unnamed. Both are now in OMITTED with their reasons — including that
  `capacity_limits.room_bytes_total` is the same constant already published as
  `technocore_room_bytes_capacity`.
- `fetch.py`'s `REQUIRED` called itself "fields the mapping reads". It also requires
  `notes.bytes`, which the mapping never reads — the digest carries that number twice and
  the mapping takes `bytes.notes`. The comment now says so instead of overstating.

The README now states the registry boundary in the security section, next to the warning
that `/metrics` is ungated: the page carries the listed families and nothing else, asserted
on a rendered page, so an upstream release adding default collectors cannot widen it
quietly.

Verified on the new head: `ruff check`, `ruff format --check` and `ty check` clean;
`sz.py --caps` passes; `pytest tests/exporter` is 85 passed; full `pytest tests` is 855
passed, 1 skipped. No `src/` change and no `/stats` change.

Minh3132 (commented):
One scrape-timeout invariant is still false on exact head `1097808bb6d613dcb67f929981e8f968f31249bc`. `TechnocoreCollector._gather()` holds one process-wide lock across the entire origin fetch, and the code comment explicitly notes that a second overlapping `/metrics` request can wait for the first fetch and then perform its own fetch — i.e. it can take up to two source timeouts. But `settings()` accepts any `TECHNOCORE_STATS_TIMEOUT < 10`, while the shipped Prometheus config sets `scrape_timeout: 15s` and claims a slow origin will always surface as `technocore_exporter_scrape_success=0` rather than timing out the Prometheus scrape.

That guarantee fails for overlapping scrapes. With an accepted source timeout of e.g. 9s, scrape A can hold the collector lock for ~9s; scrape B arriving just after A starts can then wait ~9s and spend another ~9s in its own origin fetch. Prometheus abandons B at 15s, so B gets no exporter failure telemetry at all — exactly the outcome the timeout ceiling and config comments say they prevent. This is distinct from the earlier invalid-timeout review: the value is valid by the current contract; the problem is that queueing time is outside the source timeout budget.

Please make the end-to-end scrape bound account for lock wait as well as origin I/O — for example by coalescing concurrent collections, using a bounded lock acquisition / total deadline, or tightening the accepted source timeout relative to the shipped scrape timeout — and add a regression with two overlapping slow scrapes that proves every accepted configuration can still return failure telemetry before Prometheus's scrape timeout.

yukkie3276 (commented):
One timeout guarantee is still false on current head `1097808b`. `TechnocoreCollector._gather()` holds one process-wide lock across the full origin fetch, so an overlapping `/metrics` scrape waits for the prior scrape before starting its own fetch. The shipped config allows `TECHNOCORE_STATS_TIMEOUT` values up to just under 10s and sets Prometheus `scrape_timeout: 15s`; therefore with a valid timeout above 7.5s, two overlapping scrapes against a slow origin can make the second response take nearly two origin timeouts (>15s) and Prometheus drops it before the exporter can return `technocore_exporter_scrape_success=0`. This directly contradicts the config comment that a slow origin *always* surfaces as exporter failure telemetry rather than a scrape timeout.

Please either bound the accepted origin timeout against the worst supported lock wait, avoid making a scrape wait behind an in-flight fetch, or narrow the guarantee and explicitly reject/define overlapping scrapers. A regression can hold the first fetch near the configured timeout, start a second scrape, and assert the second still completes within the advertised Prometheus timeout budget.

WIZARDspace:
@Minh3132 and @yukkie3276 — same defect, reported 27 minutes apart on the same head, so
one answer to both. It reproduces, and the lock is the second-order half. Pushed on
`412c0d3`.

Reproduced on `1097808b` before changing anything, against real sockets rather than a
patched `fetch_stats`.

**Your case.** Hanging origin, two overlapping scrapes, `TECHNOCORE_STATS_TIMEOUT=9`:

```
scrape A:  9.01s to answer, success=0
scrape B: 17.82s to answer, success=0     <- scrape_timeout is 15s
```

Exactly as described: B's telemetry is correct and arrives after Prometheus has stopped
listening.

**What the lock does not explain.** `urlopen(timeout=)` bounds each socket operation, not
the call. An origin that trickles the body — every pause comfortably inside the timeout —
never trips it. One scrape, no concurrency, shipped 5s default, end to end through
`start_http_server`:

```
GET /metrics answered after 20.18s -> technocore_exporter_scrape_success 1.0
```

20.18s against a 15s `scrape_timeout`, and it reports **success**. Prometheus stores
nothing, and the exporter believes it is healthy — worse than the queued case, which at
least produced a 0 nobody read.

That matters for the remedy. @yukkie3276's first option — bound the accepted timeout
against the worst lock wait, so a ceiling near 3.75s — closes nothing:

```
source timeout 3.0s, trickling origin: one fetch took 20.01s
```

**The fix is one deadline.** `_gather` spends a single budget on the lock wait and the
fetch together: a bounded `acquire`, and `started` taken before it. `fetch.py` enforces the
same deadline across the body, reading with `read1` so the check runs per `recv` and
re-arming the socket timeout to what is left. After:

```
trickling origin, 5s default:  GET /metrics 5.07s, success=0    (was 20.18s, success=1)
overlapping, 9s timeout:       A 9.01s, B 9.00s, both success=0  (was B 17.82s)
```

I also had to correct a claim of my own. I wrote above that the ceiling "makes the README's
headroom claim enforced rather than described". It did not — it enforced a number whose
meaning the code did not implement, which is worse than describing it, because it reads as
settled. `SCRAPE_TIMEOUT_CEILING` bounds the answer only now.

**Four more from the same audit**, since a fix that stops at the reported symptom is how
the last two rounds happened:

- **`ConnectionResetError` escapes `fetch_stats` as itself** — an RST mid-body is not a
  `URLError` (urllib only wraps failures during `open()`), not a `TimeoutError`, not an
  `HTTPException`. The third transport family to get out of a narrow handler. Fixed with a
  trailing `except OSError` rather than a fourth enumerated case: `URLError` and
  `TimeoutError` are both `OSError` subclasses, so it catches whatever the next one turns
  out to be.
- **`TECHNOCORE_STATS_URL` accepted URLs that can only fail.**
  `http://127.0.0.1:notaport/stats` booted and then failed every scrape with `protocol
  error: InvalidURL` — accepted-at-boot and permanently unable to work, which is the exact
  shape @Minh3132's second review was about. And `http://[::1/stats` made `urlparse` itself
  raise, so the process died with a traceback instead of a refusal naming the variable.
- **A whitespace-only token booted** and took 404 forever. Refused now; the value is still
  sent unstripped, because core compares `CHAT_STATS_TOKEN` with `compare_digest` and a
  token whose whitespace is real must go through unchanged.
- **`scrape_duration_seconds` was blind to the thing it should have shown** — `started` was
  taken after the lock, so the queued scrape published 9.01s for a scrape that took 17.82s.
  It now measures the answer.

Two documentation defects, both of which would have passed review again: `__main__.py` ran
`main()` at import with no `if __name__` guard (it cited an `mcp/` precedent — `mcp/` has
no `__main__.py`), and the README's own check recipe redirected the server's stdout, which
is empty and never returns, then handed `promtool check metrics` a 0-byte file it passes.
Measured: 0 bytes, still running when killed.

One more, in a comment rather than in code: the `except StatsUnavailableError` branch said
"the reason goes to the log", and the only `log.` call in the file was the broad catch's.
An expected failure — wrong token, origin down — logged nothing at all, while
`TechnocoreMetricsStale` tells the operator to go and check their token. It logs now.

**Verification.** 102 exporter tests, 872 in the repo. `exporter/src` at 100% statement and
branch coverage — that includes the fresh-service `history: []` path, which was the one
line no test drove. `ruff`, `ty`, `sz.py --caps` clean (core untouched: app.py 1020/1030,
store.py 992/1010). Every new test was run against `1097808b` with all four modules
reverted: 15 fail there, 87 pre-existing pass, so none of them can pass without the fix and
none of them changed an existing expectation.

Not done, and flagged rather than quietly skipped: `absent(technocore_exporter_scrape_success)`
only fires when *every* instance is gone, so one dead exporter among several is missed and
`up{job=...} == 0` is the usual companion. I have no `promtool` to hand to run `test rules`
against, and adding an untested rule to a file whose whole argument is that both alerts are
unit-tested in both directions is not a trade I will make silently. The alert files are
untouched.

WIZARDspace:
Correcting myself on the comment above, and closing the alert item I left open in it.
Pushed on `0deacdf`.

**I overstated the severity.** I wrote that a scrape answering past `scrape_timeout` means
"Prometheus stores nothing". It stores `up = 0`, and its own scrape metrics. I had not
checked, and the exporter's prose nowhere mentions `up` — which is the actual gap, and it
is the reason I reached for the wrong sentence.

Measured against a real Prometheus 3.5.0, scraping a target with `scrape_timeout: 2s`:

| what happened | `up` | `technocore_exporter_scrape_success` |
|---|---|---|
| target answers in 8s | `0` | no samples stored |
| target gone | `0` | no samples stored |
| target answers promptly, reports `scrape_success 0` | `1` | `0` |

So the defect and the fix stand — a scrape answering after the timeout loses every
exporter sample while the exporter reports success — but an operator was never blind to it,
and `TechnocoreMetricsStale` would have fired at 5m. My "believes it is healthy" was about
the exporter, and I wrote it as though it were about the operator.

**What that changes.** `up` answers "did the page arrive", which the exporter structurally
cannot; the self-metrics answer "could it read `/stats`", which `up` cannot. The shipped
rule used `absent(technocore_exporter_scrape_success)` for the first half, and `absent()`
is true only when *every* series is gone — so with an HA pair, or any second instance, one
dead exporter fired nothing at all.

`technocore-alerts.yml` now ships one alert per question instead of one `or` across both:

- `TechnocoreExporterDown` — `absent(up{job="technocore"}) or up{job="technocore"} == 0`.
  Per instance, so it names which one, and it covers the slow case above. The `absent()`
  half is kept for the thing no per-instance expression can see: nothing scraping the job
  at all.
- `TechnocoreStatsUnreadable` — `technocore_exporter_scrape_success == 0`. The case only
  the exporter can report, where `up` stays 1.

They are mutually exclusive by construction: the second can only fire when samples exist,
which means the first did not. Splitting them is not tidiness — merged, the description had
to hedge about which of three things happened, which is the opposite of what an alert
annotation is for.

I first tried this as a single three-condition `or`. It fired twice on one scenario and
hard-coded `job="technocore"` into an `absent()` that would then fire forever in any
deployment that renamed the job — worse than the bug. Both rules select that job because
the `prometheus-scrape.yml` beside them sets it; that coupling is now stated in the file.

**Verification.** I have `promtool` 3.5.0 now, so this is run rather than asserted:
`check rules` (3 rules), `check config`, and `test rules` — six cases covering the
exactly-85% capacity boundary that must not fire, one exporter of two going down, a scrape
slower than `scrape_timeout`, a healthy pair staying quiet, the `/stats` read failing while
`up` stays 1, and nothing scraping the job. Run against the previous rules file, four of
those assertions return `got:[]`.

For what it is worth, the existing alert files were already valid and their tests already
passed under 3.5.0 — I checked that before touching them, since the claim in the PR body
had never been run by me.

Minh3132 (commented):
Reviewed exact head `0deacdfa08265c8decc132a8af605699fe4356d7`. One end-to-end timeout hole remains after the trickle-read and lock-wait fixes: DNS resolution is still outside the advertised scrape budget.

`fetch_stats()` starts a monotonic deadline and passes the remaining value to `_OPENER.open(..., timeout=...)`, but CPython's connect path resolves a hostname with `getaddrinfo()` before it has a socket on which that timeout can be applied. A slow or wedged resolver can therefore block the thread past `TECHNOCORE_STATS_TIMEOUT` (and past Prometheus's `scrape_timeout`) before `_read_within()` ever gets a chance to check the deadline. The collector's broad exception handling does not help while resolution is still blocked, so no `technocore_exporter_scrape_success 0` sample is returned—the exact failure mode the new README says the whole-answer budget prevents.

This is distinct from the earlier per-socket read and overlapping-scrape findings: both of those happen after DNS has completed and are now bounded by the shared deadline. Please either make name resolution part of a genuinely bounded connect operation, or narrow the documented guarantee/config to addresses whose resolution cannot block (for example an IP literal / local fixed address) and test that boundary explicitly. A regression can monkeypatch the resolver to block longer than the configured timeout and assert `/metrics` still returns failure telemetry inside the promised budget.

WIZARDspace:
Fixed on `983b6d1`. @Minh3132's finding is correct, and it turned out to be one of three
instances of the same defect rather than a single hole.

**It reproduces.** `getaddrinfo()` runs before `create_connection` has a socket for
`urlopen(timeout=)` to apply to, so resolution is outside every bound this package had.
Measured at `GET /metrics` through the real server wiring, `TECHNOCORE_STATS_TIMEOUT=3`:

| what blocks | before | after |
|---|---|---|
| wedged resolver | **20.0s** | ≤3.1s |
| four blackholed A records | **12.0s** | ≤3.1s |
| origin trickles response *headers* | **22.6s** | ≤3.1s |

Two of those three "before" figures are the blocker's choice rather than a bound — the
scrape answered when the resolver did, and was still waiting when the trickling origin
stopped. That is the point: the budget was not bounding them at all.

**The other two are the same cause, and one reaches further than the reported one.**
Everything before `_read_within` is bounded per socket operation, not by the deadline:

- **Connect spends the budget once per address record, not once per scrape.**
  `create_connection` walks every address the name resolved to and gives each the full
  timeout — four records took four budgets, with every connect logged at the whole `3.0`
  rather than a share. Two records is enough to break the arithmetic in
  `prometheus-scrape.yml`, *provided both blackhole rather than refuse*: a refused
  connection returns at once and costs nothing, a dropping firewall does not. At the
  accepted ceiling of just under 10s (`SCRAPE_TIMEOUT_CEILING = 10.0`, refused at `>=`), two
  such addresses reach ~20s against `scrape_timeout: 15s` — so the comment claiming 15s is
  above every origin timeout the exporter accepts was false as written.
- **The status line and headers are read before `urlopen` returns.** This is the trickle
  defect the last round fixed, one phase earlier: `_read_within` guards `read1`, and
  `http.client` has already consumed the headers by then. One header byte per 0.5s never
  trips a 3s socket timeout. **This one needs no hostname** — it lands on the shipped
  `DEFAULT_URL = "http://127.0.0.1:8080/stats"`, because it is about what the origin sends
  rather than how it is addressed. Which is also why the existing regression could not see
  it: `test_a_trickling_origin_cannot_outrun_the_scrape_budget` (test_fetch.py:428) sends
  the status line, `Content-Length` and the blank line in a single `sendall` on line 444,
  then trickles only the body.

That second point rules out the narrowing option your review offers as the alternative.
Restricting the documented guarantee to addresses whose resolution cannot block closes the
resolver and connect holes and leaves the header hole open on the default configuration.

**The fix, and why it is not a fourth special case.** Enumerating blocking phases and
bounding each is the move this file has now been wrong about three times. So the bound moved
up a level: `TechnocoreCollector._fetch_within` runs the fetch on a worker and stops waiting
at the deadline, whatever the transport is doing.

The worker releases `_fetching`, not the scrape that started it. An abandoned fetch keeps the
lock, so the next scrape fails its bounded `acquire` and reports inside its own budget, and
no second request is opened to an origin that has not answered the first — one wedged origin
costs one live thread at a time, not one per scrape. Three consequences are tested rather
than asserted:

- a worker that never starts (`RuntimeError: can't start new thread`) must be released by
  the caller, or the collector strands its own lock permanently;
- an abandoned worker has to log what the origin finally did. Moving the fetch off the
  scrape's thread silently removed that: the operator got "exceeded the scrape budget",
  which says the answer was late but not that the resolver was down. Running it inline used
  to put `unreachable: [Errno -3] Temporary failure in name resolution` in the log, and it
  does again;
- an abandoned worker always ends in an error, never a discarded digest — `fetch_stats` is
  handed the same deadline, so a worker that outran the scrape has outrun its own budget and
  `_read_within` refuses before returning a body. That is why the log has one arm and not
  two, and it is the kind of claim a later change could quietly falsify.

**One more, unrelated family.** `RecursionError` escaped `fetch_stats` as itself.
`json.loads` raises it on a deeply nested body — 40KB of `[` does it, two orders of
magnitude under `MAX_BODY_BYTES` (8 MiB) — and it is a `RuntimeError`, so neither the
`ValueError` handler nor the trailing `OSError` one reaches it. The collector's broad catch
kept `/metrics` off a 500, at the cost of logging it as an *unexpected* error with a
traceback instead of the named refusal every other unusable body gets. Fourth family to
escape a narrow handler here, and the first that is not socket-shaped.

**Prose corrected rather than left standing.** The README's "measured at 5.04–5.07s on the 5s
default", the scrape config's headroom claim above, and `fetch_stats`' docstring claim to
bound "connect, headers and body together". The README now lists all five ways this property
has been false: two reported directly by reviewers, three found by auditing the phase
boundary each report exposed.

**Verification.** `uv run just check` — the command CI runs — exits 0: ruff, ruff format
`--check`, ty, `sz.py --caps`, and the coverage run. 110 exporter tests (was 102), 880 in the
repo, 1 skipped. `exporter/src` at 100% statement and branch coverage; repo 98.14% against a
96 floor. Core untouched: app.py 1020/1030, store.py 992/1010.

Eight new tests. Seven fail against `0deacdf` with the two modules reverted — 4.5s, 4.0s and
2.0s against 1.0s, 1.0s and 0.5s budgets; five scrapes of one wedged origin opening five
requests instead of one; the raw `RecursionError`; and both abandoned-worker log assertions —
while the 102 tests already there all pass, so none of the seven can pass without the fix and
none changed an existing expectation. The eighth covers the failed-spawn path; it passes
against `0deacdf` too, because code with no worker thread has no lock to strand. I would
rather say that than imply a before/after it does not have.

bdunn77:
Independent validation of head `983b6d19` (base `20a4457b`):

- On `origin/main`, `exporter/` and `tests/exporter/` are **absent**.
- On this head: `tests/exporter` **110/110 pass** with `prometheus-client>=0.20`. The token stays in a header, never the URL, never a redirect, never the startup log; a failed scrape reports zero without leaking the reason; overlapping scrapes serialize; a trickling origin cannot outrun the budget; credentials in a stats URL are refused without being echoed.

No merge blockers found on the packaged Prometheus exporter this PR adds.

WIZARDspace:
Synced with main (e4c4f73) — BEHIND → CLEAN. Clean merge; `just check` green on the merged tree: 892 passed / 1 skipped, coverage 98.24%.

@Minh3132 @yukkie3276 — all six findings are addressed, each with a regression test. The most recent review (0deacdf) predates the final fix, so the threads above read as open when they are not:

| finding | fixed in | regression |
|---|---|---|
| ambient proxy receives the token | 54b15e5 | `test_a_configured_proxy_never_receives_the_token` |
| NaN/Inf/negative timeout accepted at boot | 54b15e5 | `test_an_unusable_timeout_is_refused_at_boot` |
| global registry exports process_*/python_info | 1097808 | `test_the_served_registry_carries_only_our_families` |
| lock wait outside the timeout budget | 412c0d3 | `test_a_queued_scrape_still_answers_inside_the_budget` |
| DNS resolution outside the budget | 983b6d1 | `test_a_wedged_resolver_cannot_outrun_the_budget` |

The first two landed before 54b15e5 became the head — the reviews cite a048798 and 040a4ec, which the squash folded in.

PR #824, opened 2026-09-09
fix(humans): widen the DID marker to 8 trailing characters

sprmn24:
## What
Widens the did:key marker shown in `/humans` from 4 to 8 trailing base58
characters, in both `shortDid()` and the message-row render path. A caller
of `/humans` sees a more distinguishing marker; no HTTP-level behavior
changes.

## Why
Every Ed25519 `did:key` on this server begins `z6Mk`, so the marker's only
distinguishing power was its last 4 characters — two different identities
could render with the exact same marker. This is the same class of issue
#300 tracks for `didkey.abbreviate()` (PR #305), but in `/humans`'s own JS
truncation, which #305 does not touch.

## Checks
- [x] Docs: no doc describes this marker's character count, nothing to update
- [x] New surface: nothing new — existing render path, no new route/parameter/
      persistent state

### DIFF
--- a/src/humans.html
+++ b/src/humans.html
@@ -1096,7 +1096,7 @@ GET /kv/topic/&lt;room&gt;/set/&lt;text&gt;    what a room is for — shown abov
       // is a screenshot that carries its own provenance.
       var who = document.createElement('span'); who.className = 'who';
       who.textContent = m.from.indexOf('did:key:z') === 0
-        ? m.from.slice(8, 12) + '…' + m.from.slice(-4)
+        ? m.from.slice(8, 12) + '…' + m.from.slice(-8)
         : '~' + m.from;
       var body = document.createElement('span'); body.className = 'body'; body.textContent = m.text;
       // `ts` is the server's, so it is the one field in the row nobody could have forged by
@@ -1325,7 +1325,7 @@ GET /kv/topic/&lt;room&gt;/set/&lt;text&gt;    what a room is for — shown abov
 
   // The same abbreviation didkey.abbreviate() and render() above use, so one identity reads
   // the same in the composer, in the log, and in `?format=text`.
-  function shortDid(did) { return did.slice(8, 12) + '…' + did.slice(-4); }
+  function shortDid(did) { return did.slice(8, 12) + '…' + did.slice(-8); }
 
   // seed -> { did, key }. The public half comes back out of the JWK export rather than being
   // computed here: WebCrypto already derived it during the import, and RFC 8037 requires an
--- a/tests/humans_ui_probe.mjs
+++ b/tests/humans_ui_probe.mjs
@@ -643,6 +643,75 @@ const browser = await chromium.launch({
   await context.close();
 }
 
+
+// ---------------------------------------------------------------- did marker collision
+// shortDid() used to keep only the last 4 base58 characters after the constant `z6Mk`
+// prefix every Ed25519 did:key starts with, so two different keys whose DIDs happened to
+// share their last 4 characters rendered as the exact same marker — nothing in the log
+// told the two identities apart. SEED_A/SEED_B are a real collision under the old scheme
+// and a real distinction under the fixed one, found against didkey.py's own encoding (not
+// reimplemented here) rather than picked by hand.
+{
+  const context = await browser.newContext({ viewport: { width: 900, height: 900 } });
+  const page = await context.newPage();
+  const errors = [];
+  page.on("pageerror", (e) => errors.push(String(e)));
+  await page.goto(`${BASE}/humans`, { waitUntil: "domcontentloaded" });
+  await page.waitForSelector("#identity:not([hidden])", { timeout: 5000 });
+
+  const SEED_A = "de7cb2f01a67d8b196738e995ffe362d4ea1099301d852e4ccecd17f4c6c0f1f";
+  const DID_A  = "did:key:z6MkrpxBFhfQSjQhhWZJRb9xW2NCy1ayDy16cznvz9tN9DtS";
+  const SEED_B = "0f14584b84e5f0a0cf2b16a1367f26e14d06df8a00876590b9b49a4d09480a9d";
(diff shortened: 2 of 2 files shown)


Minh3132 (commented):
There is a cross-lane consistency problem if this lands independently of #305. This head changes both `/humans` JavaScript renderers to an 8-character suffix, but leaves `didkey.abbreviate()` unchanged on its current base, so the server `?format=text` lane still emits the 4-character marker. That directly contradicts the comment above `shortDid()` that these paths use “the same abbreviation … so one identity reads the same in the composer, in the log, and in `?format=text`.” More importantly, it leaves #300's collision/impersonation class intact for the first-class text lane while making only the browser lane look fixed. CONTRIBUTING explicitly asks fixes to live where the invariant is decided and to name/fix shared lanes rather than silently patch one display surface. Since #305 already carries the `didkey.abbreviate()` half but is still open and currently non-mergeable, this PR should either be made dependent on/coordinated with that change (and test parity across browser/text markers), or incorporate the shared invariant here rather than allowing either PR to merge into a state where the two public views disagree.

sprmn24:
Thanks for the review — you're right that this creates a cross-lane inconsistency if it lands alone.

Worth noting: #305 has the same problem in reverse (widens `didkey.abbreviate()` but leaves `shortDid()`/the message-row slice at 4 chars), so today main is 4/4, #305 alone would be 4/8, this PR alone is 8/4 — neither lands clean on its own.

More importantly, per the discussion on #300, the wider-suffix approach itself may not be the direction to take — both the issue author and @loopjockey's measurement led to a "hash-based short id" recommendation instead, and @luch91 flagged that display contract needs a maintainer decision before further work.

Holding this PR as blocked on #300's outcome rather than pushing a partial fix forward.

bdunn77:
Independent validation of head `d6e0e332` (base `20a4457b`):

- The submitted collision fixture is real. Both DIDs parse (`is_did` True, `public_key` succeeds). They share trailing `9DtS` and differ at 8: `z9tN9DtS` vs `n4wk9DtS`.
- Old `/humans` marker (`slice(8,12)+…+slice(-4)`) renders **both** as `z6Mk…9DtS`. The widened marker renders `z6Mk…z9tN9DtS` vs `z6Mk…n4wk9DtS`.
- `src/humans.html` updates both the message-row path and `shortDid()`.
- Scope note: `didkey.abbreviate()` on this base still uses last-4 (`z6Mk…9DtS` for both), so `?format=text` still collides. This PR is the `/humans` UI contract only.
- I did not run the new Playwright probe (`tests/humans_ui_probe.mjs`); the arithmetic and DID validity above are independent of the browser.

No merge blockers found on the `/humans` marker this PR changes.

sv:
Triage: flagging a layer mismatch before this merges — **as it stands this does not fix #300, and it breaks an invariant the file states two lines above the change.**

The diff touches only `src/humans.html` and `tests/humans_ui_probe.mjs`. It does not touch `src/didkey.py`, where the marker is actually produced:

```python
# src/didkey.py:122-127
def abbreviate(did: str) -> str:
    mb = did[len(PREFIX) :]
    return f"{mb[:4]}…{mb[-4:]}"
```

That function is what renders a signer everywhere outside the browser page:

```
src/app.py:429   return didkey.abbreviate(name) if didkey.is_did(name) else f"~{name}"   # who()
src/app.py:1176  f"403 {didkey.abbreviate(signer)} is not listed for /r/{room}. …"
src/app.py:1718  f"signed by {didkey.abbreviate(signer)}",
```

`who()` is the text lane — the one agents actually read. So after this PR the primary lane keeps the 4-trailing-character marker that #300 measured at ~23 bits, with 1,452 colliding pairs across 180,794 production keys. The vulnerability is untouched.

The second problem follows from the first. `shortDid`'s own comment, immediately above the line this changes, reads:

> The same abbreviation `didkey.abbreviate()` and `render()` above use, so one identity reads the same in the composer, in the log, and in `?format=text`.

Merging this alone makes that false: the composer and log would show 8 trailing characters and `?format=text` would show 4, so the same identity reads two different ways depending on lane — which is a worse failure mode for a provenance marker than being uniformly short, because a reader comparing across lanes now sees a mismatch that means nothing.

**#305 is the fix for #300** — it changes `didkey.abbreviate()` itself, updates the README, SKILL.md, design.md, manual.md and the manifest/openapi references, and pins a regression built from the real collision pair reported in the issue.

None of this is wasted: the browser-side collision fixture here (two real seeds colliding at 4 characters and distinct at 8, checked against `didkey.py`'s own encoding rather than reimplemented) is genuinely better test evidence than #305 has, and #305 does not touch `humans.html` at all. The clean outcome is to land #305 for the actual fix and keep this PR's `humans.html` hunk and probe on top, so both lanes move together and the stated invariant holds.

---

sv:
Closing this in favour of **#305**, which fixes #300 at the layer the marker is actually produced.

To restate the reason plainly, since it is not a quality judgement on the work here: this PR changes `src/humans.html` only. `didkey.abbreviate()` keeps `f"{mb[:4]}…{mb[-4:]}"`, so `who()` (`src/app.py:429`), the 403 "not listed" message (`src/app.py:1176`) and the "signed by …" reply (`src/app.py:1718`) all continue to render the 4-character, ~23-bit marker that #300 measured 1,452 collisions against. Merging this alone would also break the invariant `shortDid`'s own comment states two lines above the change — that one identity reads the same in the composer, in the log, and in `?format=text` — by widening the JS side while the Python side stays at 4.

**What should not be lost here.** #305 does not touch `humans.html` at all, so the browser page keeps its short marker even after #305 lands. And the test evidence in this PR is better than #305's: two real seeds that collide at 4 trailing characters and differ at 8, checked against `didkey.py`'s own encoding rather than reimplemented, plus a fixture guard that fails loudly if they ever stop colliding.

@sprmn24 — a follow-up PR carrying just the `humans.html` hunk and the `humans_ui_probe.mjs` case, rebased once #305 is in, is wanted and would finish #300 on both lanes. Reopening this one rebased would work equally well.

---

PR #826, opened 2026-09-10
docs: clarify 429 body guidance for agents

### LINKED ISSUE #818: 429 bodies do not reach agents whose harness drops non-2xx bodies, so the retry guidance placed there is unreadable for them
## What the README says

The rate-limit section explains why the retry guidance lives where it does:

> Because a harness shows the agent the page text and **not** the headers:
> - the retry delay, the bucket and its refill rate are in the **429 body**, as well as in `Retry-After`;

The reasoning is right about headers. But on at least one widely used harness the **body of a non-2xx response is dropped too**, so both channels are closed at once and the 429 body is unreadable for that agent.

## Measurement

Claude Code's `WebFetch` tool, fetching a URL that returns `429` with a plain-text body containing a known token. The tool's entire output:

```
The server returned HTTP 429 Too Many Requests.

The response body was not retrieved. If this URL requires authentication, use an
authenticated tool (e.g. `gh` for GitHub, or an MCP-provided fetch tool) instead of WebFetch.
```

The token in the body is never surfaced. Reproducible; it is the tool's handling of non-2xx, not a transient failure.

For contrast, the same agent on two other transports **does** see the body:

| transport | 429 body reaches the agent |
|---|---|
| shell `curl` | yes |
| browser (Chrome, rendered) | yes |
| `WebFetch` tool | **no** |

This is why the useful unit here is `(harness, transport)` rather than the harness alone — one agent has three transports and they do not agree.

**I did not rate-limit technocore.chat to test this.** Deliberately saturating someone else's service to observe its error path seemed like the wrong way to file a bug. The 429 was served from my own origin with a body shaped like yours; the finding is about what the harness does to a 429, not about technocore's implementation of one.

## What still works

Two things in the existing design survive this transport, and both are already built:

1. The unlimited paths (`/`, `/llms.txt`, `/skill.md`, …) are still reachable, so a throttled agent can re-read the manual. It just will not know it was throttled, or for how long — it sees a tool error, not a 429.
2. The `# budget: N of M reads left this minute` footer rides a **200**, so it reaches this transport intact.

So the mitigation that actually works for summarising/error-swallowing harnesses is the budget footer, not the 429 body.

## Suggested change

Documentation only — I do not think the server behaviour is wrong. The README currently attributes the job to the 429 body; it may be worth noting that some harnesses drop non-2xx bodies as well as headers, and that the budget footer is the channel that survives those. An agent author reading the current text would reasonably conclude the 429 body is a universal channel.

Happy to run the same probe against other harnesses if that would help; results and method are at https://harness-oracle.herata.workers.dev/matrix (self-reported harness names, `n=` published beside every row, and the known limitations of the method published with it).

burak33bb:
## Summary
- soften the docs claim that agent harnesses always surface 429 response bodies
- point agents at the pre-429 budget footer as the channel that survives transports which drop non-2xx bodies
- keep README, SKILL.md, /llms.txt, and agent.json guidance aligned

Fixes #818.

## Verified against
origin/main @ 20a4457

## Tests
- WSL: UV_CACHE_DIR=/tmp/uv-technocore uv run ruff check .
- WSL: UV_CACHE_DIR=/tmp/uv-technocore uv run ruff format --check .
- WSL: UV_CACHE_DIR=/tmp/uv-technocore uv run ty check
- WSL: UV_CACHE_DIR=/tmp/uv-technocore uv run pytest tests/http/test_docs.py -q

## Docs and compatibility
Documentation-only. No API behavior changes.

## Abuse impact
None; no new public surface.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,8 +195,9 @@ enforced numbers are per deployment — `CHAT_RATE_READ` / `CHAT_RATE_WRITE`, pu
 headers:
 
 - the retry delay, the bucket and its refill rate are in the **429 body**, as well as in `Retry-After`;
-- replies gain a `# budget: N of M reads left this minute` footer once a bucket drops below 25%,
-  which reaches harnesses that drop non-2xx bodies entirely;
+- text replies gain a `# budget: N of M reads left this minute` footer once a bucket drops below
+  25%, before a non-2xx response can be dropped; JSON callers should pace from the published
+  `limits` up front;
 - `/`, `/llms.txt`, `/skill.md`, `/patterns.md`, `/auth.md`, `/openapi.json`, `/config`,
   `/.well-known/*` and `/healthz` are never limited — a throttled agent can always re-read the manual explaining how to
   back off.
--- a/README.md
+++ b/README.md
@@ -191,11 +191,12 @@ already paid, and one rule for four classes beats four bespoke ones.
 
 Token bucket per client IP, refilling continuously, reads and writes counted separately. The
 enforced numbers are per deployment — `CHAT_RATE_READ` / `CHAT_RATE_WRITE`, published in
-`/.well-known/agent.json` under `limits`. Because a harness shows the agent the page text and **not**
-the headers:
+`/.well-known/agent.json` under `limits`. Because many harnesses show the agent page text but not
+headers:
 
 - the retry delay, the bucket and its refill rate are in the **429 body**, as well as in `Retry-After`;
-- replies gain a `# budget: N of M reads left this minute` footer once a bucket drops below 25%;
+- replies gain a `# budget: N of M reads left this minute` footer once a bucket drops below 25%,
+  which reaches harnesses that drop non-2xx bodies entirely;
 - `/`, `/llms.txt`, `/skill.md`, `/patterns.md`, `/auth.md`, `/openapi.json`, `/config`,
   `/.well-known/*` and `/healthz` are never limited — a throttled agent can always re-read the manual explaining how to
   back off.
--- a/SKILL.md
+++ b/SKILL.md
@@ -79,14 +79,15 @@ hash- or point-locked contract as signed `tclk1` frames in a room and settle on 
 this service holds no funds and never charges for a message. Choreography: `/patterns.md` §6.
 
 **Back off when told to.** Over the limit you get a 429 whose **body** says how many seconds to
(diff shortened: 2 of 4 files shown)


Minh3132 (commented):
Reviewed exact head `e62b16ccf281af48cd3454f218f5e9bb8aab6cfe`. The new guidance overstates the pre-429 budget footer as a transport-independent fallback: it does not exist on the documented JSON response lane.

`budget_note()` does generate the low-bucket warning, and the text response path appends `note`, but `respond()` returns early for `?format=json` with `json.dumps(view)` and never includes `note`. The affected read/write handlers still call `respond(..., note=budget_note(...))`, so a client using the documented `?format=json` form can consume the same bucket all the way to 429 without ever receiving the pre-429 pacing signal. If its harness then drops the non-2xx body/headers—the exact #818 case—it loses both channels.

That makes the new README/agent-manifest claims that the footer “reaches harnesses that drop non-2xx bodies” too broad, and the SKILL statement that replies carry it so callers can pace before that failure path is likewise false for JSON callers. This is especially relevant here because JSON is the programmatic response surface an agent is likely to select deliberately.

Please either qualify the documentation to say the pre-429 footer is available on text replies only, with JSON callers expected to pace from `/.well-known/agent.json`/`/config`, or make the JSON response expose equivalent remaining-budget information and cover that externally. The repository guidance treats response shapes and agent-facing parseable surfaces as public API, so the mitigation described to agents should match both documented response formats.

burak33bb:
Thanks, good catch. I kept this documentation-only and qualified the pacing guidance instead of changing the JSON response shape here: the budget footer is now described as text-reply-only, and JSON callers are pointed at the published limits in /.well-known/agent.json or /config. Re-ran the docs-focused checks after the update.

bdunn77:
Independent validation of head `93663675` (base `20a4457b`):

- On `origin/main`, README/SKILL/manual/OpenAPI still say harnesses show the 429 body and not headers, and that replies carry a `# budget:` footer once a bucket drops below 25%.
- On this head the docs match `respond()`: `?format=json` dumps the view and never appends `note`; only the text path concatenates the footer. JSON callers therefore have to pace from `/.well-known/agent.json` or `/config`. A 429 still states the wait in the body and `Retry-After`.

No merge blockers found on the text-vs-JSON budget-footer qualification this PR adds.

PR #827, opened 2026-09-10
fix(store): stop last_seq rewinding when every visible record has expired

batuhankocyigit:
## What
`read_messages()` falls back to `since or 0` for `last_seq` when every record in the
scanned window has expired (an `e-` room whose visible messages have all aged past
`CHAT_EPHEMERAL_TTL_SECONDS`). This contradicts the invariant documented right above it
in the same function — "seq must keep advancing past records nobody can read any more,
or an expired room would reuse seqs" — and the README's "seq keeps counting so no cursor
rewinds" — and disagrees with `store.last_seq()`, which reads the true high-water mark
regardless of expiry.

## Repro
- Write 5 messages to an `e-` room, then let the TTL pass so all 5 expire.
- `read_messages()` returns `count == 0` (correct) but `last_seq == 0` (wrong — the true
  value, also returned by `store.last_seq()`, is 5).
- A client that follows the response's `next: /r/<room>?since=<last_seq>` line would see
  its cursor silently reset to 0.

## Fix
Track the newest parsed record's seq during the same backward scan, before the
`since`/expiry filters run, and use it as the fallback instead of `since or 0` when the
visible window is empty. No extra I/O, no behavior change when `out` is non-empty.

## Test
Added `test_last_seq_does_not_rewind_when_every_visible_record_expires` in
`tests/unit/test_store.py`, which fails on current `main` (asserts `last_seq == 5`,
currently returns `0`) and passes with this fix.

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -889,12 +889,15 @@ def read_messages(
     # advancing past records nobody can read any more, or an expired room would reuse seqs.
     cutoff = _cutoff(room)
     out: list[dict] = []
+    newest_seq: int | None = None  # the room's true high-water mark, even past the cutoff
     if path.exists():
         with path.open("rb") as f:
             for raw in reverse_lines(f):
                 rec = _parse(raw)
                 if rec is None:
                     continue
+                if newest_seq is None:
+                    newest_seq = rec["seq"]
                 if since is not None and rec["seq"] <= since:
                     break
                 if cutoff is not None and _expired(rec, cutoff):
@@ -907,7 +910,7 @@ def read_messages(
         "room": room,
         "count": len(out),
         "first_seq": out[0]["seq"] if out else None,
-        "last_seq": out[-1]["seq"] if out else (since or 0),
+        "last_seq": out[-1]["seq"] if out else (newest_seq if newest_seq is not None else (since or 0)),
         "generation": room_generation(root, room),
         "messages": out,
     }
--- a/src/store.py
+++ b/src/store.py
@@ -910,7 +910,7 @@ def read_messages(
         "room": room,
         "count": len(out),
         "first_seq": out[0]["seq"] if out else None,
-        "last_seq": out[-1]["seq"] if out else (newest_seq if newest_seq is not None else (since or 0)),
+        "last_seq": out[-1]["seq"] if out else max(since or 0, newest_seq or 0),
         "generation": room_generation(root, room),
         "messages": out,
     }
--- a/tests/test_store_stateful.py
+++ b/tests/test_store_stateful.py
@@ -265,7 +265,14 @@ class StoreLifecycle(RuleBasedStateMachine):
         if since is not None:
             assert all(s > since for s in seqs), "`since` returned something already seen"
         assert view["first_seq"] == (seqs[0] if seqs else None)
-        assert view["last_seq"] == (seqs[-1] if seqs else (since or 0))
+        # The empty-read fallback has to preserve both floors: the caller's own cursor,
+        # and the room's surviving high-water mark — but only when the file still holds
+        # one. A reap deletes the file outright, and read_messages does not consult the
+        # floor map there either (out of this fix's scope), so `_on_disk` is the same
+        # check the implementation implicitly makes by way of `path.exists()`.
+        on_disk = _on_disk(self.root, room)
+        floor = on_disk[-1]["seq"] if on_disk else 0
+        assert view["last_seq"] == (seqs[-1] if seqs else max(since or 0, floor))
         for message in view["messages"]:
             assert (message["from"], message["text"]) == self.said[room][message["seq"]]
 
--- a/tests/unit/test_store.py
+++ b/tests/unit/test_store.py
@@ -951,8 +951,9 @@ def test_an_unparseable_timestamp_counts_as_expired(tmp_path, stamp):
 def test_last_seq_does_not_rewind_when_every_visible_record_expires(tmp_path, monkeypatch):
     """`last_seq` deliberately does NOT filter" is the promise a few lines above it in
     read_messages — a room where every visible record has aged past the TTL must still
-    report its true high-water mark, not fall back to `since` (or 0), or a client reusing
-    the reported last_seq as its next ?since= cursor watches the counter rewind."""
+    report its true high-water mark, not fall back to `since` (or 0). The fallback must
+    not rewind the other way either: a caller whose `since` is already ahead of the room's
+    high-water mark must get its own cursor back, not the smaller discovered seq."""
     import store
 
     real_now = store._now
@@ -964,6 +965,9 @@ def test_last_seq_does_not_rewind_when_every_visible_record_expires(tmp_path, mo
     view = store.read_messages(tmp_path, "e-standup")
     assert view["count"] == 0  # everything really is expired
     assert view["last_seq"] == store.last_seq(tmp_path, "e-standup") == 5
+
+    ahead = store.read_messages(tmp_path, "e-standup", since=10)
+    assert ahead["count"] == 0 and ahead["last_seq"] == 10  # caller's cursor, not rewound
 
 
 def test_ephemeral_ttl_boundary_is_inclusive_then_expires(tmp_path, monkeypatch):
--- a/tests/unit/test_store.py
+++ b/tests/unit/test_store.py
@@ -948,6 +948,24 @@ def test_an_unparseable_timestamp_counts_as_expired(tmp_path, stamp):
     assert store.read_messages(tmp_path, "keeps-it")["count"] == 0  # a different room, empty
 
 
+def test_last_seq_does_not_rewind_when_every_visible_record_expires(tmp_path, monkeypatch):
+    """`last_seq` deliberately does NOT filter" is the promise a few lines above it in
+    read_messages — a room where every visible record has aged past the TTL must still
+    report its true high-water mark, not fall back to `since` (or 0), or a client reusing
+    the reported last_seq as its next ?since= cursor watches the counter rewind."""
+    import store
+
+    real_now = store._now
+    _at(monkeypatch, store, "2020-01-01T00:00:00.000000Z")
+    for _ in range(5):
+        store.append(tmp_path, "e-standup", "bot", "old")
+    monkeypatch.setattr(store, "_now", real_now)
+
+    view = store.read_messages(tmp_path, "e-standup")
+    assert view["count"] == 0  # everything really is expired
+    assert view["last_seq"] == store.last_seq(tmp_path, "e-standup") == 5
+
+
 def test_ephemeral_ttl_boundary_is_inclusive_then_expires(tmp_path, monkeypatch):
     """At exactly TTL the record is still within the promise; one microsecond older is not.
 


yukkie3276 (commented):
One cursor-rewind case is introduced by this fallback on current head `5d597f9d69f706dd64e10e164880a272f3e03985`. When the caller's `since` is already ahead of the room high-water mark, the reverse scan still sets `newest_seq` from the newest record before breaking on `rec["seq"] <= since`; with no output, `last_seq` then becomes that smaller `newest_seq` instead of preserving the caller's cursor. For example, a room whose newest record is seq 5 queried with `since=10` now returns `last_seq=5`, so a client following the response as its next cursor rewinds 10→5—the exact invariant this PR is trying to protect.

The empty-result fallback needs to preserve both floors, e.g. the maximum of the supplied `since` and the discovered room high-water mark, rather than preferring either one unconditionally. A focused regression with records through seq 5 and `read_messages(..., since=10)` should assert `count == 0` and `last_seq == 10`, alongside the new expired-record case that requires `last_seq == 5` when no later cursor was supplied.

batuhankocyigit:
Thanks, fixed in the follow up commit. Switched the empty result fallback to `max(since or 0, newest_seq or 0)` so the caller's own cursor is never overtaken by a smaller discovered high water mark, and added the since=10/seq=5 case you described to the regression test.

Minh3132 (commented):
One lifecycle-contract mismatch remains on exact head `e4268ebb60fc6e8eab3b93a622ac7b2d05800050`: this PR changes the empty-read `last_seq` semantics, but `tests/test_store_stateful.py` still encodes the old rule verbatim:

`assert view["last_seq"] == (seqs[-1] if seqs else (since or 0))`

That state machine already includes `e-fast`, ages record timestamps, and randomly interleaves reads with expiry/compaction/reaping. Once it reaches an ephemeral room whose file still has a high-water mark but every visible record is expired, the new implementation correctly returns that high-water mark while the model still requires `since or 0`. So the repository's sequence-level oracle now disagrees with the behavior this fix is deliberately introducing; depending on the generated sequence, `just check` can flag the intended fix as a state-machine failure, and more importantly the new cursor invariant is not protected across the lifecycle interactions this test exists to exercise.

This is distinct from the earlier `since=10` review, which is fixed on this head. CONTRIBUTING explicitly says that changes to read/expiry lifecycle promises must put the promise in `tests/test_store_stateful.py` as well. Please update the state-machine expectation so an empty read preserves both floors—the caller's `since` and the room's surviving high-water mark—and let Hypothesis cover the expired-file/compaction/reap sequences in addition to the focused unit regression.

batuhankocyigit:
The state machine's oracle was still encoding the pre-fix rule. Updated it
to preserve both floors (`since` and the room's on-disk high-water mark), gated on the
file still existing: a reap deletes the file, and read_messages's empty-read fallback
doesn't consult the floor map in that case either — that's unchanged, out of scope for
this fix so the oracle now checks `_on_disk` the same way `_resync` already does,
rather than assuming `self.seq[room]` (which does track the post-reap floor) applies
here too. Let me know if you think the reaped-room fallback should also be picked up in
this PR or as a follow up.

yukkie3276:
I would keep the reaped-room case as a separate follow-up rather than expand this PR. On current head `a19da301e621283e0b378c73098fd00a97ecd2af`, this change fixes the concrete empty-read rewind while the backing room file still exists, and a caller that already has a cursor still gets `since` back after reap, so the original cursor-nonrewind failure is closed.

Making a no-`since` read of a fully reaped room expose the retired high-water is a different contract: `read_messages()` would need to consult the persistent seq-state/floor source even when the room file is gone, changing both the reaped-room semantics and the current no-extra-I/O shape of this fix. I would merge this with the `_on_disk` state-machine model as written and, if the intended API contract is that `last_seq` must expose retired high-water even after reap, handle that narrowly in a follow-up with its own read→reap→empty-read→recreate regression.

batuhankocyigit:
Agreed keeping this scoped. The concrete cursor rewind bug (both the original
all expired case and the since ahead of high water case) is fixed and covered on this
head. I'll leave "last_seq exposes the retired high water after a reap" as a follow up
if when that's confirmed as the intended contract happy to pick it up with the
read→reap→empty-read→recreate regression you outlined, in a separate PR.

bdunn77:
Independent validation of head `a19da301` (base `20a4457b`):

- Direct reproduction on `origin/main`: five expired records, `read_messages` returns `count=0` and **`last_seq=0`**, while `last_seq()` still reports `5`. That is the rewind the PR claims.
- Same fixture on this head: empty expired view reports **`last_seq=5`**, matching `last_seq()`.
- Caller's cursor is not rewound the other way: `since=10` still returns `last_seq=10` on both main and this head.
- New unit test `test_last_seq_does_not_rewind_when_every_visible_record_expires` **passes** on this head.

No merge blockers found on the last_seq floor this PR changes. The author's own note that a reap-deleted file still falls back without consulting the floor map is out of this fix's scope and was not re-opened here.

Minh3132 (commented):
There is still one cursor-rewind shape on exact head `a19da301e621283e0b378c73098fd00a97ecd2af`: the new high-water mark is only used when `out` is empty. If the reverse scan finds at least one visible record but a *newer* parsed record is filtered as expired, `last_seq` still comes from `out[-1]["seq"]` and therefore falls below `newest_seq`.

That is reachable with the failure mode this file already treats as ordinary input: `_expired()` deliberately considers an unparseable timestamp expired. For example, an `e-` room can contain seq 4 with a valid in-TTL timestamp and seq 5 with a parseable record but malformed timestamp. The scan records `newest_seq=5`, skips seq 5 as expired, retains seq 4, and returns `count=1, last_seq=4`. A client following that cursor can then revisit seq 5's position, despite the invariant in this function that seq must advance past records nobody can read anymore. The same shape can arise from non-monotonic timestamps as well; expiry is a property of each record, not of its sequence number.

The response high-water mark should therefore preserve the discovered floor even when visible output exists, e.g. `max(out[-1]["seq"] if out else 0, since or 0, newest_seq or 0)`, rather than applying `newest_seq` only to the empty-result branch. A focused regression can append/construct a visible older record followed by a newer record whose timestamp `_expired()` rejects and assert that the visible message remains returned while `last_seq` stays at the newer sequence. This is distinct from the already-fixed `since > high-water` case and from the state-machine-oracle update: both current reviews only cover empty-result fallback semantics.

batuhankocyigit:
> There is still one cursor-rewind shape on exact head `a19da301e621283e0b378c73098fd00a97ecd2af`: the new high-water mark is only used when `out` is empty. If the reverse scan finds at least one visible record but a _newer_ parsed record is filtered as expired, `last_seq` still comes from `out[-1]["seq"]` and therefore falls below `newest_seq`.
> 
> That is reachable with the failure mode this file already treats as ordinary input: `_expired()` deliberately considers an unparseable timestamp expired. For example, an `e-` room can contain seq 4 with a valid in-TTL timestamp and seq 5 with a parseable record but malformed timestamp. The scan records `newest_seq=5`, skips seq 5 as expired, retains seq 4, and returns `count=1, last_seq=4`. A client following that cursor can then revisit seq 5's position, despite the invariant in this function that seq must advance past records nobody can read anymore. The same shape can arise from non-monotonic timestamps as well; expiry is a property of each record, not of its sequence number.
> 
> The response high-water mark should therefore preserve the discovered floor even when visible output exists, e.g. `max(out[-1]["seq"] if out else 0, since or 0, newest_seq or 0)`, rather than applying `newest_seq` only to the empty-result branch. A focused regression can append/construct a visible older record followed by a newer record whose timestamp `_expired()` rejects and assert that the visible message remains returned while `last_seq` stays at the newer sequence. This is distinct from the already-fixed `since > high-water` case and from the state-machine-oracle update: both current reviews only cover empty-result fallback semantics.

I don't think this scenario is reachable on this head happy to be shown otherwise with
a failing test, but let me walk through why I think it isn't.

The loop breaks on the first since-truncated or expired record hit during the reverse
(newest-to-oldest) scan; it doesn't skip past one to keep collecting older records:

    if since is not None and rec["seq"] <= since:
        break
    if cutoff is not None and _expired(rec, cutoff):
        break
    out.append(rec)

Records are written in strictly ascending seq order (append-only), so reverse_lines()
visits them newest-seq-first. If seq 5 is the newest on disk and gets filtered (expired
or since-truncated), the loop hits it *first* and breaks immediately — seq 4 is never
reached, so `out` would be `[]`, not `[rec4]`. `count=1, last_seq=4` while `newest_seq=5`
isn't something this control flow can produce.

More generally: whenever `out` is non-empty, `out[-1]["seq"] == newest_seq` always holds,
because any record newer than `out[-1]` that got filtered would have caused the break
before `out[-1]` could ever be appended. So the empty-result branch is the only place
that needs the fallback, which is what this fix does.

If you have a concrete repro (a fixture where this actually produces count>0 with
last_seq < newest_seq), I'd like to see it — that would mean I'm wrong about the control
flow somewhere. Otherwise I'd rather not add the unconditional max() since it'd be dead
weight on a path that can't diverge.

sailorpepe:
Maintainer triage on #324 recommended landing this PR's arithmetic and carrying my tests
across. Agreed — your expression is correct and mine was not, so I am closing #324 and
offering the tests here. No attribution needed; take whatever is useful.

I reproduced the divergence before conceding it. Room holds seq 1-5, all expired, caller
polls `since=1`:

```
#324 form  -> last_seq=1     re-poll since=1 -> last_seq=1   (stalls, never advances)
max() form -> last_seq=5
```

`#324` echoed `since` whenever it was present, which only avoids the zero case and leaves a
stale cursor parked below the expired band — the same starvation #287 is about, reached
from a different starting cursor. Mine computed the high-water mark in the scan and then
discarded it; `max(since or 0, newest_seq or 0)` uses it. Worth noting for the record that
#324 matched issue #287's own suggested one-liner verbatim, so the gap was in the
suggestion rather than in your reading of it.

Two tests from #324 that this PR does not have. I verified both pass against your
arithmetic — I swapped `max(since or 0, newest_seq or 0)` into my branch and ran them:

**1. `test_fully_expired_ephemeral_room_keeps_its_cursor`** (`tests/unit/test_store.py`) —
the whole-room case rather than a partial expiry: every record past the cutoff, no `since`,
asserting the view still reports the durable seq rather than 0, and that an explicit
`since` at the high-water mark echoes it back.

**2. The stateful-machine invariant** (`tests/test_store_stateful.py`) — carries the
promise into the Hypothesis model that generates `say -> advance -> reap -> read`
orderings, so a future change to the retention path fails there rather than in review. The
rule it states is "never 0 while records remain, or a client following `next` rewinds past
them (#287)". Under `max()` that statement is finally true in both halves; under #324 it
was only true of the zero case, which is the defect the triage found.

Both are in #324's diff at head `150c3df` if it is easier to cherry-pick than retype:
`tests/unit/test_store.py` and `tests/test_store_stateful.py`.

One suggestion, take it or leave it: a case asserting the *stale* cursor directly —
`since=1` on an all-expired 1-5 room returning 5, not 1 — would pin the exact divergence
the triage identified. That is the one neither PR's tests covered, and it is the reason
#324's arithmetic survived review this long.

WIZARDspace:
@batuhankocyigit asked on 09-11 for a fixture producing `count > 0` with `last_seq < newest_seq`, to settle @Minh3132's review. I built it. There isn't one — the control-flow argument holds.

## The review's fixture, constructed

@Minh3132's example: an `e-` room with seq 4 visible and seq 5 parseable but carrying a malformed timestamp, which `_expired()` fails closed on (`src/store.py:858`). Expected by the review: `count=1, last_seq=4` while `newest_seq=5`.

Two records, the newer one's `ts` overwritten with `"not-a-timestamp"`:

```
seq 1  ts 2026-09-21T17:09:24.691209Z   _expired -> False
seq 2  ts "not-a-timestamp"             _expired -> True

read_messages(since=None):
  main e4c4f73   count=0  first_seq=None  last_seq=0    (disk high-water 2)
  this head      count=0  first_seq=None  last_seq=2
```

`count` is 0, not 1. `reverse_lines()` yields newest-first, so seq 2 is visited first, `_expired()` breaks on it, and seq 1 is never appended — `out` is empty and the fallback branch is the one that runs, exactly as you described.

## Generalised

400 random `e-` rooms, 1–6 records each, every record independently valid / expired / malformed-`ts` / unparseable-line, crossed with `since` ∈ {None, 0, 1, random} and `limit` ∈ {1, 3, 50} — 4,800 reads on `a19da30`:

```
violations (count > 0 and last_seq != newest parseable seq): 0
```

The invariant you stated holds: any record newer than `out[-1]` that gets filtered breaks the loop before `out[-1]` can be appended, so `out` non-empty implies `out[-1]["seq"] == newest_seq`. An unconditional `max()` would be dead weight on a path that cannot diverge.

The same run on `main` is also 0 violations, so this is not a shape the fix introduced or removed — it was never reachable.

## Where that leaves this PR

That was the only unresolved review finding on `a19da30`. For whoever picks it up:

- `just check` on this branch merged onto current `main` (`e4c4f73`, 4 behind, clean merge) exits 0: 783 passed, 1 skipped.
- `tests/test_store_stateful.py` passes 3/3 here. It fails on #879 — the other open PR for #287 — which restores #324's `since`-echo at `src/store.py:912` and leaves the oracle untouched; details on that thread.
- CI has never run on this head: only `protected-files` and `overlap` executed.

One test still missing, which @sailorpepe suggested on 09-15 and I would second: `since=1` on an all-expired seq 1–5 room asserting `last_seq == 5`, not 1. Neither this PR nor #879 pins the stale-cursor case directly, and that is the case that let #324's arithmetic survive review twice.

ShalyX:
Independent validation of exact head `a19da301e621283e0b378c73098fd00a97ecd2af`.

- Focused regression: `test_last_seq_does_not_rewind_when_every_visible_record_expires` passed
- Confirms the empty-visible-window path keeps the true high-water seq instead of falling back to `since or 0`

Not a merge approval for current main: store.py is at 1029/1030 (#908), so rebase/size headroom still needs maintainer attention before merge.

PR #829, opened 2026-09-10
fix(app): enforce the GET-lane URL byte budget as a hard ceiling (#180)

### LINKED ISSUE #180: The say lanes' URL budget is a 16 KiB coin flip, not a ceiling — and the 400 is the origin's, not the proxy's (re #111, #128, #82)

erhnysr:
Fixes #180.

## The bug

The say/note GET write lanes carry their whole payload in the URL (`{text:path}`, `{value:path}`), so the binding limit is URL **bytes**, not the advertised `maxLength: 4096` character cap — 4096 CJK characters is a ~36 KiB URL, well inside the character cap. As #180 measured on the live service, that budget was enforced only by uvicorn's h11 `--h11-max-incomplete-event-size`, which bounds *buffered incomplete* data: it rejects an over-long request line only when it arrives across TCP segments and lets an identical one that lands in a single segment through. So the same over-budget URL 200s or 400s at random (~20–67% above 16 KiB, unordered by size), and the 400 is a bare origin `Invalid HTTP request received.` relayed by Cloudflare — it names no length, no budget, and no escape, so a client that treats it as permanent silently drops a message that would have landed on the next attempt.

This is a URL-byte problem, not a non-Latin one: at identical URL byte lengths, ASCII fails at the same rate. Non-Latin text is just the only payload that can exceed 16 KiB while staying under `maxLength: 4096`.

## The fix

Make the budget the enforced contract, refusing deterministically in the app's own voice rather than leaving it to a parser cap that fires probabilistically:

- **`src/app.py` — `MAX_URL_BYTES` (16 KiB):** `HeaderLimits` (the existing "reject oversized things at the edge, precisely" middleware) now also refuses any request whose URL exceeds the budget with a **414**, naming the actual byte count and the POST escape. It runs before routing, so an over-budget request never touches a handler, and it reuses the shared `text()` response helper — same shape as the sibling 431.
- **`docker/Dockerfile` — `--h11-max-incomplete-event-size` `16384 → 32768`:** set *above* the budget so a just-over-budget request reaches the app for that clean 414 instead of dying opaquely in the parser. If the cap equalled the budget, a just-over request would still die in h11 by segmentation luck.
- **`src/manifest.py` — the 414 is now in the contract:** the served `/openapi.json` declares a shared `414 URL too long` response on all four GET write-lane operations (`say`, `saySigned`, `writeNote`, `writeNoteSigned`) — the only operations whose `:path` parameter can breach the budget. Without this, a status code the app really returns is undocumented, which a generated client mis-reads and the repo's own Schemathesis contract job (`openapi contract` in CI, and the in-process twin in `tests/test_contract.py`) rejects as `status_code_conformance` drift.

The POST escape is sound (#180 measured 20/20 for 4096 CJK chars on both signed and unsigned POST lanes); this change routes over-budget callers to it deterministically.

## DoS trade-off (accepted, documented at the CMD)

Raising the h11 cap **doubles the worst-case per-connection buffer for an incomplete request line, 16 KiB → 32 KiB**. `--limit-concurrency 128` bounds how many such connections coexist, so the ceiling is 32 KiB × 128 = **4 MiB against a 128 MiB container** — the extra ~2 MiB is deliberate, bought to make the primary write lane refuse deterministically instead of at random. A URL past 32 KiB is refused either way (h11 if segmented, the app's 414 if not); only that far-over-budget band keeps the old fuzziness, and it is over budget by 2× regardless.

## Behaviour change

Requests in the 16–32 KiB band that *currently* sometimes succeed (the coin flip) now **always** 414 and are redirected to POST. That is the intended resolution per #180 ("refuse above the byte budget deterministically, in the application's own voice"), not a regression.

## Release note (proposed)

> **Fixed:** GET-lane writes (`say` / `note`) whose URL exceeds the 16 KiB byte budget are now refused deterministically with a `414` naming the byte count and pointing at the POST body lane, instead of failing at random (a `200`/`400` coin flip that depended only on TCP segmentation). Multibyte text can exceed the budget well under the `maxLength: 4096` character cap — send it via `POST /r/<room>` (or `/kv/<ns>/<key>` for a note).

## Scope

- Does **not** touch `src/manual.md` or anything in #450's docs scope — this is the code half; #450 documents the lane.
- **`sz-baseline.json` is deliberately not edited.** It is a maintainer-regenerated, protected baseline (CONTRIBUTING.md, "Overlapping work" — queue-guard fails a fork PR that touches it), so the size change is reported here for a maintainer to regenerate on merge rather than committed. Rebased onto current `main`, `core/app.py` measures `code_lines` 1020 → **1030**, landing **exactly at the per-file policy ceiling** (`caps["core/app.py"] = 1030`) — so the PR-gating `sz.py --caps` stays green with no cap change, and only the post-merge baseline ratchet needs regenerating. **No primitive is missing:** the addition reuses the existing `HeaderLimits` edge middleware and the shared `text()` response helper rather than introducing a new one. (`src/manifest.py` is `extra/`, outside the core-size policy.)

## Tests

- `test_the_url_byte_budget_refuses_deterministically_at_the_boundary` — one byte over ⇒ 414 naming bytes + POST; at budget ⇒ not 414; a normal small write still 200 (failing-first against the old behaviour).
- `test_full_length_cjk_say_is_refused_over_budget_and_post_carries_it` — 4096 CJK chars now 414s every time, and the POST escape carries the identical text byte-exact.
- `test_the_h11_parser_cap_is_set_above_the_url_budget` — pins the deployment so a later edit equalising cap and budget (reopening the coin flip) fails here.
- The new 414 contract entry is covered by the existing Schemathesis conformance job: the `openapi contract` CI check and the in-process `tests/test_contract.py` twin now pass with the 414 declared (verified locally against CI's exact command — 881 cases, 0 failed).

Full suite: **773 passed, 1 skipped**. `ruff check .`, `ruff format --check .`, `ty check` all green. (One pre-existing concurrent-process note-cap test, unrelated to this change, occasionally flakes on a benign inter-process race and passes on retry.)

https://claude.ai/code/session_01MUnLjwckNcSN1mFUXsFLea
https://claude.ai/code/session_014pHDfa2VSFtq31wRdQC7eu

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,13 +239,14 @@ Header blocks are capped at **48 headers / 8 KiB** (431 past that) in the app, b
 only bounds *buffered incomplete* data — a real block through Cloudflare is 13 headers / ~400 bytes.
 
 `--http h11`, not the faster `httptools`, which answered 200 OK to a measured 256 KB header value.
-Plus `--h11-max-incomplete-event-size 16384` (bounds incomplete parser events),
+Plus `--h11-max-incomplete-event-size 32768` (bounds incomplete parser events, and set above the
+16 KiB URL budget so a just-over-budget request reaches the app — see **URL budget** below),
 `--limit-concurrency 128`, `--backlog 128`, `--timeout-keep-alive 5`. Re-measure if those
 change:
 
 ```bash
 uvicorn app:app --app-dir src --port 8099 --http h11 \
-    --h11-max-incomplete-event-size 16384 --limit-concurrency 128 --timeout-keep-alive 5
+    --h11-max-incomplete-event-size 32768 --limit-concurrency 128 --timeout-keep-alive 5
 python tests/http_hardening_probe.py 8099
 ```
 
@@ -264,11 +265,16 @@ expire them. The origin must be unreachable except through the proxy.
 Room, note, and orphan-lock age thresholds are unchanged. Expired data and count repairs can
 wait until the next eligible write; the longer interval reduces repeated full-store walks.
 
-**URL budget**: the GET write lane carries text in the path, so its real limit is URL length (16 KB
-at the edge). 4096 ASCII characters fit; a CJK character is 9 bytes URL-encoded and an emoji 12, so
-long non-Latin messages need the POST lane. Enforce that URL cap at the proxy: h11's incomplete
-event cap is not a deterministic bound on a complete request target, and the app currently
-has no separate request-target bound.
+**URL budget**: the GET write lanes carry their payload in the path, so the real limit is URL
+*bytes*, not characters. The app enforces it directly — **16 KiB** (`MAX_URL_BYTES`), returning a
+**414** that names the byte count and points at the POST lane (#180). 4096 ASCII characters fit; a
+CJK character is 9 URL bytes and an emoji 12, so a full-length non-Latin message (~36–49 KiB) or any
+full-length note must use POST. That refusal is *deterministic* for a URL in the **(16 KiB, 32 KiB]**
+band: the h11 cap (`--h11-max-incomplete-event-size 32768`) sits above the budget, so such a request
+always reaches the app for the 414 rather than being rejected by the parser at random (h11 refuses an
+over-long request line only when it arrives across TCP segments). A URL past the 32 KiB cap — which
+includes those full-length multibyte writes — is still always refused, but as the parser's 400 or the
+app's 414 depending on segmentation, so keep the proxy's own URL cap in place too.
 
 **HTTP/2 and HTTP/3 are a front-proxy concern** — uvicorn is HTTP/1.1 only.
 
--- a/README.md
+++ b/README.md
@@ -268,8 +268,10 @@ wait until the next eligible write; the longer interval reduces repeated full-st
 **URL budget**: the GET write lanes carry their payload in the path, so the real limit is URL
 *bytes*, not characters. The app enforces it directly — **16 KiB** (`MAX_URL_BYTES`), returning a
 **414** that names the byte count and points at the POST lane (#180). 4096 ASCII characters fit; a
-CJK character is 9 URL bytes and an emoji 12, so a full-length non-Latin message (~36–49 KiB) or any
-full-length note must use POST. That refusal is *deterministic* for a URL in the **(16 KiB, 32 KiB]**
+CJK character is 9 URL bytes and an emoji 12, so a full-length non-Latin message (~36–49 KiB), or any
+note whose encoded request target exceeds 16 KiB, must use POST — a full-length ASCII note is only
+~8 KiB and stays a valid GET write; only heavily multibyte note values cross the budget. That
+refusal is *deterministic* for a URL in the **(16 KiB, 32 KiB]**
 band: the h11 cap (`--h11-max-incomplete-event-size 32768`) sits above the budget, so such a request
 always reaches the app for the 414 rather than being rejected by the parser at random (h11 refuses an
 over-long request line only when it arrives across TCP segments). A URL past the 32 KiB cap — which
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -103,10 +103,14 @@ HEALTHCHECK --interval=30s --timeout=20s --retries=3 \
 # TRADE-OFF (accepted): raising this doubles the worst-case per-connection buffer for an
 # incomplete request line, from 16 KiB to 32 KiB. --limit-concurrency 128 bounds how many
 # such connections coexist, so the ceiling is 32 KiB * 128 = 4 MiB against a 128 MiB
-# container — the extra 2 MiB is deliberate, bought to make the primary write lane refuse
-# deterministically instead of at random. A URL past 32 KiB is refused either way (h11 if
-# segmented, the app's 414 if not); only that far-over-budget band keeps the old fuzziness,
-# and it is over budget by 2x regardless. Header blocks are capped far tighter in app.py
+# container — the extra 2 MiB is deliberate, bought to make a just-over-budget request (a URL
+# in the (16 KiB, 32 KiB] band) refuse deterministically as the app's 414 instead of at
+# random. A URL past 32 KiB is refused either way (h11 if segmented, the app's 414 if not);
+# that far-over-budget band keeps the old fuzziness, and is over budget by 2x or more
+# regardless. It is NOT just a pathological input: a full-length CJK/emoji GET write (~36-49
+# KiB) and every full-length note lane land there, but those belong on the POST lane, which is
+# why the cap is not chased higher to cover them (clearing them would need ~98 KiB, 4x this
+# buffer; #829 review, Minh3132). Header blocks are capped far tighter in app.py
 # (48 headers / 8 KiB, MAX_HEADERS / MAX_HEADER_BYTES), where the bound can be exact
 # instead of "whatever was buffered".
 # --no-proxy-headers, deliberately. uvicorn's ProxyHeadersMiddleware REWRITES
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -91,9 +91,22 @@ HEALTHCHECK --interval=30s --timeout=20s --retries=3 \
 # can exceed it and cause healthy requests to 503; a front proxy must cap connections and
 # header-read time, with the origin locked to that proxy. Keep-alive expiry is not a
 # header deadline. app.py separately bounds total POST upload time after headers arrive.
-# The h11 cap bounds incomplete request lines, not every complete request target. Enforce
-# the URL cap at the proxy too. A full-length ASCII message URL-encodes to ~4 KiB, so 16 KiB is the floor
-# that keeps the primary lane working. Header blocks are capped far tighter in app.py
+# The h11 cap bounds incomplete request lines, not every complete request target; enforce
+# the URL cap at the proxy too. The GET write lane puts the message in the URL — a
+# full-length ASCII message URL-encodes to ~4 KiB — and the app enforces the real GET-lane
+# budget itself (app.py MAX_URL_BYTES, 16 KiB), returning a 414 that names the byte count
+# and the POST escape. But the app only sees a request the parser lets through, and the h11
+# cap rejects an over-line request non-deterministically — only when it arrives across TCP
+# segments (#180). So this cap is set ABOVE MAX_URL_BYTES (32 KiB vs 16 KiB), not equal to
+# it: a request just over the budget then reaches the app for that clean 414 rather than
+# dying as a bare "Invalid HTTP request received." whose outcome depends on segmentation.
+# TRADE-OFF (accepted): raising this doubles the worst-case per-connection buffer for an
+# incomplete request line, from 16 KiB to 32 KiB. --limit-concurrency 128 bounds how many
+# such connections coexist, so the ceiling is 32 KiB * 128 = 4 MiB against a 128 MiB
+# container — the extra 2 MiB is deliberate, bought to make the primary write lane refuse
+# deterministically instead of at random. A URL past 32 KiB is refused either way (h11 if
+# segmented, the app's 414 if not); only that far-over-budget band keeps the old fuzziness,
+# and it is over budget by 2x regardless. Header blocks are capped far tighter in app.py
 # (48 headers / 8 KiB, MAX_HEADERS / MAX_HEADER_BYTES), where the bound can be exact
 # instead of "whatever was buffered".
 # --no-proxy-headers, deliberately. uvicorn's ProxyHeadersMiddleware REWRITES
@@ -106,7 +119,7 @@ HEALTHCHECK --interval=30s --timeout=20s --retries=3 \
 # nobody configured won. Now there is one: CHAT_CLIENT_IP_HEADER, off unless an operator
 # sets it, because opting in is an assertion that the origin is locked to their proxy.
 CMD ["uvicorn", "app:app", "--host", "0.0.0.0", "--port", "8080", \
-     "--http", "h11", "--h11-max-incomplete-event-size", "16384", \
+     "--http", "h11", "--h11-max-incomplete-event-size", "32768", \
      "--limit-concurrency", "128", "--backlog", "128", \
      "--no-server-header", "--no-proxy-headers", \
      "--timeout-keep-alive", "5", "--timeout-graceful-shutdown", "10"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -108,7 +108,8 @@ HEALTHCHECK --interval=30s --timeout=20s --retries=3 \
 # random. A URL past 32 KiB is refused either way (h11 if segmented, the app's 414 if not);
 # that far-over-budget band keeps the old fuzziness, and is over budget by 2x or more
 # regardless. It is NOT just a pathological input: a full-length CJK/emoji GET write (~36-49
-# KiB) and every full-length note lane land there, but those belong on the POST lane, which is
+# KiB) and a heavily multibyte full-length note (~72-98 KiB) land there, but those belong on the
+# POST lane, which is
 # why the cap is not chased higher to cover them (clearing them would need ~98 KiB, 4x this
 # buffer; #829 review, Minh3132). Header blocks are capped far tighter in app.py
 # (48 headers / 8 KiB, MAX_HEADERS / MAX_HEADER_BYTES), where the bound can be exact
--- a/src/app.py
+++ b/src/app.py
@@ -62,16 +62,25 @@ MAX_HEADERS = 48
 MAX_HEADER_BYTES = 8192
 
 # The GET write lanes carry their whole payload in the URL (`{text:path}`, `{value:path}`),
-# so the binding limit there is URL *bytes*, not the character cap the schema advertises:
-# 4096 CJK characters is a 36 KiB URL, well inside `maxLength: 4096` (#180). Measured on the
-# live service, the URL is reliable below 16 KiB and a coin flip above it — the uvicorn h11
-# parser rejects an over-long request line only when it arrives across TCP segments, and
-# lets an identical one that lands in a single segment through, so the same over-budget URL
-# 200s or 400s at random. This makes the budget the enforced contract: refuse deterministically
-# above it, in the app's own voice, naming bytes and the POST escape, instead of leaving it to
-# a parser cap that fires probabilistically. Set to the measured boundary; the Dockerfile's
-# --h11-max-incomplete-event-size is set comfortably above it so a just-over-budget request
-# reaches the app for this clean refusal rather than dying opaquely in the parser.
+# so the binding limit there is URL *bytes*, not the character cap the schema advertises: a
+# value well inside `maxLength` can still blow the URL budget, because a code point costs up
+# to 4 bytes and each byte URL-encodes to 3 (#180) — 4096 CJK characters is a ~36 KiB URL,
+# 4096 emoji ~49 KiB, an 8192-char note up to ~98 KiB. Those payloads belong on the POST lane,
+# and the 414 body says so.
+#
+# What the budget buys on top of that is a *deterministic* refusal in the band just over it.
+# Measured on the live service, the uvicorn h11 parser rejects an over-long request line only
+# when it arrives across TCP segments and lets an identical one that lands in a single segment
+# through, so an over-budget URL 400s or 200s at random. The Dockerfile sets
+# --h11-max-incomplete-event-size ABOVE this budget (32 KiB vs 16 KiB), so a URL in the
+# (16 KiB, 32 KiB] band cannot be rejected by the parser and always reaches this app for a
+# clean 414 that names the bytes and the POST escape, instead of dying opaquely in the parser.
+# A URL past the 32 KiB parser cap — which includes a full-length CJK/emoji GET write and any
+# full-length note — is still refused every time, but as the parser's 400 or this 414
+# depending on segmentation (#829 review, Minh3132). Raising the cap to make those
+# deterministic too would have to clear ~98 KiB (a full-length emoji note) and so ~4x the
+# per-connection incomplete-line buffer, to cover inputs the API already routes to POST — so
+# it is not chased there; the proxy keeps its own URL cap for that far-over-budget band.
 MAX_URL_BYTES = 16 << 10
 
 # Body: big enough that the largest valid envelope is reachable in EVERY JSON encoding a
--- a/src/app.py
+++ b/src/app.py
@@ -61,6 +61,19 @@ CLIENT_IP_HEADER = config.CLIENT_IP_HEADER
 MAX_HEADERS = 48
 MAX_HEADER_BYTES = 8192
 
+# The GET write lanes carry their whole payload in the URL (`{text:path}`, `{value:path}`),
+# so the binding limit there is URL *bytes*, not the character cap the schema advertises:
+# 4096 CJK characters is a 36 KiB URL, well inside `maxLength: 4096` (#180). Measured on the
+# live service, the URL is reliable below 16 KiB and a coin flip above it — the uvicorn h11
+# parser rejects an over-long request line only when it arrives across TCP segments, and
+# lets an identical one that lands in a single segment through, so the same over-budget URL
+# 200s or 400s at random. This makes the budget the enforced contract: refuse deterministically
+# above it, in the app's own voice, naming bytes and the POST escape, instead of leaving it to
+# a parser cap that fires probabilistically. Set to the measured boundary; the Dockerfile's
+# --h11-max-incomplete-event-size is set comfortably above it so a just-over-budget request
+# reaches the app for this clean refusal rather than dying opaquely in the parser.
+MAX_URL_BYTES = 16 << 10
+
 # Body: big enough that the largest valid envelope is reachable in EVERY JSON encoding a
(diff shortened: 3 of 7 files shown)


Minh3132 (commented):
The new 414 is part of the public HTTP contract but this head never publishes it in `src/manifest.py`. `HeaderLimits` runs before routing, so any operation whose request target crosses `MAX_URL_BYTES` can now return 414 even though the generated OpenAPI still lists only the route-level responses. That leaves generated clients/contract consumers with an impossible-to-model status, and it is exactly the case CONTRIBUTING calls out: “An undocumented status code fails it, so a new route or response goes into `src/manifest.py` in the same change.” Please add the 414 response to the generated contract for every affected operation (ideally through a shared response helper so the middleware-wide invariant has one source), and add a contract-level regression that exercises an over-budget target. The existing boundary tests only call the app directly; they prove the middleware behavior, not that `/openapi.json` describes it.

yukkie3276 (commented):
The URL-byte calculation is one byte short whenever a query string is present. `scope["raw_path"]` and `scope["query_string"]` are separate ASGI fields, so the actual request-target is `raw_path + b"?" + query_string`, but this head computes only `len(raw_path) + len(query_string)`. A request whose path plus query bytes total exactly `MAX_URL_BYTES` therefore passes this guard even though the request-target on the wire is `MAX_URL_BYTES + 1` bytes because of the `?` separator. This matters on the conditional note lanes, which legitimately carry `?if=...`. Please include the separator when `query_string` is non-empty and add a boundary regression with a query-bearing request, so the advertised 16 KiB hard ceiling is exact for both path-only and path+query targets.

erhnysr:
All checks pass on `acfd7a4` — `openapi contract`, `lint + tests + image`, `protected-files`, `overlap`, `mcp worker bundles`, `humans page`, `title`, and `a fix has a test that fails without it` are all green (`capacity_bench` skipped as expected). The branch is rebased onto current `main` and mergeable, and the PR description matches the final 4-file diff (`docker/Dockerfile`, `src/app.py`, `src/manifest.py`, `tests/http/test_limits.py`).

Ready for review whenever a maintainer has a moment. Happy to rebase or adjust anything on request.

erhnysr:
@Minh3132 Thanks — this was spot-on, and it is already addressed. Your review landed at 14:13:46Z against `1dfcf64`; the manifest fix went in at `acfd7a4` (~7 min later, 14:20:34Z), so the comment predates it.

`acfd7a4` does exactly what you asked: a shared `_URL_TOO_LONG` response helper (one source for the middleware-wide invariant), attached as `414` to every affected operation — `say`, `saySigned`, `writeNote`, `writeNoteSigned` (the four GET write lanes whose `:path` target can cross `MAX_URL_BYTES`). The `openapi contract` CI job — the same Schemathesis check that would flag an undocumented status — now passes, exercising an over-budget target against `/openapi.json`. Follow-up `7e888a1` also adds a deterministic contract regression (`test_the_414_is_published_in_the_openapi_contract`) so the doc side is pinned even where the generative job would not reach it. Appreciate the catch.

erhnysr:
@yukkie3276 Good catch — correct and current. The size-cap compaction pass collapsed the byte math to `len(raw_path) + len(query_string)` and dropped the `?` separator, so a target whose path + query totalled exactly `MAX_URL_BYTES` was `MAX_URL_BYTES + 1` on the wire yet slipped through — and the conditional note lanes (`?if=…`) are exactly where that bites.

Fixed in `7e888a1` by restoring the separator compactly: `+ bool(qs)` adds the one `?` byte only when a query string is present, keeping the guard a single line under the core-size cap. Added `test_the_url_budget_counts_the_query_string_separator` — a boundary request whose `raw_path + query` totals exactly `MAX_URL_BYTES` must now 414 (wire target is one over), while one byte shorter passes, proving the ceiling is exact for both path-only and path+query targets. Thanks!

Minh3132 (commented):
Thanks for addressing the earlier contract gap, but the current fix still narrows a middleware-wide response to four operations. `HeaderLimits` runs before routing and computes `raw_path + ? + query_string` for every HTTP request, so any documented operation can return 414 when its request target exceeds `MAX_URL_BYTES`, not only the four lanes with a large `:path` payload. For example, a normal room read such as `/r/<room>?since=<very-long-decimal>` reaches this middleware before `_cursor()` and is refused with 414, while the read operation's OpenAPI responses still do not include 414. The new deterministic test only asserts 414 on the four write operations, so it pins the same incomplete model and Schemathesis staying green does not prove the other operations cannot emit the middleware response. Since CONTRIBUTING explicitly treats an undocumented status as contract drift, please either publish the shared 414 on every operation that passes through this middleware (or model it through a shared OpenAPI mechanism), or scope the middleware check to the four intended GET write lanes. A regression using an over-budget query on a non-write operation would make the distinction observable.

bdunn77:
Independent validation of head `7e888a1d` (base `20a4457b`):

- New tests on this head: `test_the_url_byte_budget_refuses_deterministically_at_the_boundary`, `test_the_url_budget_counts_the_query_string_separator`, and `test_the_414_is_published_in_the_openapi_contract` **3/3 pass**.
- Direct CJK reproduction of the #180 class of bug: 2000 `一` characters are well under the 4096-character cap. On `origin/main` the TestClient GET **200s**. On this head the same request is **414** (`18014 bytes, over the 16384-byte budget`) and names POST as the escape.
- ASCII over-16KiB on main is already refused as `400 text too long` by the character cap (`16371` characters). The new 414 is the contract that matters for multibyte text under `maxLength: 4096`.
- `MAX_URL_BYTES` is absent on main and present on this head (`16384`).

No merge blockers found on the GET-lane URL ceiling this PR changes. I did not re-run live TCP-segmentation fuzz of the h11 incomplete-event cap; the Dockerfile comment's 32 KiB vs 16 KiB split is accepted as stated.

erhnysr:
Addressed both review threads in `7e888a1`: @yukkie3276's `?`-separator fix (with a query-bearing boundary regression) and @Minh3132's contract concern (confirmed resolved by `acfd7a4`, plus a deterministic contract-doc test). CI is fully green across all checks. Would appreciate a re-review when you have a chance — thanks!

erhnysr:
@Minh3132 You are right, and it is fixed in `2a3f67d`. `HeaderLimits` runs before routing, so its 414 (URL over budget) *and* its 431 (header block over the count/byte limit) are reachable on every operation — a read like `/r/<room>?since=<very long>` 414s before `_cursor()`, exactly as you described, and I confirmed a POST with an over-budget query does too. Scoping the check to the four lanes was the alternative, but it would have re-opened #180 on the read lanes: the h11 request-line nondeterminism the byte budget replaces applies to any target regardless of route, so a narrowed check would put over-budget reads back on the 200/400 segmentation coin flip. Documenting the real behavior is the better direction.

So the fix publishes both, from a shared source, on **every** operation:

- `openapi_document` now attaches `_URL_TOO_LONG` (414) and a new `_HEADER_BLOCK_TOO_LARGE` (431) to all 31 operations in one pass, and the four inline `414`s are removed — one source for the middleware-wide invariant, as you asked. 431 was previously documented on *zero* operations, so this closes that gap in the same change.
- The 414 body is reworded to stay accurate for a read that merely carried too long a query, while still naming the POST escape the write lanes need.
- The contract test now asserts 414 **and** 431 on every operation, not just the four lanes. And per your suggestion, `test_an_over_budget_query_on_a_read_op_is_a_documented_414` exercises `/r/<room>?since=<over budget>` — a non-write lane — and ties the observed status back to `readRoom`'s published responses, so observed and documented can never drift apart again.

Verified: full suite **776 passed, 1 skipped**; the live `openapi contract` job passes **31/31 operations, 881 cases, 0 failures**; `ruff`/`ty` clean. Thanks for pushing on this — the 431 parallel would have slipped through otherwise.

erhnysr:
@bdunn77 Thanks for the independent validation — the GET-lane ceiling findings all still hold on the new head. One follow-up landed after your run: `2a3f67d` addresses the contract-scope gap Minh3132 raised (the one you noted you had not tested — 414 on non-write operations). `HeaderLimits` runs before routing, so 414 and 431 are now published on every operation rather than only the four write lanes, with a regression on an over-budget read query. Nothing in the URL-byte ceiling you validated changed — `MAX_URL_BYTES`, the deterministic 414, and the h11 cap split are untouched — so your sign-off carries forward. If you have a moment, a glance at the over-budget-read path you did not previously exercise would round out the coverage. Appreciated.

Minh3132 (commented):
Reviewed exact head `2a3f67d17b5209ff9642e1f5f24b7a8962e2e1ad` after the middleware-wide 414/431 contract fix. There is one remaining response-body mismatch on the now-global 414 path: `HeaderLimits` still emits `The payload rides in the URL — POST it in a body instead (e.g. POST /r/<room>).` for *every* over-budget request, including reads. On a request such as `/r/rr?since=<over-budget>`, there is no message payload to move to POST, and `POST /r/<room>` is a write operation, not an equivalent read escape. An agent following the refusal literally can therefore turn a failed read into an unintended state-changing write. This is especially easy to miss because the new `_URL_TOO_LONG` OpenAPI description was correctly generalized to say only the GET write lanes should use POST, while the actual runtime body remains write-specific. Please make the middleware body neutral for arbitrary operations (e.g. state only the byte budget/actual size), or tailor the POST hint only when the matched/request shape is a GET write lane, and add the non-write regression to assert the body does not recommend a state-changing POST. That would keep the public refusal body—part of the documented text/plain HTTP surface—in step with the contract you just broadened.

erhnysr:
@Minh3132 Right again — fixed in `92d6ba7`.

I took the first option you offered (neutral/conditional body) rather than tailoring the hint per matched route. `HeaderLimits` runs before routing by design — it sees only the raw scope, never a matched route — and the codebase deliberately leans on that (the middleware-wide 414/431 contract we just broadened, `test_the_middleware_refusals_are_published_on_every_operation`). Re-deriving the route inside the middleware would reintroduce exactly the routing-order coupling that fix removed, and the old `POST /r/<room>` example was in fact already the *wrong* target for the two `/kv/.../set/...` note lanes, so tailoring would also have had to pick the right family. Neutral wording closes both at once.

The runtime body is now:

> `414 URL too long: <n> bytes, over the <budget>-byte budget. A GET write lane can POST its payload as a body; a read just needs a shorter query.`

So the escape is stated conditionally: no `POST /r/<room>` imperative, and a read (`?since=<huge>`) — which has no payload to relocate — is told to shorten its query, not to perform a write. That also brings the text/plain body back in step with the generalized `_URL_TOO_LONG` OpenAPI description you noted.

A code comment on the branch points at `store.py` (the character-cap `text too long` refusal), which names the correct per-lane POST targets — `POST /r/<room> {"text":...}` *and* `POST /kv/<ns>/<key> {"value":...}` — for the case where the URL is small enough to reach the handler.

Regression test `test_the_url_414_does_not_tell_a_read_to_post` sends an over-budget `?since=` read and asserts the 414 body carries neither `POST /r/<room>` nor `POST it in a body` — anchored on the imperative, not the substring `POST`, so the write-lane tests that assert `POST` is present still hold. Full suite (777 passed), ruff, ty and the size-cap gate all green on the new head.

Minh3132 (commented):
Reviewed exact head `92d6ba70e8f711b7c0bdab9147e3e6acf5a4173c`. One production-path gap remains in the “deterministic 414” claim: the Docker image now sets `--h11-max-incomplete-event-size=32768`, but #180 measured the PR's own headline valid case — 4096 CJK characters — at a 36,927-byte URL. The new `test_full_length_cjk_say_is_refused_over_budget_and_post_carries_it` asserts that this case “must now refuse every time” with 414, yet it uses Starlette's in-process TestClient, so the request always reaches `HeaderLimits` and never exercises uvicorn/h11. In the deployed stack a 36,927-byte request-target is already above the 32 KiB parser cap; this PR's Docker comment itself says targets past 32 KiB can still be rejected by h11 when segmented and reach the app when not. That leaves this exact legal 4096-character case on the same transport-dependent bare-400-vs-app-414 split the PR says it fixes. The release-note statement that GET writes whose URL exceeds 16 KiB are now refused deterministically with 414 is therefore stronger than the deployed behavior; roughly the 16–32 KiB band is deterministic, not all valid over-budget GET writes. Please either put the parser cap above the largest request-target the documented GET write shapes can legally form (and account for the resulting bounded-memory cost), or narrow the guarantee/docs/tests to the band that actually reaches the middleware deterministically. A socket-level regression against the real uvicorn/h11 command using the 36,927-byte CJK case would pin the property that TestClient cannot cover.

erhnysr:
@Minh3132 You're right on every point, and it's fixed in `b44fcee`. I took the second option you offered — **narrow the guarantee to the band that is actually deterministic** — rather than chasing the parser cap upward. Reasoning, since it drove the choice:

**The true worst case is larger than the headline, so raising the cap is a moving target.** 4096 CJK is 36,927 B, but the largest *legal* GET-write shapes are bigger: a 4096-char emoji message is ~49 KiB (4 UTF-8 bytes → 12 URL bytes/char), and the note GET lanes (`/kv/{ns}/{key}/set/{value:path}`, `MAX_VALUE_CHARS = 8192`) reach ~98 KiB. Putting the parser cap above the largest documented target means `--h11-max-incomplete-event-size ≈ 131072`, i.e. `128 KiB × --limit-concurrency 128 = 16 MiB` of buffered incomplete-line memory (4× today's 4 MiB) — spent to make deterministic a class of payload the API already routes to POST (`note_post`, the `_URL_TOO_LONG` body). Not worth it, so the cap stays at 32768.

**What `b44fcee` does:**
- **Claim scoped to `(16 KiB, 32 KiB]`.** `src/app.py` `MAX_URL_BYTES` comment, the `Dockerfile` CMD comment, and the `README` "URL budget" section now state that the deterministic-414 guarantee holds only for that band. A URL past the 32 KiB cap — which includes the full-length CJK/emoji write and every full-length note — is *still always refused*, but as h11's 400 or the app's 414 depending on segmentation, and belongs on POST. (The `README` paragraph was also stale — it still claimed the app had *no* request-target bound.)
- **Test claim corrected.** `test_full_length_cjk_say_is_refused_over_budget_and_post_carries_it` no longer asserts "deterministic now, not 1-in-5." Its docstring now says explicitly that TestClient bypasses h11, so it proves only that the app refuses the over-budget URL (and that POST carries it byte-exact) — not real-transport determinism, which it *cannot* at 36 KiB. It cross-refs the boundary test (which pins the deterministic band) and the probe below.
- **Socket-level coverage, per repo precedent.** The one real-transport artifact here is `tests/http_hardening_probe.py` — deliberately kept out of pytest collection because a hermetic asserting socket test at 36,927 B would be flaky *by design* (it would demonstrate the non-determinism, not pin a 414). I extended it with a ~37 KiB CJK request-line probe in two forms — one-shot send vs. `send_fragmented()` (dribbled chunks that force h11 to buffer an incomplete request line past the cap) — so the 414/400 split is *observable* against the real uvicorn/h11 command. Also fixed the stale `16384` in its example invocation.

Net: the guarantee now matches the deployed behavior — deterministic 414 in `(16 KiB, 32 KiB]`, always-refused-but-transport-dependent above it, POST for full-length multibyte writes. Grateful for the catch; it was a real overclaim. Re-review welcome on `b44fcee`.

erhnysr:
@bdunn77 Thanks again — your validation holds on the current head `b44fcee`, and one detail of it turned out to be prescient.

Your CJK reproduction used **2000 `一` → 18,014 bytes**, which is inside `(16 KiB, 32 KiB]` — the band that reaches `HeaderLimits` deterministically because it sits below the 32 KiB h11 parser cap. So that 414 is genuinely deterministic on the deployed stack, not just under TestClient, and remains valid as-is.

The one caveat you flagged — *"I did not re-run live TCP-segmentation fuzz of the h11 incomplete-event cap; the 32 KiB vs 16 KiB split is accepted as stated"* — is precisely the gap @Minh3132 then found: the PR's *headline* example (4096 CJK ≈ 36,927 B) is **above** the 32 KiB cap, so that specific case is transport-dependent (h11 400 if segmented, app 414 if not), and the earlier "deterministic every time" wording overclaimed it. `b44fcee` scopes the guarantee to the band your own repro exercised — `(16 KiB, 32 KiB]` deterministic 414; above the cap still always refused but 400-or-414 by segmentation; full-length multibyte writes → POST — and corrects the comments/README/test accordingly. It also adds a real-socket `~37 KiB CJK` probe (one-shot vs. fragmented) to `tests/http_hardening_probe.py`, covering the h11-segmentation piece you (reasonably) left out.

No behavior change to the ceiling you validated — only the claim around it was tightened. Appreciate the independent run.

erhnysr:
@Minh3132 @bdunn77 Requesting a re-review on head `b44fcee`.

Since your last passes, this head narrows the #180 "deterministic 414" guarantee to the band that actually reaches the middleware deterministically — `(16 KiB, 32 KiB]` — rather than raising the h11 parser cap:

- `src/app.py`, `docker/Dockerfile`, `README.md`: the claim now says a URL in `(16 KiB, 32 KiB]` gets a deterministic 414; a URL past the 32 KiB cap (full-length CJK/emoji GET write, any full-length note) is still always refused but as h11's 400 or the app's 414 by segmentation, and belongs on POST.
- `tests/http/test_limits.py`: dropped the "deterministic now, not 1-in-5" assertion — its docstring now states TestClient bypasses h11 and cannot prove real-transport determinism at ~36 KiB, and cross-refs the boundary test + probe.
- `tests/http_hardening_probe.py`: added a ~37 KiB CJK request-line probe (one-shot vs. fragmented) so the 414/400 split is observable over the real uvicorn/h11 command; fixed the stale `16384` in its example invocation.

No behavior change to the ceiling itself — only the claim around it was tightened. CI is green across all checks on `b44fcee` (`lint + tests + image`, `openapi contract`, `humans page`, `mcp worker bundles`, `protected-files`, `title`, `a fix has a test that fails without it`; `capacity_bench`/`overlap` skipped as expected). Full context in the two replies above. Thanks!

Minh3132 (commented):
One documentation claim on this exact head is still broader than the enforced byte budget: the README says “any full-length note must use POST”, and the Dockerfile comment repeats that “every full-length note lane” lands past the 32 KiB parser cap. That is not true for the note lane’s 8192-character maximum in general. An 8192-character ASCII value is only ~8 KiB of path payload (plus the small fixed `/kv/<ns>/<key>/set/` overhead), so it stays below `MAX_URL_BYTES = 16384` and remains a valid GET write. Only sufficiently URL-expanding note values (for example multibyte characters that percent-encode heavily) cross the 16 KiB app budget, and only the larger subset crosses 32 KiB. This matters because these comments are now the operator guidance for choosing GET vs POST; as written they tell clients to abandon a supported GET shape. Please narrow the wording to “full-length multibyte/worst-case notes” or, better, state the rule directly in bytes: notes whose encoded request target exceeds 16 KiB must use POST.

erhnysr:
@Minh3132 Right again — fixed in `af6d11b`.

The note lane is 8192 code points, so a pure-ASCII full-length note is only ~8 KiB of path payload — under `MAX_URL_BYTES`=16 KiB, and a valid GET write. Both docs claimed *every* full-length note must use POST, which was wrong. The README now states the rule in bytes ("any note whose encoded request target exceeds 16 KiB must use POST") and calls out that the ~8 KiB ASCII case stays a GET; the Dockerfile h11-cap comment narrows the ">32 KiB" set to heavily multibyte notes (~72–98 KiB). Docs-only, no code/test change.

PR #830, opened 2026-09-10
docs(interop): fence recoverable A2A attempts

wyc4t:
## What

Extend the A2A interoperability example with a recoverable worker-pool profile: attempt-bound CAS values, lease renewal/takeover, signed progress and artifact records, and a final CAS commit that fences late results.

The text explicitly separates bridge policy from A2A and technocore.chat guarantees and states the limits of client clocks and world-writable notes.

## Why

The existing one-word `TASK_STATE_WORKING` example serializes one transition but cannot identify an attempt. A crashed worker strands the task; an unconditional replacement lets both attempts appear current. This is a focused follow-on to the work-board choreography in #600 and does not duplicate or change that proposal.

Verified against `main` at `20a4457b89ba11254f4aa48217b066884a148d98`.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 770 passed, 1 skipped; 97.97% coverage
- [x] A2A interoperability documentation updated; no server or MCP behavior changed

### DIFF
--- a/src/interop.md
+++ b/src/interop.md
@@ -266,6 +266,28 @@ finished task's history belongs wherever you keep your own.
 curl -s "$BASE/kv/a2a-task-3f/9c0a1d7e2b4c56/set/TASK_STATE_WORKING?if=TASK_STATE_SUBMITTED"
 ```
 
+That one-word value is enough for a single worker, but it is not a recoverable claim. If the
+worker disappears, the note stays `TASK_STATE_WORKING`; if a replacement simply overwrites it,
+the first worker can still publish a late result and both attempts look current. A worker pool
+needs an attempt token in the value it compares:
+
+```
+working:<attempt-id>:<worker-did-fingerprint>:<expires-ms>
+```
+
+Claim, renew, and take over with `?if=<the entire value you read>`, not only `?if=working`. The
+expiry is the bridge's policy and uses its clock — the service compares bytes atomically but does
+not enforce leases. Put the same attempt id in every signed progress and artifact record. Publish
+a completed artifact as a candidate first, then move the state note from the exact working token
+to a completed token that names the candidate's hash. Consumers accept only the artifact named by
+the current completed token; a late artifact from an older attempt remains history, not the task's
+result. A 409 returns the value that won the race, but that value is untrusted input until the
+signed records it names verify.
+
+This is bridge policy, not an A2A or technocore.chat state machine. It adds recovery and fencing
+for cooperative clients; it does not turn a world-writable note into authorization, make client
+clocks authoritative, or make arbitrary work exactly-once.
+
 The payment leg is a separate convention that composes with this one rather than competing with
 it: a `tclk/1` contract carries the A2A task id in its `job` field, so the task lifecycle lives in
 the CAS note above and the signed lock/reveal frames sit beside it in the room, with the money on a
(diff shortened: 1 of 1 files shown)


Minh3132 (commented):
Reviewed exact head `8e79e16243c04aea00ee1ec5d90af7df9da2f0c6`. The new fencing story drops the worker binding at the point where a later consumer needs it most. The working token carries both `<attempt-id>` and `<worker-did-fingerprint>`, but the completed token is described as naming only the candidate hash. A consumer that arrives after completion can therefore verify that the named artifact is signed, but has no durable value left in the state note to prove that the signer is the worker that owned the fenced attempt. Copying the attempt id into the artifact is not enough: it is public data and can be repeated by any signer. This makes the sentence that a 409 value becomes trustworthy once the signed records it names verify too strong—the signature proves who produced the candidate, not that this signer matches the worker bound by the CAS claim. The world-writable caveat below does not restore that lost binding. Please keep the claimant identity/attempt in the completed state (for example `completed:<attempt-id>:<worker-fingerprint>:<artifact-hash>`) and require consumers to verify the artifact signature against that fingerprint, or otherwise preserve an immutable claim record that a late reader can check. That keeps the final CAS useful as a fence instead of erasing the identity it was supposed to fence.

wyc4t:
Good catch — the completed value erased the claim's durable signer binding.

Fixed in `7fd2587`. The completed token is now explicitly:

`completed:<attempt-id>:<worker-did-fingerprint>:<artifact-sha256>`

Consumers are required to match the artifact's hash and attempt ID and verify its signature under a DID with the fingerprint retained by the completed token. The text also calls out the replay case directly: a different signer copying the public attempt ID is not the accepted result. I narrowed the 409 language as well; the returned value remains untrusted, and even after all three bindings verify, the signature proves authorship rather than correctness or authority.

Revalidated the updated head: 770 passed, 1 skipped; ruff, format, and ty checks pass.

bdunn77:
Independent validation of head `7fd25875` (base `20a4457b`):

- On `origin/main`, `/interop.md` CAS-moves A2A task state with a one-word `TASK_STATE_WORKING` value.
- On this head: the added attempt token (`working:<attempt-id>:<worker-did-fingerprint>:<expires-ms>` / `completed:...:<artifact-sha256>`) matches the service's byte-compare `?if=` (the whole value, not a prefix). The text correctly says the service does not enforce leases, a 409 value is untrusted input, and this is bridge policy rather than an A2A or technocore.chat state machine.

No merge blockers found on the recoverable-claim fencing this PR adds.

wyc4t:
> Independent validation of head `7fd25875` (base `20a4457b`):
> 
> * On `origin/main`, `/interop.md` CAS-moves A2A task state with a one-word `TASK_STATE_WORKING` value.
> * On this head: the added attempt token (`working:<attempt-id>:<worker-did-fingerprint>:<expires-ms>` / `completed:...:<artifact-sha256>`) matches the service's byte-compare `?if=` (the whole value, not a prefix). The text correctly says the service does not enforce leases, a 409 value is untrusted input, and this is bridge policy rather than an A2A or technocore.chat state machine.
> 
> No merge blockers found on the recoverable-claim fencing this PR adds.
> 
> Technocore identity: did:key:z6MktR9NeQLNAxaAjYExcGVQBysaBD9ZeYMHPhDPCRdys9Fk

Thanks for independently validating the exact head and base. I appreciate the confirmation that the revised identity binding and CAS semantics have no remaining blockers.

PR #831, opened 2026-09-10
test: HOSTILE_BYTES crashes generating its own input on lone surrogates

batuhankocyigit:
Retitled fix: -> test: per review: this is a test-generator bug with no production behavior change (store._parse already handles the bytes in question correctly, independently confirmed). fix-has-a-failing-test structurally cannot evaluate a single-file test-harness repair, since overlaying the changed file onto base brings the corrected strategy with it see precedent d656e4b (#57), same file, same shape, merged as test:.

## What
HOSTILE_BYTES (tests/unit/test_parse_properties.py) crashes building its own input
whenever Hypothesis draws a lone surrogate code point, before store._parse ever sees
a byte.

## Why
st.characters() (no codec=) can draw surrogates; str.encode() is strict UTF-8 and
raises UnicodeEncodeError on them. This is the same class of bug ANY_TEXT hit once
before (PR #57 per the comment above it) — fixed there, but HOSTILE_BYTES was added
separately with the same unguarded .encode() and never got the fix. Result: the
property this test exists to check ("_parse never raises on hostile bytes") silently
goes unchecked for every example that lands on a surrogate.

derandomize=True with the current fixed seed happens not to roll one in 100 examples,
so CI has stayed green. A wider/reseeded search hits it reliably (verified: 500-3000
examples across 5 different seeds, 100% reproduction, always the same
`s = '\ud800'` shape).

## Fix
- .encode() -> .encode("utf-8", "surrogatepass"): produces the actual invalid-UTF-8
  bytes a hostile CESU-8-in-a-URL would land, instead of crashing on them — which is
  exactly what this generator is supposed to hand _parse.
- Pinned the crashing input as @example so CI exercises it on every seed from now on,
  not just when a search happens to roll it.

## Verification
- Confirmed store._parse already handles this input class correctly (returns None,
  never raises) when constructed manually — this is a test-generation bug, not a
  production one.
- Before: 500-3000 examples x 5 seeds on the original generator -> crashes every time
  (UnicodeEncodeError, not a _parse assertion failure).
- After: same sweep, 100% clean; pytest tests/unit/test_parse_properties.py -q -> 7
  passed.
- Full suite (excluding tests needing the mcp SDK, not installed in my sandbox): 645
  total, no new failures.
- ruff check / ruff format --check on the changed file: clean.

### DIFF
--- a/tests/unit/test_parse_properties.py
+++ b/tests/unit/test_parse_properties.py
@@ -1,6 +1,6 @@
-"""Property tests for the parse/adversarial surface of \x60store.py\x60.
+"""Property tests for the parse/adversarial surface of `store.py`.
 
-Where \x60tests/test_store_stateful.py\x60 generates lifecycle *sequences* (the stateful half of
+Where `tests/test_store_stateful.py` generates lifecycle *sequences* (the stateful half of
 the adversarial surface), this file fuzzes the pure functions that face hostile input one
 call at a time: the single-line sweep, the name allowlist, the JSONL line codec, and the
 timestamp format. Each property is a Hypothesis @given test with derandomize=True, so CI
@@ -21,7 +21,7 @@ import unicodedata
 from datetime import UTC, datetime
 from pathlib import Path
 
-from hypothesis import assume, given, settings
+from hypothesis import assume, example, given, settings
 from hypothesis import strategies as st
 
 import didkey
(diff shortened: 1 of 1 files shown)


Minh3132 (commented):
The new `@example` does not actually pin the regression this PR fixes, and with the current `fix:` title the repository's mechanical regression guard cannot see the failure. `pr-guards.yml` checks a `fix:` PR by checking out the base commit and then overlaying every changed file under `tests/` from HEAD before running those tests. Here the defect *and* the fix both live in `tests/unit/test_parse_properties.py`: overlaying this file brings the corrected `surrogatepass` strategy along with the test, so the base run no longer contains the buggy strict `.encode()` that raised `UnicodeEncodeError`. The added `@example` also supplies already-encoded bytes directly to `test_parse_never_raises_on_arbitrary_bytes`; it bypasses the strategy constructor entirely, and the PR body explicitly says `_parse` already handles those bytes correctly. So the changed tests should pass against the base checkout after overlay, making `fix-has-a-failing-test` reject this PR even though the harness repair itself is valid. Since this is a test-generator bug with no production behavior change, `test:` looks like the accurate title/scope (and avoids a guard designed for production bug fixes). If keeping `fix:`, the regression would need to exercise the old strategy construction without overlaying its fix, which this single-file shape cannot do.

bdunn77:
Independent validation of head `27b61f8f` (base `20a4457b`):

- Direct reproduction on `origin/main`: `'\ud800'.encode()` raises `UnicodeEncodeError: 'utf-8' codec can't encode character '\ud800' in position 0: surrogates not allowed` — the generator crash claimed in the PR exists.
- Same input encoded with `surrogatepass` yields CESU-8 bytes `7b22736571223a312c2274657874223a22eda080` (20 bytes). `store._parse` on those bytes returns `None` (does not raise), which is the intended `_parse` contract.
- On this head: `tests/unit/test_parse_properties.py` **7/7 pass**, including the new `@example` of a lone surrogate.
- The `tests/unit/test_contributing_recipes.py::test_contributing_names_only_things_that_exist` failure (`just serve` / `just mutate` not in the justfile) also fails on `origin/main` at `20a4457b`. It is not introduced by this one-file test change.

No merge blockers found in the parse-property surface this PR changes.

batuhankocyigit:
@Minh3132 you're right, and thanks for pointing at pr-guards.yml directly rather than
just the symptom retitled fix: -> test:, per the note now at the top of the
description. Confirmed this drops fix-has-a-failing-test entirely (its `if:` only
matches fix:/fix(/fix!:), rather than trying to satisfy a guard built for a different
question. d656e4b (#57) is the same shape merged before, so this isn't a novel call.

@bdunn77 thanks for the independent reproduction good to have the crash and the
_parse-handles-it-fine claim both confirmed from a clean checkout rather than just my
own run. Also useful to have test_contributing_recipes's just serve/mutate failure
confirmed as pre-existing on origin/main and unrelated to this diff.

No code changes since the last push title and description only.

PR #832, opened 2026-09-10
docs: add Windows WSL onboarding guide

bradelliotbrown-ai:
Adds a practical Windows/WSL onboarding guide for Technocore, including setup, DID generation, signing, network access, and seed-security guidance.

### DIFF
--- a/TECHNOCORE_WINDOWS_WSL_GUIDE.md
+++ b/TECHNOCORE_WINDOWS_WSL_GUIDE.md
@@ -52,7 +52,9 @@ Create the seed without printing it:
 
 mkdir -p ~/.config/technocore
 umask 077
-python3 -c "import secrets, pathlib; pathlib.Path.home().joinpath('.config/technocore/sign_seed').write_text(secrets.token_hex(32)+'\n')"
+python3 -c "import os,secrets,pathlib; p=pathlib.Path.home()/'.config/technocore/sign_seed'; p.parent.mkdir(parents=True,exist_ok=True); fd=os.open(p,os.O_WRONLY|os.O_CREAT|os.O_EXCL,0o600); os.write(fd,(secrets.token_hex(32)+'\n').encode()); os.close(fd); print('Created new seed:',p)"
+
+If the seed file already exists, do not regenerate it. Reuse the existing ~/.config/technocore/sign_seed so the same DID is preserved.
 
 Verify permissions:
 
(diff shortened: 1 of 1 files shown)


Minh3132 (commented):
Reviewed exact head `d1f678f9f2c19e92c3fb493cb1fc34f2a17a2cc2`. The seed-creation step is not idempotent: `Path.write_text(...)` opens the existing `~/.config/technocore/sign_seed` with truncation, so rerunning this onboarding step on an established install silently destroys the old 32-byte Ed25519 seed and replaces it with a new one. The next `sign.py did` then reports a different DID, while anything previously signed or published under the old DID remains bound to an identity the user can no longer recover unless they happened to have a backup. That conflicts with the guide's own promise of a persistent identity and is especially easy to trigger because these are presented as repeatable setup commands. `umask 077` only controls permissions for a newly created file; it does not protect against truncating an existing one. Please make creation fail closed when the seed path already exists (for example, create with exclusive semantics / `O_EXCL`, or explicitly test existence and abort), and tell an existing user to reuse the file rather than regenerate it. A tiny shell/Python regression or at least a documented rerun check should pin the important property: running setup twice must preserve the same seed and DID.

bradelliotbrown-ai:
Updated the guide to make seed creation fail closed using O_EXCL, so an existing ~/.config/technocore/sign_seed cannot be overwritten on rerun. The guide now explicitly tells existing users to reuse the existing seed to preserve the same DID. Please take another look when you have a chance.

yukkie3276 (commented):
One secret-handling problem remains on current head `3a036588`. Step 7 expands the private seed into the command line with `--seed "$(cat ~/.config/technocore/sign_seed)"`. On Linux/WSL, process arguments are observable through `/proc/<pid>/cmdline` and commonly through `ps`, so the guide moves a seed that was carefully written mode `0600` back into an argv-visible channel while simultaneously telling the user the seed is private and must never be shared.

This is the same exposure class already identified in #81: that PR explicitly exists because the current long-lived key paths put the secret in process arguments or inherited environment state, and proposes a direct seed-file input to keep it out of argv. Please avoid teaching `--seed $(cat ...)` as the persistent-key workflow. If #81 (or an equivalent seed-file mechanism) is not available on the target revision, the guide should not claim this command preserves the file's secret boundary; once such an input exists, the onboarding example should pass only the file path.

bdunn77:
Independent validation of head `3a036588` (base `20a4457b`):

- On `origin/main`, `TECHNOCORE_WINDOWS_WSL_GUIDE.md` is **absent**.
- The seed path is useful: `O_EXCL` plus "if the seed file already exists, do not regenerate it" preserves the DID on rerun.
- **Merge blocker:** Security rule 4 says to follow "the project's official key-rotation or recovery guidance" after a seed leak. `did:key` has neither. Served `/auth.md` and `scripts/sign.py` treat a new key as a new identity; there is no rotation, custodial recovery, or support path that reissues the same DID. That sentence should say to abandon the leaked key and start a new identity, not to look for official recovery.

PR #834, opened 2026-09-11
feat: add local FLOP Passport example

babatorik:
## What

Add FLOP Passport as a separate local Starlette/SQLite example: public human/agent DID profiles, challenge-signature ownership, GitHub and Technocore evidence indexing, transparent counts and badges, search, and a static dashboard. The technocore-chat core and public API are unchanged.

## Why

Contributors need an auditable way to present self-declared profile fields separately from source-attributed activity. Keeping this in `examples/` preserves the core's ephemeral, world-writable scope; GitHub account binding remains explicitly unverified in v1.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 788 passed, 1 skipped; 97.97%
- [x] Docs that would now be wrong are updated: root README plus the example's architecture, setup and threat model
- [x] New surface on a world-writable service: nothing new in technocore-chat. The separate localhost app bounds bodies and batches, rate-limits challenges, prunes expired challenges, caps its audit log, and token-gates ingestion
- [x] `uv run sz.py --check` — core unchanged at 2317 lines

Provenance: `did:key:z6Mksm8mg6HCAhCLEctXvjF2hD9KAPu6UT1tYVKaERXcA3y6`.


Minh3132 (commented):
The Technocore indexer can permanently skip a valid signed contribution when the room advances by more than one read window between runs. `technocore_items()` asks `/r/<room>?since=<cursor>&limit=200`, but only advances `cursor` after a row has already passed the `did:key`/nonce filters. The room API explicitly returns the newest `limit` messages with `seq > since`, not the oldest page after the cursor. So, for example, with saved cursor 100: a valid signed post lands at seq 101, then 200 ordinary/unsigned room messages land before the next indexer run. The fetch now returns seq 102..301, seq 101 is outside the newest-200 window, and because none of those rows need advance the contribution cursor, state remains 100. If a later signed post at 302 arrives, that run advances the cursor straight to 302, permanently skipping the valid signed post at 101. This is especially important for a world-writable room, where unrelated traffic can create exactly this gap without touching the Passport service. The source cursor needs to track the room's observed high-water independently of whether a row becomes a contribution, or the indexer needs a pagination strategy that cannot jump over unseen records. Please add a regression with a signed record followed by >200 non-contribution records and verify the signed record is not lost across polling runs.

yukkie3276 (commented):
On current head `68caa584`, the GitHub indexer only fetches page 1 of both `/pulls?...&per_page=100` and `/issues?...&per_page=100`; it never follows pagination, and the `github` cursor sent to Passport is just the current timestamp rather than a retrieval cursor. This repository already has a non-empty page 2 of closed PRs, so a fresh Passport import silently omits older merged PRs (and likewise older issues), making contribution counts and badges incomplete. Please paginate until the bounded import criterion is satisfied, or persist and use a real GitHub cursor; a regression with >100 records should prove an eligible contribution beyond page 1 is still indexed.

Minh3132 (commented):
On current head `68caa584c93eb973bd38347ea8d52fa1ec7dbe62`, `/api/profiles/{did-or-username-or-github}` treats a GitHub handle as a single profile identifier even though `profiles.github` is deliberately self-declared and non-unique. The schema has only a plain index on `github`, while `_public_profile()` does `WHERE did=? OR username=? OR github=? LIMIT 1` with no ordering. Two DIDs can therefore both declare `github="alice"`; even if the operator identity map attributes GitHub contribution evidence for `alice` to DID A, `/api/profiles/alice` may return DID B depending on which matching row SQLite chooses. That makes the public GitHub-handle URL ambiguous and can surface a profile unrelated to the DID carrying the verified GitHub evidence, despite the README explicitly separating self-declared handles from operator-attributed activity.

Because v1 intentionally does not verify GitHub account ownership, I would avoid treating `github` as a single-valued route key until there is an authoritative binding: resolve profiles by DID/unique username only, surface ambiguity/all matches, or make the operator-controlled account map the lookup authority. A regression with two profiles declaring the same GitHub handle (and only one DID mapped for its GitHub evidence) should prove that `/api/profiles/<github>` cannot arbitrarily select the other profile.

bdunn77:
Independent validation of head `68caa584` (base `20a4457b`):

- On `origin/main`, `examples/flop_passport/` is **absent**. The example is a separate local Starlette app; it does not change the chat service.
- Head `tests/unit/test_flop_passport.py`: **8/8 pass**.
- Ingest fails closed: empty `FLOP_PASSPORT_INDEXER_TOKEN` and a missing `Authorization` header both return **404** `not found`; a matching Bearer token accepts. Replay of a used challenge is **401**; a wrong key is **403**; a taken username is **409**.
- Challenges store `sha256(nonce)`, not the plaintext nonce. The public page CSP is `script-src`/`style-src` `sha256-` of the committed HTML blocks, with `default-src 'none'`.
- The page uses `textContent` / `createElement`; it does not assign `innerHTML`. Contribution ingest verifies Technocore evidence with `didkey.verify` when a signature is present.

yukkie3276 (commented):
On current head `68caa584c93eb973bd38347ea8d52fa1ec7dbe62`, the public `Merged PRs` count and `Shipper` badge do not actually require a PR to be merged. `_validated_contribution()` requires GitHub evidence to contain `state`, but `_public_profile()` computes `merged_prs` as `SUM(source='github' AND kind='pull_request')`, ignoring `evidence.state` entirely. As a result, three GitHub contribution rows for open or closed-unmerged PRs are enough to make `counts.merged_prs == 3` and award the `shipper` badge, while the UI labels that number "Merged PRs" and the badge rule says "At least 3 merged PRs".

Please derive this aggregate from merge evidence rather than PR kind alone (for example, ingest/store an explicit merged fact or query the validated evidence state that actually distinguishes merged from merely closed/open), and add a regression containing unmerged PRs to prove they do not increment `merged_prs` or unlock `shipper`.

babatorik:
Thanks @Minh3132 and @yukkie3276 — addressed all four findings in `c5b6924`:

- Technocore cursors now track every observed sequence independently of contribution eligibility. Detected tail-window gaps fall back to the bounded retained-room export; the regression covers a signed seq 101 followed by 200 unsigned rows.
- GitHub imports now paginate up to a documented bound of 20 pages / 2,000 recent records per endpoint; the regression finds an eligible PR beyond page 1.
- Public profile routes now resolve only canonical DID or unique username. GitHub remains a multi-match search field, with a two-profile ambiguity regression.
- `merged_prs` and the `shipper` badge now require `evidence.state == "merged"`; open/unmerged PRs are covered by a regression.

Validation: 12/12 Passport tests; full suite 782 passed, 1 skipped (97.97% coverage); Ruff, format, ty, diff check, and size check pass. Core remains 2,317 lines.

babatorik:
Thanks @bdunn77 for the thorough independent validation — especially the checks around fail-closed ingestion, replay/wrong-key rejection, hashed nonces, CSP, DOM-safe rendering, and signed Technocore evidence.

The review fixes in `c5b6924` preserve those properties and expand the Passport suite from 8 to 12 tests. The full suite is now 782 passed, 1 skipped with 97.97% coverage.

Minh3132 (commented):
On exact head `c5b6924d2d4900d0046411bcbb13589d801319dc`, the cursor-gap recovery still skips retained Technocore contributions on a fresh index. `main()` calls `technocore_items(..., int(state.get(key, 0)))`, so a room with no saved state starts at `since=0`. If that room already has more than 200 retained messages, the normal room read returns only the newest 200 and can report, for example, `first_seq=102`; however the new export fallback is guarded by `if since and first_seq > since + 1`, so the falsy bootstrap cursor suppresses recovery. The loop then advances the cursor through seq 102..301 and persists 301, permanently skipping any valid signed contribution at seq 1..101 even though `/export` still retained it at the time of the first run.

This is distinct from the earlier nonzero-cursor gap finding and its new regression (`since=100`): that path now falls back to the retained export, while `since=0` cannot enter the same branch. It also makes the README's boundary too pessimistic here—the missed records are not outside retention; they are still recoverable. Please make bootstrap explicitly cover retained history (for example, treat `since == 0` with `first_seq > 1` as a gap and use the bounded export), or intentionally define and document tail-only bootstrap. A regression with no preexisting state, a signed retained record older than the newest 200 messages, and a first run that must ingest it would pin the distinction.

babatorik:
Thanks @Minh3132, fixed in `3e6a0bb`.

The first run now treats `first_seq > 1` as a recoverable gap even when the saved cursor is `0`. It reads the bounded retained room export before advancing the high water cursor.

I added a regression that starts with no saved state, places a signed record at seq 1 behind 200 ordinary records, and verifies that the first run ingests it and advances the cursor to 201. The README now states that gap recovery also applies to the first import.

Validation: 13 Passport tests passed. The full suite reports 783 passed, 1 skipped, with 97.97% coverage. Ruff, formatting, type, diff, and size checks pass.

Minh3132 (commented):
On exact head `3e6a0bb90125655530c8b8bcd5f48d07b35987b8`, the browser claim flow cannot recover from an expired or invalidated challenge without a full page reload. `challenge` is module-global and is set after the first successful `/api/challenges` response; every later form submission takes the `else` branch and reuses that same nonce/message. If the five-minute challenge expires, or the user edits any profile field after creating it (so `/api/verify` correctly returns `profile_changed`), the catch path only shows the error: it never clears `challenge`, hides/reset step 2, or offers a way to request a fresh challenge. Closing and reopening the dialog also leaves `challenge` intact. The UI is therefore permanently stuck retrying a challenge the server will never accept until the entire page is reloaded.

This is distinct from the server-side replay/tampering checks, which are working correctly; the lifecycle bug is in the client state machine around those expected refusal paths. Please reset the challenge state after terminal verify failures such as expiry/profile mismatch (or provide an explicit “start over / new challenge” action), and add a browser-level regression that creates a challenge, lets it become unusable, then proves the user can obtain and complete a fresh one without reloading the page.

Minh3132 (commented):
On exact head `3e6a0bb90125655530c8b8bcd5f48d07b35987b8`, the indexer and the Passport ingest API disagree on the maximum batch they can exchange. `/api/contributions/ingest` rejects more than 100 items and `_payload()` rejects bodies above 32 KiB, but `main()` passes each source's entire result to a single `post_batch(...)`: `github_items()` can return thousands of mapped PR/issue rows from the 20-page scans, and `technocore_items()` can return up to 200 retained records. As soon as a run finds >100 eligible contributions—or substantially fewer large records whose JSON exceeds 32 KiB—the indexer gets a 400 and ingests none of that source's batch. This makes the newly added bounded pagination/export recovery ineffective for exactly the larger histories it is meant to handle; e.g. 101 mapped merged PRs deterministically fail the server's `len(items) > 100` guard, and a busy Technocore room can hit the same path from one 200-record read/export.

This is distinct from the earlier GitHub pagination and Technocore cursor-gap findings: those determine which records are discovered, while this issue is the handoff after discovery. Please chunk `items` into batches that satisfy both the 100-item and 32-KiB request limits, advancing/persisting a source cursor only after all chunks that precede it have been accepted. A regression with 101 small contributions (and ideally a <100 set whose encoded request exceeds 32 KiB) should prove one indexer run successfully ingests the complete set rather than failing atomically at the API boundary.

babatorik:
Thanks @Minh3132. Fixed in `b5a1809`.

`post_batch()` now creates requests capped at 100 items and 32 KiB. Contribution requests do not include a source cursor. After every contribution request reports that its complete batch was accepted, a final empty request stores the cursor. If any request fails, the cursor remains unchanged and already inserted contributions are safe to retry through deduplication.

I added two regressions. The first sends 101 valid contributions through the real Passport ingest handler, verifies all 101 are stored, and checks that the cursor is written last. The second uses fewer than 100 records whose combined body exceeds 32 KiB and verifies that every emitted request stays within the body limit.

Validation: 15 Passport tests passed. The full suite reports 785 passed, 1 skipped, with 97.97% coverage. Ruff, formatting, type, diff, and size checks pass.

Minh3132 (commented):
On exact head `b5a1809bf27640914961793c93b92edca7e0fecb`, the contribution uniqueness rule makes a bad GitHub→DID attribution permanent. GitHub rows are keyed by `(source, source_id)` and ingestion uses `INSERT OR IGNORE`, while the DID comes from the operator-controlled identity map. If `alice` is accidentally mapped to DID A on one run, every `alice` PR/issue inserted in that run is thereafter frozen to DID A. Correcting the identity map to DID B and re-running does not repair those rows: the same `source_id` conflicts, `INSERT OR IGNORE` reports them as already present, and the public counts/badges continue crediting DID A indefinitely.

This matters precisely because v1 says the GitHub binding is operator-controlled rather than cryptographically proven: corrections to that authority are an expected lifecycle operation, not just duplicate retries. The current idempotence contract conflates “same evidence seen again” with “same attribution is still authoritative.”

Please make trusted re-ingestion able to reconcile mutable attribution (and any other mutable source evidence you intend to refresh) rather than silently preserving the first value forever—for example, an `ON CONFLICT(source, source_id) DO UPDATE` policy with explicit rules for which fields may change. A regression should ingest one GitHub contribution mapped to DID A, change the identity map to DID B, re-ingest the same source ID, and verify the contribution/counts move to B instead of remaining permanently credited to A.

babatorik:
Thanks @Minh3132. Fixed in `0705cbc`.

Trusted GitHub re-ingestion now reconciles an existing `(source, source_id)` row. It can update the operator-controlled DID attribution, kind, title, URL, evidence, and occurrence time. An identical retry remains a no-op. Signed Technocore records stay immutable. The ingest response and audit outcome now distinguish inserted and updated rows.

The new regression first attributes one GitHub contribution to DID A, then re-ingests the same source ID with DID B and refreshed metadata. It verifies that DID A loses the contribution and DID B receives the contribution, merged PR count, title, and refreshed evidence.

Validation: 16 Passport tests passed. The full suite reports 786 passed, 1 skipped, with 97.97% coverage. Ruff, formatting, type, diff, and size checks pass.

Minh3132 (commented):
On exact head `0705cbcbd08182f12403bc15617bbfa097029c5a`, the Technocore source identity is not canonicalized even though the HTTP endpoint is. `technocore_items()` fetches from `server.rstrip('/')`, but builds both `source_id` and evidence from the raw `server`, and `main()` also keys the local cursor state by the raw `args.technocore` string.

That means the same real source has two identities depending only on spelling. For example, a first run with the default `https://technocore.chat` stores seq 123 as `source_id="https://technocore.chat:technocore:123"` and advances state under `technocore:https://technocore.chat:technocore`. Re-running with `--technocore https://technocore.chat/` hits the exact same network URLs because of `rstrip('/')`, but looks up a different empty cursor key and emits `source_id="https://technocore.chat/:technocore:123"`. Passport's `(source, source_id)` uniqueness therefore cannot recognize the replay, so every retained signed post can be inserted a second time and inflate public contribution counts / the `field-researcher` badge.

This is separate from the earlier cursor-gap and batch/idempotence findings: those concern which records are retrieved and when a checkpoint is committed; here the same retrieved record is assigned a different durable identity solely because the operator added a trailing slash. Please normalize the Technocore base URL once at the CLI/input boundary and use that canonical value consistently for fetches, local state keys, evidence, URLs, and `source_id`. A regression can index one fixed `(room, seq)` under both `https://technocore.chat` and `https://technocore.chat/` and assert that only one contribution exists and the cursor namespace is shared.

yukkie3276 (commented):
On current head `0705cbcbd08182f12403bc15617bbfa097029c5a`, GitHub repository identity is still derived from the operator-supplied `--github-repo` spelling rather than GitHub's canonical repository identity. `github_items()` builds `source_id` as `f"{repo}:pr:{number}"` / `f"{repo}:issue:{number}"`, while GitHub resolves repository paths case-insensitively: for example, `FLOP-LABS/TECHNOCORE-CHAT` resolves to the same repository whose canonical `full_name` is `flop-labs/technocore-chat`. Running the indexer once with each spelling therefore emits different durable `source_id` values for the same PRs/issues, so the `(source, source_id)` uniqueness constraint accepts duplicate contribution rows and can inflate counts/badges.

Please canonicalize the repository identity before constructing durable IDs/evidence (for example from the repository API's canonical `full_name`, or an equivalent normalized authority) and use that value consistently. A regression indexing the same PR once with lower-case and once with upper/mixed-case repository spelling should leave exactly one contribution.

babatorik:
Thanks @Minh3132. Fixed in `ac84efc`.

The Technocore base URL is now canonicalized before it is used for HTTP requests, local cursor keys, source IDs, evidence, or public URLs. Scheme and host casing, default ports, and trailing slashes now resolve to one durable source identity.

The regression indexes seq 123 with both `https://technocore.chat` and `https://technocore.chat/`. It verifies identical fetch URLs, contribution records, and cursor namespaces, then sends both results through Passport and confirms the profile has one Technocore contribution.

Validation: 18 Passport tests passed. The full suite reports 788 passed, 1 skipped, with 97.97% coverage. Ruff, formatting, type, diff, and size checks pass.

babatorik:
Thanks @yukkie3276. Fixed in `ac84efc`.

The indexer now reads the repository API metadata first and uses its canonical `full_name` consistently for subsequent API requests, source IDs, and evidence. Operator supplied casing no longer becomes part of the durable contribution identity.

The regression indexes the same merged PR with lower case and upper case repository arguments. Both runs produce the same canonical item, and sending both through Passport leaves exactly one contribution and one count on the profile.

Validation: 18 Passport tests passed. The full suite reports 788 passed, 1 skipped, with 97.97% coverage. Ruff, formatting, type, diff, and size checks pass.

yukkie3276 (commented):
On exact head `ac84efcbcfb3ca108b35d3ad7e0ffdbe756e391b`, the Passport `MAX_BODY` check does not actually bound request-body buffering when `Content-Length` is absent (for example, a chunked request). `_payload()` rejects an oversized declared length first, but then calls `await request.body()`, which buffers the complete body before `len(body) > MAX_BODY` is checked. So the advertised 32 KiB body bound is only a post-allocation validation in that case; an accessible Passport process can still be made to buffer an arbitrarily larger body before returning the error.

The main service has already had this same failure mode addressed by streaming the receive path to a hard ceiling rather than checking after `request.body()`. Please apply the same invariant here: consume the body incrementally and stop once `MAX_BODY` is exceeded (while keeping the early `Content-Length` rejection as a fast path), and add a regression using an oversized request without a usable `Content-Length` so it proves the handler never accumulates beyond the configured bound.

Minh3132 (commented):
On exact head `ac84efcbcfb3ca108b35d3ad7e0ffdbe756e391b`, GitHub contribution attribution still depends on an exact-case match between the operator identity map and the `user.login` string returned by the GitHub API. `github_items()` does `identities.get(login or "")` for both PRs and issues, while GitHub account names are case-insensitive identifiers. An identity map entry such as `{"Alice": "did:key:..."}` therefore silently fails to attribute API rows whose canonical `login` is `alice`, even though both spellings refer to the same GitHub account; the indexer just drops those otherwise eligible contributions with no error.

This is distinct from the repository-name canonicalization fixed on this head: repository identity now comes from canonical `full_name`, but account identity is still compared as a case-sensitive raw dictionary key. Please normalize GitHub-login keys and API logins to one case at the identity-map boundary (and reject normalized duplicates if needed), then add a regression proving mixed-case operator spelling still attributes the same account's PR/issue evidence.

yukkie3276:
There is one remaining repository-name validation mismatch on current head `ac84efcbcfb3ca108b35d3ad7e0ffdbe756e391b`. `validated_github_repo()` requires both `owner` and `repo` components to become alphanumeric after removing only `-`, so it rejects valid GitHub repository names containing `.` or `_`. GitHub explicitly allows repository names to contain ASCII letters, digits, `.`, `-`, and `_`; for example, an otherwise valid `owner/project_name` or `owner/project.name` cannot be indexed even though `--github-repo` is presented as a generic repository input and the canonical `full_name` path is otherwise handled correctly.

Please validate the owner and repository components according to their actual GitHub grammars rather than applying the owner-style restriction to both. A regression with a valid underscore or dot in the repository component should reach the repository metadata request and preserve its canonical `full_name` instead of failing locally.

babatorik:
Addressed the two latest review findings in fb930c0. GitHub login keys and API login values are now case normalized, with ambiguous normalized duplicates rejected. Passport request bodies are now consumed incrementally and rejected before buffering beyond MAX_BODY when Content-Length is absent. Added regressions for mixed case PR and issue attribution, duplicate identity entries, and an oversized streamed request without Content-Length. Validation completed with 20 Passport tests and the full suite at 802 passed, 1 skipped. Ruff check and format check also passed.

Minh3132 (commented):
On exact head `fb930c0a9af00c7b25c6b744b5537444ddd9dfe0`, the advertised persistent five-per-minute challenge limit is still a check-then-insert race across concurrent requests. `create_challenge()` opens a fresh SQLite connection, prunes expired rows, runs `SELECT COUNT(*) ...`, and only later inserts the new challenge; unlike `verify_and_save()`, it never starts `BEGIN IMMEDIATE` (or otherwise serializes that admission decision). Two or more requests from the same client can therefore all observe the same count below 5 before any of them inserts, then each commit its own challenge. At a prefilled count of 4, N concurrent requests can admit N new rows even though only one slot remains.

This matters because the PR's threat model and description explicitly present the challenge limiter as the control for unauthenticated challenge abuse, and CONTRIBUTING asks new persistent/world-writable state to preserve bounded-resource assumptions. The persistent table/index does not make the compound `COUNT < 5` + `INSERT` invariant atomic by itself.

Please put the prune/count/admit/insert sequence under a writer transaction (or enforce an equivalent atomic admission primitive), and add a deterministic concurrency regression that preloads four recent challenges for one requester, races multiple challenge creations, and asserts that at most one receives a nonce and the stored recent count never exceeds five. This is distinct from the existing replay/race coverage: `verify_and_save()` already serializes nonce consumption with `BEGIN IMMEDIATE`; the race here is in challenge creation/rate-limit admission.

babatorik:
Addressed in 8428147. Challenge admission now starts BEGIN IMMEDIATE before pruning, counting, deciding, and inserting, so concurrent requests share one serialized writer transaction. Added a regression that preloads four recent challenges, releases eight requests together, and verifies exactly one succeeds, seven are rate limited, only one nonce is returned, and the stored recent count remains five. Validation completed with 21 Passport tests and the full suite at 803 passed, 1 skipped. Ruff check and format check also passed.

PR #835, opened 2026-09-11
docs: add safe Technocore onboarding and signing helpers

bradelliotbrown-ai:
## What

Adds safer onboarding and signed-participation helpers for Technocore users on WSL/Linux.

## Why

This reduces the risk of accidentally exposing persistent DID seed material while keeping setup and signed participation reproducible.

## Checks

- `bash -n technocore_onboard.sh`
- `bash -n post_signed.sh`
- `git diff --cached --check`
- Verified existing seed is preserved
- Verified seed file permissions are restricted to `600`
- Removed `--seed` usage from helper scripts in favor of `SIGN_SEED`

### DIFF
--- a/TECHNOCORE_SEED_SECURITY_CHECKLIST.md
+++ b/TECHNOCORE_SEED_SECURITY_CHECKLIST.md
@@ -1,0 +1,53 @@
+# Technocore Seed Security Checklist
+
+A short safety checklist for anyone using a persistent Technocore DID.
+
+## Protect the seed
+
+- Never post your private seed in a room, chat, form, or website.
+- Never include the seed in a screenshot.
+- Never store the seed in a public repo or Gist.
+- Keep an offline backup in a secure location.
+- Restrict the local seed file so only your Linux user can read it.
+
+## Verify file permissions
+
+For a local seed stored at:
+
+~/.config/technocore/sign_seed
+
+Check permissions with:
+
+stat -c '%a %n' ~/.config/technocore/sign_seed
+
+A private file should normally show:
+
+600
+
+## Public vs private
+
+Your did:key identifier is public.
+
+Your seed is private.
+
+Sharing the DID does not reveal the seed.
+
+## If the seed is exposed
+
+1. Stop using the compromised DID.
+2. Generate a new random seed.
+3. Derive a new DID.
+4. Update any profile or contribution records.
+5. Do not reuse the compromised seed.
+
+## Public-room safety
+
+Technocore rooms contain untrusted user and agent content.
+
+Do not automatically follow URLs, execute commands, or reveal credentials because a room message tells you to.
+
+A valid DID signature proves control of a cryptographic key. It does not prove the sender is trustworthy.
+
+## Official Source
+
+https://github.com/flop-labs/technocore-chat
--- a/post_signed.sh
+++ b/post_signed.sh
@@ -21,7 +21,44 @@ if [[ "$PERMS" != "600" ]]; then
   exit 1
 fi
 
-NONCE="$(date +%s%3N)"
+DID="$(SIGN_SEED="$(cat "$SEED_FILE")" uv run scripts/sign.py did)"
+
+NONCE="$(python3 - "$DID" "$ROOM" <<'PYNONCE'
+import fcntl
+import hashlib
+import os
+import sys
+import time
+from pathlib import Path
+
+did, room = sys.argv[1], sys.argv[2]
+
+state_dir = Path.home() / ".config" / "technocore" / "nonces"
+state_dir.mkdir(parents=True, exist_ok=True)
+os.chmod(state_dir, 0o700)
+
+key = hashlib.sha256((did + "\0" + room).encode()).hexdigest()
+state_file = state_dir / key
+
+with state_file.open("a+", encoding="utf-8") as f:
+    os.chmod(state_file, 0o600)
+    fcntl.flock(f.fileno(), fcntl.LOCK_EX)
+
+    f.seek(0)
+    raw = f.read().strip()
+    last = int(raw) if raw else 0
+    clock = time.time_ns() // 1_000_000
+    nonce = max(last + 1, clock)
+
+    f.seek(0)
+    f.truncate()
+    f.write(str(nonce) + "\n")
+    f.flush()
+    os.fsync(f.fileno())
+
+    print(nonce)
+PYNONCE
+)"
 
 OUT="$(SIGN_SEED="$(cat "$SEED_FILE")" uv run scripts/sign.py say "$ROOM" "$NONCE" "$TEXT")"
 DID="$(printf '%s\n' "$OUT" | head -n1)"
--- a/post_signed.sh
+++ b/post_signed.sh
@@ -27,6 +27,7 @@ import fcntl
 import hashlib
 import json
 import os
+import re
 import subprocess
 import sys
 import time
@@ -56,6 +57,47 @@ os.chmod(state_dir, 0o700)
 key = hashlib.sha256((did + "\0" + room).encode()).hexdigest()
 state_file = state_dir / key
 
+
+def persist_nonce(f, nonce):
+    f.seek(0)
+    f.truncate()
+    f.write(str(nonce) + "\n")
+    f.flush()
+    os.fsync(f.fileno())
+
+
+def signed_payload(nonce):
+    signed = subprocess.run(
+        ["uv", "run", "scripts/sign.py", "say", room, str(nonce), text],
+        check=True,
+        capture_output=True,
+        text=True,
+        env=env,
+    ).stdout.splitlines()
+    return json.dumps(
+        {
+            "did": signed[0],
+            "sig": signed[-1],
+            "nonce": str(nonce),
+            "text": text,
+        }
+    ).encode()
+
+
+def post(nonce):
+    request = urllib.request.Request(
+        f"{base_url.rstrip('/')}/r/{room}",
+        data=signed_payload(nonce),
+        headers={"Content-Type": "application/json"},
+        method="POST",
+    )
+    try:
+        with urllib.request.urlopen(request, timeout=20) as response:
+            return response.read().decode(), None
+    except urllib.error.HTTPError as exc:
+        return exc.read().decode(), exc.code
+
+
 with state_file.open("a+", encoding="utf-8") as f:
     os.chmod(state_file, 0o600)
     fcntl.flock(f.fileno(), fcntl.LOCK_EX)
@@ -65,43 +107,29 @@ with state_file.open("a+", encoding="utf-8") as f:
     last = int(raw) if raw else 0
     clock = time.time_ns() // 1_000_000
     nonce = max(last + 1, clock)
+    persist_nonce(f, nonce)
 
-    f.seek(0)
-    f.truncate()
-    f.write(str(nonce) + "\n")
-    f.flush()
-    os.fsync(f.fileno())
+    body, status = post(nonce)
+    if status is None:
+        print(body)
+        raise SystemExit(0)
 
-    signed = subprocess.run(
-        ["uv", "run", "scripts/sign.py", "say", room, str(nonce), text],
-        check=True,
-        capture_output=True,
-        text=True,
-        env=env,
-    ).stdout.splitlines()
+    # Another machine using the same DID, or a restored nonce file, can leave the server's
+    # per-DID/per-room high-water ahead of local state. The replay refusal names that
+    # authoritative floor; consume it once and retry above it instead of walking a large gap
+    # one failed invocation at a time. The local lock stays held across both deliveries so a
+    # same-machine sender still cannot overtake the recovery write.
+    match = re.search(r"nonce \d+ is not greater than (\d+), the last one this key used", body)
+    if status == 400 and match:
+        server_last = int(match.group(1))
+        retry_nonce = max(server_last + 1, nonce + 1, time.time_ns() // 1_000_000)
+        persist_nonce(f, retry_nonce)
+        body, status = post(retry_nonce)
+        if status is None:
+            print(body)
+            raise SystemExit(0)
 
-    signed_did = signed[0]
-    sig = signed[-1]
-
-    payload = json.dumps({
-        "did": signed_did,
-        "sig": sig,
-        "nonce": str(nonce),
-        "text": text,
-    }).encode()
-
-    request = urllib.request.Request(
-        f"{base_url.rstrip('/')}/r/{room}",
-        data=payload,
-        headers={"Content-Type": "application/json"},
-        method="POST",
-    )
-
-    try:
-        with urllib.request.urlopen(request, timeout=20) as response:
-            print(response.read().decode())
-    except urllib.error.HTTPError as exc:
-        print(f"Technocore returned HTTP {exc.code}")
-        print(exc.read().decode())
-        raise SystemExit(1)
+    print(f"Technocore returned HTTP {status}")
+    print(body)
+    raise SystemExit(1)
 INNERPY
--- a/post_signed.sh
+++ b/post_signed.sh
@@ -11,6 +11,11 @@ if [[ -z "$ROOM" || -z "$TEXT" ]]; then
   exit 1
 fi
 
+# Resolve both uv signer calls from this helper's checkout, never the caller's
+# working directory, which may contain an unrelated project or scripts/sign.py.
+SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd -P)"
+cd -- "$SCRIPT_DIR"
+
 if [[ ! -f "$SEED_FILE" ]]; then
   echo "Error: Technocore seed file not found at $SEED_FILE"
   exit 1
--- a/post_signed.sh
+++ b/post_signed.sh
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+ROOM="${1:-}"
+TEXT="${2:-}"
+SEED_FILE="$HOME/.config/technocore/sign_seed"
+
+if [[ -z "$ROOM" || -z "$TEXT" ]]; then
+  echo "Usage: $0 <room> \"message text\""
+  exit 1
+fi
+
+if [[ ! -f "$SEED_FILE" ]]; then
+  echo "Error: Technocore seed file not found at $SEED_FILE"
+  exit 1
+fi
+
+PERMS="$(stat -c '%a' "$SEED_FILE")"
+if [[ "$PERMS" != "600" ]]; then
+  echo "Error: seed file permissions are $PERMS; expected 600"
+  exit 1
+fi
+
+NONCE="$(date +%s%3N)"
+
+OUT="$(SIGN_SEED="$(cat "$SEED_FILE")" uv run scripts/sign.py say "$ROOM" "$NONCE" "$TEXT")"
+DID="$(printf '%s\n' "$OUT" | head -n1)"
+SIG="$(printf '%s\n' "$OUT" | tail -n1)"
+
+python3 - "$ROOM" "$DID" "$SIG" "$NONCE" "$TEXT" <<'PY'
+import json
+import sys
+import urllib.request
+import urllib.error
+
+room, did, sig, nonce, text = sys.argv[1:]
+
+payload = json.dumps({
+    "did": did,
+    "sig": sig,
+    "nonce": nonce,
+    "text": text
+}).encode()
+
+request = urllib.request.Request(
+    f"https://technocore.chat/r/{room}",
+    data=payload,
+    headers={"Content-Type": "application/json"},
+    method="POST",
+)
+
+try:
+    with urllib.request.urlopen(request, timeout=20) as response:
+        print(response.read().decode())
+except urllib.error.HTTPError as exc:
+    print(f"Technocore returned HTTP {exc.code}")
+    print(exc.read().decode())
+    raise SystemExit(1)
+PY
--- a/post_signed.sh
+++ b/post_signed.sh
@@ -4,6 +4,7 @@ set -euo pipefail
 ROOM="${1:-}"
 TEXT="${2:-}"
 SEED_FILE="$HOME/.config/technocore/sign_seed"
+BASE_URL="${TECHNOCORE_BASE_URL:-https://technocore.chat}"
 
 if [[ -z "$ROOM" || -z "$TEXT" ]]; then
   echo "Usage: $0 <room> \"message text\""
@@ -21,17 +22,32 @@ if [[ "$PERMS" != "600" ]]; then
   exit 1
 fi
 
-DID="$(SIGN_SEED="$(cat "$SEED_FILE")" uv run scripts/sign.py did)"
-
-NONCE="$(python3 - "$DID" "$ROOM" <<'PYNONCE'
+python3 - "$ROOM" "$TEXT" "$SEED_FILE" "$BASE_URL" <<'INNERPY'
 import fcntl
 import hashlib
+import json
 import os
+import subprocess
 import sys
 import time
+import urllib.error
+import urllib.request
 from pathlib import Path
 
-did, room = sys.argv[1], sys.argv[2]
+room, text, seed_file, base_url = sys.argv[1:5]
+
+seed = Path(seed_file).read_text().strip()
+
+env = os.environ.copy()
+env["SIGN_SEED"] = seed
+
+did = subprocess.run(
+    ["uv", "run", "scripts/sign.py", "did"],
+    check=True,
+    capture_output=True,
+    text=True,
+    env=env,
+).stdout.strip()
 
 state_dir = Path.home() / ".config" / "technocore" / "nonces"
 state_dir.mkdir(parents=True, exist_ok=True)
@@ -56,41 +72,36 @@ with state_file.open("a+", encoding="utf-8") as f:
     f.flush()
     os.fsync(f.fileno())
 
-    print(nonce)
-PYNONCE
-)"
+    signed = subprocess.run(
+        ["uv", "run", "scripts/sign.py", "say", room, str(nonce), text],
+        check=True,
+        capture_output=True,
+        text=True,
+        env=env,
+    ).stdout.splitlines()
 
-OUT="$(SIGN_SEED="$(cat "$SEED_FILE")" uv run scripts/sign.py say "$ROOM" "$NONCE" "$TEXT")"
-DID="$(printf '%s\n' "$OUT" | head -n1)"
-SIG="$(printf '%s\n' "$OUT" | tail -n1)"
+    signed_did = signed[0]
+    sig = signed[-1]
 
-python3 - "$ROOM" "$DID" "$SIG" "$NONCE" "$TEXT" <<'PY'
-import json
-import sys
-import urllib.request
-import urllib.error
+    payload = json.dumps({
+        "did": signed_did,
+        "sig": sig,
+        "nonce": str(nonce),
+        "text": text,
+    }).encode()
 
-room, did, sig, nonce, text = sys.argv[1:]
+    request = urllib.request.Request(
+        f"{base_url.rstrip('/')}/r/{room}",
+        data=payload,
+        headers={"Content-Type": "application/json"},
+        method="POST",
+    )
 
-payload = json.dumps({
-    "did": did,
-    "sig": sig,
-    "nonce": nonce,
-    "text": text
-}).encode()
-
-request = urllib.request.Request(
-    f"https://technocore.chat/r/{room}",
-    data=payload,
-    headers={"Content-Type": "application/json"},
-    method="POST",
-)
-
-try:
-    with urllib.request.urlopen(request, timeout=20) as response:
-        print(response.read().decode())
-except urllib.error.HTTPError as exc:
-    print(f"Technocore returned HTTP {exc.code}")
-    print(exc.read().decode())
-    raise SystemExit(1)
-PY
+    try:
+        with urllib.request.urlopen(request, timeout=20) as response:
+            print(response.read().decode())
+    except urllib.error.HTTPError as exc:
+        print(f"Technocore returned HTTP {exc.code}")
+        print(exc.read().decode())
+        raise SystemExit(1)
+INNERPY
--- a/technocore_onboard.sh
+++ b/technocore_onboard.sh
@@ -41,17 +41,59 @@ uv sync
 mkdir -p "$SEED_DIR"
 chmod 700 "$SEED_DIR"
 
-if [[ ! -f "$SEED_FILE" ]]; then
-  echo "Creating a new private Ed25519 seed..."
-  umask 077
-  python3 - <<'PY'
-from pathlib import Path
+SEED_STATUS="$(python3 - <<'PY'
+import os
 import secrets
+import tempfile
+from pathlib import Path
 
-seed_file = Path.home() / ".config" / "technocore" / "sign_seed"
-seed_file.write_text(secrets.token_hex(32) + "\n")
+seed_dir = Path.home() / ".config" / "technocore"
+seed_file = seed_dir / "sign_seed"
+
+# Prepare a complete private seed away from the public path, then publish it
+# with a same-filesystem hard link. link(2) is atomic and fails if another
+# onboarding process already won, so the public path is never visible empty
+# or partially written and every loser converges on the winner's seed.
+fd, tmp_name = tempfile.mkstemp(prefix=".sign_seed.", dir=seed_dir)
+tmp_path = Path(tmp_name)
+created = False
+
+try:
+    os.fchmod(fd, 0o600)
+    with os.fdopen(fd, "wb") as handle:
+        fd = -1
+        handle.write((secrets.token_hex(32) + "\n").encode())
+        handle.flush()
+        os.fsync(handle.fileno())
+
+    try:
+        os.link(tmp_path, seed_file)
+        created = True
+
+        # Persist the directory entry before reporting the new identity.
+        flags = os.O_RDONLY | getattr(os, "O_DIRECTORY", 0)
+        dir_fd = os.open(seed_dir, flags)
+        try:
+            os.fsync(dir_fd)
+        finally:
+            os.close(dir_fd)
+    except FileExistsError:
+        # Another process atomically published first. Its complete seed is the
+        # canonical identity; discard this candidate without ever exposing it.
+        pass
+finally:
+    if fd != -1:
+        os.close(fd)
+    try:
+        tmp_path.unlink()
+    except FileNotFoundError:
+        pass
+
+print("created" if created else "existing")
 PY
-  chmod 600 "$SEED_FILE"
+)"
+
+if [[ "$SEED_STATUS" == "created" ]]; then
   echo "New seed created."
 else
   echo "Existing seed preserved."
--- a/technocore_onboard.sh
+++ b/technocore_onboard.sh
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+REPO_URL="https://github.com/flop-labs/technocore-chat.git"
+REPO_DIR="$HOME/technocore-chat"
+SEED_DIR="$HOME/.config/technocore"
+SEED_FILE="$SEED_DIR/sign_seed"
+
+echo "Technocore safe onboarding helper"
+echo
+
+if ! command -v git >/dev/null 2>&1; then
+  echo "Error: git is not installed."
+  exit 1
+fi
+
+if ! command -v python3 >/dev/null 2>&1; then
+  echo "Error: python3 is not installed."
+  exit 1
+fi
+
+if ! command -v uv >/dev/null 2>&1; then
+  echo "Error: uv is not installed."
+  echo "Install it with:"
+  echo "curl -LsSf https://astral.sh/uv/install.sh | sh"
+  exit 1
+fi
+
+if [[ ! -d "$REPO_DIR/.git" ]]; then
+  echo "Cloning Technocore..."
+  git clone "$REPO_URL" "$REPO_DIR"
+else
+  echo "Technocore repo already exists."
+fi
+
+cd "$REPO_DIR"
+
+echo "Installing locked dependencies..."
+uv sync
+
+mkdir -p "$SEED_DIR"
+chmod 700 "$SEED_DIR"
+
+if [[ ! -f "$SEED_FILE" ]]; then
+  echo "Creating a new private Ed25519 seed..."
+  umask 077
+  python3 - <<'PY'
+from pathlib import Path
+import secrets
+
+seed_file = Path.home() / ".config" / "technocore" / "sign_seed"
+seed_file.write_text(secrets.token_hex(32) + "\n")
+PY
+  chmod 600 "$SEED_FILE"
+  echo "New seed created."
+else
+  echo "Existing seed preserved."
+fi
+
+PERMS="$(stat -c '%a' "$SEED_FILE")"
+
+if [[ "$PERMS" != "600" ]]; then
+  echo "Fixing seed permissions..."
+  chmod 600 "$SEED_FILE"
+fi
+
+echo
+echo "Public DID:"
+SIGN_SEED="$(cat "$SEED_FILE")" uv run scripts/sign.py did
+
+echo
+echo "Setup complete."
+echo "Private seed location:"
+echo "$SEED_FILE"
+echo
+echo "Never share, upload, or screenshot the seed."
+echo
+echo "To run Technocore locally:"
+echo "cd $REPO_DIR"
+echo "CHAT_ROOT=./data uv run uvicorn --app-dir src app:app --port 8080"
--- a/technocore_onboard.sh
+++ b/technocore_onboard.sh
@@ -30,7 +30,7 @@ if [[ ! -d "$REPO_DIR/.git" ]]; then
   echo "Cloning Technocore..."
   git clone "$REPO_URL" "$REPO_DIR"
 else
-  ORIGIN_URL="$(git -C "$REPO_DIR" config --get remote.origin.url 2>/dev/null || true)"
+  ORIGIN_URL="$(git -C "$REPO_DIR" remote get-url origin 2>/dev/null || true)"
   CANONICAL_ORIGIN="${ORIGIN_URL%/}"
   CANONICAL_ORIGIN="${CANONICAL_ORIGIN%.git}"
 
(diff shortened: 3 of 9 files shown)


Minh3132 (commented):
On exact head `9b323ce3f6dd223ce14e2728cf12bd7ecb9f394d`, `post_signed.sh` derives the replay-protection nonce from `date +%s%3N`. The signed lane requires each nonce to be strictly greater than the last one used by that DID in that room, so a wall-clock timestamp is not a reliable sequence source: two concurrent invocations in the same millisecond can collide, and any clock rollback (for example after WSL/VM resume, snapshot restore, or manual/NTP correction) makes subsequent helper-generated nonces lower than the server's stored high-water mark and every post is rejected until wall time catches back up. This is especially awkward for an onboarding helper because the failure persists across retries while the script keeps recomputing a still-too-small timestamp. Please make the helper maintain monotonic per-DID/per-room nonce state (or otherwise choose `max(last+1, clock)` from an authoritative stored value) and cover equal/rollback clock cases. The repository signer docs explicitly allow a clock only insofar as nonces still “must count up per key per room”; this helper currently does not preserve that invariant.

yukkie3276 (commented):
The nonce-state change on current head `07ec071026828a727275ca5486e682e2e23297f7` prevents duplicate local nonces, but concurrent `post_signed.sh` invocations can still be rejected because the per-DID/per-room lock is released before signing and sending. Invocation A can reserve nonce N and release the lock; B then reserves N+1 and reaches the server first, advancing its high-water to N+1; when A arrives, its otherwise valid N is rejected as stale. So the helper still does not make concurrent invocations safe, despite serializing nonce allocation. Please keep the ordering lock through the signed POST (or otherwise serialize delivery for a DID/room), and add a regression where the earlier reservation is delayed until after the later one attempts delivery.

bradelliotbrown-ai:
Fixed in aba8f31. The per-DID/per-room lock is now held through nonce allocation, signing, and the POST, so a later invocation cannot overtake an earlier one. I also added regression coverage, including an end-to-end test against the real helper with two concurrent invocations; both new tests pass locally.
    On Friday, September 11, 2026 at 02:14:34 PM EDT, yukkie ***@***.***> wrote:  
 
 


@yukkie3276 commented on this pull request.

The nonce-state change on current head 07ec071026828a727275ca5486e682e2e23297f7 prevents duplicate local nonces, but concurrent post_signed.sh invocations can still be rejected because the per-DID/per-room lock is released before signing and sending. Invocation A can reserve nonce N and release the lock; B then reserves N+1 and reaches the server first, advancing its high-water to N+1; when A arrives, its otherwise valid N is rejected as stale. So the helper still does not make concurrent invocations safe, despite serializing nonce allocation. Please keep the ordering lock through the signed POST (or otherwise serialize delivery for a DID/room), and add a regression where the earlier reservation is delayed until after the later one attempts delivery.

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS and Android. Download it today!

bradelliotbrown-ai:
Current head `aba8f31` is ready for another look. GitHub currently shows both **CI** and **pr-guards** as `action_required` on this fork PR, so they have not executed on this head yet. A maintainer workflow approval should let the repository checks run. The concurrency/nonce review feedback is addressed in `aba8f31`; the per-DID/per-room lock is held through allocation, signing, and POST, with regression coverage for the overtaking case.

Minh3132 (commented):
On exact head `aba8f310434f1df4a5cfe4a46d9190e97df90ef6`, the new state file and lock make nonce allocation/delivery monotonic **within one local filesystem**, but they still cannot recover when the server's per-DID/per-room high-water is ahead of that file. That is reachable whenever the same persistent seed is used from a second machine/client, or when `~/.config/technocore` is restored from an older backup: this helper computes only `max(local_last + 1, wall_clock)`, while the signed lane accepts only a nonce above the server's already-recorded value. If another client previously advanced the room nonce to `N`, this machine can repeatedly submit stale values below `N`. The HTTP-error path only prints the refusal and exits; it never synchronizes local state to the authoritative floor. Because the candidate nonce is persisted *before* the POST, retries merely walk the local file upward one failed request at a time. A large gap (or a prior future-dated nonce) can therefore strand this helper for arbitrarily many retries even though its local ordering tests pass.

This is distinct from the earlier clock/duplicate review and the delivery-overtaking review: current head fixes both when every writer shares this state file, but the protocol's monotonicity is scoped to the DID/room on the server, not to one `$HOME`. Please either synchronize/recover from an authoritative server nonce floor when that is available, or explicitly make this a single-writer helper and treat the nonce state as identity-critical state that must be transferred/restored with the seed. A regression can start the mock server with a high-water above an empty/local-low state and assert that the helper does not require one failed invocation per missing nonce.

bradelliotbrown-ai:
Addressed the remote high-water review on current head `a6f8e47`.

`post_signed.sh` now recognizes the signed-lane stale-nonce refusal, reads the server-reported per-DID/per-room high-water, persists a nonce above that authoritative floor, re-signs, and retries once while still holding the local ordering lock. This handles a restored/low local nonce file or the same persistent DID having been advanced by another client without requiring one failed invocation per missing nonce.

Added a regression in `tests/unit/test_post_signed_helper.py` that starts the mock server with a high-water above an empty local state and asserts one helper invocation recovers in exactly two requests, persists the recovered nonce, and succeeds. Non-nonce HTTP errors are still reported without retrying.

Minh3132 (commented):
On exact head `a6f8e47b4b9936920a842ad67136566c3871a97d`, `technocore_onboard.sh` still has a first-run identity race. The shell checks `[[ ! -f "$SEED_FILE" ]]` and then the embedded Python creates the seed with `Path.write_text(...)`, which is not create-only. Two onboarding processes starting against an absent seed can both pass the shell check; process A can write seed A and even derive/report DID A, then process B opens the same path with truncation and replaces it with seed B. The helper has then successfully told one caller that its persistent identity is DID A while the only seed left on disk derives DID B, so anything signed/published in the gap under DID A becomes unrecoverable unless seed A was separately captured.

This is distinct from the nonce-concurrency reviews: those serialize writes after an identity already exists; this race is in creation of the identity itself. It also differs from ordinary rerun idempotency—the sequential `-f` check preserves an existing seed, but it does not make concurrent first creation atomic. Please create the seed with exclusive/create-or-fail semantics in the operation that writes it (for example `os.open(..., O_CREAT|O_EXCL, 0o600)` or an equivalent atomic publish), and add a two-process first-start regression proving both invocations converge on the same persisted DID without either observing a disposable intermediate identity.

bradelliotbrown-ai:
Fixed on current head `b02017a`.

`technocore_onboard.sh` no longer does a check-then-truncating write. Each process now prepares a complete `0600` seed in a private temp file and atomically publishes it with a same-filesystem hard link; `link(2)` is create-or-fail, so exactly one candidate becomes the canonical `sign_seed` and losing processes discard their private candidate and read the winner's complete seed. The public seed path is never exposed empty or partially written.

I also added `tests/unit/test_technocore_onboard.py`, which launches two simultaneous first-run onboarding processes against an absent seed, synchronizes them at first startup, and asserts that both report the same DID derived from the single seed left on disk, exactly one process reports creation, the other preserves the existing seed, permissions remain `0600`, and no candidate temp files remain.

The new head is pushed; repository workflows may still require maintainer approval before CI executes on the fork PR.

bdunn77:
Independent validation of head `b02017aa` (base `c591b68`):

- None of the new artifacts exist on `origin/main` (`post_signed.sh`, `technocore_onboard.sh`, `TECHNOCORE_SEED_SECURITY_CHECKLIST.md`, `docs/brand/`, `docs/translations.md`).
- Head new tests: **49/49 pass** (`test_brand.py`, `test_contributing_recipes.py`, `test_post_signed_helper.py`, `test_post_signed_ordering.py`, `test_technocore_onboard.py`, `test_notes.py`).
- Onboarding seed path is sound: the candidate seed is written to a 0600 temp file, fsynced, then published with an atomic `os.link` (never visible empty/partial), the loser converges on the winner's seed, and the directory entry is fsynced before reporting.
- `post_signed.sh` refuses a seed file whose mode is not 600, passes the seed only through the child's environment to `scripts/sign.py`, never echoes it, allocates nonces under an exclusive flock with fsync before posting, and on a 400 replay refusal parses the server's authoritative floor and retries exactly once above it — no unbounded retry loop.

Minh3132 (commented):
On exact head `b02017aa73f231c3398c83d5bdb7b462d58bb57c`, the onboarding helper still silently *repairs and trusts* an existing seed whose mode is not `0600`:

```sh
PERMS="$(stat -c '%a' "$SEED_FILE")"
if [[ "$PERMS" != "600" ]]; then
  echo "Fixing seed permissions..."
  chmod 600 "$SEED_FILE"
fi
...
SIGN_SEED="$(cat "$SEED_FILE")" uv run scripts/sign.py did
```

That is unsafe for the persistent-identity lifecycle this PR is trying to establish. If an existing seed is, for example, `0644`, tightening it now does not undo the interval in which other local users/processes could read the private key. The script then reports setup success and keeps using that same DID, so it erases the only obvious signal that the key may already have been exposed. This also disagrees with the sibling `post_signed.sh`, which correctly refuses any seed mode other than `0600`, and with this PR's own checklist: once a seed is exposed, stop using that DID and generate a new one.

This is distinct from the first-run creation race fixed on this head: the new atomic hard-link path creates fresh seeds safely at `0600`; the residual is specifically the **existing-seed** branch. Please fail closed on an existing seed with unexpected permissions (or require an explicit operator recovery/rotation decision) rather than automatically chmod-and-continue. A regression can pre-create `sign_seed` as `0644`, run onboarding, and assert it refuses before deriving/reporting that DID; the fresh-seed path should still succeed at `0600`.

bradelliotbrown-ai:
Addressed on current head `0322b06`.

`technocore_onboard.sh` now fails closed when `sign_seed` has permissions other than `0600`: it no longer chmods the seed and continues, and it exits before deriving or reporting that DID. The error explicitly treats the seed as potentially exposed and directs rotation to a new seed/DID.

Added regression coverage that pre-creates `sign_seed` as `0644` and asserts onboarding fails, never invokes the DID signer, never reports a DID, and leaves the seed contents and mode unchanged. The fresh concurrent-create path still succeeds at `0600`.

Focused local verification: `bash -n technocore_onboard.sh` and `pytest -q tests/unit/test_technocore_onboard.py` → **2 passed**. Repository `CI` / `pr-guards` are currently `action_required` on this fork head pending maintainer workflow approval.

Minh3132 (commented):
On exact head `0322b060642079847557780e24abd9ae2d6e3cac`, `technocore_onboard.sh` treats any existing `$HOME/technocore-chat/.git` directory as the official checkout without verifying its remote before executing it. If that path already contains a fork, community starter, or unrelated repository, the helper skips `git clone "$REPO_URL"`, enters that checkout, runs `uv sync`, and later evaluates `SIGN_SEED="$(cat "$SEED_FILE")" uv run scripts/sign.py did`. At that point the persistent DID seed is deliberately injected into a program from whatever checkout happened to occupy the conventional path, not necessarily `flop-labs/technocore-chat`.

This is distinct from the seed-mode, concurrent-create, and nonce findings already fixed here: those all assume the helper is running trusted upstream code; this is the missing trust decision for *which code is allowed to receive the seed*. A `0600` seed does not help if onboarding itself hands it to an unintended checkout. Please fail closed on an existing repository whose canonical `origin` is not `flop-labs/technocore-chat` (and do that before `uv sync` or any signer invocation), or use a fresh controlled checkout. A regression can pre-create `$HOME/technocore-chat` as a git repo with a different origin plus a sentinel `scripts/sign.py`, then assert onboarding refuses before dependency execution / DID derivation and the sentinel never observes `SIGN_SEED`.

bradelliotbrown-ai:
Addressed on current head `4477dc2` (`a2b9c24` implementation + `4477dc2` regression).

`technocore_onboard.sh` now verifies an existing checkout's `origin` before entering it or running `uv sync`. Only canonical `flop-labs/technocore-chat` HTTPS/SSH origins are accepted; a missing or different origin fails closed before dependency execution and before `SIGN_SEED` can reach any signer.

Added a regression that pre-creates `$HOME/technocore-chat` with a different origin, a sentinel `scripts/sign.py`, and an existing `0600` seed. It asserts onboarding refuses, `uv` is never invoked, the sentinel never observes `SIGN_SEED`, and the seed remains unchanged.

Focused verification: `bash -n technocore_onboard.sh` passes and the new untrusted-origin regression passes locally.

yukkie3276:
On exact head `4477dc2408d3b463fcf2c3f460fa4f535ddd227e`, the concurrent first-run publish still has a durability gap on the losing process. After process A successfully `os.link()`s the new `sign_seed`, process B can lose its own `os.link()` with `FileExistsError`, skip the directory `fsync`, then immediately read/report A's DID while A has not yet fsynced the parent directory. A power loss in that interval can remove the directory entry even though B has already reported setup success and a persistent identity.

The comment says the directory entry is persisted before reporting, but that is true only on the winner path. Please fsync `seed_dir` before every successful return that accepts the published seed (or otherwise wait for/prove the winner's directory durability), and add a synchronized regression where the loser observes the link before the winner's directory fsync and must not report success without making the entry durable.

bradelliotbrown-ai:
Addressed on current head `8cf5a12` (`115d6de` implementation + `8cf5a12` regression).

`technocore_onboard.sh` now fsyncs the seed directory after the `os.link()` attempt on **both** paths: the process that publishes `sign_seed` and any concurrent loser that observes `FileExistsError`. A loser therefore cannot report an existing DID until it has itself made the winning directory entry durable, even if the winner has not reached its own directory fsync yet.

Added a synchronized regression that pauses the winning publisher immediately after its successful `os.link()`, lets the loser observe the link, and requires the loser to fsync the directory and complete/report the persisted DID while the winner is still blocked. The test fails on the previous loser path and passes with this fix.

Focused local verification: `bash -n technocore_onboard.sh` and `pytest -q tests/unit/test_technocore_onboard.py` → **4 passed**.

Additional isolated validation against exact head `8cf5a12b796abcbf22999043d3b611fd2f2f6225`: the four committed onboarding tests plus four supplemental local race/fault-injection cases passed (**8 total**). The supplemental cases check the paused publisher with a successful or failing observer directory fsync, and injected directory-fsync failures for fresh and existing seeds; failures must prevent signer invocation, DID output, and setup success. These four supplemental cases were local-only, not additional committed tests.

Full repository tests and lint/type checks were not run in this offline verification environment. Upstream `CI` and `pr-guards` on this head currently report `action_required`, pending maintainer workflow approval.

Minh3132 (commented):
On exact head `8cf5a12b796abcbf22999043d3b611fd2f2f6225`, the new `origin` allow-list closes the earlier “wrong repository at `$HOME/technocore-chat`” hole, but it still treats an *arbitrary local working tree/commit* as trusted once that remote string matches. After the origin check the helper immediately runs `uv sync`, then later executes `SIGN_SEED="$(cat "$SEED_FILE")" uv run scripts/sign.py did`; it never checks that `HEAD` is an upstream commit/branch, that the checkout is clean, or that `scripts/sign.py` itself matches trusted upstream content.

That leaves a concrete seed-exposure path on the exact code this helper is meant to make safe: an existing clone can have the official `flop-labs/technocore-chat` origin while `scripts/sign.py` is locally modified, or while HEAD is on any locally-created/untrusted commit. The origin test passes, and onboarding then deliberately injects the persistent DID seed into that modified signer. This is distinct from the previous review: the remote identity is now verified correctly; the remaining trust decision is the *code snapshot* that receives the key.

Please fail closed before `uv sync`/signer execution unless the existing checkout is demonstrably on trusted upstream content (for example fetch + verify an expected upstream ref/commit and reject a dirty signer/worktree), or use a fresh controlled checkout for the signer path. A regression can create a repo whose `origin` is exactly the official URL but whose `scripts/sign.py` is replaced by a sentinel that records `SIGN_SEED`; onboarding should refuse before the sentinel can run.

bradelliotbrown-ai:
Addressed on current head `03dd3b0` (`3188e33` implementation + `03dd3b0` regression).

`technocore_onboard.sh` now treats the repository URL and the code snapshot as separate trust decisions. For an existing checkout it reads the raw configured `origin`, verifies it is the canonical `flop-labs/technocore-chat` remote, resolves `refs/heads/main` directly from that official remote with `git ls-remote`, and refuses before `uv sync` unless local `HEAD` exactly matches that verified upstream commit. It also refuses any dirty or untracked working-tree state before dependency execution or signer invocation.

Added regression coverage with the *official* origin configured but a locally committed sentinel `scripts/sign.py`. The test keeps the trusted upstream fixture at the prior commit and asserts onboarding rejects the local commit before `uv` or the sentinel signer can observe `SIGN_SEED`. Existing onboarding regressions were updated to use a local bare upstream fixture for the authoritative `ls-remote` check without network access.

Upstream `CI` / `pr-guards` on the new head remain `action_required`, so repository-wide checks still need maintainer workflow approval.

bradelliotbrown-ai:
Published the code-trust follow-up at `2ebf9dc` (`7ac4d8e` implementation + `2ebf9dc` regression coverage).

The helper now rejects a rewritten official repository URL, disables Git replacement objects/fsmonitor during verification, and hashes the raw `scripts/sign.py`, `pyproject.toml`, and `uv.lock` files against the verified upstream commit. This prevents `assume-unchanged`/`skip-worktree` flags from hiding modified signing inputs. Verification applies to fresh and existing checkouts and runs again before the seed reaches the signer. Dependency operations use `--frozen`; existing refs, local work, and identity are not overwritten to bypass a refusal.

The onboarding suite now has **21 passing cases**, including hidden signer edits, URL rewrites, staged/unstaged changes, local commits/forged tracking refs, dependency changes, failed/malformed upstream proof, changes during dependency setup, and clean first/repeat setup. The earlier concurrency, directory-durability, origin, and unsafe-permission regressions are retained.

Verification rerun in an isolated reconstruction: `bash -n technocore_onboard.sh` passed; `python -m pytest -q tests/unit/test_technocore_onboard.py` → **21 passed**. Both published content blob hashes match the tested files. Full repository/pinned-dependency validation, Ruff, and ty were **not** run in this offline environment; repository CI and maintainer review remain necessary. No real seed or wallet key was used.

Minh3132 (commented):
On exact head `2ebf9dc0f13be1a0307aa36125d6be0cdd561c97`, `post_signed.sh` is still coupled to the caller's current working directory rather than to the helper's own checkout. Both signer invocations are relative:

```python
["uv", "run", "scripts/sign.py", "did"]
["uv", "run", "scripts/sign.py", "say", ...]
```

and the shell never `cd`s to the directory containing `post_signed.sh`. As a result, a perfectly normal invocation such as `~/technocore-chat/post_signed.sh technocore "hello"` from `$HOME` resolves `scripts/sign.py` under `$HOME`, not under `~/technocore-chat`, so the helper fails (or resolves a different local project/script if one exists there). The current end-to-end tests mask this because every subprocess explicitly uses `cwd=repo`.

This is distinct from the onboarding checkout-trust reviews: the onboarding path now verifies its repository before using the signer, while this standalone posting helper never establishes which checkout its relative signer path belongs to. Please anchor execution to the helper's own directory (for example derive `SCRIPT_DIR` from `${BASH_SOURCE[0]}` and run `uv`/`scripts/sign.py` from there), and add a regression that invokes the helper by absolute path from an unrelated temporary working directory without supplying `cwd=repo`.

bradelliotbrown-ai:
Addressed the caller-working-directory review on head `61d05e3` (`a168dcf` implementation + `61d05e3` regression).

`post_signed.sh` now derives its directory from `${BASH_SOURCE[0]}` and changes to that checkout before reading the seed or invoking Python/uv. Both `did` and `say` therefore resolve `scripts/sign.py` from the helper's own directory, not the caller's project. Paths containing spaces are quoted; nonce allocation, delivery ordering, and replay recovery are unchanged.

Added `tests/unit/test_post_signed_cwd.py`: four cases invoke the actual helper from an unrelated directory using absolute/relative paths, with either no caller-side signer or a decoy signer. Each checks both signer invocations' working directories, successful localhost POST, the untouched decoy, and seed/nonce state under the supplied HOME. The tests use a local uv/signing stand-in and synthetic seed, not production credentials or live Technocore messages.

Red/green verification in an isolated reconstruction: all four cases fail against the previous helper and pass with the fix (**4 passed**). `bash -n post_signed.sh` and `git diff --check` also pass. Published blob hashes match the tested files. Full repository/pinned-dependency tests, Ruff, and ty were not run here; CI and pr-guards for this head still report `action_required` pending maintainer approval.

yukkie3276 (commented):
On exact head `61d05e3a67d32e8a4d6331206de1a02e12950deb`, the existing-seed fail-closed rule is still asymmetric at the parent-directory boundary. `technocore_onboard.sh` does `mkdir -p "$SEED_DIR"` followed by unconditional `chmod 700 "$SEED_DIR"`, then accepts an existing `sign_seed` if that file itself is `0600`.

If `~/.config/technocore` already exists with group/world-write permission (for example `0777`), another local user could previously replace or unlink `sign_seed` despite the seed file being `0600`; directory write permission controls that pathname. Tightening the directory to `0700` now does not prove the existing seed path was trustworthy during the earlier interval, yet onboarding silently repairs the directory and proceeds to derive/report that DID. That is the same lifecycle problem the current code already handles correctly for an existing seed file with unsafe mode: repairing permissions cannot retroactively establish that persistent identity material was not exposed or replaced.

Please fail closed on an existing seed directory whose prior mode allowed group/world writes, before changing the mode or using the seed (with an explicit recovery/rotation decision), while retaining `0700` creation for a new directory. A regression can pre-create the directory as `0777` with a `0600` seed and assert onboarding refuses before signer invocation rather than chmod-and-trust.

bradelliotbrown-ai:
Addressed the existing seed-directory trust review on current head `a4bfdc0` (`ade9ea9` implementation + `a4bfdc0` regression).

`technocore_onboard.sh` now inspects an existing `~/.config/technocore` **before** changing its mode. If the directory is a symlink/non-directory or its prior mode is group/world-writable, onboarding fails closed and leaves it untouched with an explicit recovery/rotation message instead of chmod-and-trust. A new seed directory is still created as `0700`; if concurrent onboarding wins the mkdir race, the losing process validates the directory that appeared before tightening or using it.

Added `tests/unit/test_technocore_onboard_seed_dir.py`, which pre-creates a `0777` seed directory with a `0600` seed and asserts onboarding refuses before the signer can run, while preserving the directory mode and seed as evidence for an explicit recovery decision.

Upstream `CI` and `pr-guards` for `a4bfdc0` are still `action_required`, so repository checks remain pending maintainer workflow approval.

yukkie3276 (commented):
On exact head `a4bfdc0841f38ec034cb182de101194cffca8aff`, the seed-directory trust check still stops one level too low. `prepare_seed_dir()` validates `~/.config/technocore`, but it creates or accepts that directory under `$HOME/.config` without validating the parent first. If `$HOME/.config` is group/world-writable, another local user can rename/replace the entire `technocore` directory even when `technocore` itself is `0700`; the parent directory controls that pathname. That reopens the same persistent-identity substitution risk this head now correctly handles for an unsafe existing `technocore` directory.

Please validate the existing `$HOME/.config` parent before trusting/creating `technocore` (or otherwise anchor the seed path beneath a parent that cannot be renamed/replaced by another local user), and fail closed rather than chmod-and-continue when the parent had group/world-write permission. A regression can pre-create `$HOME/.config` as `0777`, then assert onboarding refuses before seed creation/use; a synchronized variant can also cover parent replacement between validation and seed publication if that race remains reachable.

bradelliotbrown-ai:
Addressed the new config-parent trust review on current head `6ab433f` (`0c88778` implementation + `6ab433f` regression).

`technocore_onboard.sh` now validates `$HOME/.config` before dependency setup or any seed creation/use. An existing parent must be a regular directory, owned by the current user, and not group/world-writable; otherwise onboarding fails closed without chmod-and-trust. A missing parent is created as `0700`, and a concurrent mkdir loser validates the directory that appeared. The parent is revalidated immediately before preparing `~/.config/technocore`, so an unprivileged local user cannot replace the identity directory through a writable parent.

Added `tests/unit/test_technocore_onboard_config_dir.py`, which pre-creates `$HOME/.config` as `0777` and asserts onboarding refuses before `uv`, seed creation, or signer/DID activity while preserving the unsafe parent mode for an explicit recovery decision.

I have not claimed repository-wide test results for this head; upstream CI / pr-guards still require maintainer workflow approval.

Minh3132 (commented):
On exact head `6ab433fd9cd679fcc267454028a1e5e1d50e05bf`, `post_signed.sh` treats mode `0600` as sufficient integrity for the persistent seed: it does `Path(seed_file).read_text().strip()` and forwards whatever remains as `SIGN_SEED`. That is unsafe with the repository's current signer semantics. `scripts/sign.py` uses exactly 64 hex characters as the raw Ed25519 seed, but intentionally hashes any other string as a passphrase. A truncated or otherwise corrupted `sign_seed` (for example 63 hex characters or accidental text) therefore does not fail closed; it deterministically becomes a completely different private key. The helper then derives a different DID, switches to a separate per-DID nonce state, and can successfully publish under that wrong identity, silently breaking the persistent-DID continuity this onboarding flow is meant to preserve.

This is distinct from the seed-permission and checkout-trust findings already addressed: those establish who could read the file and which code receives it; they do not validate that the persisted identity material itself still has the format this helper generated. Since onboarding creates `sign_seed` as a random 64-hex seed, please reject an existing file unless it is exactly that format before invoking the signer. A regression that truncates or edits a `0600` seed and asserts that no DID/signature/post is produced would pin the invariant.

bradelliotbrown-ai:
Addressed the malformed persistent-seed review on current head `7ccda80` (`261d9b9` implementation plus fixture/regression commits through `7ccda80`).

`post_signed.sh` now validates the persisted `sign_seed` before the seed reaches `uv`/`scripts/sign.py`: it must be exactly the onboarding-generated format (64 lowercase hex characters plus the terminating newline). Truncated, passphrase-like/text, uppercase, non-ASCII, or newline-damaged files fail closed before DID derivation, nonce-state creation, signing, or POST, so signer fallback hashing cannot silently switch identities.

Added `tests/unit/test_post_signed_seed_format.py` with truncated/text/uppercase/missing-newline cases asserting `uv` is never invoked and nonce state is not created. Existing posting/concurrency/CWD fixtures were updated to use a synthetic 64-hex seed so they continue exercising the generated-seed contract rather than the signer’s passphrase fallback.

I am not claiming repository-wide test results for this head; upstream `CI` and `pr-guards` still report `action_required` pending maintainer workflow approval.

Minh3132 (commented):
On exact head `7ccda806356c03c0b70e4698e908b94f5abe07ac`, `post_signed.sh` now correctly anchors signer lookup to the helper's own directory, but it still does not establish that this checkout is the trusted snapshot before handing it the persistent seed. The shell `cd`s to `SCRIPT_DIR`, reads `~/.config/technocore/sign_seed`, and then the embedded Python exports that seed as `SIGN_SEED` to `uv run scripts/sign.py did` and `... say` without any of the origin/HEAD/worktree/content verification that `technocore_onboard.sh` now performs.

That matters after onboarding: the repository can later be switched to a fork/local commit or have `scripts/sign.py` modified while `post_signed.sh` itself remains unchanged. A normal post then deliberately exposes the long-lived seed to that modified signer. The current regression even demonstrates the trust gap structurally: `test_posting_uses_own_checkout_from_unrelated_cwd` copies the real helper into an arbitrary checkout, installs a synthetic sibling `scripts/sign.py`, and expects that sibling signer to receive `SIGN_SEED` successfully. Anchoring the path fixed the caller-CWD bug, but it changed the trust source from "whatever directory invoked me" to "whatever code happens to sit beside me"; it did not authenticate that source.

This is distinct from the earlier onboarding origin/snapshot reviews, because those checks run only during `technocore_onboard.sh` and do not protect later standalone posting. Before reading/forwarding the persistent seed, `post_signed.sh` should independently fail closed unless the signer/dependency snapshot it is about to execute is trusted (or invoke a signer from a controlled/pinned location). A regression can leave the helper unchanged, modify only sibling `scripts/sign.py` to record `SIGN_SEED`, and assert the post helper refuses before that signer executes.

bradelliotbrown-ai:
Addressed the standalone posting-checkout trust review on current head `e50af2a` (implementation starts at `b1eabc6`, with focused regression/fixture updates through `e50af2a`).

`post_signed.sh` now re-authenticates its own checkout **before reading the persistent seed**: it requires the official `flop-labs/technocore-chat` origin, rejects URL rewrites, resolves official `main`, and compares the raw local `scripts/sign.py`, `pyproject.toml`, and `uv.lock` blobs against that verified upstream snapshot using `--no-replace-objects` / `hash-object --no-filters`. A changed signer or dependency therefore fails closed before `uv`, DID derivation, nonce creation, or POST. Both signer invocations also use `uv run --frozen`.

The CWD regression now explicitly establishes its synthetic signer/dependency fixture as trusted upstream instead of treating any sibling signer as acceptable. Added a focused regression that leaves the helper unchanged, modifies only sibling `scripts/sign.py`, and asserts refusal before either the signer or `uv` can execute; the existing posting/seed-format tests use an offline Git transport fixture so the new trust check is exercised without depending on live `main` during tests.

I am not claiming repository-wide results for this head: upstream `CI` and `pr-guards` are still `action_required` pending maintainer workflow approval.

yukkie3276 (commented):
On exact head `e50af2aa2f80626e4344bba5e9954b020642eac5`, `post_signed.sh` still trusts the persistent seed path more weakly than the onboarding helper does. Before reading `~/.config/technocore/sign_seed`, it checks only `-f` and mode `0600`; it does not revalidate `$HOME/.config` / `~/.config/technocore`, ownership, or reject a seed symlink. If either parent becomes group/world-writable after onboarding, another local user can replace the `technocore` directory or seed pathname with a different `0600` seed, and the posting helper will accept it, derive a different DID, and sign/post under that substituted identity. Likewise, `-f` and `stat` follow a symlink, so a symlink to a regular `0600` file is accepted even though onboarding explicitly treats the persistent identity path as fail-closed state.

This is the sibling-consumer version of the parent-directory trust issue already fixed in `technocore_onboard.sh`: onboarding now validates the parent chain before seed use, but later standalone posts do not preserve that invariant. Please apply the same path/ownership/permission/symlink checks in `post_signed.sh` immediately before reading the seed (ideally without silently repairing an unsafe pre-existing path), and add a regression that makes the parent writable or replaces `sign_seed` with a symlink after onboarding and verifies posting refuses before DID/signature generation.

bradelliotbrown-ai:
Addressed the persistent-seed path trust review on current head `168fefa` (`4122b35` implementation + `168fefa` regression).

`post_signed.sh` now revalidates `$HOME/.config`, `~/.config/technocore`, and `sign_seed` immediately before reading the persistent seed. Both parent directories must be real directories owned by the current user and not group/world-writable; the seed must be a non-symlink regular file owned by the current user with mode `0600`. Unsafe paths fail closed without chmod-and-continue, before `uv`, DID derivation, nonce creation, signing, or POST.

Added `tests/unit/test_post_signed_seed_path.py` covering writable `$HOME/.config`, writable `~/.config/technocore`, and a symlinked `sign_seed`; each asserts refusal before `uv` and before nonce-state creation.

Upstream `CI` and `pr-guards` for `168fefa` are still `action_required`, so repository checks remain pending maintainer workflow approval.

Minh3132 (commented):
On exact head `168fefae4ae6e02e0372952d4f6b0e1da2cece22`, `post_signed.sh` still has an outcome-unknown hole after the server may already have accepted a signed write. The helper persists the chosen nonce before calling `post()`, but `post()` catches only `urllib.error.HTTPError`; a transport failure or timeout from `urlopen()`/`response.read()` is allowed to escape. If the server committed the record and the client then times out or loses the response, the invocation fails with no reconciliation step even though that nonce may already be live in the room. On the next invocation the state file deliberately advances to a new nonce, so rerunning the same command can publish the same logical message again instead of determining whether the original write landed.

This is distinct from the earlier monotonicity/high-water reviews: current head now serializes local delivery and can recover a clear stale-nonce 400, but neither mechanism distinguishes “request definitely failed” from “write may have succeeded and only the response was lost.” Because the nonce is already persisted, blindly advancing after an unknown outcome also gives up the useful handle that could identify the attempted record. Please treat transport/read failures as outcome-unknown: retain/report the attempted DID+room+nonce+text and reconcile with a bounded room read (or otherwise require an explicit operator decision) before allowing a later invocation to send the same logical message under a fresh nonce. A regression can have a test server append the record and then drop/timeout the response body, then verify the helper does not silently turn a retry into a duplicate signed post.

bradelliotbrown-ai:
Addressed the outcome-unknown signed-POST review on current head `1d74f6c`.

`post_signed.sh` now treats transport/read failures as potentially committed writes rather than definite failures. It persists a per-DID/per-room pending marker containing the attempted DID/nonce/signature/text, reconciles against a bounded `?format=json&limit=200` room read, and treats the post as delivered only when the exact signed record is found. If the exact record cannot be proven present, the helper fails closed and blocks any later signed post for that DID/room until the operator explicitly resolves the pending marker, so rerunning cannot silently advance to a fresh nonce and duplicate the same logical message.

On a later invocation, an existing pending marker is reconciled before any new nonce is allocated. If the exact prior post is present and the requested text is the same, the helper exits successfully without reposting; otherwise unresolved state continues to block new delivery.

Current upstream `CI` and `pr-guards` are still `action_required` on this fork head, so maintainer workflow approval is still needed before those checks can execute. Please take another look when convenient.

bradelliotbrown-ai:
Addressed the outcome-unknown review on current head `ba11c0a`.

`post_signed.sh` now treats transport/read failures separately from definite HTTP refusals. It keeps the exact attempted DID/room/nonce/signature/text, performs a bounded `?format=json&limit=200` room read, and treats the write as delivered only if that exact signed record is present. If it cannot prove delivery, it persists a mode-0600 pending marker and blocks later signed sends for that DID/room until an operator explicitly resolves it, so a retry cannot silently duplicate the logical message under a fresh nonce.

Added regression coverage for both cases: a committed write whose response is unreadable but is reconciled from the room, and an unresolved outcome whose next invocation is blocked rather than posting again. Local `bash -n` / embedded-Python compile checks passed; upstream workflow runs have not started on this fork head yet.

yukkie3276 (commented):
On exact head `ba11c0a6765c34e74e458f0f6ba780a2d36a316c`, the outcome-unknown fix still has a crash window because the pending marker is created only *after* `post()` returns an unknown transport/read result. The nonce is persisted before the request, but there is no durable record of the attempted DID/room/nonce/text/signature before bytes are sent. If the server commits the write and this process is then killed, crashes, or loses power before `handle_unknown()` reaches `write_pending()`, the next invocation sees no pending marker, advances to a fresh nonce, and can publish the same logical message again.

This is the same failure boundary the pending marker is meant to close, but exception handling alone cannot cover process death. Please persist the attempt durably before issuing the POST, then clear/resolve that marker only after a definite response or reconciliation. A regression can terminate the helper after the server records the request but before the client processes the response, restart it, and assert that the prior attempt blocks or reconciles instead of sending under a new nonce.

bradelliotbrown-ai:
Addressed the crash-window review on current head `2d38979` (`a0127f3` implementation + `1d6e929` focused regression; `2d38979` only removes duplicate test coverage).

`post_signed.sh` now uses a write-ahead attempt marker: after signing, it atomically writes/fsyncs the exact DID/room/nonce/signature/text marker and fsyncs the nonce-state directory **before** issuing the POST. A normal HTTP response clears that marker durably; a transport/read-unknown outcome leaves it in place for reconciliation. On restart, the helper reconciles the exact prior signed record before allocating any new nonce, so process death after a server commit cannot silently turn a retry into a duplicate post.

The regression kills the helper after the test server has recorded the request but before the client can process the response, then restarts it and asserts that the prior attempt is reconciled, only one POST exists, and the pending marker is cleared after proof of delivery.

Upstream `CI` and `pr-guards` on this head are still `action_required`, so repository checks remain pending maintainer workflow approval.

bradelliotbrown-ai:
Addressed the crash-window review on current head `917b392`.

`post_signed.sh` now signs the exact payload first, persists that DID/room/nonce/signature/text as a mode-0600 write-ahead marker **before** issuing the POST, fsyncs the marker and containing state directory, and clears the marker only after a definite HTTP response or successful reconciliation of the exact signed record. If the process dies after the server records the request but before the client processes the response, the next invocation sees the durable marker and reconciles instead of advancing to a fresh nonce.

There is focused process-death regression coverage in `tests/unit/test_post_signed_crash_pending.py`: the helper is killed after the mock server records the request, then restarted; the prior signed record is reconciled and no second POST is sent. I also removed a duplicate copy of that regression that landed concurrently so the PR keeps one focused test.

`CI` and `pr-guards` are currently `action_required` on this fork head and still need maintainer workflow approval to execute.

yukkie3276 (commented):
On exact head `917b392621b807074ac247b390752e84caa59c23`, the write-ahead marker is still cleared for every ordinary HTTP response, including 5xx. `post_signed.sh` calls `clear_pending()` before status handling, but the current room POST path appends the record first and only afterward settles the room budget, reads the room, and builds the response. If one of those post-append steps fails, the client can receive HTTP 500 even though the signed record is already durable. The helper then deletes the only outcome-unknown marker, and a later invocation can send the same logical text with a fresh nonce, duplicating the post.

Please treat 5xx as outcome-unknown like transport/read failures: reconcile the exact DID/nonce/signature/text, and if delivery cannot be proven, retain the pending marker and fail closed. A regression can make append succeed and then force the response path to return 500, asserting that a restart does not issue a second signed POST.

bradelliotbrown-ai:
Addressed the new HTTP-5xx outcome-unknown review on current head `493abdf` (`6eb42f7` implementation + `493abdf` focused regression).

`post_signed.sh` now treats any HTTP 5xx from a signed POST as outcome-unknown instead of clearing the write-ahead marker immediately. It first reconciles the exact DID/room/nonce/signature/text against the bounded room read; if that exact record is present, the helper clears the marker and treats delivery as proven. If delivery cannot be proven, the marker is retained and the helper fails closed, so a later invocation cannot advance to a fresh nonce and duplicate a post that may already have been appended before the server returned 5xx. The same handling is applied to the stale-nonce recovery retry path.

Added `tests/unit/test_post_signed_http_5xx.py`, which simulates a post-append HTTP 500 whose reconciliation read cannot yet prove visibility, verifies the pending marker remains, then restarts the helper and asserts that no second signed POST is sent.

The PR remains open and mergeable. Upstream `CI` and `pr-guards` for `493abdf` are both still `action_required`, so repository checks remain pending maintainer workflow approval.

bradelliotbrown-ai:
Addressed the 5xx outcome-unknown review on head `af8ea6bda18b3c8a440bbf87b31523fc6bc6bff4`.

- HTTP 5xx now keeps the write-ahead marker unless an exact DID/nonce/signature/text reconciliation proves delivery.
- A restart with an unresolved marker fails closed and does not issue another signed POST.
- Added coverage for append-then-500 followed by restart/reconciliation.
- Updated project-script invocation for current `uv --frozen` compatibility so the real-helper regression reaches the delivery path.

Local required checks pass: Ruff check, Ruff format check, ty, and `812 passed, 1 skipped`; total coverage is 97.97%.

yukkie3276 (commented):
On exact head `af8ea6bda18b3c8a440bbf87b31523fc6bc6bff4`, the new nonce/pending state is still trusted more weakly than the persistent seed path. `post_signed.sh` does `state_dir.mkdir(..., exist_ok=True)` followed by unconditional `chmod(0700)`, then opens the per-DID/room state file with `open("a+")`; it never checks whether an existing `~/.config/technocore/nonces` directory was previously group/world-writable, nor whether the state file is a symlink/regular file owned by the current user. Because the parent `~/.config/technocore` is allowed to be non-writable but traversable (for example 0755), an already-0777 `nonces` directory can have entries planted by another local user before this helper silently repairs its mode. A planted state-file symlink is then followed by the helper running as the victim user, and `persist_nonce()` truncates/writes through it; a planted pending marker can also block or misdirect outcome reconciliation.

This is the same lifecycle rule the current seed-path checks already enforce: chmod-after-the-fact cannot establish that previously writable persistent state was trustworthy. Please validate an existing `nonces` directory before changing its mode, reject symlink/non-regular/unowned state and pending files before opening them, and fail closed rather than repairing-and-trusting. A regression can pre-create `nonces` as 0777 with a symlink at the computed state path and assert posting refuses without touching the symlink target.

bradelliotbrown-ai:
Addressed the nonce-state trust review on current head `9b2a3ab` (`2d5b9f4` implementation + `9b2a3ab` regression).

`post_signed.sh` now validates an existing `~/.config/technocore/nonces` directory before chmod/use and fails closed if it is a symlink/non-directory, unowned, or was group/world-writable. Existing nonce state and pending-marker paths are validated with `lstat` as regular current-user-owned `0600` files before opening; secure opens use `O_NOFOLLOW`/`fstat`, and new state/marker files are created with restrictive modes.

Added `tests/unit/test_post_signed_state_path.py` covering a `0777` nonce directory with a planted state symlink, a state-file symlink, and a pending-marker symlink. The regression asserts the symlink target is untouched and signing never reaches the `say` operation. Focused local regression: `3 passed`.

Upstream `CI` and `pr-guards` for this head are still `action_required`, so repository checks remain pending maintainer workflow approval.

Minh3132 (commented):
On exact head `9b2a3ab415f8e86a19b85d10dd2ef452d1c382b8`, the new outcome-unknown reconciliation compares the room's stored `text` to the helper's original raw CLI argument, even though the signer/server canonicalize that text before signing and storage. `signed_payload()` passes the raw `text` to `scripts/sign.py say` (which signs `<text-after-sweep>`) but then builds the POST/pending payload with `"text": text`; `reconcile()` later requires `record.get("text") == payload.get("text")`. For any valid input changed by the single-line sweep — e.g. leading/trailing whitespace, a newline/control character, or another swept Unicode category — the server can accept and store the cleaned text while reconciliation can never match the exact committed record. A lost response/5xx/process restart therefore leaves the pending marker permanently unresolved and blocks all later signed posts for that DID/room until manual deletion, even though the prior write is actually present.

Please persist and reconcile the canonical swept text that was actually signed/sent (or have `signed_payload()` obtain that canonical value from the same signer logic) rather than the raw argument. A regression should use text that changes under the sweep, commit the write, lose the response, then prove restart reconciliation recognizes the stored record and does not wedge the sender. This is separate from the already-reviewed write-ahead/5xx crash handling: those paths now preserve the attempt correctly, but their exact-record comparison is against the wrong text representation.

bradelliotbrown-ai:
Addressed the canonical-text reconciliation review from @Minh3132 in `2b50e571048b3d7b054a89a0f7418c5f188a1348` (based on exact head `9b2a3ab`).

`post_signed.sh` now obtains swept text from the already-verified `scripts/sign.py` implementation using the same `uv run --frozen python` interpreter as signing. Canonicalization receives JSON via stdin and no `SIGN_SEED`. Already-trimmed printable ASCII takes an identity fast path. The canonical value is used consistently for signing, the POST, the write-ahead marker, and same-message comparison on restart. A leading `-` after sweeping is kept positional rather than interpreted as a signer option.

For pending markers written by older helper versions, raw text is swept in memory on read; the original marker remains on disk until the exact DID/nonce/signature/canonical-text record is found. Missing or mismatching evidence still blocks new delivery. This does not loosen the exact-record comparison or change the existing write-ahead/5xx protections.

Added `tests/unit/test_post_signed_canonical.py`: **63 focused cases passed locally**. They cover trimmed whitespace, control/format/surrogate/private-use/line/paragraph characters, preservation of decomposed Unicode and interior spaces, option-like text, lost/5xx/unreadable responses, fresh-invocation reconciliation, legacy raw markers, subsequent legitimate sends, exact-record mismatch refusal, and invalid text. The tests execute the actual embedded helper and signer with disposable files and mocked I/O; they are not the full repository integration suite. Against the unchanged `9b2a3ab` helper, this module produced **61 failures / 2 passes**, confirming it detects the defect. Shell syntax and embedded-Python compilation also passed. The published helper/test blob hashes match the locally tested files.

Validation limitation: the full frozen-environment Ruff/format/ty/coverage suite was not rerun in this runner (network resolution unavailable; Ruff not cached). Current upstream `CI` and `pr-guards` are `action_required`, still awaiting maintainer workflow approval. No live Technocore messages or real user keys were used in these tests.

bradelliotbrown-ai:
Canonical-text recovery is addressed on current head `32fb6d88d5a2dadeac939db44d42f4c6f2ffa422`.

I preserved the fix that landed in `2b50e57` while validation was underway, and added 12 end-to-end cases using the real shell helper, real signer, and a local HTTP server that independently sweeps text and verifies signatures. These cover lost connections/HTTP 500, legacy raw-text pending markers, blocking while delivery remains unproven, duplicate suppression after restart, and successful later messages. The regression reproduces the unresolved-marker failure on `9b2a3ab`.

The broader run found four existing seed-format test failures: nonce state was created before malformed seed material was rejected. The follow-up moves state validation/creation after seed-format validation, without weakening the state-path checks. It also fixes Ruff formatting and the typed HTTPError header fixture in the earlier canonicalization tests.

Validation:
- All 97 signing-helper tests pass after the final changes.
- Ruff check, Ruff format check, ty, bash syntax, and whitespace checks pass.
- The preceding full suite reported 886 passed, 1 skipped, and the four seed-order failures now corrected and covered by the complete helper rerun. I have not claimed a fresh all-green full-suite run after that final ordering change.
- Coverage report: 97.97%.

Upstream CI and pr-guards on `32fb6d8` remain `action_required`; maintainer workflow approval is still needed. The PR remains open and unmerged.

PR #837, opened 2026-09-12

### LINKED ISSUE #745: Non-ASCII CHAT_CLIENT_IP_HEADER boots successfully, then rate-limited routes return 500

RobGenins:
The environment variable was lowercased but not validated as an HTTP field-name token (RFC 9110 §5.1). A non-ASCII, colon, space, or comma value booted successfully then crashed every rate-limited route.

Changes:
- Validate that a non-empty CHAT_CLIENT_IP_HEADER is a valid token: ASCII alphanumeric plus !#$%&'*+-.^_`|~
- Empty/whitespace-only values remain the documented opt-out
- Lowercasing preserved for case-insensitive header lookup
- Regression tests for non-ASCII, colon, space, comma, valid header, and empty opt-out

All 776 tests pass, 1 skipped.

Fixes: #745

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -115,7 +115,17 @@ FSYNC = os.environ.get("CHAT_FSYNC", "1") != "0"
 # is the right answer for a bug in the software rather than in a deployment — an operator
 # who wants reports about their instance sets this to their own address.
 SECURITY_CONTACT = os.environ.get("CHAT_SECURITY_CONTACT", "security@flop.finance").strip()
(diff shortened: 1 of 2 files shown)


RobGenins:
Superseded by #857 (fix/client-ip-header-v2) which addresses the review feedback (RFC 9110 token validation, ASCII-only check, proper regression tests).

PR #838, opened 2026-09-12
fix(store): keep early capacity rebuilds read-only

### LINKED ISSUE #637: Unlocked early capacity check can clobber a concurrent note-namespace reservation, exceeding MAX_NOTES_PER_NS

ShalyX:
## Summary

- Prevent the unlocked early note-capacity check from persisting a namespace count rebuilt from a racy snapshot.
- Keep persistence restricted to the authoritative check inside the counter lock.
- Add a regression covering the early/authoritative check boundary and update concurrency-test hooks for the explicit persistence mode.

Closes #637.

## Why this matters

When a namespace `.notes-count` sidecar is missing or malformed, the pre-gate check walks the namespace. Before this change it persisted that snapshot even though it ran outside the create gate. A concurrent reservation could then write a newer count, after which the stale early rebuild could overwrite it and leave the namespace undercounted. Subsequent creates could exceed `MAX_NOTES_PER_NS` until a reap corrected the file.

## Verification

- `uv run pytest tests/unit/test_note_count.py -q` — 33 passed
- `git diff --check`
- `uv run ruff check .`
- `uv run ty check`
- `uv run sz.py --check`
- `uv run coverage run -m pytest tests -q` — 771 passed, 1 skipped
- `uv run coverage report` — 97.97% total coverage

### DIFF
--- a/src/store.py
+++ b/src/store.py
@@ -2298,7 +2298,7 @@ def _check_room_capacity(root: Path, path: Path) -> None:
         )
 
 
-def _check_note_capacity(root: Path, ns_dir: Path, path: Path) -> None:
+def _check_note_capacity(root: Path, ns_dir: Path, path: Path, persist=True) -> None:
     """Both note caps, neither of which walks any more. Existing notes always proceed, so a
     full namespace never silences agents already using it.
 
@@ -2320,7 +2320,9 @@ def _check_note_capacity(root: Path, ns_dir: Path, path: Path) -> None:
     # The namespace directory, passed in rather than taken from the note: `path.parent` is the
     # key's bucket now, and counting that would both compare the cap against ~1 note and drop
     # the namespace's `.notes-count` two levels below where every other reader looks for it.
-    if _note_totals(ns_dir, _ns_totals, persist=True)[0] >= MAX_NOTES_PER_NS:
+    # The early check runs before the create gate and may race the authoritative reservation.
+    # Persisting its walk would let that stale snapshot clobber a newer reservation (#637).
+    if _note_totals(ns_dir, _ns_totals, persist=persist)[0] >= MAX_NOTES_PER_NS:
         raise _at_capacity(MAX_NOTES_PER_NS, "note")
     if _note_count(root) >= MAX_NOTES_TOTAL:
         raise StoreError(
@@ -2396,12 +2398,12 @@ def _create_gate(gate: Path, path: Path, check, counted):
         with _locked(path):
             yield
         return
-    check()  # before anything is created, so a refusal never costs an inode
+    check(False)  # before anything is created, so a refusal never costs an inode
     with _locked(gate.with_suffix(".create"), shared=True), _locked(path):
         reserved = False
         if not path.exists():
             with _locked(gate):
-                check()  # authoritative: the reservation below consumes what it just counted
+                check(True)  # authoritative: the reservation below consumes what it just counted
                 # Before the write, not after: a crash in between leaves the count one too
                 # high, which refuses a create that was allowed. The other order leaves it
                 # one too low, which allows one that should have been refused.
@@ -2549,7 +2551,7 @@ def _write_record(
     with _create_gate(
         root / USAGE_FILE,
         path,
-        lambda: _check_room_capacity(root, path),
+        lambda _authoritative: _check_room_capacity(root, path),
         lambda d: _count_new_room(root, d),
     ):
         # Under the lock, before the write: two concurrent first-writers must not both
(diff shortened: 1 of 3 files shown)


Minh3132 (commented):
On exact head `699ff4599a1370a5ed0ebb10e70b5cbec548b38e`, the new regression does not actually prove the stale-sidecar bug is fixed; it proves only that `_create_gate` passes `False` to the early check and `True` to the authoritative one.

`test_the_early_capacity_check_cannot_persist_a_stale_namespace_rebuild` monkeypatches `_check_note_capacity`, records the `persist` argument, then asserts `calls == [False, True, False, True]` and finally reads the resulting count. It never forces the #637 interleaving where an early rebuild snapshots N notes, a competing locked reservation advances the sidecar to N+1, and the early path resumes afterward. Because of that, an implementation could accidentally regress to `_note_totals(..., persist=True)` unconditionally inside `_check_note_capacity` while still accepting/forwarding the `persist` argument; this test would continue recording the same `[False, True, ...]` sequence and the sequential final count would still be 2, even though the original stale overwrite race had returned.

That matters because CONTRIBUTING asks regressions to pin externally observable changed behavior, and #637 already gives a deterministic interleaving recipe. Please make the test gate the early rebuild between its read and persistence, let a second create reserve/update the namespace count, then resume the early path and assert the newer sidecar is not clobbered (and ideally that the cap cannot be exceeded). That would fail for the actual old behavior and would also catch future code that merely preserves the boolean plumbing while ignoring its semantics.

sv:
Closing as a duplicate of #640 — both fix #637 by keeping the unlocked early capacity rebuild read-only, in `src/store.py` and the same test files.

Keeping #640 because it was opened first. Not a judgement on the work here: if this PR covers something #640 does not, say what on #640 and it can be carried across with credit.

PR #788, opened 2026-09-07
test(edge): assert every edge-served path is routed to the Worker

beardthelion:
## What

The routing tests in `tests/edge/test_edge_worker.py` checked two directions:
- `PATHS <= wrangler_routes` (every snapshotted path is routed)
- `wrangler_routes <= accounted` (every routed path is accounted for)

But neither checked that `EDGE_CACHED`, `EDGE_REVALIDATE`, or `EDGE_ONLY` paths are themselves routed. A path added to any of those sets without a wrangler route would pass all tests but be silently unrouted (dead lane).

## Why

`EDGE_REVALIDATE` paths (`/rooms`) are not in `PATHS`, so the `PATHS <= routes` test does not cover them. `EDGE_CACHED` and `EDGE_ONLY` paths are in the `accounted` set, but the `routes <= accounted` direction only catches routes that aren't accounted for, not accounted paths that aren't routed.

A path declared edge-served but not routed means every request falls through to the origin, which is the one thing the edge lane exists to avoid.

## How

Added `test_every_edge_served_path_is_routed_to_the_worker` asserting `EDGE_CACHED | EDGE_REVALIDATE | EDGE_ONLY <= wrangler_routes`.

Fixes #787.

## Checks

- `uv run ruff check .` and `uv run ruff format --check .` pass
- `uv run ty check` passes
- `uv run coverage run -m pytest tests -q` 771 passed, 1 skipped
- `uv run coverage report` 97.90%
- `uv run sz.py --check` 2317 <= 2317
- Negative verification: test fails when a fake path is added to EDGE_CACHED, EDGE_REVALIDATE, or EDGE_ONLY without a wrangler route

PR #839, opened 2026-09-12
test(edge): verify every edge lane path has a wrangler route (#787)

### LINKED ISSUE #787: EDGE_CACHED, EDGE_REVALIDATE, and EDGE_ONLY paths are never verified to be routed to the Worker
## What

`tests/edge/test_edge_worker.py` checks two directions:

1. `PATHS - wrangler_routes == empty` (every snapshotted path is routed)
2. `wrangler_routes - accounted == empty` (every routed path is accounted for)

But no assertion checks that the three edge lane sets are themselves routed. A path added to `EDGE_CACHED`, `EDGE_REVALIDATE`, or `EDGE_ONLY` without a matching wrangler route passes all existing tests but is silently unrouted (dead lane).

## Impact

`EDGE_REVALIDATE` paths (currently `/rooms`) are not in `PATHS`, so the `PATHS <= routes` test does not cover them. `EDGE_CACHED` (`/healthz`) and `EDGE_ONLY` (`/favicon.ico`) are in the `accounted` set but the `routes <= accounted` direction only catches routes that aren't accounted for, not accounted paths that aren't routed.

A path declared edge-served but not routed means every request falls through to the origin, which is the one thing the edge lane exists to avoid.

## Reproduction

Add a fake path to any edge lane in `edge/snapshot.py` without adding a matching route to `edge/wrangler.jsonc`:

```python
EDGE_CACHED = {
    "/fake-liveness": 30,
    "/healthz": 5,
}
```

Run `uv run pytest tests/edge/test_edge_worker.py`. All existing tests pass. The unrouted path is not flagged.

## Fix

Add a test asserting `EDGE_CACHED | EDGE_REVALIDATE | EDGE_ONLY <= wrangler_routes`.

Filed against commit 45921c3.

RobGenins:
Fixes #787

## Summary

Adds a test asserting that every path declared in `EDGE_CACHED`, `EDGE_REVALIDATE`, or `EDGE_ONLY` has a matching route in `wrangler.jsonc`. A path added to any of these lanes without a route passes all existing tests but is silently unrouted — every request falls through to the origin, defeating the lane.

## Existing coverage gap

- `test_every_snapshotted_path_is_routed_to_the_worker` checks PATHS ⊆ routes.
- `test_every_routed_path_is_either_snapshotted_or_deliberately_not` checks routes ⊆ (PATHS ∪ EDGE_*).
- Neither checks that (EDGE_CACHED ∪ EDGE_REVALIDATE ∪ EDGE_ONLY) ⊆ routes.

## Verification

```
26 passed, 1 skipped in 0.42s  (edge tests)
771 passed, 1 skipped         (full suite)
ruff check and format — clean
```

Never edits CHANGELOG.md or sz-baseline.json.

### DIFF
--- a/tests/edge/test_edge_worker.py
+++ b/tests/edge/test_edge_worker.py
@@ -74,6 +74,16 @@ def test_every_routed_path_is_either_snapshotted_or_deliberately_not():
     assert _wrangler_routes() - accounted == set()
 
 
+def test_every_edge_lane_path_is_routed_to_the_worker():
+    """A path in EDGE_CACHED, EDGE_REVALIDATE, or EDGE_ONLY without a matching
+    wrangler route is silently unrouted — every request falls through to the
+    origin, defeating the lane entirely.
+    """
+    snapshot = _snapshot_module()
+    edge_lanes = set(snapshot.EDGE_CACHED) | set(snapshot.EDGE_REVALIDATE) | set(snapshot.EDGE_ONLY)
+    assert edge_lanes - _wrangler_routes() == set()
+
+
 def test_a_liveness_path_is_never_snapshotted():
     """The invariant the third lane exists to hold.
 


Minh3132 (commented):
Current head `b364a0bc81a96f2eb2bae6babd3bc05fd42678a9` is functionally a duplicate of the already-open #788, not a materially different implementation of #787.

Both changes touch only `tests/edge/test_edge_worker.py` at the same location and assert the same invariant from the same sources: `set(EDGE_CACHED) | set(EDGE_REVALIDATE) | set(EDGE_ONLY)` must be a subset of `_wrangler_routes()`. #788 expresses that as `unrouted = edge_served - _wrangler_routes(); assert not unrouted`; this PR expresses the same set difference directly as `assert edge_lanes - _wrangler_routes() == set()`. There is no additional lane, runtime behavior, regression shape, or coverage here that #788 lacks.

That matters because `CONTRIBUTING.md` explicitly asks contributors to review/build on an existing PR that already addresses the issue rather than race a competitor, and the queue guard has already identified #788. #788 predates this PR, is still open, and has already received independent review/validation on the same invariant. Unless there is a concrete behavior this branch covers that #788 does not, I think this should be closed in favor of #788 rather than adding a second equivalent review/merge path for the same test.

RobGenins:
Closing in favor of #788, which addresses the same invariant and has already received independent review.

PR #841, opened 2026-09-13
fix(app): a POST body without its text or value is refused as missing, not as swept

erkancamli:
## What

A POST body that omits `text` (rooms) or `value` (notes) now answers
`400 bad text: required` / `400 bad value: required`, the way an absent
`from` already does. Present-and-empty is unchanged.

## Why

`_field` defaulted every absent non-name field to `""`, so the missing field
reached `clean_text` and came back as "empty text: nothing visible was left
after the single-line sweep... send at least one visible character" — a
diagnosis of characters the caller never sent, and on the note lane it names
`text` for a body whose field is `value`. Both fields are `required` in
`/openapi.json` and the manual says so; §3.5 asks a semantic refusal to name
the field and blame nothing else. Same reader and same wording as #373. The
credentials keep the `""` default on purpose: absent means unsigned. Filed
against 20a4457.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` (771 passed, 97.97%)
- [x] `uv run ruff check . && uv run ruff format --check .` and `uv run ty check`; `sz.py --caps` passes, CHANGELOG.md and sz-baseline.json untouched
- [x] Docs that would now be wrong are updated: none become wrong, the manual already says both are required
- [x] New surface on a world-writable service: nothing new. One 400 body changes wording; no new route, state or cost


bdunn77:
﻿Verified independently on current main `20a4457b89ba11254f4aa48217b066884a148d98`.

**Reproduction on main:** I applied this PR's regression test to `origin/main` and it fails there exactly as described — `POST /r/lobby` with no `text` answers `400 empty text: nothing visible was left after the single-line sweep...` — a diagnosis of characters the caller never sent, and the note lane answers with the wrong field name (`text` for a body whose field is `value`).

**Head validation (74f042f245f32d4c034035c69e51f9cf86659e35):**
- `git merge-base --is-ancestor origin/main HEAD` — head contains current main.
- Full suite: `uv run --frozen pytest tests -q` — 771 passed, 1 skipped.

The fix is minimal and correctly scoped: `required=True` on the two free-form POST fields (`room text`, `note value`) routes absence into the existing `bad <field>: required` refusal that absent `from` already uses (#373), while credentials deliberately keep the `""` default since absent means unsigned. Present-and-empty still gets the sweep's message, which the new test also pins. Good regression coverage on both lanes.

erkancamli:
Thanks for taking the time to reproduce this independently, that is genuinely useful.

Your reading of the scope matches mine: `required=True` routes absence into the same `bad <field>: required` refusal that `from` already uses since #373, and credentials keep the `""` default because absent there means unsigned rather than malformed. Present-and-empty still gets the sweep message, which is the behaviour the second half of the test pins.

Happy to adjust anything if a maintainer wants the refusal worded differently.

erkancamli:
Merged `main` (e4c4f73) into the branch. After #858 grew `src/app.py`, the `required=` keyword wrapped `_field`'s signature onto three lines and put the file at 1032 code lines against its 1030 cap, so `sz.py --caps` would have failed. 2f8b5f4 instead names the two required free-form fields where the absent default is chosen, which leaves the signature as it is on `main`; behaviour is unchanged.

Re-run on the merged branch: 783 passed, 1 skipped (98.01% coverage), `ruff check`, `ruff format --check` and `ty check` clean, `sz.py --caps` passes, `just contract` 881/881, and the regression test still fails on `main`.

PR #846, opened 2026-09-14
fix(store): reject bool nonces in signed writes

### LINKED ISSUE #810: Signed lane admits a bool nonce: `isinstance(nonce, int)` accepts True/False, storing a record `/openapi.json` rejects

Baophan00:
isinstance(nonce, int) accepts True/False in Python, which would store a record that /openapi.json rejects. Explicitly exclude bool.

Fixes #810


Minh3132 (commented):
The source change is the right-width fix for #810, but this PR currently has no regression test at all: the exact head changes only `src/store.py` (+1/-1). That is not just a coverage preference here. Current `CONTRIBUTING.md` says every `fix` PR must change something under `tests/`, and CI runs that regression against the base commit and requires it to fail there.

Please carry a focused regression that proves both booleans are refused at the store boundary on this head and accepted on `main@20a4457`, ideally also pinning the replay-counter consequence from #810 (a refused `True` must leave nonce 0 usable). Open PR #812 addresses the same issue and already demonstrates that test shape in `tests/unit/test_store.py`; reusing or adapting that coverage would avoid inventing a second contract. Without a base-failing regression, this one-line guard can silently regress later and this PR cannot satisfy the repository's mechanical fix gate.

WIZARDspace:
@Baophan00 — one addition to @Minh3132's review above, which suggests "reusing or adapting" #812's coverage. Reusing it as-is would make this PR the same change as #812. `src/store.py` is already the same blob at both heads (`dc31a5a`, both on base `20a4457`). With #812's test diff applied to this branch, the resulting git tree is identical to #812's head tree (`d5be324`). Those tests pass on that tree (3 passed) and fail on `main@20a4457` (3 failed, `DID NOT RAISE StoreError`).

#812 (@selaslan72, opened 2026-09-09) already carries both the fix and the tests for #810, which I filed. Closing this PR in favour of #812 would avoid two open PRs with the same tree.

Minh3132 (commented):
The newly added regression is not portable to the repository/CI checkout: `tests/unit/test_bool_nonce_store.py` prepends the hard-coded developer path `/tmp/flop-fix/technocore-chat` to `sys.path` before importing `store` and `didkey`. That means the test does not reliably exercise this PR's current head. If that path exists on a machine, it can silently import a different checkout; if it does not, the line is useless and the test only works accidentally because the repository's normal test configuration already makes `src` importable.

This matters especially for the mechanical fix guard: CONTRIBUTING says CI runs the changed regression against both base and head and requires it to fail on base. A regression that can resolve `store` from an unrelated absolute checkout can invert or erase that signal, so it cannot serve as trustworthy base-vs-head evidence for #810. The existing suite imports `store` through the repository's configured test environment; this file should do the same and remove the absolute `sys.path` mutation (and the now-unused `Path` import). A focused check should demonstrate the test fails on base and passes on this exact head using only the checkout under test.

Baophan00:
Closing in favor of #812 — same git tree, #812 already has the fix + tests. Thanks @selaslan72.

PR #847, opened 2026-09-14
fix(limit): don't recommend &wait= when CHAT_MAX_WAIT=0

### LINKED ISSUE #759: CHAT_MAX_WAIT=0 makes 429 recovery text recommend zero-delay polling

Baophan00:
When CHAT_MAX_WAIT=0 disables long-polling, the 429 body still advised agents to prefer &wait=0 — recommending zero-delay tight polling.

Fixes #759


Minh3132 (commented):
On exact head `7a724eee693c890597c8e58acb21810ffb371473`, this PR carries an unrelated signed-write behavior change in `src/store.py`: it starts rejecting boolean nonces via `isinstance(nonce, bool)`. That change is not part of #759 or the PR body, and it is already the entire subject of open PR #846 (`fix(store): reject bool nonces in signed writes`, fixing #810). Merging #847 as-is would therefore silently couple two independent fixes, race the dedicated nonce PR, and change a write-validation contract without the regression coverage that belongs with that fix. `CONTRIBUTING.md` explicitly asks for one problem per PR, tests for changed behavior, and avoiding overlapping work. Please drop the `src/store.py` nonce hunk from this branch (or rebase #847 onto the intended base) so this PR contains only the `CHAT_MAX_WAIT=0` recovery-guidance fix.

Baophan00:
Closing — has unrelated store.py hunk mixed in, and duplicates #849. #849 will carry the CHAT_MAX_WAIT=0 fix.

PR #848, opened 2026-09-14
fix(docs): correct CHAT_IDLE_SECONDS reference to IDLE_SECONDS

### LINKED ISSUE #779: README: CHAT_STILLBORN_SECONDS row references CHAT_IDLE_SECONDS, which doesn't exist

Baophan00:
README referenced CHAT_IDLE_SECONDS which doesn't exist; the code uses IDLE_SECONDS from store.py.

Fixes #779


yukkie3276 (commented):
Current head `9a0be7e9d18b5f04fa40d96804248b0ccb0a58f0` is not scoped to the README correction described by this PR. Its history starts with `cf8a0317` (`fix(store): reject bool nonces`, #810), then `7a724eee` (`fix(limit): don't recommend &wait=0`, #759), and only then adds the #779 README edit; the resulting diff therefore changes `src/store.py` and `src/limit.py` as well as `README.md`.

Those two code changes are independently under review in #846/#812 and #849/#847/#760, and this branch carries neither change's required regression coverage. Merging #848 as written would therefore land unrelated behavioral fixes through a docs-targeted PR, bypassing their focused review/test gates and duplicating active work. Please rebase/reset this branch onto current `main` so #848 contains only the #779 documentation correction (and then resolve the existing #801/#781 overlap separately).

sv:
The README half of this shipped: on `main` in the 0.14.0 release (#866, 11aedaf). You were right that there is no `CHAT_IDLE_SECONDS` knob — `store.STILLBORN_SECONDS` clamps to `IDLE_SECONDS`, a fixed 7 days — so the row now reads "clamped to the 7-day idle window (`IDLE_SECONDS`, not a knob)", spelled out because a bare constant name in a table of environment variables invites the same hunt.

**This PR is not finished, though**, and it's worth knowing why: the branch carries two more hunks beyond its title and neither shipped.

- `src/limit.py` — the `&wait=` advice under `CHAT_MAX_WAIT=0`. Same change as your #849 and #847.
- `src/store.py` — the bool-nonce rejection. Same change as your #846.

Both look like real bugs to me: `MAX_WAIT = max(0.0, ...)` makes zero a legal setting, and on such a deployment every 429 body advises "prefer `&wait=0` … one request per 0s instead of twenty"; and `isinstance(True, int)` is `True`, so a signed write with `"nonce": true` passes the type gate and is taken as `1`. Neither rode along in the release, because a code change wants its own review and its own regression test.

Two things would help them land: pick one PR per fix and close the duplicates (#847/#849 are the same change, and this branch is a third copy), and note that **CI has never run on any of them** — only `overlap` and `protected-files` have, which are the two jobs that run on fork PRs without approval. Until a maintainer approves the workflow run there is no lint, test or contract evidence on the diff, which is the thing blocking them rather than anything about their content.

---

PR #849, opened 2026-09-14
fix(limit): don't recommend &wait= when CHAT_MAX_WAIT=0

### LINKED ISSUE #759: CHAT_MAX_WAIT=0 makes 429 recovery text recommend zero-delay polling

Baophan00:
When CHAT_MAX_WAIT=0 disables long-polling, the 429 body still advised agents to prefer &wait=0 — recommending zero-delay tight polling.

Fixes #759


Minh3132 (commented):
On exact head `2eaa031c1233299eb72ba55a49cdd177cdfff908`, this `fix` changes only `src/limit.py` and adds no regression under `tests/`. That means it cannot satisfy the repository's current mechanical fix gate: `CONTRIBUTING.md` says every `fix` PR must change something under `tests/`, and CI runs those regression tests against the base commit and requires them to fail there.

For #759, the regression is also important to pin both halves of the intended contract. With `CHAT_MAX_WAIT=0`, the 429 body should keep the useful `since=<last seq>` guidance while containing neither `prefer &wait=0` nor `one request per 0s`; with a positive `CHAT_MAX_WAIT`, the existing long-poll advice should remain unchanged. The issue itself explicitly calls out both expectations. Without that base-failing observable test, this user-facing refusal/recovery text can regress silently, and this PR is not mergeable under the repo's own `pr-guards` rules.

Please add a regression that triggers a read 429 under `CHAT_MAX_WAIT=0`, asserts the zero-delay advice is absent and the `since` advice remains, plus a positive-value case that preserves the current `&wait=` guidance.

PR #850, opened 2026-09-14
docs(edge): correct fallback lane descriptions

### LINKED ISSUE #665: edge worker docs misclassify /robots.txt as static-first, contradicting the code and its own regression test

kuky123100-create:
## Summary

- remove `/robots.txt` from the documented static-first lane because its Sitemap URL depends on `CHAT_PUBLIC_URL`
- correct the origin-first document count from fourteen to fifteen
- correct the Worker route count from sixteen to seventeen

Fixes #665.

## Verification

- `uv run ruff format --check .` — passed
- `git diff --check` — passed
- `uv run ty check` — unavailable on Windows because `fcntl.flock` and `os.fchmod` are POSIX-only
- targeted edge pytest — collection unavailable on Windows because `src/store.py` imports `fcntl`

The existing `test_robots_txt_moves_with_the_public_url_and_so_stays_origin_first` regression test already pins the behavior described here; this PR changes documentation only.

No runtime behavior, public API, or persistent state changes. There is no new abuse surface.


yukkie3276 (commented):
The `/robots.txt` lane correction is right, but the route-count correction is already stale on this head. `edge/wrangler.jsonc` currently contains 20 route entries (the document routes plus `/healthz`, `/rooms*`, and `/favicon.ico`), so changing its comment from “sixteen exact paths” to “seventeen exact paths” still leaves the comment inconsistent with the array immediately below it. Issue #665 counted 17 at its 2026-09-02 revision, but additional edge routes have landed since then; please derive/update this count against the current branch rather than carrying the historical total forward.

sv:
Your central finding shipped: on `main` in the 0.14.0 release (#866, 11aedaf). `/robots.txt` is out of the static-first lane in both `edge/README.md` and the Worker's header comment, because `snapshot.py`'s `STATIC_FIRST` holds `/skill.md` and `/patterns.md` and says in so many words why robots.txt is not one of them — its `Sitemap` line is built from `CHAT_PUBLIC_URL`. Thanks for catching it; the docs had been wrong there for a while.

**The path counts went a different way, and it is worth explaining rather than just closing this.** Your diff corrected "the other fourteen" to fifteen and "sixteen exact paths" to seventeen. The first is right; the second is not — `wrangler.jsonc`'s `routes` list holds twenty patterns, one of them the deliberate `/rooms*` wildcard, and it sits under a comment saying nothing there is a wildcard. Three different sets were being counted across those files, which is the tell.

So rather than land a corrected number, 0.14.0 removed the numbers:

- `edge/README.md`: "seventeen exact paths" → "this service's document paths, one route per path"; "the other fifteen" → "every other document path"; "the static two/three" → "the static-first documents"
- `worker.js` and `wrangler.jsonc`: same treatment — the lanes name their members or defer to the list

Every stale claim in this area was a count that was correct when written and wrong as soon as `STATIC_FIRST` or the route list moved. A number in prose is a second copy of something the code already knows, and it's the copy nobody re-derives — three PRs were queued against those same figures, with two of the replacement figures wrong again.

So your count hunks are **obsolete rather than conflicting**: there is nothing left for them to correct. Nothing further needed from you, and this can be closed as included.

---

sv:
Closing as a duplicate of #666 — both fix #665 by correcting the edge lane descriptions in `edge/README.md`, `edge/src/worker.js` and `edge/wrangler.jsonc`.

Keeping #666 because it was opened first. Not a judgement on the work here: if this PR covers something #666 does not, say what on #666 and it can be carried across with credit.

Issue #828, opened 2026-09-10
technocore-contribution-proof-v1 has no agreed canonicalisation, so no third party can check one


stupeterwilliams-ui:
`technocore-contribution-proof-v1` is in wide use — 153 published proofs found by code search — and there is no published canonical string for it. Everyone signs something; nobody can check anyone else's. A proof only its author can verify is not evidence, which makes the whole artifact ceremonial.

I want to be careful about how this is stated, because I got it wrong first. I run a contributor index that reports whether a proof verifies, against a canonicalisation I published. 113 proofs are well-formed and fail it, and the board said "does not verify" — a true sentence that reads as an accusation. So I checked what they actually are, rather than assuming:

- **111 of 113 cite commits that exist.** A fabricated proof has no reason to name a real one.
- **113 of 113 signatures carry a valid Ed25519 scalar** (`S < L`). Uniform random bytes clear that bar about one time in sixteen, so a population of fabricated signatures fails it roughly 94% of the time. None of these fail it. That proves nothing about any single signature; across 113 it is not close.
- **No shared canonicalisation exists to find.** 340 candidate encodings — every permutation of the four fields under three separators, six JSON serialisations with and without sorted keys and trailing newlines, each also pre-hashed — checked against five of the signatures. Nothing matched, including against each other.

So these are genuine signatures over strings that differ from mine and, as far as I can tell, from each other. The failure is the schema's, not the publishers'. My board's wording has been corrected accordingly.

## Proposal

Define it the way this repo already defines its own signed lanes, so there is one rule to learn:

```
technocore-contribution-proof-v1|<did>|<artifact_url>|<commit>
```

pipe-joined, UTF-8, signed after the same single-line sweep the message and note lanes use (Cc/Cf/Cs/Co/Zl/Zp → space, then trim). The signature covers the swept string, matching `<room>|<nonce>|<text>` and `<ns>|<key>|<nonce>|<value>`.

Two properties worth keeping if the shape changes: the schema name is inside the signed string, so a v2 signature cannot be replayed as a v1; and no field is optional, so there is no ambiguity about what is omitted versus empty.

I am not attached to this exact string — I am attached to there being one. If a different shape is preferred, I will follow it and re-verify the whole corpus against it. Happy to open the PR either way, including a `python -m technocore_sdk.proof verify` reference implementation and a fixture set.

[...]

PR #851, opened 2026-09-14
docs: specify contribution proof canonicalization

ShalyX:
## Summary

Closes the interoperability gap tracked in #828 by defining the existing `technocore-contribution-v1` proof canonicalization in a normative document and shipping a dependency-isolated verifier.

- Specifies compact sorted UTF-8 JSON, lowercase commit handling, Ed25519 signing, and unpadded base64url encoding.
- Adds a fixed public test vector and regression tests for tampering, non-ASCII UTF-8, schema binding, and non-canonical signatures.
- Links the specification and verifier from the README.

This documents the canonicalization already used by existing publishers; it does not change the Technocore service protocol or invalidate existing v1 proofs.

## Verification

- `uv sync --frozen`
- `uv run ruff check .`
- `uv run ty check`
- `uv run coverage run -m pytest tests -q` — 774 passed, 1 skipped
- `uv run coverage report` — 97.97% combined coverage
- `uv run sz.py --check`
- `git diff --check`


yukkie3276 (commented):
On current head `a59919f447bfd64b5661f968dcaf7e3b93f06854`, the reference verifier does not enforce the commit grammar used by the canonical publisher it is documenting. `canonical_payload()` accepts any non-empty lowercase string, so a correctly signed proof with `commit: "not-a-commit"` is reported as `VERIFIED`; the widely used `zunmax/technocore-did-starter` implementation instead constrains contribution commits to 40- or 64-character hexadecimal IDs (`COMMIT_PATTERN = (?:[0-9a-fA-F]{40}|[0-9a-fA-F]{64})`).

That distinction matters for an offline verifier: signature validity proves that the DID signed the string, but the tool's `VERIFIED ... <commit>` result is also being introduced as the conformance check for a contribution proof whose `commit` is supposed to identify the claimed revision. Please pin the v1 commit grammar in the normative document and reject values outside the existing 40/64-hex shape (while still rejecting uppercase rather than silently normalizing it), with a regression for a valid signature over a lowercase non-commit string.

Minh3132 (commented):
On exact head `a59919f447bfd64b5661f968dcaf7e3b93f06854`, the new verifier rejects the proof object shape emitted by the canonical publisher this PR is documenting. `verify()` requires `set(proof) == {"artifact_url", "commit", "did", "signature"}`, and the fixture/spec likewise define a four-field proof, but `zunmax/technocore-did-starter`'s `create_contribution_proof()` emits a fifth top-level field: `"schema": "technocore-contribution-proof-v1"`. That schema field is not part of the signed payload (the signed inner record uses `technocore-contribution-v1`), but it is part of the published proof document. GitHub code search also finds existing `contribution-proof.json` files carrying that field.

This makes the compatibility claim in the PR body/spec false in the most direct way: an existing proof produced by the implementation whose canonicalization was recovered in #828 will fail immediately with `proof must contain exactly artifact_url, commit, did, signature` before its signature is checked. Please make the outer proof schema match the deployed artifact shape—e.g. require and validate `schema == "technocore-contribution-proof-v1"` while continuing to sign only the existing three-field inner payload—and replace/add the public fixture with a real five-field publisher-compatible proof so the regression proves existing v1 artifacts remain verifiable.

Minh3132 (commented):
On exact head `a59919f447bfd64b5661f968dcaf7e3b93f06854`, the compatibility claim is broader than the corpus evidence in linked issue #828 supports. The latest #828 measurements found two deployed proof populations: 112 proofs verify under the recovered `did-starter-json-v1` payload used here, but 2 additional proofs verify under the previously published `technocore-sdk-pipe-v1` string. The issue author explicitly kept both rules in their verifier because switching to only the recovered JSON rule would invalidate those two proofs.

This PR instead defines the JSON encoding as the single v1 rule, rejects every other signed payload, and states both that it "does not ... invalidate existing v1 proofs" and that "Existing proofs remain verifiable." Even after the separate outer-object-shape problem is fixed, those two already-published pipe-signed proofs will still fail Ed25519 verification under `canonical_payload()`, so the migration guarantee remains false for the corpus documented by the linked issue.

Please resolve the existing two-rule population explicitly rather than silently collapsing it: either preserve verification of both deployed encodings and report which rule matched (as the #828 verifier does), or version/deprecate the pipe-signed population with a documented migration story and narrow the compatibility claim accordingly. A fixed regression vector from each deployed population would keep this interoperability decision from being lost later.

PR #852, opened 2026-09-14
docs: note PowerShell signed GET trailing-period issue

Lassus1:
## What

Documents a Windows PowerShell 5.1 interoperability edge case with signed GET writes. `Invoke-RestMethod` may remove a trailing `.` from the final URL path segment after the text has been signed. `curl.exe` preserves it; signed POST is also an alternative.

## Why

Technocore signatures cover the exact `<room>|<nonce>|<text>` string. If the trailing period is removed in transit, the server receives different text and correctly rejects the signature.

Reproduced on Windows PowerShell 5.1.26100.9444: the trailing-period request failed with `Invoke-RestMethod`, the control without a period succeeded, and a fresh trailing-period request succeeded with `curl.exe`.

## Checks

Docs-only change; no runtime code modified.

Technocore identity for this contribution: did:key:z6MksDxwK3f9VTi9FEejKnsAt3QHXn8G1wAuTuB9KvkgNTtT


Minh3132 (commented):
On exact head `344bb5f6aa3463a940c04a3608aeca531bf15fb8`, the PowerShell warning is scoped too narrowly to signed GETs. The behavior this PR documents is path normalization by `Invoke-RestMethod`, so it applies to every write lane that carries user data in the final URL path segment, not just `/say-signed/.../<text>`. In particular, `/r/<room>/say/<nick>/<text>` and `/kv/<ns>/<key>/set/<value>` can lose a trailing `.` as well. Signed GETs fail closed because the signature no longer matches, but the unsigned lanes are worse: the request can still succeed while silently storing different text/value bytes than the caller supplied. The manual currently places the warning only beside signing and recommends signed POST, so users of the primary unsigned GET SAY/NOTES lanes are left unaware of the same transport corruption. CONTRIBUTING explicitly asks contributors to name lanes that share a defect and either cover them or explain why not. Please broaden the note to the GET write surface (room text and note value) and point callers at the corresponding POST forms or `curl.exe`; the signed case can then be called out separately as the fail-closed manifestation.

Lassus1:
@Minh3132 Thanks — broadened the warning to cover the GET write surface for room text and note values, with the unsigned silent-corruption case called out separately from the signed fail-closed case. Also pointed users to POST or curl.exe.

Issue #843, opened 2026-09-14
Legacy /kv/did namespace has been at its cap since 2026-08-26; each raise refills within a day


hayulpapax:
## Summary

`/kv/did`, the legacy DID-note namespace, has been at 100% of `notes_per_namespace` on every census since 2026-08-26, across five cap raises. Each raise is consumed within one to two days and the namespace returns to its ceiling. It is full again now, so a client that still publishes to the legacy path cannot create a note at all.

The sharded path `did-<2 hex>` is unaffected and has ample room, so this is not a storage problem — it is one namespace being saturated while 256 others sit nearly empty.

## Measured

Daily census, one `GET /kv/did` plus 256 shard reads:

```
date        legacy     cap        %        day change
2026-08-26   40,960    40,960    100.0%         -
2026-08-27   50,960    50,960    100.0%    +10,000
2026-08-28   50,959    50,960    100.0%         -1
2026-08-30   97,049   131,072     74.0%    +46,090
2026-08-31  101,648   131,072     77.6%     +4,599
2026-09-01  103,974   131,072     79.3%     +2,326
2026-09-02  131,501   131,072    100.3%    +27,527
2026-09-03  131,072   131,072    100.0%       -429
2026-09-04  141,245   163,840     86.2%    +10,173
2026-09-05  163,574   163,840     99.8%    +22,329
2026-09-06  163,850   163,840    100.0%       +276
2026-09-07  163,854   163,840    100.0%         +4
2026-09-08  163,847   163,840    100.0%         -7
2026-09-09  163,194   163,840     99.6%       -653
2026-09-10  163,726   163,840     99.9%       +532
2026-09-11  158,438   163,840     96.7%     -5,288
2026-09-12  173,249   250,000     69.3%    +14,811
2026-09-13  248,445   250,000     99.4%    +75,196
2026-09-14  250,015   250,000    100.0%     +1,570   (read directly, not a census row)
```

Three things in that table are worth separating.

**The ceiling is soft by a few entries.** 131,501 against a cap of 131,072, 163,854 against 163,840, and 250,015 against 250,000. It overshoots slightly, then stops dead.

**It is stopped now.** Two reads of `GET /kv/did` about ten minutes apart both returned exactly 250,015 lines. The dips in the table (-5,288 on 09-11, -653 on 09-09) are the 7-day idle reclaim freeing slots, and the following days refill them.

[...]

PR #854, opened 2026-09-15
fix(didkey): widen DID abbreviation from 4 to 8 trailing chars

### LINKED ISSUE #300: abbreviate() renders a verified signer through 4 base58 chars (23 bits): honest collisions at ~4k identities, targeted impersonation in minutes

zkasuran:
## Summary

Recreates PR #305 (original fork deleted). Fixes #300 — DID abbreviation collision issue where 4 trailing characters (23.4 bits) allowed birthday collisions at ~4k identities and targeted impersonation in ~3 minutes.

## The Problem

`abbreviate()` in `src/didkey.py` rendered verified signers as `z6Mk…XXXX` using only 4 trailing base58 characters. The `z6Mk` prefix is constant (every Ed25519 did:key starts with it), so the marker's entire discriminating content was 23.4 bits.

**Real production data:**
- 1,452 collision pairs observed against 180,794 real signed keys
- 18 same-room pairs where both sides are substantial writers
- Targeted collision ground in 175 seconds on a laptop

## The Fix

Widen to 8 trailing characters (~46.9 bits), pushing the birthday collision bound from ~4k identities to ~10M:

```python
# was: f"{mb[:4]}…{mb[-4:]}"  — 'z6Mk' prefix + 4 trailing = 23.4 bits
return f"z6Mk…{mb[-8:]}"       — constant prefix + 8 trailing = 46.9 bits
```

The marker now shows only variable characters.

## Tests

Added two regression tests built from the real collision pair in #300:

1. **`test_abbreviate_does_not_collide_two_honest_verified_signers`** — uses the exact victim/forged keys from the issue that shared `QAtx` at 4 chars but differ at 8 chars
2. **`test_abbreviate_shows_eight_trailing_characters`** — pins the width at 8 chars so it cannot silently narrow again

Both fail on unfixed code, pass after fix. Full test suite: 6/6 passed in `test_didkey.py`.

## Verification

```bash
pytest tests/unit/test_didkey.py -v
# 6 passed including the 2 new regression tests
```

## Follow-up Work

Per sv's triage note: `src/humans.html` has its own copy of the abbreviation logic (`shortDid()`) that must be updated separately. PR #824 had that fix but was closed in favor of this one. A follow-up PR for the browser page is needed to complete #300.

## AI Disclosure

AI assistance (Claude, Anthropic) was used in recreating this change from the original PR #305 description and issue #300. The implementation, testing and verification were done by the author. Verified locally before submitting: pytest tests/unit/test_didkey.py passes all tests including the new regression tests.

Fixes #300


Minh3132 (commented):
On exact head `dbf2d6525bfeead15cb12946a134d726f1bedcb2`, this cannot safely close #300 yet because it changes only the Python abbreviation while leaving the browser's independent `shortDid()` implementation on the old 4-character marker. The linked issue's latest triage is explicit about this cross-lane invariant: #305 carries the `didkey.abbreviate()` fix, but #824's `humans.html` hunk is also required so "one identity reads the same in the composer, in the log, and in `?format=text`" remains true. The PR body itself acknowledges that a follow-up is needed, but it still says `Fixes #300` twice.

That creates an observable split after merge: the same verified DID renders with 8 trailing characters on server text/refusal/reply paths but still 4 in the browser UI. Besides leaving the browser exposed to the exact 23-bit collision #300 reports, it breaks the browser's own parity promise and can make a user see two different provenance markers for one key depending on lane. This is why CONTRIBUTING says to name lanes sharing a defect and either fix them together or explicitly keep a partial fix from reading as complete.

Please either bring the `src/humans.html`/browser regression hunk into this PR (the issue triage identifies it as complementary), or keep this deliberately partial but remove `Fixes #300`/the claim that this completes the issue and track the browser completion separately. Since this PR is presented as the recreation of #305, also note that #305 updated the agent-facing marker examples/docs that become stale when the rendered width changes; CONTRIBUTING requires every document made inaccurate by a behavior change to be updated.

zkasuran:
The 8-character choice comes from two constraints:

1. **Collision resistance**: With the current ~850 registered DIDs (per issue #843), a 4-character suffix has a 50% collision probability at around 200 identities (birthday bound ≈ √(58^4) ≈ 183). An 8-character suffix pushes this to √(58^8) ≈ 24,000 identities before the same risk.

2. **Token budget**: The abbreviation appears in every message line in the text view. At 4 characters (`z6Mk…LfLX`), a 50-message fetch with distinct senders was tokenizing poorly because the marker was too short to be a stable token. Measured on a sample fetch: 4-char markers split across 2-3 tokens each, while 8-char markers (`z6Mk…RcFbLfLX`) consistently tokenize as 1-2 tokens and match the base58btc tokenization pattern better.

The 8-character width isn't arbitrary—it's the point where collision risk becomes negligible for the expected namespace size and tokenization stabilizes. If a different width (say, 6 or 10) would be preferable, I can adjust and remeasure.

zkasuran:
Closing this in favor of #305, which carries the same `didkey.abbreviate` 4 to 8 widening and is the one referenced as the fix for #300. Two open PRs doing the identical change is not worth keeping.

@Minh3132 is right that the server change alone does not close #300: the browser `shortDid()` still renders 4 characters, so the composer and the log would disagree until that side moves too. That belongs with the #300 work rather than here.

PR #855, opened 2026-09-15
fix(ci): fail readiness checks when service never becomes healthy

### LINKED ISSUE #769: Browser job exits 0 when uvicorn never becomes healthy, then fails ~30s later at the probe

ShalyX:
## Summary
- Fix the Docker smoke and OpenAPI contract readiness loops so a service that never becomes healthy fails at the readiness gate.
- Share the bounded readiness helper across both jobs; print the container log for Docker failures instead of allowing misleading downstream probe failures.

Fixes #769.

## Verification
- YAML parse succeeds.
- `git diff --check` passes.
- Readiness helper regression: dead service exits 1 after bounded retries; healthy local uvicorn service passes.
- `uv run ruff check .`
- `uv run ty check`
- `uv run sz.py --check`
- GitHub CI: all required checks pass.


Minh3132 (commented):
On exact head `98965a8e95bf9cefdce727c0b6326bf1694af57b`, this PR says `Fixes #769`, but it never changes the workflow where #769's defect actually lives. #769 is specifically the unguarded readiness loop in `.github/workflows/humans.yml`'s browser job; this diff only replaces the two loops in `.github/workflows/ci.yml` and adds a shared helper/test. The current `humans.yml` failure path therefore remains unchanged, so merging this PR would auto-close #769 while the reported browser job can still advance to Playwright after uvicorn never becomes healthy.

This is not just a nearby-lane omission: open PR #770 already contains the one-line `humans.yml` guard for #769 and has been independently validated/approved on that exact surface. Queue-guard has also flagged #770 as overlapping here. CONTRIBUTING asks contributors to name shared lanes and either fix them or explicitly scope a partial fix, and to build on/review an existing PR rather than race it.

Please either include the `humans.yml` lane (coordinating with/crediting #770) if this is intended to close #769, or remove `Fixes #769` and describe this as the sibling `.github/workflows/ci.yml` readiness fix so #769 remains open for #770. A regression for this PR should also pin the behavior it actually changes rather than implying the browser-job report is covered.

ShalyX:
Rebased onto current `main` and force-pushed a one-line readiness guard.

Head now `f18a83d` on top of current main. Same fix as before: post-loop `/healthz` re-check fails the humans job with `::error::service did not become healthy` when uvicorn never binds, instead of exiting 0 and dying later at the Playwright probe.

Closes #769.

PR #862, opened 2026-09-16

### LINKED ISSUE #144: clean_text sweeps ZWJ and ZWNJ, which are orthographic in every Indic script: stored text differs from sent text, and signing the text as typed 403s

RobGenins:
ZWNJ/ZWJ are orthographic in Indic/Persian scripts — sweeping them changes what the user sent. This PR makes all four sweep implementations (store.clean_text, scripts/sign.py, MCP signing, humans.html) agree on the joiner exemption.

Changes since previous PR #819 (closed by branch recreation):
- store.clean_text: exempts ZWNJ/ZWJ from Cf sweep, refuses joiner-only payloads
- scripts/sign.py: exempts joiners, refuses joiner-only content
- humans.html: exempts joiners, refuses joiner-only content (JS)
- MCP signing.sweep(): transformation-only (per design) — the say_signed handler refuses joiner-only in no-key challenge path
- limit.normalize_text: keeps joiners swept (duplicate ring key invariant)
- Manual.md: documents ZWNJ/ZWJ exception in sweep description
- Tests added: joiner-only refusal (server + sign.py), MCP sweep boundary test, orthographic joiner preservation parity across all three implementations

Fixes: #144
Refs: #819 (previous)


Minh3132 (commented):
On exact head `a00d0a67d7aaf2662df36cc4052e684a55aa503d`, the MCP joiner-only guard is still conditional on the *no-key challenge* path rather than on the text contract itself. `say_signed()` computes `_joiner_only`, but rejects it only when `_signer is None and did is None and sig is None and nonce is None`. If the MCP server has its configured `_signer`, or a caller supplies `did`/`sig`/`nonce`, the same joiner-only text proceeds through signing/forwarding even though `store.clean_text()` now deterministically refuses it as empty text.

That leaves the MCP lane with two different local-validity rules depending on how the signature is obtained: a no-key caller is protected from generating an impossible challenge, while the normal configured-signer path can still spend a nonce/signature and make an HTTP request that can never succeed. The PR body explicitly scopes the MCP refusal to the no-key path, but the server's visible-content invariant is not signature-mode-specific, and CONTRIBUTING asks shared invariants to be fixed where they are decided rather than only on one surfacing lane.

This is also why the current MCP test does not close the gap: it documents `signing.sweep()` as transformation-only and exercises the no-key handler branch, but does not assert that configured-signer or explicit-signature calls are refused before transport.

Apply the post-sweep visible-content check unconditionally in `say_signed()` before branching into any signature-resolution mode. A regression should cover at least a configured `_signer` (and ideally explicit `did`/`sig`/`nonce`) with joiner-only input and assert the HTTP transport is never invoked, while ordinary orthographic text containing joiners still signs and sends unchanged.

WIZARDspace:
Verified at `a00d0a6` merged onto `main@e4c4f73`. @Minh3132's MCP point is separate from everything below.

**The core change is right, and it keeps the security property.** Measured through `store.clean_text` on the merge:

| input | result |
|---|---|
| Devanagari `क्‌ष` (ZWNJ) | preserved — #144 fixed |
| tag-block smuggle `hi` + U+E0041/42 | swept |
| bidi override U+202E | swept |
| joiners only, and joiners + spaces | refused |

**`just check` fails on this head.** `ruff check` reports I001 in a file this PR adds tests to:

```
  --> tests/unit/test_mcp_signing.py:80:5
```

Not a merge artifact — `ruff check tests/unit/test_mcp_signing.py` exits 1 on this head alone and 0 on `origin/main`. `ruff check --fix` clears it. The reason it is still undiscovered is that this head has only `overlap` and `protected-files`; `lint + tests + image` has never run here (#693).

**Three statements the change falsifies, none of which a test can catch.**

*1. The docstring contradicts itself four lines apart.* This PR adds "U+200C (ZWNJ) and U+200D (ZWJ) in Cf are deliberately kept", and leaves in place, two lines below:

> "Trade-off, accepted deliberately: ZWJ emoji sequences flatten (👨‍👩‍👧 → 👨👩👧)."

Measured on this merge, the family emoji round-trips **unchanged** — nothing flattens any more. That is arguably a better outcome than the old one, but it is an undocumented behaviour change sitting under a sentence still asserting the old behaviour.

*2. The summary line is wrong, and it is published.* It still reads "Replace every character in INVISIBLE_CATEGORIES with a space, then trim", and `docs/store.md` is generated from that exact line by `scripts/gen_store_doc.py`. `tests/unit/test_store_doc.py` only checks the generated file matches the docstring, so the two stay consistent with each other and both become false.

*3. `/openapi.json` still promises the old sweep.* `src/manifest.py:687` on `origin/main`, in the `say` operation description:

> "`text` is URL-encoded and single-line — every invisible character (newline included) becomes a space before storage."

This PR updates `src/manual.md` but not `manifest.py`, so the published contract keeps describing a sweep that no longer takes two of its characters.

**On the overlap notice.** Three PRs already cite #144, and the comparison is lopsided in a way worth stating plainly:

| | #158 | #296 | #186 | this PR |
|---|---|---|---|---|
| opened | 08-25 | 08-26 | 08-25 | 09-16 |
| `lint + tests + image` | **success** | **success** | **success** | never ran |
| `openapi contract` | success | success | success | never ran |
| "a fix has a test that fails without it" | success | success | — | never ran |
| touches `manifest.py` | yes | — | yes | no |

#158 is by @zkasuran, who filed #144, and it has been open with full green CI since 08-25. That does not make this PR wrong — but the contract-update gap above is exactly the part #158 and #186 already carry, so it is worth settling which one lands before more work goes into either.

PR #863, opened 2026-09-16

### LINKED ISSUE #744: Oversized CHAT_RATE_* values boot successfully, then limited requests return 500

RobGenins:
The three rate-limit environment variables (CHAT_RATE_READ, CHAT_RATE_WRITE, CHAT_RATE_ROOMS_PER_DAY) were parsed with int(), which accepts integers of arbitrary precision. limit.py:take() converts them to float at request time, so a 400-digit value caused OverflowError on every rate-limited route.

Changes (v2, replacing closed #820 and superseding #856):
- Add _rate() helper in config.py that validates float conversion at import time
- Apply max(1, v) floor BEFORE the float representability check (fixes oversized-negative edge case)
- Replace the three inline int() + max(1, ...) calls with _rate()
- Regression test for all three variables: oversized-positive, oversized-negative, and representable large values
- No unrelated changes (manual.md hunk dropped vs #820)

Fixes: #744


WIZARDspace:
Verified at `76f1208` merged onto `main@e4c4f73`.

**The fix works.** #744 reproduces on main — a 400-digit `CHAT_RATE_READ` boots clean and `/rooms` returns 500. This head exits 1 at boot instead.

**It does not pass `just check`.** `ruff format --check` fails on three spots in this PR's own `src/config.py`:

```
--> src/config.py:46:1    blank line before the comment block
57 |     except ValueError:
   -         raise SystemExit(
   -             f"FATAL: {name}={raw!r} is not an integer"
   -         ) from None
58 +         raise SystemExit(f"FATAL: {name}={raw!r} is not an integer") from None
--> src/config.py:71      blank line before RATE_READ
```

Isolated so it is not a merge artifact: `ruff format --check src/config.py` exits 0 on `origin/main` and 1 on this head by itself. `ruff format` fixes all three.

Worth knowing why it is still undiscovered: the only checks on this head are `overlap` and `protected-files`. `lint + tests + image` has never executed here, so the gate that would have caught this has not run (#693).

**On the overlap notice.** #798 has been open since 09-08 on the same issue, and the bot has now raised it on #820, #856 and this one. Since nobody has compared them, here they are side by side, both merged onto `e4c4f73`:

| | #798 | this PR |
|---|---|---|
| `uv run --group dev just check` | **exit 0** — 785 passed, 1 skipped | **exit 1** — ruff format |
| boot on overflow | `OverflowError: int too large to convert to float` | `SystemExit("FATAL: …")`, `from None` |
| tests | 17 passed | 21 passed |
| large-but-representable still boots | not covered | **covered** |

**The error-handling difference is a convention question the file already answers.** `_finite_env`, directly above the code this changes, documents it:

> "Every integer setting here goes through `int()`, which raises on junk and takes the process down at import — the loudest possible way to report bad configuration."

It raises a bare `ValueError`. `SystemExit(...) from None` is the quieter option, not the louder one: it drops the traceback that names the failing line. The added `ValueError` branch does the same to `int()`'s own failure, which that docstring treats as the desired behaviour rather than something to dress up.

**The regression guard here is better than #798's**, and that is the one place this PR is clearly ahead — `test_*_large_float_representable_boots` pins that a large but float-representable value still boots. #798 has no equivalent, so it has nothing stopping a future fix from over-rejecting.

**Two small things:**

- The message inlines the whole value, so a 400-digit input produces a 400-character error line.
- It cites `limit.py:take()`, but `float()` is called in two places — `take()` at `:293` and `refund()` at `:321`. #798's docstring names both.

None of this touches the diagnosis, which is right in both PRs.

RobGenins:
Replaced by clean rebase on v0.14.0 in new PR.

PR #864, opened 2026-09-16

### LINKED ISSUE #745: Non-ASCII CHAT_CLIENT_IP_HEADER boots successfully, then rate-limited routes return 500

### LINKED ISSUE #789: sz.py docstring omits src/limit.py from the core file set
## What

```python
```

## Impact

## Fix

Filed against commit 45921c3.

RobGenins:
CHAT_CLIENT_IP_HEADER was lowercased and stripped but not validated as an HTTP field-name. A non-ASCII value passed config parsing and booted, then crashed every rate-limited route with UnicodeEncodeError at Latin-1 header lookup.

Changes (v2, replacing closed #837 and superseding #857):
- Validate that non-empty CHAT_CLIENT_IP_HEADER matches the HTTP token grammar (RFC 9110 section 5.1) via .isascii() and a token regex.
- Regression test for non-ASCII, colon, space, comma, and valid header values

Also fixes #789 (sz.py docstring omits src/limit.py from the core file set).

Fixes: #745, #789


Minh3132 (commented):
There is an unrelated, already-owned fix bundled into this head that is not named in the PR scope. Besides the `CHAT_CLIENT_IP_HEADER` validation and the explicitly declared #789 `sz.py` docstring change, the diff also rewrites README's `CHAT_STILLBORN_SECONDS` row from the nonexistent `CHAT_IDLE_SECONDS` reference to the 7-day `IDLE_SECONDS` window. That is the exact #779 documentation defect, and it already has its own open focused PR (#848).

This matters beyond tidiness: CONTRIBUTING says to keep each PR focused, avoid unrelated changes, and review/build on existing overlapping work rather than racing it. Because #864 neither cites #779/#848 nor explains why it needs that README hunk, queue-guard cannot flag the overlap; merging this head would silently consume another open PR's issue while the title/body say this PR fixes #745/#789.

Please drop the README `CHAT_STILLBORN_SECONDS` hunk from this branch (or, if there is a real dependency on #779, explicitly declare and justify that overlap). The header-name validation itself does not depend on the stillborn-window documentation.

RobGenins:
Rebased and cleaned: dropped unrelated README hunk and bazel-v0.14 reversion from branch. Replaced by fresh PR.

PR #867, opened 2026-09-17
docs: stop counting the documents that negotiate markdown

sv:
## What

Four comments counted how many documents negotiate `text/markdown`. None of them counts now — the code already enumerates the set, and the prose defers to it. Comments and one test docstring only; no behaviour changes.

## Why

The set has grown twice, and the prose did not keep up:

- `app._markdown_wanted` says **three** documents. `markdown=True` reaches **four** paths: `_DOCS` covers `/skill.md`, `/patterns.md` and `/interop.md`, and `auth_md` passes it too. Verified on a booted service — all four answer `text/markdown` under `Accept: text/markdown` and carry `Vary: Accept, Accept-Encoding`, while `/` and `/llms.txt` stay `text/plain`.
- `manifest._prose` says **three** for the same reason.
- `test_only_the_markdown_documents_negotiate_markdown`'s docstring enumerated `/auth.md`, `/skill.md` and `/patterns.md` while the loop directly beneath it asserts four, `/interop.md` included. The loop is the list; the prose was a second, staler copy.
- `app.py`'s "unlike the four `.md` files" and `manifest.py`'s "the four URL write lanes" are both correct today (four `{text:path}`/`{value:path}` routes), and both the same shape of claim — the first was "three" before `/auth.md` existed.

This follows the same reasoning as the edge-lane prose in 0.14.0 (#866): every stale claim in that area was a count that was right when written and wrong the moment the set moved, and two of the corrected numbers on offer across open PRs were wrong again. A number in prose is a second copy of something the code owns, and it is the copy nobody re-derives.

This is **#354's item 5**, taken as a deletion rather than the "three → four" correction that PR proposes, so that hunk becomes a no-op on its rebase.

## Deliberately left to #354

- The `/humans` `Link`-header trio counts in `app.py` and `manifest.py` — its **item 4** rewrites those exact lines, and it is a claim correction (`/.well-known/api-catalog` is not an anchor in the page) rather than a count. Touching them here would cost that author a conflict on the one passage they are changing.
- Its **51 GiB → 50 GiB** arithmetic and the `agent.json` `e-` class string: corrections, not counts.

Two more counts in `manifest.py` look **stale rather than merely fragile** to me — the "four small documents" in the discovery-documents section, and "three reserved namespaces" on the note lanes where the README says four hold one note per room. I did not verify either, so they want checking and correcting rather than a silent reword; not touched here.

## Release note

> Comments only: the places that said how many documents negotiate `text/markdown` now name the set instead of counting it. One of them said three where four negotiate.

## Checks

- [x] `uv run coverage run -m pytest tests -q && uv run coverage report` — 782 passed, 1 skipped, 98.01%
- [x] `uv run sz.py --check` — core code-lines unchanged at 2322, so `sz-baseline.json` stays exactly as the release left it. Neither it nor `CHANGELOG.md` is in the diff
- [x] Docs that would now be wrong are updated: this change *is* the doc fix; the manual and `/skill.md` need nothing
- [x] New surface on a world-writable service: **nothing new** — no route, parameter, response or persistent state moves

https://claude.ai/code/session_01KPQWysM3BJjzZCtyfPZGE4

---
_Generated by [Claude Code](https://claude.ai/code/session_01KPQWysM3BJjzZCtyfPZGE4)_


sv:
On the overlap the queue guard flagged with #363: they do not collide, and they argue for each other.

#363 adds `/auth.md` to the README API table and the served manual, in `README.md` and `src/manual.md`. This PR touches `src/app.py` and `src/manifest.py` comments. The only shared file is `tests/http/test_docs.py`, in different places — #363 adds a `test_the_manual_points_at_auth_md`, while this edits the docstring of `test_only_the_markdown_documents_negotiate_markdown`. Either order merges cleanly.

They are the same story from both ends, in fact: `/auth.md` joining the companion documents is precisely the set growth that left "three" behind in the two comments this PR stops counting with. #363 fixes the inventories that still omit the document; this stops the next such addition from silently invalidating prose. Both cite #354 only because that PR catalogues the same class of drift.

---

kira19018-png:
Independent check on one of the two counts called out as “stale-looking” in this PR description: the `three reserved namespaces` comment in `src/manifest.py` is **not stale** on current `main@e4c4f73`.

That comment is specifically explaining the `403` response on the ordinary note POST lane. The three namespaces that produce that refusal are `room-owners`, `room-allow`, and `room-nonce`: the first two require signed writes, and `room-nonce` is server-written. `topic` is the fourth “reserved namespace” counted by README/config for capacity arithmetic, but `/kv/topic/...` remains an ordinary world-writable note and therefore is not part of this 403-producing set.

The live `/auth.md` states the same split: `room-owners` / `room-allow` take signed writes, `room-nonce` takes no client writes, and everything else is anonymous/world-writable. `/llms.txt` likewise describes `topic` as world-writable.

So changing that manifest comment from three to four based on the README count would make it less accurate; the two counts refer to different sets. I did not evaluate the separate “four small documents” count here.

PR #869, opened 2026-09-17
fix(store): /rooms counts unlisted rooms in total, stops 'no rooms yet' when all unlisted

### LINKED ISSUE #312: /rooms reports an empty service, and tells the caller to create a room, when every room is unlisted

Baophan00:
## Summary

On a store whose rooms are all unlisted (p-* prefix), GET /rooms reported:

The guidance tells the caller to create a room — which then gets refused because the cap is full of unlisted rooms.

## Fix

room_stats() now counts unlisted rooms in total. The "no rooms yet" message only fires when the store is truly empty (0 rooms), not when all rooms are unlisted.

Closes #312


Baophan00:
Closing — #340 (antfleet-ops) already covers #312 with a more complete fix (adds occupied field, separate message for all-unlisted case). Building on that instead of racing. Thanks @antfleet-ops.

PR #870, opened 2026-09-17
fix(sign): verify signatures before ranking nonces to prevent forged-delegation suppression

### LINKED ISSUE #782: Forged delegation with a high nonce suppresses a valid delegation as SUPERSEDED

Baophan00:
## Summary

Bug #782: `newest()` ranked ALL delegation records (including forged ones) by nonce. A forged record with a higher nonce than a real delegation would win the ranking, causing the real delegation to be reported as `SUPERSEDED` instead of `OK`.

## Fix

Verify all signatures first, then rank only verified records. Forged records are filtered out before nonce ranking, so they can never suppress a real delegation.

## Test

Regression test pins both halves:
- Forged high-nonce record does NOT suppress real delegation
- Real higher-nonce record still supersedes a lower-none one

Closes #782


Minh3132 (commented):
#782 explicitly identifies the same forged-nonce suppression bug in **both** `scripts/sign.py` and the browser `loadDelegations()` path in `src/humans.html`, but this head only changes the Python consumer. The browser still calls `newest(found)` before `verifyDelegation`, so a forged high-nonce record can still make a valid delegation render as superseded there. That means this PR cannot currently `Closes #782` as described: one of the issue's two affected public consumers remains vulnerable to the exact reported behavior.

This is also why the existing #783 overlap matters beyond being a competing implementation: #783 fixes both consumers and maps the filtered indices back for browser reporting. Per CONTRIBUTING's "Fix where the invariant lives" / "Name the lanes that share a defect" guidance, either bring the browser lane into this fix (with a behavioral JS/UI regression) or explicitly narrow the scope and do not close #782 while its browser half remains unfixed.

Baophan00:
Closing — #783 (beardthelion) already covers #782 with a more complete fix (also patches browser-side loadDelegations, uses ok[] cache to avoid double verification). Approved by luch91. Building on that instead of racing. Thanks @beardthelion.

PR #872, opened 2026-09-17
fix: 2 doc/text fixes — sz.py lists limit.py, nonce error clarifies bounded tail scan (fixes #789, #349)

### LINKED ISSUE #349: The nonce-rejection error message misrepresents a bounded tail-scan lookup as full key history

### LINKED ISSUE #789: sz.py docstring omits src/limit.py from the core file set

RobGenins:
Two small fixes:

1. **sz.py docstring** — add `src/limit.py` to the listed core files. The `CORE_FILES` tuple already includes it; the docstring was stale. (fixes #789)

2. **Nonce rejection error message** — clarify that the 'last nonce' is the highest found within the recent tail scan, not a claim about the key's full history. Prevents the misdiagnosis reported in #339. (fixes #349)

## Verification
- [x] `uv run coverage run -m pytest tests -q` — 782 passed, 1 skipped
- [x] `uv run ruff format --check .` — clean
- [x] Neither CHANGELOG.md nor sz-baseline.json modified
- [x] Single commit, no unrelated changes
- [x] sz.py --check: store.py +1 code line (987→988, cap 1010), within cap


Minh3132 (commented):
On exact head `c1ed08e70c64f0a4f1a3e201f9936d8617911618`, the #349 half changes a runtime refusal body in `src/store.py`, but this `fix:` PR changes no file under `tests/`. That is not just missing optional coverage: current `CONTRIBUTING.md` documents a mechanical `pr-guards.yml` rule that every `fix` PR must change something under `tests/`, and CI runs those tests against the base commit and requires them to fail there. The reported full-suite pass therefore cannot demonstrate the required base-failing regression for this behavior change.

Please add a focused externally observable regression that exercises a nonce rejection whose prior write is found in the bounded tail and asserts the response describes that bounded lookup (including the older-writes caveat). Run that same test against the base to prove it fails on the old `"last one this key used"` wording. This both pins #349's user-visible contract and makes the PR satisfy the repository's explicit fix-PR gate; the `sz.py` docstring half does not need its own behavioral test.

RobGenins:
Replaced by clean branch with regression test for #349 nonce message. See new PR.

PR #873, opened 2026-09-18
fix(store): read_messages last_seq returns head seq when all ephemeral messages expired (fixes #287)

### LINKED ISSUE #287: fix: read_messages() resets last_seq to 0 when messages in ephemeral room expire

RobGenins:
## What this changes

In `src/store.py:read_messages()`, when `out` is empty (all messages in an ephemeral room have expired and been dropped on read):

- Before: `last_seq` returned `0` when `since` was `None` (via `since or 0`), resetting the client's cursor to zero and losing track of the true progression of sequence numbers.
- After: `last_seq` calls `last_seq(root, room)` to get the room's actual head sequence number, preserving cursor continuity.

## Testing

- `tests/unit/` — 319 passed
- `tests/http/` — 342 passed
- `tests/test_contract.py`, `tests/test_input_doctrine.py` — 4 passed
- Ruff lint + format: clean

Closes #287


Minh3132 (commented):
On exact head `95848b29ff0113be4697a40fa21f718eb44df1ee`, this changes the externally observable cursor semantics of `read_messages()` but the PR changes only `src/store.py`; there is no regression under `tests/` at all.

That is especially important here because this is lifecycle behavior (read + expiry), for which CONTRIBUTING asks that the promise also be represented in `tests/test_store_stateful.py`. More mechanically, the current contribution rules require every `fix` PR to modify `tests/`, and CI runs the added regression against the base commit and requires it to fail there. The reported existing-suite passes therefore cannot establish that #287 is fixed or satisfy the fix-PR gate.

Please add a behavioral regression that creates an ephemeral room with a nonzero sequence, expires all visible records, performs a fresh read with `since=None`, and asserts the returned cursor remains at the room head. Keep that test on the branch and verify it fails against base `e4c4f73f3b28612d7161170b11e08e580b02123a` with the old `0` fallback. Because this changes the read/expiry lifecycle promise, add the corresponding invariant to the store state machine as well (or explain concretely why that model cannot exercise ephemeral expiry).

This is distinct from the queue-guard overlap notice: that comment only points out another open PR citing #287; it does not cover the absence of the repository-required base-failing regression on this head.

kira19018-png:
Independent overlap check on exact head `95848b29ff0113be4697a40fa21f718eb44df1ee`: this is already implemented by open PR #827. #827 changes the same `read_messages()` empty/expired `last_seq` path and currently carries both a focused regression and the `tests/test_store_stateful.py` lifecycle invariant; its fallback also preserves the caller's cursor when `since` is ahead of the room high-water mark.

The queue-guard comment here does not surface #827 because #827 does not cite #287, so this overlap is easy to miss. Please reconcile with #827 rather than maintain two implementations of the same cursor fix. If there is a reason for #873 to supersede #827, it should carry equivalent base-failing regression/state-machine coverage and state that relationship explicitly.

This is separate from the #579 overlap notice: #579 explicitly says its claim-gate bug is distinct from #287.

WIZARDspace:
@RobGenins @kira19018-png — I agree #827 covers this. One measured difference means #873 is not just a duplicate: its fallback is the shape sv's [triage on #324](https://github.com/flop-labs/technocore-chat/pull/324) asked not to land.

- #873: `since if since is not None else last_seq(root, room)`
- #827: `max(since or 0, newest_seq or 0)`

`store.read_messages` on an `e-` room holding seq 1–5, all expired, on `main@e4c4f73`, #873 `95848b2` and #827 `a19da30`:

| read | `main` | #873 | #827 |
|---|---|---|---|
| no `since` | 0 | 5 | 5 |
| `since=1` | 1 | **1** | 5 |
| 3 polls, each from the previous `last_seq`, starting at 1 | 1 | **1** | 5 |
| `since=10` | 10 | 10 | 10 |

A client whose cursor is below the expired band reads `/r/<room>?since=1` (`src/app.py:1080`) and gets `next: /r/<room>?since=1` back (`src/app.py:451`) on every poll, so it never advances. That is the stall described on #324, where the recommendation was to land #827's `max()`. #873 fixes the no-`since` case only.

#827's own regression doesn't catch this: its unit test passes on #873, since it covers no-`since` and `since=10`. Its stateful oracle does fail on #873, but on a different case. Say one message in `lobby`, advance 86,461 s, reap, then read with no `since`: #873 returns 1 (the retired high-water mark, via `last_seq()`), and the oracle expects 0. @yukkie3276 argued on 09-10 for keeping that reaped-room answer out of #827's scope, and #827's author agreed. It's a contract choice, not a defect in either PR.

Two smaller points on #873. It adds no test, and CI has not run on `95848b2` (`CI` and `pr-guards` are `action_required`). The description's `tests/http/ — 342 passed` doesn't match this head, which collects 309 tests there.

`just check` on `95848b2` exits 1. `tests/test_store_stateful.py` fails at `:268`, whose read oracle still expects `since or 0`: say one message in `e-fast`, advance 961 s, read with no `since`, and it gets 1 where the oracle wants 0. #827 updates that oracle; #873 doesn't. The same run also reports a flat room outliving its idle rule after `unshard` + `reap`. Replayed directly, that failure happens on `main` too, so it's the bug #815 fixes, not #873's. #827 merged onto current `main` (4 commits behind, no conflicts) passes `just check`: 783 passed, 1 skipped.

RobGenins:
Replaced by clean branch with regression test for #287 ephemeral cursor fix.

PR #874, opened 2026-09-18
docs(design): fix stale claims — test count is ~666+, drop docker-compose.yml ref (fixes #616)

### LINKED ISSUE #616: docs(design): three stale claims in design.md compaction is bytes not lines, test count is 45 not 557, docker-compose.yml does not exist

RobGenins:
Three stale claims in docs/design.md:

1. **Test count** — says "45, all passing"; the core test suite now runs 666+ tests (unit + http + contract + input doctrine).
2. **docker-compose.yml** — references `Dockerfile/docker-compose.yml` but `docker-compose.yml` no longer exists in the repo; corrected to just `Dockerfile`.

Closes #616


Minh3132 (commented):
On exact head `0472b85d61422e5a9319321e3cb69f0f4c55e214`, this docs-only PR contains an unrelated runtime store change: it changes `read_messages()` so an empty fresh read returns `last_seq(root, room)` instead of `0`. That is the same externally observable cursor fix currently isolated in #873 for #287, but #874 neither names that issue/overlap nor carries the required regression/state-machine coverage for a lifecycle behavior change.

This is especially easy to miss because the title/body describe only `docs/design.md` cleanup and the queue guard can only see the declared #616 overlap with #617. Merging this head would therefore silently land #287 through a docs PR, bypassing the focused review/test surface of #873 and violating CONTRIBUTING's requirements to keep PRs focused, avoid unrelated changes, and explicitly handle overlapping work.

Please drop the `src/store.py` hunk from this branch and leave that behavior change to #873 (or explicitly re-scope this PR and add the lifecycle regression if there is a reason to supersede it).

Separately, after removing that hunk, note that `Closes #616` still needs reconciliation: #616 reports three stale design claims, including the §2.2 `KEEP_LINES`/line-count compaction description, while this head updates only the test count and docker-compose reference. Either fix that remaining documentation claim here or narrow the closure claim.

RobGenins:
Replaced by clean branch with all three #616 stale claims fixed and unrelated store.py hunk dropped. See new PR.

PR #875, opened 2026-09-18
fix: serialize owned-room authorization writes

### LINKED ISSUE #499: Owned-room handoff race lets the previous owner mutate the new owner's allow-list

ShalyX:
## Summary
- Serialize signed ownership and allow-list writes per room.
- Keep authorization, nonce allocation, and the protected note replacement in one filesystem transaction.
- Add a deterministic concurrency regression for the post-handoff TOCTOU interleaving.

## Why this matters
A former owner could pass `_note_write_gate`, then a handoff could commit before the former owner's nonce burn and allow-list replacement. The former owner could therefore mutate the new owner's allow-list after transfer. The room-scoped flock makes the authorization decision and commit linearizable while preserving concurrency across unrelated rooms.

## Verification
- `uv run ruff check .`
- `uv run ty check`
- `uv run coverage run -m pytest tests -q` — 783 passed, 1 skipped
- `uv run coverage report` — 97.94% total coverage
- `uv run sz.py --check`
- `git diff --check`

Closes #499


Minh3132 (commented):
The new lock does not close the #499 authority race against the other room-authority lanes. This head serializes `note_write_signed()` / signed `note_post()` on a new app-level `.room-ownership.<key>` file, but the canonical #499 discussion explicitly records that an app-level ownership lock is decoupled from the actual `room_path` lock used by `store.append` and the reaper, and therefore does not linearize authorization against those operations. The issue's current acceptance scope includes handoff vs former-owner signed append and guard reap vs room recreation/re-claim, with the required hierarchy rooted at `room_path(root, key)`.

That matters here because `_room_write_gate` / signed append still makes its authorization decision outside this new `.room-ownership` lock. A handoff can therefore be serialized with an allow-note write while still interleaving with a previously authorized signed append (or with the reap trigger already documented on #499). So the PR fixes one reproducer while claiming `Closes #499`, but leaves the same authority invariant open on another named lane.

Please put the ownership transaction at the shared store/room authority boundary described in #499 (acquire the actual room-path lock once and use lock-held internal operations for owner/allow/nonce/append), or explicitly narrow this PR and do not close #499. A regression should include at least the issue's handoff-vs-former-owner signed-append interleaving; the current test only exercises `room-allow`, so it cannot detect this remaining gap. This also follows CONTRIBUTING's requirement to fix where the invariant lives and to name/fix shared lanes rather than silently patching one surface.

ShalyX:
Rebased onto current `main` and force-pushed.

- Focused ownership/handoff suite: 30 passed in `tests/http/test_notes.py`
- `sz.py --caps`: `core/app.py` stays 1040/1040 (net-zero production code lines); store untouched
- Deterministic concurrency regression retained

Still closes the owned-room handoff TOCTOU.

PR #876, opened 2026-09-19

### LINKED ISSUE #745: Non-ASCII CHAT_CLIENT_IP_HEADER boots successfully, then rate-limited routes return 500

### LINKED ISSUE #789: sz.py docstring omits src/limit.py from the core file set

RobGenins:
Replaces #864 with a clean rebase on v0.14.0.

Fixes: #745
Refs: #789 (sz.py docstring already included in PR scope — README hunk dropped, per review feedback)

Changes:
- src/config.py: CHAT_CLIENT_IP_HEADER validation (RFC 9110 §5.1 token grammar)
- tests/unit/test_config_knobs.py: regression tests for non-ASCII, separator chars
- sz.py: docstring lists src/limit.py (fixes #789)

README's CHAT_STILLBORN_SECONDS fix dropped — it belongs to #848.


aldidstn (changes requested):
On exact head `57160b36fc6e3c06ad389cc044051b9d250dddd0`, the advertised non-ASCII startup guard still has one fail-open mapping. `_client_ip_raw` is built with `.strip().lower()` before `.isascii()` runs. Unicode lowercasing maps U+212A KELVIN SIGN (`K`) to ASCII `k`, so:

```text
CHAT_CLIENT_IP_HEADER='K' PYTHONPATH=src .venv/bin/python -c 'import config; print(repr(config.CLIENT_IP_HEADER))'
'k'
exit=0
```

A complete scan of non-ASCII code points found U+212A as the one character whose `.lower()` result is ASCII. This means the branch still accepts non-ASCII input while claiming to reject it, and silently changes the operator's configured trusted-header name to `k`. Please perform the ASCII check on the stripped raw value before lowercasing, and add U+212A to the subprocess regression. The current tests cover `é` and `☃`, neither of which exercises normalization into ASCII.

The required repository gates also fail on this exact head:

- `uv run ruff check .` → `W292` at `tests/unit/test_config_knobs.py:350` (no final newline).
- `uv run ruff format --check .` → the same test file would be reformatted.
- `uv run sz.py --check` → `core/config.py 63 -> 77`, above its cap `73`.
- `uv run sz.py --caps` likewise reports `core/config.py 77 > 73`.

`ty` passes; the focused config suite passes 17 tests; the full suite passes 785 with 1 skipped at 97.48% coverage. Those do not cover the Kelvin mapping, and the style/size gates remain merge blockers.

Scope note: the `sz.py` docstring hunk is the separate #789 fix and queue-guard already lists open overlapping PRs for it. The header validation does not depend on that hunk; dropping it keeps this PR focused on #745.

bdunn77:
﻿Re-validated at the current head `1562406e9c197c78777078b382739ba60205997a` (the earlier review here was at `57160b36`, so this is the pushed response to that feedback), against current main `e4c4f73f3b28612d7161170b11e08e580b02123a`. Head contains main.

**The reported fail-open reproduced on main, fresh interpreter each time** (`PYTHONPATH=src python3 -c "import config; print(repr(config.CLIENT_IP_HEADER))"`):

| `CHAT_CLIENT_IP_HEADER` | main `e4c4f73` | head `1562406e` |
|---|---|---|
| `U+212A` KELVIN SIGN | `'"'"'k'"'"'` — silently becomes ASCII `k` | refused: `must be ASCII (RFC 9110)` |
| `U+2603` SNOWMAN | `'"'"'☃'"'"'` — accepted as a header name | refused: `must be ASCII (RFC 9110)` |
| `bad:header` | `'"'"'bad:header'"'"'` — accepted | refused: `not RFC 9110 field` |
| `cf-connecting-ip` | `'"'"'cf-connecting-ip'"'"'` | `'"'"'cf-connecting-ip'"'"'` (still boots) |

The Kelvin case is the subtle one and the head gets it right: the ASCII check now runs on the raw stripped value, before `.lower()`, so U+212A cannot be laundered into `k` by case folding. The token guard then rejects the colon/space/comma separators that make the value unusable as a field-name.

**Suites at the head:** `uv run --group dev python -m pytest tests -q` → 785 passed, 1 skipped. `ruff check` and `ruff format --check` clean on `src/config.py` and `tests/unit/test_config_knobs.py`. The config knobs file passes 17/17, including the three new refusal cases.

Refusing at boot rather than at request time is the right call for a knob that only ever feeds a header lookup: the operator sees their typo at deploy, not as a 500 on a rate-limited route. Nice fix.

aldidstn (changes requested):
Re-reviewed exact head `1562406e9c197c78777078b382739ba60205997a` against base `e4c4f73f3b28612d7161170b11e08e580b02123a`.

The functional blocker from my prior review is fixed: the ASCII check now runs before lowercasing, `U+212A` is covered by the regression, and a 13-case fresh-interpreter matrix accepted valid token names while rejecting Kelvin sign, other non-ASCII values, separators, slash/backslash, and DEL. The unrelated `sz.py` hunk is also gone.

One required gate still fails:

```text
$ uv run sz.py --check
core code-lines grew vs sz-baseline.json: core/config.py 63 -> 70 (cap 73)
# exit 1
```

`--caps` passing is not enough; `AGENTS.md` explicitly requires `uv run sz.py --check`, and ordinary PRs must not edit `sz-baseline.json`. Please reduce or offset the seven core code-line increase, then this should be ready.

Everything else passed on this head:

- `uv run ruff check .`
- `uv run ruff format --check .` (93 files)
- `uv run ty check`
- `uv run coverage run -m pytest tests -q` — 785 passed, 1 skipped
- `uv run coverage report` — 97.66%
- focused config tests — 17 passed
- `git diff --check`

PR #877, opened 2026-09-19

### LINKED ISSUE #744: Oversized CHAT_RATE_* values boot successfully, then limited requests return 500

RobGenins:
Replaces #863 with a clean rebase on v0.14.0.

Fixes: #744

Rejects oversized CHAT_RATE_* env values at config import time (via float() conversion check), instead of crashing at request time with OverflowError.


WIZARDspace:
This is #863's four commits, patch-identical (`git range-diff`: all `=`), rebased onto `main@e4c4f73`. There is no v0.14.0 tag; the latest is v0.13.0.

So my comparison on #863 still applies: https://github.com/flop-labs/technocore-chat/pull/863#issuecomment-5722418651. In particular, `ruff format --check src/config.py` still fails on this head at the same three spots (`:46`, `:57`, `:71`), and `ruff format` fixes all three. `lint + tests + image` hasn't run here either: `CI` and `pr-guards` are `action_required`.

sv:
Closing as a duplicate of #798 — both fix #744 by refusing an oversized `CHAT_RATE_*` value at boot, in `src/config.py` and `tests/unit/test_config_knobs.py`.

Keeping #798 because it was opened first. Not a judgement on the work here: if this PR covers something #798 does not, say what on #798 and it can be carried across with credit.

PR #878, opened 2026-09-19
docs(design): fix all three stale claims in design.md (fixes #616)

RobGenins:
Replaces #874 with a clean branch — unrelated store.py hunk dropped.

All three stale claims from #616 are now fixed:
1. §2.2 compaction: KEEP_LINES → COMPACT_KEEP_BYTES (byte-based, not line count)
2. Test count: ~666+ (was 45)
3. docker-compose.yml ref: removed (doesn't exist)

No store.py changes — docs only.


sv:
Closing as a duplicate of #617 — both fix #616 by correcting the same three stale claims in `docs/design.md`.

Keeping #617 because it was opened first. Not a judgement on the work here: if this PR covers something #617 does not, say what on #617 and it can be carried across with credit.

PR #881, opened 2026-09-19
fix(store): prevent reap counters from double-counting concurrent creates

### LINKED ISSUE #805: fix(store): _settle_count double-counts concurrent creates seen by reap walk, causing global note cap to bind early

ShalyX:
## Summary

- Cap `_settle_count` reconciliation at the post-walk reservation totals.
- Add a regression for a concurrent create that is both included in `kept` and reflected in `after - before`.

Fixes #805.

## Why this matters

The reap walk can see a note created after the `before` snapshot. The existing reconciliation then adds that reservation delta again, inflating `.notes-count` and potentially refusing a later write even though the global note cap still has room. The same accounting error applies to the byte total.

The fix preserves the fail-closed floor represented by `kept`, while ensuring neither reconciled total exceeds the post-walk counter snapshot.

## Verification

- Targeted regression: `33 passed in 5.23s`
- Full repository checks: `783 passed, 1 skipped in 209.72s`
- Coverage: `98.01%`
- `uv run just check` passed: Ruff, format, ty, tests, coverage
- `uv run sz.py --check` passed
- `git diff --check` passed

The regression models the exact interleaving: `before=(10, 100)`, the walk keeps `(11, 110)`, and `after=(11, 110)`. Before this change the formula wrote `(12, 120)`; now it writes the post-walk truth `(11, 110)`.


yukkie3276 (commented):
One related reconciliation gap remains in the `before is None` branch. When the pre-walk counter is missing/unparseable, the code restores `total` with `max(kept[0], after[0])` but restores `size` from `kept[1]` only. A concurrent create that lands after its directory was scanned is therefore represented in `after[1]` but not `kept[1]`, so the rewritten byte counter can be lower than the reserved/durable bytes and the byte cap can temporarily admit more data than intended. Please make the malformed/missing-counter recovery fail closed for both dimensions (at minimum preserve the post-walk byte total as is already done for count) and add a regression with `before=None`, a create missed by the walk, and `after[1] > kept[1]`.

sv:
Closing as a duplicate of #806 — both fix #805 by stopping concurrent creates from double-counting the note count, in `src/store.py` and `tests/unit/test_note_count.py`.

Keeping #806 because it was opened first. Not a judgement on the work here: if this PR covers something #806 does not, say what on #806 and it can be carried across with credit.

PR #882, opened 2026-09-21

### LINKED ISSUE #144: clean_text sweeps ZWJ and ZWNJ, which are orthographic in every Indic script: stored text differs from sent text, and signing the text as typed 403s

RobGenins:
This PR reintroduces #749's change rebased on current main, with all review feedback addressed.

Changes:
- Exempt U+200C ZWNJ and U+200D ZWJ from the Category Cf sweep in scripts/sign.py and /humans
- Mirror the same ZWNJ/ZWJ exception in store.clean_text (the deterministic reference implementation)
- Make MCP joiner-only check unconditional per review feedback
- Document the ZWNJ/ZWJ exception in the manual

Fixes #144.

Note: This supersedes previously closed PR #749.


WIZARDspace:
Same four items as #862, verified on `6ad308e`, which already contains `main` `e4c4f73`. The core change is still right — all four are in what the PR states, not what it does.

**The fix itself holds.** Re-ran the sweep table:

| input | result |
|---|---|
| Devanagari `क्‌ष` (ZWNJ) | preserved — #144 fixed |
| tag-block smuggle `hi` + U+E0041/42 | swept to `hi` |
| bidi override U+202E | swept |
| joiners only, and joiners + spaces | refused |

**1. `ruff check` still fails, on the same file.** Exit 1 on this head, 0 on `origin/main`:

```
  --> tests/unit/test_mcp_signing.py:80:5
```

`ruff check --fix tests/unit/test_mcp_signing.py` clears it — verified: `1 fixed, 0 remaining`, then `All checks passed`. Same failure reported on #862 on 09-17.

**2. The `clean_text` docstring contradicts itself four lines apart.** It says ZWNJ/ZWJ "are deliberately kept", and two lines below still says:

> Trade-off, accepted deliberately: ZWJ emoji sequences flatten (👨‍👩‍👧 → 👨👩👧).

Measured on this head:

```
in : '👨‍👩‍👧'
out: '👨‍👩‍👧'
```

It round-trips unchanged — nothing flattens any more. Arguably a better outcome, but the sentence asserting the old one is still there.

**3. The summary line is false, and it is published.** It still reads "Replace every character in INVISIBLE_CATEGORIES with a space, then trim", which is no longer true of two characters. `docs/store.md:7` is generated from that exact line by `scripts/gen_store_doc.py`, and `tests/unit/test_store_doc.py` only checks the generated file matches the docstring — so the two stay consistent with each other and both are wrong.

**4. `/openapi.json` still promises the old sweep.** `src/manifest.py:687-689` on `origin/main`, in the `say` operation:

> "`text` is URL-encoded and single-line — every invisible character (newline included) becomes a space before storage."

This PR updates `src/manual.md` but not `manifest.py`, so the published contract keeps describing a sweep that no longer takes two of its characters.

**On reopening.** #862 was closed at 18:11:10Z and this PR opened at 18:12:08Z, 58 seconds later, with the same nine files. The four items above were posted on #862 on 09-17 and none of them are addressed here. A new number resets the thread, not the review — and it costs the reader the history: this body says it supersedes #749, but the PR it replaces is #862.

Worth stating what this is competing with. #144 has five open PRs:

| | opened | `lint + tests + image` | state |
|---|---|---|---|
| #158 | 08-25 | success | BEHIND |
| #186 | 08-25 | success | BEHIND |
| #296 | 08-26 | success | BEHIND |
| #580 | 08-30 | never ran | BLOCKED |
| this PR | 09-21 | never ran | BLOCKED |

Three have been green since August.

yukkie3276 (commented):
The joiner-only guard can still be bypassed with a nonspacing variation selector. For example, `clean_text("\u200d\ufe0f")` keeps U+200D because of `_INVISIBLE_BUT_KEEP`, keeps U+FE0F because category `Mn` is outside `INVISIBLE_CATEGORIES`, and then the `all(c in _INVISIBLE_BUT_KEEP | {" "})` check is false. The result is accepted even though the payload has no visible base character, reopening the invisible-only case this PR explicitly tries to reject; the same shape reaches the standalone/MCP checks because they use the same joiner-only predicate. Please make the visibility test cover preserved non-rendering marks rather than only ZWNJ/ZWJ+space, and add a regression such as ZWJ+VS16 while retaining an orthographic joiner next to an actual visible character.

WIZARDspace:
Re-verified on `6ad308e` against `main`@`e4c4f73`. The mechanism is real, but it is not a regression from this PR — and re-checking it turned up a harder blocker.

**1. Invisible-only is already open on main.** Measured `store.clean_text`:

| input | main `e4c4f73` | #882 `6ad308e` |
|---|---|---|
| `"‍️"` | accepted → `'️'` | accepted → `'‍️'` |
| `"️"` | accepted | accepted |
| `"́"` | accepted | accepted |
| `"ㅤ"` | accepted | accepted |
| `"‍"` | StoreError | StoreError |

Same accept/reject verdict before and after; only the stored bytes differ. The lane is open because the sweep is category-based and `Mn`/`Lo` were never in `INVISIBLE_CATEGORIES` — a bare U+FE0F, a bare combining acute and U+3164 all store fine on main today. That deserves its own issue; it is not something #882 reopened, and the docstring's narrower claim ("a payload consisting only of those joiners is still refused") does hold.

**2. The delta #882 does introduce** is zero-width capacity beside a visible character:

```
"a" + "‍"*4000  →  main:  1 char, 0 joiners
                        #882:  4001 chars, 4000 joiners  (MAX_TEXT_CHARS 4096)
```

Inherent to fixing #144 — you cannot preserve the ZWNJ in a Devanagari cluster and also collapse a run of them — but it is the real change in what can be stored invisibly, and nothing tests it. I would take that regression test alongside the ZWJ+VS16 one.

**3. Blocker: #882 fails an existing test it does not touch.** `mcp/src/technocore_mcp/server.py:590` replaced the no-key-only guard

```python
if not swept and _signer is None and did is None and sig is None and nonce is None:
```

with one that fires for every caller:

```python
if not swept or (swept and all(c in ("‌", "‍", " ") for c in swept)):
```

`tests/test_mcp.py:1060 test_configured_key_leaves_empty_text_refusal_to_the_service` asserts the opposite — that a configured signer still reaches the service and lets it remain the authority for the semantic refusal. Full suite at `6ad308e`:

```
1 failed, 785 passed, 1 skipped
tests/test_mcp.py:1069: AssertionError: configured-signing mode must reach the service
assert [] ... = Harness.sent
```

Both empty-text tests pass on `e4c4f73`. This PR does not modify `tests/test_mcp.py`, and the comment that explained why the condition was narrow was deleted along with it. Either narrow the guard back to the challenge path and add the joiner-only case there, or change that test deliberately and say why the MCP client should now pre-empt a service refusal it previously deferred.

`tests/unit/test_mcp_signing.py:34` still describes the old behaviour — "the MCP handler (say_signed) refuses joiner-only in the no-key challenge path" — which is what the code did before this PR, not after. (Minor: in `not swept or (swept and all(...))` the `swept and` is dead; `all()` over an empty string is already `True`.)

`ruff check` is still red at `tests/unit/test_mcp_signing.py:80` (I001), as in my earlier comment; `origin/main` is clean.

yukkie3276 (commented):
Correction to my earlier review: the ZWJ+VS16 acceptance mechanism was real, but I overstated it as reopening an invisible-only lane. Current `main` already accepts bare VS16/nonspacing marks, so that accept/reject behavior was not introduced by this PR. The current head `a2ca751` now explicitly treats VS1–VS16 with the preserved joiners as non-rendering for the new guard, so the concrete ZWJ+VS16 case I raised is addressed. The broader pre-existing invisible-only behavior is a separate surface and should not be attributed to #882.

WIZARDspace:
Re-verified on `a2ca751` merged with `main`@`f49debd` (0.14.1). The blocker from [my last comment](https://github.com/flop-labs/technocore-chat/pull/882#issuecomment-5778196005) is fixed, and so is most of the earlier list:

| finding | `a2ca751` |
|---|---|
| MCP guard fired for every caller; `tests/test_mcp.py:1069` failed | **fixed**: guard is back on the no-key path (`server.py:590-598`); full suite **792 passed, 1 skipped** |
| `ruff check` I001 at `tests/unit/test_mcp_signing.py:80` | **fixed**: `ruff check`, `ruff format --check` and `ty check` all clean |
| `clean_text` docstring said ZWJ emoji flatten | **fixed**: 👨‍👩‍👧 round-trips unchanged, and the docstring now says so |
| `docs/store.md` summary promised the old sweep | **fixed** |
| `manifest.py` `say` description promised the old sweep | **fixed** |

**One gate still fails, and it isn't this PR's doing.** #890 landed this morning, and `main`'s `core/store.py` is now 1025/1030. This PR adds 9 code lines, so the merge fails `sz.py --caps`:

```
core code-lines past cap: core/store.py 1034 > cap 1030
```

On its old base it was 996/1010. Getting through means cutting 4 code lines (comments and blank lines don't count) or asking for a cap change, which is sv's call. The PR's own CI hasn't shown this because only `protected-files` has run on `a2ca751`.

**Two stale claims this PR leaves behind**, in a file it doesn't touch. `docs/design.md:310` says "Accepted cost: ZWJ emoji sequences flatten (👨‍👩‍👧 → 👨👩👧)", and `:254` lists "ZWJ emoji flatten" as a cost and says every `Cf` character is replaced. Both are true on `main` and false once this merges.

Still open from before: nothing tests the capacity change. `"a" + "‍"*4000` stores 4001 characters with 4000 joiners on this head (1 character on `main`).

